PR #1, opened 2018-01-22
Make the SFFCorrector bspline more robust

barentsen:
This PR makes the `SFFCorrector` more robust by fitting its smoothing bspline against an arbitrary index instead of time.  This helps in the event of a big time gaps in the data, which make the bspline fit fail due to the knots being linearly spaced in time.

This PR very slightly changes the outcome of the SFF algorithm and, as a result, makes on of the unit tests fail.  This needs to be investigated.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -755,9 +755,9 @@ class SFFCorrector(object):
                 # First, fit a spline to capture the long-term varation
                 # We don't want to fit the long-term trend because we know
                 # that the K2 motion noise is a high-frequency effect.
-                self.bspline = self.fit_bspline(time[i], flux[i])
+                self.bspline = self.fit_bspline(np.arange(len(flux[i])), flux[i])
                 # Remove the long-term variation by dividing the flux by the spline
-                self.normflux = flux[i] / self.bspline(time[i] - time[i][0])
+                self.normflux = flux[i] / self.bspline(np.arange(len(flux[i])))
                 # Bin and interpolate normalized flux to capture the dependency
                 # of the flux as a function of arclength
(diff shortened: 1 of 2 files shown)


mirca:
@barentsen according to my investigations, this doesn't help :(

Before this PR:
screen shot 2018-01-21 at 9 48 40 pm

After this PR:
screen shot 2018-01-21 at 9 47 41 pm

barentsen:
Huh, I'm surprised it has such a big impact!

The alternative solution is to automatically detect those gaps in `time` that are longer than the interval between the knots of a bspline, and make sure the gap is a window boundary.

Possible pseudocode:
```
window_boundaries = []
for idx in range(len(time) - 1):
    if (time[idx + 1] - time[idx]) > interval_between_knots:
        window_boundaries.append(idx)
```

barentsen:
Also note: in `flatten()` we ignore the `time` values when passing the fluxes to the savgol filter.  If it is a bad idea here then it is probably also a bad idea there.

barentsen:
I guess we'd really want to implement the `breakpoints()` method that is already in SFFCorrector but not implemented.

*or* do we simply want to use `savgol` inside SFF?  Are you aware of a reason we (or Vanderburg) are using `bspline` in SFF rather than `savgol`?

mirca:
> I guess we'd really want to implement the ``breakpoints()``

Yes!!

> Are you aware of a reason we (or Vanderburg) are using bspline in SFF rather than savgol?

There is no particular reason. The goal is to solely remove the long term variation, which can be done with a savgol. I will try that shortly.

barentsen:
My attempt to improve `SFFCorrector` was not effective.  Closing for now until we revisit improvements.

PR #3, opened 2018-01-26
Added error catching for KeplerTargetPixelFile when passed the wrong file type.

christinahedges:
If passed the wrong file (e.g. a light curve) KeplerTargetPixelFile would not give a helpful error message. I've added a simple test to find if the file is a target pixel file.

I think the most robust test is to look for the 'FLUX' and 'QUALITY' keys in the fits file, as other keys can be missing and may not interfere with the rest of the functions.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -536,7 +536,7 @@ class KeplerLightCurveFile(object):
 
     @hdu.setter
     def hdu(self, value, keys=['SAP_FLUX','SAP_QUALITY']):
-        '''Raises a ValueError exception if self.hdu does not appear to be a Target Pixel File.
+        '''Raises a ValueError exception if value does not appear to be a Light Curve file.
         '''
         for key in keys:
             if ~(np.any([value[1].header[ttype] == key for ttype in value[1].header['TTYPE*']])):
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -530,6 +530,22 @@ class KeplerLightCurveFile(object):
         self.quality_bitmask = quality_bitmask
         self.quality_mask = self._quality_mask(quality_bitmask)
 
+    @property
+    def hdu(self):
+        return self._hdu
+
+    @hdu.setter
+    def hdu(self, value, keys=['SAP_FLUX','SAP_QUALITY']):
+        '''Raises a ValueError exception if self.hdu does not appear to be a Target Pixel File.
+        '''
+        for key in keys:
+            if ~(np.any([value[1].header[ttype] == key for ttype in value[1].header['TTYPE*']])):
+                raise ValueError("File {} does not have a {} column, "
+                         "is this a light curve file?".format(self.path, key))
+        else:
+            self._hdu = value
+
+
     def get_lightcurve(self, flux_type, centroid_type='MOM_CENTR'):
         if flux_type in self._flux_types():
             return KeplerLightCurve(self.hdu[1].data['TIME'][self.quality_mask],
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -535,16 +535,16 @@ class KeplerLightCurveFile(object):
         return self._hdu
 
     @hdu.setter
-    def hdu(self, value, keys=['SAP_FLUX','SAP_QUALITY']):
+    def hdu(self, value, keys=['SAP_FLUX', 'SAP_QUALITY']):
         '''Raises a ValueError exception if value does not appear to be a Light Curve file.
         '''
         for key in keys:
-            if ~(np.any([value[1].header[ttype] == key for ttype in value[1].header['TTYPE*']])):
+            if ~(np.any([value[1].header[ttype] == key
+                for ttype in value[1].header['TTYPE*']])):
                 raise ValueError("File {} does not have a {} column, "
                          "is this a light curve file?".format(self.path, key))
         else:
             self._hdu = value
-
 
     def get_lightcurve(self, flux_type, centroid_type='MOM_CENTR'):
         if flux_type in self._flux_types():
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -56,17 +56,23 @@ class KeplerTargetPixelFile(TargetPixelFile):
                  **kwargs):
         self.path = path
         self.hdu = fits.open(self.path, **kwargs)
-        self._test_file()
         self.quality_bitmask = quality_bitmask
         self.quality_mask = self._quality_mask(quality_bitmask)
 
-    def _test_file(self, keys=['FLUX', 'QUALITY']):
-        """Tests if the file has FLUX and QUALITY data.
-        """
+    @property
+    def hdu(self):
+        return self._hdu
+
+    @hdu.setter
+    def hdu(self, value, keys=['FLUX','QUALITY']):
+        '''Raises a ValueError exception if self.hdu does not appear to be a Target Pixel File.
+        '''
         for key in keys:
-            if ~(np.any([self.hdu[1].header[ttype] == key for ttype in self.hdu[1].header['TTYPE*']])):
-                raise ValueError("File {} does not have a {} extension, "
+            if ~(np.any([value[1].header[ttype] == key for ttype in value[1].header['TTYPE*']])):
+                raise ValueError("File {} does not have a {} column, "
                          "is this a target pixel file?".format(self.path, key))
+        else:
+            self._hdu = value
 
     def _quality_mask(self, bitmask):
         """Returns a boolean mask which flags all good-quality cadences.
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -64,11 +64,12 @@ class KeplerTargetPixelFile(TargetPixelFile):
         return self._hdu
 
     @hdu.setter
-    def hdu(self, value, keys=['FLUX','QUALITY']):
+    def hdu(self, value, keys=['FLUX', 'QUALITY']):
         '''Raises a ValueError exception if value does not appear to be a Target Pixel File.
         '''
         for key in keys:
-            if ~(np.any([value[1].header[ttype] == key for ttype in value[1].header['TTYPE*']])):
+            if ~(np.any([value[1].header[ttype] == key
+                for ttype in value[1].header['TTYPE*']])):
                 raise ValueError("File {} does not have a {} column, "
                          "is this a target pixel file?".format(self.path, key))
         else:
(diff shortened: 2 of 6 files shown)


lightkurve/targetpixelfile.py
```diff
@@ -51,13 +51,23 @@ class KeplerTargetPixelFile(TargetPixelFile):
...
         self.hdu = fits.open(self.path, **kwargs)
+        self._test_file()
         self.quality_bitmask = quality_bitmask
         self.quality_mask = self._quality_mask(quality_bitmask)
 
+    def _test_file(self, keys=['FLUX', 'QUALITY']):
+        """Tests if the file has FLUX and QUALITY data.
+        """
+        for key in keys:
+            if ~(np.any([self.hdu[1].header[ttype] == key for ttype in self.hdu[1].header['TTYPE*']])):
```
barentsen: A slightly shorter version could be `~(np.any([ttype[1] == key for ttype in f[1].header['TTYPE*'].cards]))`, but yours is equally good I think.

lightkurve/targetpixelfile.py
```diff
@@ -51,13 +51,23 @@ class KeplerTargetPixelFile(TargetPixelFile):
...
+                raise ValueError("File {} does not have a {} extension, "
```
barentsen: I'd change `extension` into `column` in the error message.   An `extension` has a different meaning in FITS speak.

mirca:
I slightly lean towards not including this in the code for a few reasons:
(i) it catches a rather silly mistake made by the user (i.e., trying to load a lightcurve file using a tpf class). What I mean by this is that, I think it should be the user's job to check the file she/he is inputing.
(ii) it makes the code a little less readable. Imagine a developer/user is reading through the code; starting from the constructor, I think, the first two lines make sense (i.e., save the path and open the file), however, "self._test_file()", although with an intuitive name, does something very obscure to check whether or not the file is valid.
(iii) if we merge this, then we should be consistent and copy this code over to ``KeplerLightCurveFile``, which I think is not a good idea.

Having said that, I think it's up to @barentsen.

lightkurve/targetpixelfile.py
```diff
@@ -51,13 +51,23 @@ class KeplerTargetPixelFile(TargetPixelFile):
...
     def __init__(self, path, quality_bitmask=KeplerQualityFlags.DEFAULT_BITMASK,
                  **kwargs):
         self.path = path
         self.hdu = fits.open(self.path, **kwargs)
+        self._test_file()
         self.quality_bitmask = quality_bitmask
         self.quality_mask = self._quality_mask(quality_bitmask)
 
+    def _test_file(self, keys=['FLUX', 'QUALITY']):
+        """Tests if the file has FLUX and QUALITY data.
```
barentsen: For clarity, I'd change the docstring into `Raises a ValueError exception if `self.hdu` does not appear like a Target Pixel File.`.

barentsen:
I think it is a good idea to print a friendly error message if the file is not a TPF.  In particular, I expect it will be common for users to confuse a lightcurve file with a tpf, so it would be particularly good to have a custom error message in such case that says: `You are trying to load a lightcurve file using the KeplerTargetPixelFile class.  Please use the KeplerLightCurveFile class instead!`

Right now, if you accidentally load a lightcurve as a tpf, you get this monster of an exception which isn't user-friendly:

```
In [3]: f = KeplerTargetPixelFile("ktwo201885041-c14_llc.fits")
/home/gb/bin/anaconda/lib/python3.6/site-packages/astropy/io/fits/column.py in _get_index(names, key)
   1825         try:
-> 1826             indx = names.index(key.rstrip())
   1827         except ValueError:

ValueError: 'QUALITY' is not in list

KeyError                                  Traceback (most recent call last)
<ipython-input-3-9ac2e6474268> in <module>()
----> 1 f = KeplerTargetPixelFile("/home/gb/tmp/wolf/ktwo201885041-c14_llc.fits")

/home/gb/dev/lightkurve/lightkurve/targetpixelfile.py in __init__(self, path, quality_bitmask, **kwargs)
     57         self.hdu = fits.open(self.path, **kwargs)
     58         self.quality_bitmask = quality_bitmask
---> 59         self.quality_mask = self._quality_mask(quality_bitmask)
     60 
     61     def _quality_mask(self, bitmask):

/home/gb/dev/lightkurve/lightkurve/targetpixelfile.py in _quality_mask(self, bitmask)
     71         elif isinstance(bitmask, str):
     72             bitmask = KeplerQualityFlags.OPTIONS[bitmask]
---> 73         return (self.hdu[1].data['QUALITY'] & bitmask) == 0
     74 
     75     def header(self, ext=0):

/home/gb/bin/anaconda/lib/python3.6/site-packages/astropy/io/fits/fitsrec.py in __getitem__(self, key)
    483 
    484         if isinstance(key, string_types):
--> 485             return self.field(key)
    486         elif isinstance(key, (slice, np.ndarray, tuple, list)):
    487             # Have to view as a recarray then back as a FITS_rec, otherwise the

/home/gb/bin/anaconda/lib/python3.6/site-packages/astropy/io/fits/fitsrec.py in field(self, key)
    681         # NOTE: The *column* index may not be the same as the field index in
    682         # the recarray, if the column is a phantom column
--> 683         column = self.columns[key]
    684         name = column.name
    685         format = column.format

/home/gb/bin/anaconda/lib/python3.6/site-packages/astropy/io/fits/column.py in __getitem__(self, key)
   1431     def __getitem__(self, key):
   1432         if isinstance(key, string_types):
-> 1433             key = _get_index(self.names, key)
   1434 
   1435         x = self.columns[key]

/home/gb/bin/anaconda/lib/python3.6/site-packages/astropy/io/fits/column.py in _get_index(names, key)
   1833                 indx = names.index(_key)
   1834             elif count == 0:
-> 1835                 raise KeyError("Key '{}' does not exist.".format(key))
   1836             else:              # multiple match
   1837                 raise KeyError("Ambiguous key name '{}'.".format(key))

KeyError: "Key 'QUALITY' does not exist."
```

mirca:
Ok, then I propose instead of creating `_test_tpf`, we just use nice Python properties:

```
@property
def hdu(self):
    return self._hdu

@hdu.setter
def hdu(self, value):
    for key in keys:
        if ~(np.any([value[1].header[ttype] == key for ttype in value[1].header['TTYPE*']])):
            raise ValueError("File {} does not have a {} extension, "
                     "is this a target pixel file?".format(self.path, key))
```

christinahedges:
@mirca has the nicest solution. Very pretty. Are we happy with the error message:

'ValueError: File data/kplr006922244-2010078095331_llc.fits does not have a FLUX column, is this a target pixel file?'

barentsen:
:+1: LGTM!

christinahedges:
I've added the same beautiful python code to KeplerLightCurveFile and had it look for SAP_FLUX and SAP_QUALITY

lightkurve/lightcurve.py
```diff
@@ -530,6 +530,22 @@ def __init__(self, path, quality_bitmask=KeplerQualityFlags.DEFAULT_BITMASK,
         self.quality_bitmask = quality_bitmask
         self.quality_mask = self._quality_mask(quality_bitmask)
 
+    @property
+    def hdu(self):
+        return self._hdu
+
+    @hdu.setter
+    def hdu(self, value, keys=['SAP_FLUX','SAP_QUALITY']):
+        '''Raises a ValueError exception if self.hdu does not appear to be a Target Pixel File.
```
barentsen: Change `does not appear to be a Target Pixel File.` into `does not appear to be a Kepler light curve file.`

lightkurve/targetpixelfile.py
```diff
@@ -56,17 +56,23 @@ def __init__(self, path, quality_bitmask=KeplerQualityFlags.DEFAULT_BITMASK,
...
-    def _test_file(self, keys=['FLUX', 'QUALITY']):
-        """Tests if the file has FLUX and QUALITY data.
-        """
+    @property
+    def hdu(self):
+        return self._hdu
+
+    @hdu.setter
+    def hdu(self, value, keys=['FLUX','QUALITY']):
+        '''Raises a ValueError exception if self.hdu does not appear to be a Target Pixel File.
```
barentsen: Change `self.hdu` into `value`

lightkurve/lightcurve.py
```diff
@@ -530,6 +530,22 @@ def __init__(self, path, quality_bitmask=KeplerQualityFlags.DEFAULT_BITMASK,
         self.quality_bitmask = quality_bitmask
         self.quality_mask = self._quality_mask(quality_bitmask)
 
+    @property
+    def hdu(self):
+        return self._hdu
+
+    @hdu.setter
+    def hdu(self, value, keys=['SAP_FLUX','SAP_QUALITY']):
```
mirca: white space after comma

lightkurve/lightcurve.py
```diff
@@ -530,6 +530,22 @@ def __init__(self, path, quality_bitmask=KeplerQualityFlags.DEFAULT_BITMASK,
...
+        '''Raises a ValueError exception if value does not appear to be a Light Curve file.
+        '''
+        for key in keys:
+            if ~(np.any([value[1].header[ttype] == key for ttype in value[1].header['TTYPE*']])):
+                raise ValueError("File {} does not have a {} column, "
+                         "is this a light curve file?".format(self.path, key))
+        else:
+            self._hdu = value
+
+
```
mirca: no need for this blank line here

lightkurve/targetpixelfile.py
```diff
@@ -51,13 +51,29 @@ class KeplerTargetPixelFile(TargetPixelFile):
...
+    @property
+    def hdu(self):
+        return self._hdu
+
+    @hdu.setter
+    def hdu(self, value, keys=['FLUX','QUALITY']):
+        '''Raises a ValueError exception if value does not appear to be a Target Pixel File.
+        '''
+        for key in keys:
+            if ~(np.any([value[1].header[ttype] == key for ttype in value[1].header['TTYPE*']])):
```
mirca: you can break the line here after ``key``

mirca:
@christinahedges Could you add a unit test for this?

mirca:
Not for this PR, but later we could then add a class called "FileValidator" in order to avoid the duplication of code introduced.

christinahedges:
Probably worth it at some point if you want in the future validate e.g. a TESS tpf and let it pass even if they don't have the same header columns.

lightkurve/tests/test_targetpixelfile.py
```diff
@@ -6,7 +6,15 @@
...
 filename_tpf_one_center = get_pkg_data_filename("data/test-tpf-non-zero-center.fits")
+TABBY_Q8 = ("https://archive.stsci.edu/missions/kepler/lightcurves"
+            "/0084/008462852/kplr008462852-2011073133259_llc.fits")
 
+def test_load_bad_file():
+    '''Test if a light curve can be opened without exception.'''
+    try:
+        tpf = KeplerTargetPixelFile(TABBY_Q8)
+    except ValueError:
+        assert True
```
mirca: You could do:
```
with pytest.raises(ValueError) as exc:
    tpf = KeplerTargetPixelFile(TABBY_Q8)

assert('part of the error message you wrote here' in exc.value.args[0])
```

PR #4, opened 2018-01-26
Yet another box period search

mirca:
``fast_box_period_search`` is 2x faster than ``box_period_search`` and it seems surprisingly more
accurate (although I didn't do many experiments to support this).

screen shot 2018-01-26 at 11 52 37 am
screen shot 2018-01-26 at 11 52 49 am

my notes on bls:
[bls.pdf](https://github.com/KeplerGO/lightkurve/files/1669172/bls.pdf)

todo: add docstrings

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1212,23 +1212,26 @@ def box_period_search(lc, min_period=0.5, max_period=30, nperiods=2000,
     return log_posterior, trial_periods, trial_periods[np.argmax(log_posterior)]
 
 
-def fast_box_period_search(lc, niters=5, min_period=0.5, max_period=30, nperiods=2000,
+def fast_box_period_search(lc, niters=2, min_period=0.5, max_period=30, nperiods=2000,
                            period_scale='linear'):
 
     t = np.linspace(-.5, .5, len(lc.time))
+    prior_to = oktopus.prior.UniformPrior(lb=-.5, ub=.5)
+    prior_width = oktopus.prior.UniformPrior(lb=.05, ub=.2)
+    m, n = np.nanmean(lc.flux), len(lc.flux)
 
     def lnlikelihood(args, amplitude=None, depth=None, data=None):
         to, width = args
-        ll = ((data - amplitude) ** 2 * ((t < to) + (t >= to + width))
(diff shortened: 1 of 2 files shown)


barentsen:
wowwwww

barentsen:
Impressive!!  Two questions:
- In the numerical optimization, do you know how many iterations there are?  I ask to understand how much slower or faster it would be to do a grid search.
- The advantage of a grid search may be that you'd be able to deal with a multi-modal case of two planets in the same period, but I guess planets like that don't even exist.

barentsen:
Your PDF is so much easier to understand and follow than the BLS paper and its funny weights.

mirca:
> In the numerical optimization, do you know how many iterations there are? I ask to understand how much slower or faster it would be to do a grid search.

It takes on average 3 iterations

barentsen:
Do you mean there are 3 iterations between the numerical and the analytical step, or the numerical optimization only requires 3 steps to find the maximum likelihood?  I'm wondering about the latter.

mirca:
> Do you mean there are 3 iterations between the numerical and the analytical step, or the numerical optimization only requires 3 steps to find the maximum likelihood? I'm wondering about the latter.

The latter.

mirca:
The marginalized loglikehood as a function of ``to`` and ``w`` respectively
screen shot 2018-01-26 at 2 48 13 pm
screen shot 2018-01-26 at 2 48 20 pm

mirca:
@barentsen I plan to merge this provided tests pass. How does that sound?

lightkurve/lightcurve.py
```diff
@@ -1224,3 +1226,114 @@ def box(amplitude, depth, to, width):
...
+    log_posterior = []
+    for p in tqdm(trial_periods):
+        folded = lc.fold(period=p)
+        amplitude_star, depth_star = m, 1e-4
+        width_star = .1
+        if np.nanmean(folded.flux[t < 0]) > np.nanmean(folded.flux[t > 0]):
+            to_star = .25
+        else:
+            to_star = -.25
+        for i in range(niters):
```
barentsen: Do you want to add a few lines of comments inside these for loops, just to guide the untrained eye where in the algorithm we are?

barentsen:
Thanks!  LGTM :+1:

PR #5, opened 2018-01-26
Added `__repr__` for KeplerTargetPixelFile, Lightcurve, KeplerLightCurve and KeplerLightCurveFile

christinahedges:
lightkurve now prints useful information to the user if you print `KeplerTargetPixelFile`, `Lightcurve`, `KeplerLightCurve` and `KeplerLightCurveFile`:

e.g.
screen shot 2018-01-26 at 2 43 42 pm

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -42,6 +42,9 @@ class LightCurve(object):
             self.flux_err = np.nan * np.ones_like(self.time)
         self.meta = meta
 
+    def __repr__(self):
+        return('LightCurve Object')
+
     def stitch(self, *others):
         """
         Stitches LightCurve objects.
@@ -472,6 +475,9 @@ class KeplerLightCurve(LightCurve):
         self.cadenceno = cadenceno
         self.keplerid = keplerid
 
+    def __repr__(self):
+        return('KeplerLightCurve Object (ID: {})'.format(self.keplerid))
+
     def correct(self, method='sff', **kwargs):
         """Corrects a lightcurve for motion-dependent systematic errors.
 
@@ -529,6 +535,9 @@ class KeplerLightCurveFile(object):
         self.hdu = pyfits.open(self.path, **kwargs)
         self.quality_bitmask = quality_bitmask
         self.quality_mask = self._quality_mask(quality_bitmask)
+
+    def __repr__(self):
+        return('KeplerLightCurveFile Object (ID: {})'.format(self.hdu[0].header['KEPLERID']))
 
     @property
     def hdu(self):
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -533,11 +533,12 @@ class KeplerLightCurveFile(object):
                  **kwargs):
         self.path = path
         self.hdu = pyfits.open(self.path, **kwargs)
+        self.keplerid = self.hdu[0].header['KEPLERID']
         self.quality_bitmask = quality_bitmask
         self.quality_mask = self._quality_mask(quality_bitmask)
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


barentsen:
Nice!!  Let's also print the quarter/campaign?

christinahedges:
@mirca and I were just discussing this, I think it might be better to have a more verbose __str__ but keep the __repr__ as a simple one line?

__str__ might pass back e.g. the time of the observation, campaign, Kepler/K2, module as well as the ID

barentsen:
Ohh, right, I forgot that `repr` and `str` are different.   Right, let's make `str` super-verbose and repr very simple!

lightkurve/lightcurve.py
```diff
@@ -530,6 +536,9 @@ def __init__(self, path, quality_bitmask=KeplerQualityFlags.DEFAULT_BITMASK,
         self.quality_bitmask = quality_bitmask
         self.quality_mask = self._quality_mask(quality_bitmask)
 
+    def __repr__(self):
+        return('KeplerLightCurveFile Object (ID: {})'.format(self.hdu[0].header['KEPLERID']))
```
mirca: I think there is a property called ``keplerid``, if there is not, let's add it.

lightkurve/lightcurve.py
```diff
@@ -42,6 +42,9 @@ def __init__(self, time, flux, flux_err=None, meta={}):
             self.flux_err = np.nan * np.ones_like(self.time)
         self.meta = meta
 
+    def __repr__(self):
+        return('LightCurve Object')
+
```
mirca: I think we don't need this for now

christinahedges: I don't agree I think this is user friendly and a quick way to distinguish between the three classes of light curve that you can have.

barentsen: Right now it says `<lightkurve.lightcurve.LightCurve object at 0x7f3dd738a780>`.  You could argue that is more informative than `LightCurve Object`.

christinahedges: We did argue, I think it's more useful for a normal user. I am worried that people will be confused when there are three possible objects that their light curve can be. I think printing a very simple 'LightCurve', 'KeplerLightCurve' or 'KeplerLightCurveFile' will help people debug why they can't run their favourite methods on their lightcurve.

barentsen: Do you mean that users will glance over `<lightkurve.lightcurve.LightCurve object at 0x7f3dd738a780>` because it looks scary, whereas `LightCurve Object` is more readable.  I think developers would deffo prefer the former, but I can see that new users may prefer the latter?

christinahedges: Yeah that's exactly my point, it just makes it less scary for users. I reasoned that most of our users won't be developers. We should try to be as user friendly as possible, right?

christinahedges:
@barentsen if you are happy with this can you please merge? Thx

lightkurve/lightcurve.py
mirca: I think that's not the purpose of ``__repr__``. I still would prefer ``<lightkurve.lightcurve.LightCurve object at 0x7f3dd738a780>``.

reference:
https://stackoverflow.com/questions/1436703/difference-between-str-and-repr-in-python

christinahedges: We're not losing information so much as improving readability. I still think this is more helpful to our average user. And if we're doing `__repr__` like this for all the other classes it makes sense they all look like this.

barentsen:
I am going to merge it without `LightCurve.__repr__` for now.  I can see good arguments for and against.  Let's discuss in real life!

PR #8, opened 2018-01-31
Changed tpf.plot() to return only ax and accept ax as keyword.

christinahedges:
tpf.plot() now returns just 'ax' instead of figure and ax, in line with how lc.plot() works. tpf.plot() also now accepts an axis, just as lc.plot() does. This allows users to put target pixel file plots into their existing figures, which I find to be useful. 

I have also added these minor tweaks

- The colorbar can now be turned off with a keyword. I find this to be useful if I'm e.g. showing the source is uncontaminated.
- The color for the aperture mask can now be specified.

These tweaks have let me create this nice plot.

rollingbandcheck

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -316,13 +316,16 @@ class KeplerTargetPixelFile(TargetPixelFile):
 
         return col_centr, row_centr
 
-    def plot(self, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
-             **kwargs):
+    def plot(self, ax=None, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
+            colorbar=True, mask_color='pink', **kwargs):
         """
         Plot a target pixel file at a given frame (index) or cadence number.
 
         Parameters
         ----------
+        ax : matplotlib.axes._subplots.AxesSubplot
+            A matplotlib axes object to plot into. If no axes is provided,
+            a new one be generated.
         frame : int
             Frame number. The default is 0, i.e. the first frame.
         cadenceno : int, optional
@@ -332,8 +335,18 @@ class KeplerTargetPixelFile(TargetPixelFile):
             If True, background will be added to the pixel values.
         aperture_mask : ndarray
             Highlight pixels selected by aperture_mask.
+        colorbar : bool
+            Whether or not to show the colorbar
+        mask_color : str
+            Color to show the aperture mask
         kwargs : dict
             Keywords arguments passed to `lightkurve.utils.plot_image`.
+
+
+        Returns
+        -------
+        ax : matplotlib.axes._subplots.AxesSubplot
+            The matplotlib axes object.
         """
         if cadenceno is not None:
             try:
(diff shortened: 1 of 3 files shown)


lightkurve/targetpixelfile.py
```diff
@@ -316,13 +316,16 @@ def centroids(self, aperture_mask='pipeline'):
...
+    def plot(self, ax=None, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
+            colorbar=True, mask_color='pink', **kwargs):
         """
         Plot a target pixel file at a given frame (index) or cadence number.
 
         Parameters
         ----------
+        ax : matplotlib.axes._subplots.AxesSubplot
+            A matplotlib axes object to plot into. If no axes is provided,
+            a new one be generated.
```
mirca: ``a new one be generated `` ``--->`` ``a new one will be generated`` (?)

lightkurve/targetpixelfile.py
```diff
@@ -332,8 +335,18 @@ def plot(self, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
             If True, background will be added to the pixel values.
         aperture_mask : ndarray
             Highlight pixels selected by aperture_mask.
+        colorbar : bool
```
mirca: How about ``show_colorbar``?

lightkurve/targetpixelfile.py
```diff
@@ -332,8 +335,18 @@ def plot(self, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
...
+            Whether or not to show the colorbar
+        mask_color : str
+            Color to show the aperture mask
         kwargs : dict
             Keywords arguments passed to `lightkurve.utils.plot_image`.
+
+
```
mirca: Remove one of these blank lines :)

lightkurve/targetpixelfile.py
christinahedges: 👍

lightkurve/targetpixelfile.py
christinahedges: Sure

lightkurve/targetpixelfile.py
```diff
@@ -350,19 +363,25 @@ def plot(self, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
...
                              extent=(self.column, self.column + self.shape[2],
-                             self.row, self.row + self.shape[1]), **kwargs)
-
+                             self.row, self.row + self.shape[1]), colorbar=colorbar,
+                             **kwargs)
+        else:
+            ax = plot_image(pflux, ax=ax, title='Kepler ID: {}'.format(self.keplerid),
+                             extent=(self.column, self.column + self.shape[2],
+                             self.row, self.row + self.shape[1]), colorbar=colorbar,
+                             **kwargs)
```
mirca: Looks like the following code is sufficient here (?)
```
ax = plot_image(pflux, ax=ax, title='Kepler ID: {}'.format(self.keplerid),
                extent=(self.column, self.column + self.shape[2], self.row, self.row + self.shape[1]),
                colorbar=colorbar, **kwargs)
```

lightkurve/utils.py
```diff
@@ -201,10 +205,19 @@ def plot_image(image, scale='linear', origin='lower',
             Label for the color bar.
         title : str or None
             Title for the plot.
+        colorbar : bool
+            Whether or not to show the colorbar
         kwargs : dict
             Keyword arguments to be passed to `matplotlib.pyplot.imshow`.
+
+
```
mirca: bl

lightkurve/targetpixelfile.py
christinahedges: Yas

lightkurve/utils.py
```diff
@@ -201,10 +205,19 @@ def plot_image(image, scale='linear', origin='lower',
...
+        Returns
+        -------
+        ax : matplotlib.axes._subplots.AxesSubplot
+            The matplotlib axes object.
         """
-        fig, ax = plt.subplots()
+        if ax is None:
+            fig, ax = plt.subplots()
```
mirca: Since `fig` is not used anywhere let's replace it by ``_``

lightkurve/utils.py
christinahedges: taken
  
care of

mirca:
This is awesome!! 👍 👍 👍 👍

mirca:
Travis is happy therefore everyone is happy! 🎉

barentsen:
:+1:

PR #13, opened 2018-02-04
Add a cdpp() convenience function

barentsen:
I propose adding a `cdpp(flux)` convenience function, so that one can do `from lightkurve import cdpp` without having to instantiate a `LightCurve` object.

### DIFF
--- a/lightkurve/convenience.py
+++ b/lightkurve/convenience.py
@@ -1,0 +1,28 @@
+import numpy as np
+
+from .lightcurve import LightCurve
+
+
+__all__ = ['cdpp']
+
+
+def cdpp(flux, **kwargs):
+    """A convenience function which wraps LightCurve.cdpp().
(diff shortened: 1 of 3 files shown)


mirca:
LGTM

lightkurve/convenience.py
```diff
@@ -0,0 +1,28 @@
+import numpy as np
+
```
mirca: Maybe we could remove this blank line?

lightkurve/tests/test_convenience.py
```diff
@@ -0,0 +1,12 @@
+import numpy as np
+from numpy.testing import assert_almost_equal
+
```
mirca: Also this one?

mirca: nvm

PR #16, opened 2018-02-07
Added WCS functions to TPF

### LINKED ISSUE #14: Consider adding a `KeplerTargetPixelFile.overlay_on_sky()` function
This would not necessarily replace `k2-pix`, because the latter provides a command-line interface etc...  Instead, perhaps `lightkurve` could provide the very basic functionality which `k2-pix` imports.

christinahedges:
> I've been dreaming about a tpf.get_wcs() 
_@barentsen_

These dreams have been addressed in this pull request. KeplerTargetPixelFile now has two functions `get_wcs` and `get_radec`.

* `tpf.get_wcs()` returns an astropy.wcs.WCS object built from the fits file
* `tpf.get_radec()` returns two, 3D arrays (same shape as flux) containing RA and Dec. There is one value for every pixel, for every time stamp. This takes into account the motion polynomials using POS_CORR1/POS_CORR2 keywords from the fits file. This allows you to plot all the flux values on to each RA and Dec value on the sky. Will allow us to line up TPFs with other sky catalogs.  You can use them to plot all the cadences 'correcting' for motion like this:

test

There is a unit test to check these run smoothly, perhaps more unit tests would be better?

Fixes #14


lightkurve/targetpixelfile.py
```diff
@@ -109,6 +110,61 @@ def get_prf_model(self):
...
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            ok = []
+            #Find extensions with valid WCS solution
+            vals = ['CRPIX', 'CRVAL']
+            for idx, hdu in enumerate(self.hdu):
+                w = WCS(hdu.header)
+                if (np.any(np.asarray([v in k for k in hdu.header.keys() for v in vals]))):
+                    ok.append(idx)
+            #Use the first extention with a valid solution.
```
barentsen: Why not use the WCS for the FLUX extension instead of trying all keywords?

lightkurve/targetpixelfile.py
```diff
@@ -109,6 +110,61 @@ def get_prf_model(self):
...
+            ok = ok[0]
+            w = WCS(self.hdu[ok].header)
+            return (w)
+
+    def get_radec(self):
```
barentsen: I wonder if we want to have a `cadenceno` argument and return two more easy-to-use 2D arrays (ra, dec) for a specified cadence?  In this scenario, I'd give the function a more verbose name such as `get_coordinate_matrices()`?

lightkurve/targetpixelfile.py
christinahedges: Because sometimes the WCS is missing, and I wanted to be future proof for TESS. I can change it?

lightkurve/targetpixelfile.py
christinahedges: I like `get_coordinate_matrices()`. I can have that be passed a cadenceno, and if it's none it returns the whole lot?

PR #17, opened 2018-02-07
Added tpf.date() and lc.date() for human readable dates

### LINKED ISSUE #15: tpf should have a method to output the time in human date.
It would be useful for people to be able to see light curves with dates instead of MJD. We can use astropy.time.Time and datetime to create a tpf.time_readable or something similar. These objects also plot up nicely.

christinahedges:
Using the same procedure as k2flix, I've added `tpf.date` and `lc.date` as properties to KeplerTargetPixelFile and KeplerLightCurveFile. These return arrays of strings with the date of the observations. e.g.:

screen shot 2018-02-06 at 4 37 43 pm

Fixes #15


lightkurve/lightcurve.py
```diff
@@ -608,6 +609,15 @@ def time(self):
         """Time measurements"""
         return self.hdu[1].data['TIME'][self.quality_mask]
 
+    @property
+    def date(self):
+        """Returns the human-readable date for all good-quality cadences."""
+        bjd = self.time + 2454833.
+        jd = bjd - self.hdu[1].data['TIMECORR'][self.quality_mask]
+        jd += (0.25 + 0.62 * (5 - self.hdu[1].header['TIMSLICE'])) / 86400.
```
mirca: @barentsen curiosity here: does this factor of ``86400`` has anything to do with the cadence type (long, short, super-long)?

christinahedges:
For @mirca the time stuff comes from here:

https://archive.stsci.edu/kepler/manuals/archive_manual.pdfs

Section 2.3.1

christinahedges:
Can we merge this?

mirca:
We can! But you'll need to pull from upstream/master and resolve the conflicts first.

PR #20, opened 2018-02-08
Overload math operators in LightCurve

mirca:
cc @christinahedges


mirca:
This is fine w/ @gully. Merging.

PR #23, opened 2018-02-16
Add a `KeplerTargetPixelFile.from_archive()` method

barentsen:
This PR adds a `from_archive()` static method to `KeplerTargetPixelFile` to enable users to access data without having to download it manually from the archive.  Example use:

```
tpf = KeplerTargetPixelFile.from_archive("Kepler-10", quarter=11, cadence="long")
```

This PR involved a lot of design decisions so a careful code review will be necessary.

Note: as part of this PR, I have marked tests that require internet access with the `@pytest.mark.remote_data` decorator.  This means that, if you have the `pytest-remotedata` package installed, the tests will be skipped by default unless you add the `--remote-data` argument to `py.test`.  (I am not sure whether Travis should skip these tests by default or not.)


lightkurve/tests/test_mast.py
```diff
@@ -0,0 +1,44 @@
...
+`--remote-data` argument is passed to py.test.  This allows tests to pass
+if no internet connection is available.
+"""
+import pytest
+
+from ..mast import search_kepler_tpf_products, ArchiveError
+from .. import KeplerTargetPixelFile
+
+
+@pytest.mark.remote_data
```
mirca: This is cool!!

lightkurve/mast.py
```diff
@@ -0,0 +1,91 @@
...
+    """Download data products from MAST.
+
+    Downloads will be cached using astroquery.mast's caching system.
+
+    Returns
+    -------
+    paths : astropy.Table.Column
+        Detailing the paths of the downloaded or cached products.
+    """
+    dl = Observations.download_products(products, mrp_only=False)
```
mirca: Could you add a comment mention why we want ``mrp_only`` to always be ``False``

mirca:
This looks all good to me!

lightkurve/mast.py
barentsen: Good thinking!!

PR #25, opened 2018-02-20
Adding TESS light curve classes

mirca:
This PR implements the following:

```
from lightkurve import TessLightCurveFile

tlc = TessLightCurveFile("https://archive.stsci.edu/missions/tess/ete-6/tid/00/000/004/104/"
                         "tess2019128220341-0000000410458113-0016-s_lc.fits")
sap = tlc.SAP_FLUX
flat, trend = sap.flatten(window_length=1001, return_trend=True)
plt.figure(figsize=[17, 4])
plt.plot(sap.time, sap.flux, 'ko', markersize=2)
plt.plot(trend.time, trend.flux)
```
in which ``tlc.SAP_FLUX`` is an object from the ``TessLightCurve`` class.

Additionally, this PR does the following secondary changes:

1. removes ``box_period_search`` in favor of ``iterative_box_period_search`` (which will also be removed in favor of ``transit_periodogram`` which will be available in ``astropy`` soon.)
2. removes ``SPLDCorrector`` since it was not well enough tested.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -611,6 +611,9 @@ class TessLightCurve(LightCurve):
 
 
 class LightCurveFile(object):
+    """Defines a generic class to handle light curve files.
+    """
+
     def __init__(self, path, **kwargs):
         self.path = path
         self.hdu = pyfits.open(self.path, **kwargs)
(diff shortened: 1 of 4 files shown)


lightkurve/lightcurve.py
```diff
@@ -565,7 +574,87 @@ def to_fits(self):
...
+    def __repr__(self):
+        return('TessLightCurveFile(TICID: {})'.format(self.ticid))
+
+    def to_fits(self):
+        raise NotImplementedError()
+
+
+class LightCurveFile(object):
+    """Defines a generic class to handle light curve files.
+    """
```
barentsen: Add a docstring for path?  And kwargs?

PR #27, opened 2018-02-21
[WIP]: Add TessTargetPixelFile class

mirca:
This PR allows one to do:
```
from lightkurve import TessTargetPixelFile
filename = ("https://archive.stsci.edu/missions/tess/ete-6/tid/00/000/"
            "004/104/tess2019128220341-0000000410458113-0016-s_tp.fits")
tpf = TessTargetPixelFile(filename)
```


barentsen:
@mirca Do you mind if we get rid of the `tpf.n_good_cadences` property?  I think it has lost its purpose now that we automatically apply the quality mask on everything, one can simply do e.g. `len(tpf.flux)` or `tpf.time.shape` and get the same answer?

(Now that we're close to the 1.0 release, now is a good time to remove features we don't need or want to support.)

mirca:
@barentsen agreed, go for it! =D

PR #29, opened 2018-02-22
Bugfix for lc.correct(), doesn't fail with NaNs and can turn off warnings

### LINKED ISSUE #28: lc.correct() won't work if there are NaNs in the lc object.
`lc.correct(method='sff')`

will break if there are NaNs in the lc object. `lc.remove_nans()` will not work, as it doesn't remove the nans from the centroids in the header.

christinahedges:
Previously `lc.correct()` would break if there were any NaNs in the light curve. It also makes a nasty warning message which ruins the progress bar because sometimes the `np.polyfit` in SFFCorrector doesn't work very well. I've added a way to turn off those warnings, but it's not very neat. 

@barentsen is there a nicer way I can accomplish this?

Fixes #28

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -542,7 +542,7 @@ class KeplerLightCurve(LightCurve):
         elif self.mission.lower() == 'k2':
             return('KeplerLightCurveFile(EPIC: {})'.format(self.keplerid))
 
-    def correct(self, method='sff', warn=False, **kwargs):
+    def correct(self, method='sff', **kwargs):
         """Corrects a lightcurve for motion-dependent systematic errors.
 
         Parameters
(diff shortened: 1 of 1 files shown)


barentsen:
Could you post the warning message?

christinahedges:
`/Users/ch/K2/repos/lightkurve/lightkurve/lightcurve.py:1002: RankWarning: Polyfit may be poorly conditioned
  self.rot_col[~self.outlier_cent], polyorder)
/Users/ch/K2/repos/lightkurve/lightkurve/lightcurve.py:1002: RankWarning: Polyfit may be poorly conditioned
  self.rot_col[~self.outlier_cent], polyorder)
/Users/ch/K2/repos/lightkurve/lightkurve/lightcurve.py:1002: RankWarning: Polyfit may be poorly conditioned
  self.rot_col[~self.outlier_cent], polyorder)`

lightkurve/lightcurve.py
```diff
@@ -541,33 +542,48 @@ def __repr__(self):
...
-                                                  **kwargs)
+            if warn:
+                self.corrector = SFFCorrector()
+                corrected_lc = self.corrector.correct(time=self.time[ok], flux=self.flux[ok],
+                                                      centroid_col=self.centroid_col[ok],
+                                                      centroid_row=self.centroid_row[ok],
+                                                      **kwargs)
+            else:
+                with warnings.catch_warnings():
+                    warnings.simplefilter("ignore")
```
barentsen: let's be more specific and change it to
```
warnings.simplefilter("ignore", action=RankWarning)  # ignore warning messages related to polyfit being poorly conditioned
```

lightkurve/lightcurve.py
```diff
@@ -541,33 +542,48 @@ def __repr__(self):
...
             Corrected lightcurve
         """
+        ok = np.isfinite(self.flux)
         if method == 'sff':
-            self.corrector = SFFCorrector()
-            corrected_lc = self.corrector.correct(time=self.time, flux=self.flux,
-                                                  centroid_col=self.centroid_col,
-                                                  centroid_row=self.centroid_row,
-                                                  **kwargs)
+            if warn:
```
barentsen: One way to avoid the duplication of code may be something like this:
```
if not verbose:
    action = "default"
else:
   action = "ignore"
```
and then
```
warnings.simplefilter(action)
```

lightkurve/lightcurve.py
```diff
@@ -541,33 +542,48 @@ def __repr__(self):
         elif self.mission.lower() == 'k2':
             return('KeplerLightCurveFile(EPIC: {})'.format(self.keplerid))
 
-    def correct(self, method='sff', **kwargs):
+    def correct(self, method='sff', warn=False, **kwargs):
```
barentsen: I'd call this argument `verbose` instead of `warn`.  But it doesn't seem appealing to put a `verbose` parameter in all functions, so I think we shouldn't have this argument and just suppress the polyfit warnings if we are confident that they are useless.

christinahedges: I think the polyfit warns will always be useless but I am worried that there may be other warnings from SFFcorrector we may want to see

lightkurve/lightcurve.py
```diff
@@ -541,33 +542,48 @@ def __repr__(self):
...
             Dictionary of keyword arguments to be passed to the function
             defined by `method`.
 
+
         Returns
         -------
         new_lc : KeplerLightCurve object
             Corrected lightcurve
         """
+        ok = np.isfinite(self.flux)
```
barentsen: I'd rename `ok` into `not_nan` just to self-document what this really is for.

Issue #32, opened 2018-02-24
The Kepler Two-Wheel Engineering Test TPFs crash due to absent "mission" keyword


gully:
```python
pathname = 'https://archive.stsci.edu/pub/k2/eng/long_cadence/kplr060021426-2014044044430_lpd-targ.fits'
tpf = KeplerTargetPixelFile(pathname)
lc = tpf.to_lightcurve(aperture_mask=tpf.pipeline_mask)
```

 results in:

```python
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-37-8299474187b3> in <module>()
----> 1 lc = tpf.to_lightcurve(aperture_mask=tpf.pipeline_mask)

~/GitHub/lightkurve/lightkurve/targetpixelfile.py in to_lightcurve(self, aperture_mask)
    435                                 campaign=self.campaign,
    436                                 quarter=self.quarter,
--> 437                                 mission=self.mission,
    438                                 cadenceno=self.cadenceno)
    439 

~/GitHub/lightkurve/lightkurve/targetpixelfile.py in mission(self)
    368     def mission(self):
    369         """Mission name"""
--> 370         return self.header(ext=0)['MISSION']
    371 
    372     def to_fits(self):

~/anaconda3/lib/python3.6/site-packages/astropy/io/fits/header.py in __getitem__(self, key)
    143         else:
    144             keyword = key
--> 145         card = self._cards[self._cardindex(key)]
    146         if card.field_specifier is not None and keyword == card.rawkeyword:
    147             # This is RVKC; if only the top-level keyword was specified return

~/anaconda3/lib/python3.6/site-packages/astropy/io/fits/header.py in _cardindex(self, key)
   1652 
   1653         if not indices:
-> 1654             raise KeyError("Keyword {!r} not found.".format(keyword))
   1655 
   1656         try:

KeyError: "Keyword 'MISSION' not found."
```

gully:
Still causes problem-- The "mission" appears in the `__repr__`, which explodes if you try to repr `lc`.

barentsen:
Thanks! I snuck a fix for the `__repr__` bug into #35 and am adding `__repr__` unit tests as part of #36, so we can close this!

PR #33, opened 2018-02-24
A quick fix for Issue #32-- simply put None for mission, if not available

gully:
A teeny-tiny pull request to allow the Kepler "two-wheeled concept engineering" data to be used in its current form.  Useful for our K2SFF head-to-head comparison.


gully:
Merging this so that I can use it right away for my tutorial!  Closes #32 .

barentsen:
This makes sense!

PR #35, opened 2018-02-25
Refactor lightcurve.py into lightcurve.py, lightcurvefile.py, and correctors.py

barentsen:
`lightcurve.py` is getting 1,500 lines long and rather tricky to edit, because it is easy to confuse `LightCurve`-based classes with `LightCurveFile`-based classes.

This PR addresses the confusion by moving the `LightCurveFile`-based classes into a new `lightcurvefile.py` module, and moving the `Corrector` classes into `correctors.py`.  The unit tests for the latter have been moved into `test_correctors.py`.

The downside is that there are circular dependencies between `lightcurve.py` and `lighcurvefile.py`.  This is ok as long as we use e.g. `from . import lightcurve as lc` instead of `from .lightcurve import KeplerLightCurve` at the top of `lightcurvefile.py`, or we would encounter an `ImportError` due to the circular dependency (which can be obscure to debug).

Thoughts?


mirca:
This all looks good to me! 👍

Issue #34, opened 2018-02-24
Support for one-fell-swoop masking and user-defined outlier rejection


gully:
Currently, the `remove_outliers()` method in `lightcurve` does not support user-defined masks.  With some discussion, Geert and I think that we could support direct masking, e.g.:

```python
lc_masked = lc[mask]
```

where mask is a boolean mask.  Geert points out that we'd have to implement dunder methods: 
https://docs.python.org/3/reference/datamodel.html#object.__getitem__

A backup approach could be to support an `input_mask` kwarg in `remove_outliers`:

```python
def remove_outliers(self, sigma=5., return_mask=False, input_mask=None, **kwargs):
        # [...]
        new_lc = copy.copy(self)
        if input_mask is not None:
            outlier_mask = input_mask
            #some checks to make sure the input_mask is the correct shape...
        else:
            outlier_mask = sigma_clip(data=new_lc.flux, sigma=sigma, **kwargs).mask
        new_lc.time = self.time[~outlier_mask]
        new_lc.flux = self.flux[~outlier_mask]
```

gully:
Related problem, a bug actually:

```python
lc.flux.shape, lc.quality.shape
```

yields
`((294,), (435,))`

since `lc.quality` is not masked in remove_outliers

barentsen:
Over in #37 slicing has been implemented using `__getitem__` and `remove_outliers()` has been edited to make use of slicing, which got rid of the big.  Please review and re-open the issue if you have any concerns!

barentsen:
Just for clarity: user-defined masks are now easy to apply, e.g. you can do
`new_lc = lc[np.abs(lc.flux - np.mean(lc.flux)) > 3*np.std(lc.flux)]`

PR #37, opened 2018-02-25
Implement `LightCurve.__getitem__()` to enable slicing and masking

barentsen:
This PR adds a `__getitem__` method to `LightCurve` and its sub-classes so that you can now slice and mask `LightCurve` objects just like an array.  For example, this test now passes:

```
lc = KeplerLightCurve(time=[1, 2, 3], flux=[1, 1, 10], quality=[0, 0, 200])
assert_array_equal(lc[lc.flux < 5].time, [1, 2])
assert_array_equal(lc[lc.flux < 5].quality, [0, 0])
```

An extra bonus is that this allowed me to considerably simplify the implementation of `remove_nans()` and `remove_outliers()` as part of this PR.  It also resolves the bug of `remove_outliers` not masking `quality` as was reported in #34.

To simplify the slicing, I changed `quality`, `centroid_col`, and `centroid_row` to behave just like `flux_err`: if any of these arrays are None, the constructor will convert them to an array of NaNs with the same length as `time`.  To do this, I replaced the `_validate_inputs` method with a more generic `_validate_array` method which is called by the constructor for each data array.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -38,21 +38,30 @@ class LightCurve(object):
     """
 
     def __init__(self, time, flux, flux_err=None, meta={}):
-        self.time, self.flux, self.flux_err = self._validate_inputs(time, flux, flux_err)
+        self.time = np.asarray(time)
+        self.flux = self._validate_array(flux, name='flux')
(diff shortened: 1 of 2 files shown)


mirca:
👍

barentsen:
Yay. :fireworks:

PR #38, opened 2018-02-25
Rename `stitch()` to `append()`

barentsen:
This PR proposes to rename `stitch()` to `append()`, consistent with e.g. Python lists and numpy arrays.

We might consider adding a `sitch()` function that takes care of e.g. removing the jumps between different quarters of campaigns, but right now all the function was doing was appending data, so I feel like it should be called "append".

Thoughts?


mirca:
LGTM!

PR #44, opened 2018-02-26
Fixed normalize bug

christinahedges:
These are the wrong way round. This won't normalize the errors, and it breaks the plotting routines.
```
lc.flux = lc.flux / np.nanmedian(lc.flux)	
lc.flux_err = lc.flux_err / np.nanmedian(lc.flux)
```
It's odd we don't have a unit test for this?


barentsen:
Good catch!!

Issue #46, opened 2018-02-27
KeplerCBVCorrector should taken in a LightCurve object as well as a LightCurveFile


christinahedges:
Currently KeplerCBVCorrector only accepts a LightCurveFile as an input. It should be able to work from a LightCurve generated from a tpf file e.g. `tpf.to_lightcurve()` still has a channel and a quarter.

christinahedges:
Additionally, we should consider editing the LightCurve object, rather than returning a new LightCurve object with no meta data.

We could use it as part of the `correct` method.

christinahedges:
This issue is duplicates in #56

PR #49, opened 2018-02-27
Added a plot_cbvs function and added errors to correct function

christinahedges:
I had to change the CBV corrector slightly for a tutorial. I changed the following:

* Moved how we obtain the CBVs to its own function `_get_cbv_data`
* Added the errors back into the `LightCurve` that the corrector returns (Although we think it shouldn't do this in future see #46 )
* Added a `plot_cbvs` function. This is useful for the tutorial.

### DIFF
--- a/lightkurve/correctors.py
+++ b/lightkurve/correctors.py
@@ -100,6 +100,19 @@ class KeplerCBVCorrector(object):
         """
         return self._opt_result
 
+    def _get_cbv_data(self, cbvs=[1, 2]):
+        '''Gets the CBV data for a channel and module
+        '''
+        module, output = channel_to_module_output(self.lc_file.channel)
+        cbv_file = pyfits.open(self.get_cbv_url())
+        cbv_data = cbv_file['MODOUT_{0}_{1}'.format(module, output)].data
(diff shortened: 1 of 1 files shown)


lightkurve/correctors.py
```diff
@@ -209,6 +216,37 @@ def get_cbv_url(self):
...
+        ax : matplotlib.pyplot.Axes.AxesSubplot
+            Matplotlib axis object
+        '''
+        if ax is None:
+            _, ax = plt.subplots(1)
+        cbv_array, time = self._get_cbv_data(cbvs)
+        for idx, cbv in enumerate(cbv_array):
+            ax.plot(time, cbv+idx/10., label='{}'.format(idx + 1))
+        ax.set_yticks([])
+        ax.set_xlabel('Time (MJD)')
```
barentsen: Are you sure this is MJD, not BKJD?

christinahedges: It says TIME_MJD in the CBV file and it has values  ~55000 so I thought it was MJD

christinahedges:
screen shot 2018-02-27 at 10 29 59 am

barentsen:
Oh, ok, let's merge!

PR #52, opened 2018-02-27
lightkurve Docs Tutorials

christinahedges:
Tutorials for the lightkurve docs. There are two sections, basics of how to use the package and common astronomy tasks.


barentsen:
For remote folks: we merged the 6 tutorials (!!) that were ready.  Christina will open a new one for the 3 that were under construction.  All can do with further edits, but they're excellent for getting started!

PR #60, opened 2018-03-01
Made plots not break if errors are all Nans

christinahedges:
Now plots won't break if all the errors are NaN's or zeros. This was a common gotcha for why `lc.plot()` produced a blank frame.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -408,8 +408,8 @@ class LightCurve(object):
             flux, flux_err = normalized_lc.flux, normalized_lc.flux_err
         else:
             flux, flux_err = self.flux, self.flux_err
-        if flux_err is None:
-            ax.plot(self.time, flux, marker='o', color=color,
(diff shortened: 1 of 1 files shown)


lightkurve/lightcurve.py
```diff
@@ -408,8 +408,8 @@ def plot(self, ax=None, normalize=True, xlabel='Time - 2454833 (days)',
             flux, flux_err = normalized_lc.flux, normalized_lc.flux_err
         else:
             flux, flux_err = self.flux, self.flux_err
-        if flux_err is None:
-            ax.plot(self.time, flux, marker='o', color=color,
+        if (flux_err is None) or (np.nansum(flux_err)==0):
```
barentsen: We can get rid of the `flux_err is None` test because the constructor now guarantees that `flux_err` is a NaN-array of length `len(time)` when `flux_err=None` is passed.

How about
```if np.any(np.isfinite(flux_err)):``

barentsen:
LGTM!  I'm going to merge in 60 secs.

PR #61, opened 2018-03-01
Added a describe function

christinahedges:
Since we're making new TPFs out of data from all kinds of sources I've found it useful to have a print function that gives me a clear/pretty overview of what's in the target pixel file. I've added a `describe` function to LightCurve and TargetPixelFile which describes the attributes in the object in a logical order.

screen shot 2018-03-01 at 11 46 05 am

screen shot 2018-03-01 at 11 45 34 am

Comments/suggestions/improvements welcome!

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -111,7 +111,7 @@ class LightCurve(object):
         attrs = {}
         for attr in dir(self):
             if not attr.startswith('_'):
-                res = getattr(self,attr)
+                res = getattr(self, attr)
                 if callable(res):
                     continue
                 if attr == 'hdu':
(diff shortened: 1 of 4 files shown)


christinahedges:
Now uses astropy.tables instead of pandas to print prettily.
screen shot 2018-03-01 at 1 08 56 pm
screen shot 2018-03-01 at 1 09 01 pm

lightkurve/lightcurve.py
```diff
@@ -102,6 +102,60 @@ def __div__(self, other):
...
+    def properties(self):
+        '''Print out a description of each of the non-callable attributes of a
+        LightCurve object.
+
+        Prints in order of type (ints, strings, lists, arrays and others)
+        Prints in alphabetical order.'''
+        attrs = {}
+        for attr in dir(self):
+            if not attr.startswith('_'):
+                res = getattr(self,attr)
```
barentsen: space after comma =)

PR #62, opened 2018-03-02
Flatten has better behavior around discontinuities.

christinahedges:
When the spacecraft is down there is a break in the 'time' axis of a light curve and usually a sharp increase or decrease in flux as the spacecraft settles. This causes a problem for the savgol filter in flatten.

I've fixed this by having flatten 'split' the light curve into different groups anywhere where 5 or more time points are missing.

I've also made it so that flatten returns a light curve that's the same shape as the original. Before it would remove nans and then give that shorter light curve back.

screen shot 2018-03-01 at 4 06 35 pm

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -158,15 +158,24 @@ class LightCurve(object):
             Trend in the lightcurve data
         """
         lc_clean = self.remove_nans()  # The SG filter does not allow NaNs
-        trend_signal = signal.savgol_filter(x=lc_clean.flux,
-                                            window_length=window_length,
-                                            polyorder=polyorder, **kwargs)
-        flatten_lc = copy.copy(lc_clean)
-        flatten_lc.flux = lc_clean.flux / trend_signal
-        flatten_lc.flux_err = lc_clean.flux_err / trend_signal
+        #Find breaks in time
+        dt = lc_clean.time[1:]-lc_clean.time[0:-1]
+        cut = np.where(dt > 5*np.nanmedian(dt))[0]+1
+        low = np.append([0], cut)
+        high = np.append(cut, len(lc_clean.time))
+        trend_signal = np.zeros(len(lc_clean.time))
+        for l, h in zip(low, high):
+            trend_signal[l:h] = signal.savgol_filter(x=lc_clean.flux[l:h],
+                                                     window_length=window_length,
+                                                     polyorder=polyorder, **kwargs)
+
+        trend_signal = np.interp(self.time, lc_clean.time, trend_signal)
+        flatten_lc = copy.deepcopy(self)
+        flatten_lc.flux /= trend_signal
+        flatten_lc.flux_err /= trend_signal
 
         if return_trend:
-            trend_lc = copy.copy(lc_clean)
+            trend_lc = copy.deepcopy(self)
             trend_lc.flux = trend_signal
             return flatten_lc, trend_lc
         return flatten_lc
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -131,7 +131,8 @@ class LightCurve(object):
                 new_lc.centroid_row = np.append(new_lc.centroid_row, others[i].centroid_row)
         return new_lc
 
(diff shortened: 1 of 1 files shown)


lightkurve/lightcurve.py
```diff
@@ -131,7 +131,7 @@ def append(self, others):
                 new_lc.centroid_row = np.append(new_lc.centroid_row, others[i].centroid_row)
         return new_lc
 
-    def flatten(self, window_length=101, polyorder=3, return_trend=False, **kwargs):
+    def flatten(self, window_length=101, polyorder=2, return_trend=False, **kwargs):
```
barentsen: Note: this may change the behavior of existing tutorials.  But I guess it is more robust?

lightkurve/lightcurve.py
```diff
@@ -158,15 +158,24 @@ def flatten(self, window_length=101, polyorder=3, return_trend=False, **kwargs):
...
         lc_clean = self.remove_nans()  # The SG filter does not allow NaNs
-        trend_signal = signal.savgol_filter(x=lc_clean.flux,
-                                            window_length=window_length,
-                                            polyorder=polyorder, **kwargs)
-        flatten_lc = copy.copy(lc_clean)
-        flatten_lc.flux = lc_clean.flux / trend_signal
-        flatten_lc.flux_err = lc_clean.flux_err / trend_signal
+        #Find breaks in time
+        dt = lc_clean.time[1:]-lc_clean.time[0:-1]
+        cut = np.where(dt > 5*np.nanmedian(dt))[0]+1
```
barentsen: add a space around all math operators above (-, +, *)

lightkurve/lightcurve.py
```diff
@@ -158,15 +158,24 @@ def flatten(self, window_length=101, polyorder=3, return_trend=False, **kwargs):
...
+        low = np.append([0], cut)
+        high = np.append(cut, len(lc_clean.time))
+        trend_signal = np.zeros(len(lc_clean.time))
+        for l, h in zip(low, high):
+            trend_signal[l:h] = signal.savgol_filter(x=lc_clean.flux[l:h],
+                                                     window_length=window_length,
+                                                     polyorder=polyorder, **kwargs)
+
+        trend_signal = np.interp(self.time, lc_clean.time, trend_signal)
+        flatten_lc = copy.deepcopy(self)
```
barentsen: Should we use `deepcopy` in all functions?

lightkurve/lightcurve.py
christinahedges: You're right, I can put it back, but I think it might be a little high?

barentsen:
Wow, nice improvement in the example you posted!!  The implementation may need tuning if some day people start using the package for irregularly-sampled data, but let's not worry about such uses now.

mirca:
This is cooool!! Thanks @christinahedges

lightkurve/lightcurve.py
christinahedges: Probably, we only didn't see it in the previous case because it was copying something that was only a local variable.

christinahedges:
> irregularly-sampled data

Good point, but in that case they're probably going to wan to do their own improvements.

lightkurve/lightcurve.py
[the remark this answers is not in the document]
christinahedges: I was thinking of just putting it in the documentation. But you're right, we can expose it and have users tune it. By setting it to a nan I bet you can turn off this functionality........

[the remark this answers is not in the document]
mirca: Oh good, let's mention that ``By setting it to a nan I bet you can turn off this functionality`` in the docstrings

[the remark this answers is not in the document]
christinahedges: Ok there's now a break_tolerance keyword which if passed None (or np.nan) will turn off the new behavior. @barentsen this will address your concern about non-uniform time, user just turns it off

barentsen:
This is fantastic work.  It looks ready to be merged.

One small change in code, one giant leap for lightcurves.

mirca:
This is so sick!! Thanks a lot @christinahedges

PR #63, opened 2018-03-02
Fixed plot marker bug and cycling issue, added style support.

### LINKED ISSUE #9: LightCurve.plot should have prettier defaults
That's what I get with the current arguments:
screen shot 2018-02-01 at 4 34 49 pm

christinahedges:
Fixes #9 
There were a few problems with plotting which was preventing me from producing really nice tutorials/notebooks:

**marker bug** has where plot wouldn't accept a different marker style has been fixed.

**color cycling** has been fixed. Before if an additional line was added to a plot, they would always be plotted in black. Now they will be plotted in the color cycling order.
screen shot 2018-03-01 at 5 17 51 pm

**style sheets** now you can use the plt.style.context choices to make prettier plots by passing `context='ggplot'` to any of the plot functions, or your favorite style.
screen shot 2018-03-01 at 5 18 02 pm

**[Hopefully] the default style has not been changed by this PR**

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -8,6 +8,7 @@ from astropy.nddata import Cutout2D
 from astropy.table import Table
 from astropy.wcs import WCS
 from matplotlib import patches
+import matplotlib.pyplot as plt
 import numpy as np
 from tqdm import tqdm
 
@@ -541,7 +542,7 @@ class KeplerTargetPixelFile(TargetPixelFile):
         return col_centr, row_centr
 
     def plot(self, ax=None, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
-             show_colorbar=True, mask_color='pink', **kwargs):
+             show_colorbar=True, mask_color='pink', context='fast', **kwargs):
         """
         Plot a target pixel file at a given frame (index) or cadence number.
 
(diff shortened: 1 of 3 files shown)


christinahedges:
**before**
screen shot 2018-03-01 at 5 23 18 pm
**after**
screen shot 2018-03-01 at 5 22 59 pm

Ok there is a slight difference it now plots with linestyle='solid' and no marker as default. Do people feel they strongly want the scatter plot default back? I can force it to be that way by default if that's what we think is the most useful plot.

christinahedges:
This is how you can now have prettier plots:
screen shot 2018-03-01 at 5 25 58 pm

lightkurve/targetpixelfile.py
```diff
@@ -563,6 +564,8 @@ def plot(self, ax=None, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
             Whether or not to show the colorbar
         mask_color : str
             Color to show the aperture mask
+        context : str
+            matplotlib.pyplot.style.context, default is ggplot
```
mirca: default looks like to be `'fast'`?

christinahedges: Although the previous default was to do a scatter plot. I think I prefer it as a line plot...

mirca: what I mean is that the docstring should say "`default is 'fast'`"

barentsen:
LGTM! :+1:

Happy for this to be merged after fixing the docstring bug Ze identified.

(I do wonder if we want line or scatter plots as the default, but at this time I'm happy for the author of the PR to get the choice!)

christinahedges:
I'm going with plotting should be a line plot by default. Scatter plots mess up the ylimits a lot of the time.

PR #64, opened 2018-03-02
Add a `FoldedLightCurve` class to fix the default plot labels of a folded light curve

barentsen:
This PR introduces a `FoldedLightCurve` class for the purpose of fixing the default plot labels (cf. #6).

It's not clear to me that there are other differences which justify the difference between `LightCurve` and `FoldedLightCurve`, so perhaps the use of inheritance is overkill here, but it's a nifty solution to #6 nonetheless.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -268,9 +268,9 @@ class LightCurve(object):
         # fold time domain from -.5 to .5
         fold_time[fold_time > 0.5] -= 1
         sorted_args = np.argsort(fold_time)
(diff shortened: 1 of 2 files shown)


christinahedges:
Can FoldedLightCurve have a property 'phase' which just returns time?

barentsen:
Good idea!!

christinahedges:
Looks excellent, the user won't even know! 😱 😄 Merge it!

barentsen:
Added a `phase` property which returns `time`.  Also updated the unit test.  Let's merge as soon as the tests complete!

PR #65, opened 2018-03-03
Add `restore_trend` as an option to SFFCorrector

barentsen:
This PR provides the ability to put the long-term trend back into an SFF-corrected lightcurve, which is essential for supernovae.  Someone please review!

### DIFF
--- a/lightkurve/correctors.py
+++ b/lightkurve/correctors.py
@@ -272,7 +272,7 @@ class SFFCorrector(object):
 
     def correct(self, time, flux, centroid_col, centroid_row,
                 polyorder=5, niters=3, bins=15, windows=1, sigma_1=3.,
-                sigma_2=5.):
+                sigma_2=5., restore_trend=True):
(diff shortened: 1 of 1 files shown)


lightkurve/correctors.py
```diff
@@ -357,6 +362,8 @@ def correct(self, time, flux, centroid_col, centroid_row,
                 flux[i] = corrected_flux
 
             flux_hat = np.append(flux_hat, flux[i])
+            if restore_trend:
+                flux_hat += self.trend
```
mirca: Not sure if I follow the maths here. `iter_trend` is being divided out in  ``self.normflux = flux[i] / iter_trend``,
so maybe instead of ``flux_hat += self.trend`` it should be ``flux_hat *= self.trend``?.

barentsen: I think you're absolutely right!

barentsen:
I'm going to merge this, but with `restore_trend=False` by default so the behavior doesn't change for now.

PR #67, opened 2018-03-04
Make the 'fast' matplotlib style optional

barentsen:
`LightCurve.plot()` & `KeplerTargetPixelFile.plot()` recently got `context='fast'` as a default matplotlib style parameters, but this style is only available in matplotlib 2.1 and later (released just a few months ago).  This PR makes sure plotting doesn't crash on earlier versions of matplotlib, because 2.1 is not our minimum requirement right now.


barentsen:
Updated the PR to change the parameter name from `context` into `style`, which really is the matplotlib term for these stylesheets.

PR #77, opened 2018-03-08
Halo TV-min 

benjaminpope:
I have implemented a method for the targetpixelfile class which accesses halophot to run a Total Variation Minimization photometry algorithm, as 

lc, weightmap = tpf.halo()

I think you might need to have it automatically install github.com/hvidy/halophot/, and I am a poor fool and I don't know how to make it do that... 

All the best,

Ben


christinahedges:
I am sooooo excited to take a look at this! Thanks ben! :D :D :D :D :D

barentsen:
This PR proposes to add a new file called `build/lib/lightkurve/targetpixelfile.py`.  That doesn't seem right.  The changes should go in `lightkurve/targetpixelfile.py`.

barentsen:
I am going to close this PR because the changes proposed are in the `build` directory, which is not under version control.  It will be easier to propose this from a new branch.

PR #97, opened 2018-04-05
`from_archive` now supports radius searches, multiple campaigns and returning data from any cadence type

### LINKED ISSUE #101: Short Cadence File ID from Table -1 error
In lightcurvefile, function from_archive, on line 134
the line should read 
products = Table(products[month-1])
not month+1

products is a table that is zero-based, but the user specifies the month in a one-based counting scheme.  Therefore you should subtract the one, not add.

### LINKED ISSUE #102: Error in logic for Short Cadence Data Retrieval
Short cadence retrieval is more complicated than lightkurve has currently implemented.
There can be times that there is only one month of short cadence in a quarter and your code is written such that it assumes you either have all three months or no short cadence.  
For example.
sc=KeplerLightCurveFile.from_archive(5034333, cadence='short',quarter=3, month=1)
sc=KeplerLightCurveFile.from_archive(5034333, cadence='short',quarter=3, month=2)
sc=KeplerLightCurveFile.from_archive(5034333, cadence='short',quarter=3, month=3)
all returns the same file (in fact month=99 also works).  This could be very confusing to users who were hoping to use the lack of a file to know that there was no data there. 5034333 was only observed one month in quarter 3.  This star is not the only case. Early on in Kepler people just wanted a little bit of short cadence data to see if it was worth getting more, or one month was all they needed for their science.

I fear that you will need a look-up table to solve this problem since the month is encoded in that time string in the file name.

christinahedges:
Here are the major changes.`from archive` now
- can return multiple hits, but by default only returns one
- allows radius searching
- allows multiple campaigns, quarters and months to be searched with lists
- allows multiple cadence types to be searched with cadence='any' (or 'all')
- allows a SkyCoord to be passed as the target
- has a verbose argument to silence the output

The behavior is the same for TPFs and LCFs. Here's a demo:

Note this fixes #101 and fixes #102

If you ask for a file with multiple quarters it will warn the user
screen shot 2018-04-05 at 11 56 58 am

Otherwise it gives back the file
screen shot 2018-04-05 at 12 00 47 pm

If you ask for multiple quarters it will give back a list of tpfs
screen shot 2018-04-05 at 12 00 37 pm

If you do a radius search it will give you the nearest files. You can increase the limit for more files.
screen shot 2018-04-05 at 12 02 26 pm

You can do both at the same time, a radius search with multiple quarters. The order that they are returned by is by target, then by quarter, then by month.

screen shot 2018-04-05 at 12 03 35 pm

You can also suppress all that nasty output with the verbose keyword.
screen shot 2018-04-05 at 12 04 41 pm


christinahedges:
I can't reproduce the problem that appveyor is having on my machine... @barentsen do you have any idea why those things are breaking?

lightkurve/data/short_cadence_month_lookup.csv
```diff
@@ -0,0 +1,50 @@
+Quarter,Month,Date
```
barentsen: What is the exact meaning of `Date`?  I suggest calling it e.g. `StartTime` if that is what it is?

lightkurve/lightcurvefile.py
```diff
@@ -110,37 +110,31 @@ def from_archive(target, cadence='long', quarter=None, month=None,
             KIC/EPIC ID or object name.
         cadence : str
             'long' or 'short'.
-        quarter, campaign : int
+        quarter, campaign : int, list or 'all'
```
barentsen: Let's say `list of int` instead of  `list` to be more precise.

lightkurve/lightcurvefile.py
```diff
@@ -110,37 +110,31 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
             Kepler Quarter or K2 Campaign number.
-        month : 1, 2, or 3
+        month : 1, 2, 3, list or 'all'
```
barentsen: => `list of int`

lightkurve/lightcurvefile.py
```diff
@@ -110,37 +110,31 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+            Target Pixel Files for each quarter, each covering one month.
             Hence, if cadence='short' you need to specify month=1, 2, or 3.
+        radius : float
+            Search radius in arcseconds. Default is 1 arcsecond.
+        targetlimit : None or int
+            Limit the number of returned target pixel files. If none, no limit
+            is set
         kwargs : dict
-            Keywords arguments passed to `KeplerLightCurveFile`.
+            Keywords arguments passed to `KeplerTargetPixelFile`.
```
barentsen: I think it should be `KeplerLightCurveFile` instead of `KeplerTargetPixelFile`.

lightkurve/lightcurvefile.py
```diff
@@ -110,37 +110,31 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
         cadence : str
             'long' or 'short'.
-        quarter, campaign : int
+        quarter, campaign : int, list or 'all'
             Kepler Quarter or K2 Campaign number.
-        month : 1, 2, or 3
+        month : 1, 2, 3, list or 'all'
             For Kepler's prime mission, there are three short-cadence
-            lightcurve files for each quarter, each covering one month.
+            Target Pixel Files for each quarter, each covering one month.
```
barentsen: => `Lightcurve files` instead of `Target Pixel Files`?

lightkurve/lightcurvefile.py
```diff
@@ -97,10 +97,10 @@ def __init__(self, path, quality_bitmask='default', **kwargs):
 
     @staticmethod
     def from_archive(target, cadence='long', quarter=None, month=None,
-                     campaign=None, **kwargs):
-        """Fetch a Light Curve File from the Kepler/K2 data archive at MAST.
+                     campaign=None, radius=1., targetlimit=1, verbose=True, **kwargs):
+        """Fetch a Lightcurve File from the Kepler/K2 data archive at MAST.
```
barentsen: Let's say `Fetch *one or more* Lightcurve files from ...` to make it clear that it can return multiple.

lightkurve/lightcurvefile.py
```diff
@@ -97,10 +97,10 @@ def __init__(self, path, quality_bitmask='default', **kwargs):
...
 
-        Raises an `ArchiveError` if a unique file cannot be found.  For example,
+        Raises an `ArchiveError` if a unique LCF cannot be found.  For example,
```
barentsen: We'll need a bit more text here to explain when multiple files are and aren't returned.

lightkurve/lightcurvefile.py
```diff
@@ -201,6 +195,10 @@ def channel(self):
     def keplerid(self):
         return self.header(ext=0)['KEPLERID']
 
+    @property
+    def cadence(self):
```
barentsen: Let's call this `obsmode` instead of `cadence` for consistency with header keyword conventions.

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
             target_name = 'ktwo{:09d}'.format(target)
         else:
             raise ValueError("Not in the Kepler KIC or EPIC ID range")
-        obs = Observations.query_criteria(target_name=target_name,
+        obs = Observations.query_criteria(objectname=target_name,
```
barentsen: Is there any point left in having a separate query for targets that look like EPIC/KIC IDs and those that don't?  I wonder if we can get rid of all this and just run the call to `query_criteria` that gets executed beneath the `except ValueError:` block right now?  What do you think?

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
+        obs.sort('distance')
     except ValueError:
         # If `target` did not look like a KIC or EPIC ID, then we let MAST
         # resolve the target name to a sky position.
         try:
             obs = Observations.query_criteria(objectname=target,
-                                              radius='1 arcsec',
+                                              radius='{} arcsec'.format(radius),
                                               project=["Kepler", "K2"])
+            #Remove KeplerFFI data set.
```
barentsen: Always add a space behind a `#`

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
+            obs = obs[(obs['obs_collection']=='Kepler') | (obs['obs_collection']=='K2')]
```
barentsen: Always add spaces around `==`

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
         except ResolverError as exc:
             raise ArchiveError(exc)
-    return Observations.get_product_list(obs)
+
+    #Make sure the final table is in DISTANCE order.
+    obsids = np.asarray(obs['obsid'])
+    t = (Observations.get_product_list(obs))
+    l = [np.where(t['parent_obsid'] == o)[0]  for o in obsids]
+    l = [item for sublist in l for item in sublist]
+    t = t[l]
```
barentsen: Please add a comment explaining the purpose of the above block (I don't understand it).  Also replace `t` and `l` by more verbose variable names to help clarify.

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
+
+    return t
 
 
-def search_kepler_tpf_products(target, cadence='long', quarter=None,
-                               campaign=None):
-    """Returns a table of Kepler or K2 Target Pixel Files for a given target.
+def search_kepler_products(target, filetype='Target Pixel', cadence='long', quarter=None,
```
barentsen: The difference between `search_products`, `search_kepler_products`, and `get_kepler_products` is not clear.  Can we give both functions a better name to explain how they differ?

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
+                               campaign=None, radius=1):
+    """Returns a table of Kepler or K2 Target Pixel Files or Lightcurve Files
+     for a given target.
 
     Parameters
     ----------
     cadence: 'short' or 'long'
         Specify short (1-min) or long (30-min) cadence data.
+    filetype : 'Target Pixel' or 'Lightcurve'
```
barentsen: Let's use `TargetPixelFile` and `LightCurveFile` as the permissible options to be consistent with the terminology elsewhere in lightkurve.  (Unless this makes the code significantly more complicated.)

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
+        Search radius in arcseconds
 
     Returns
     -------
     products : astropy.Table
         Table detailing the available Target Pixel File products.
     """
-    products = search_kepler_products(target)
+    if filetype not in ['Target Pixel', 'Lightcurve']:
+        logging.exception("Choose a filetype of 'Target Pixel' or 'Lightcurve'")
```
barentsen: This should raise a `ValueError` exception.

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
+                if q is not None:
+                    mask |= np.array([desc.lower().endswith('q{}'.format(q)) or
+                                      desc.lower().endswith('c{:02d}'.format(q))
+                                      for desc in products['description']])
+        else:
+            mask = np.ones(np.shape(products)[0], dtype=bool)
+        mask &= np.array([desc.lower().endswith('fits') or
+                          desc.lower().endswith('fits.gz')
+                          for desc in products['dataURI']])
+        mask &= np.array([suffix in desc for desc in products['description']])
```
barentsen: The above statements need a few comments to clarify their purpose.

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
+                           "for target {}. Please specify quarter/month "
+                           "or campaign number."
+                           "".format(len(products), target))
+
+    #Otherwise download all the files
+    if verbose:
+        print('Found {} File(s)'.format(np.shape(products)[0]))
+    if not verbose:
+        old_stdout = sys.stdout
+        sys.stdout = open(os.devnull, "w")
```
barentsen: Is there a reason to re-assign `sys.stdout` *and* do `astropylog.setLevel`?

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
+        Limit the number of returned target pixel files. If none, no limit
+        is set
+    kwargs : dict
+        Keywords arguments passed to `KeplerTargetPixelFile`.
 
     Returns
     -------
-    products : astropy.Table
-        Table detailing the available Target Pixel File products.
+    path : str or list of strs
```
barentsen: The return value is `return np.asarray(path)` so I presume it always returns an `array` now, not a `str` or a `list`?

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
+    products['dates'] = np.asarray(dates)
+    return products
 
-def search_kepler_lightcurve_products(target, cadence='long', quarter=None,
-                                      campaign=None):
-    """Returns a table of Kepler or K2 lightcurve files for a given target.
+def get_kepler_products(target, filetype='Target Pixel', cadence='long',
+                        quarter=None, month=None, campaign=None, radius=1.,
+                        targetlimit=1, verbose=True, **kwargs):
+    """Fetch files from from the Kepler/K2 data archive at MAST.
```
barentsen: This function now seems to return an array of paths, so we have to change the description (i.e. nothing is being fetched?)

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
 
-    This only returns products produced by the official Kepler pipeline,
-    which is not necessarily the best choice for every use case.
+    Raises an `ArchiveError` if a unique TPF cannot be found.  For example,
+    this is the case if a target was observed in multiple Quarters and the
+    quarter parameter is unspecified.
```
barentsen: We'll have to update this description to explain that multiple files can be returned now.

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
+        astropylog.setLevel('INFO')
+    else:
+        astropylog.setLevel('ERROR')
+
+    #If we are asking for all cadences (long and short) and didn't specify a month, override it.
+    if (cadence in ['any', 'both']) & (month is None):
+        month = np.arange(3) + 1
+
+    #anys and all should be a list of numbers
+    if month in ['all','any']:
```
barentsen: Always put a space behind a `,` in a list.

christinahedges:
I've found a weird MAST portal error with the 'split' campaigns.

If you look for a TPF using the MAST api where the target was observed in a split campaign (e.g. C9-1 and C9-2) there appears to be an error with the first part. The URLs get confused and so they can't download the first one. 

This is the best MWE I can come up with 


screen shot 2018-04-06 at 3 55 34 pm

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
+        #all use the correct month
+        finalmask = np.ones(len(products), dtype=bool)
+        for c in np.unique(products[scmask]['qoc']):
+            ok = (products['qoc'] == c) & (scmask)
+            mask = np.zeros(np.shape(products[ok])[0], dtype=bool)
+            for m in month:
+                udate = (table['Date'][np.where((table['Month'] == m) & (table['Quarter'] == c))[0][0]])
+                mask |= np.asarray(products['dates'][ok]) == udate
+            finalmask[ok] = mask
+        products = products[finalmask]
```
barentsen: I wonder if we can simplify this somehow.  This is going to be tricky to understand and maintain in the future.

lightkurve/mast.py
```diff
@@ -61,86 +69,242 @@ def search_kepler_products(target):
...
+    if np.any(scmask):
+        #Error check the user if there's multiple months and they didn't ask
+        #for a specific one
+        if month is None:
+            raise ArchiveError("Found {} different Target Pixel Files "
+                               "for target {} in Quarter {}. "
+                               "Please specify the month (1, 2, or 3)."
+                               "".format(len(products), target, quarter))
+        #Get the short cadence date lookup table.
+        table = ascii.read('/Users/ch/K2/repos/lightkurve/lightkurve/data/short_cadence_month_lookup.csv')
```
barentsen: Change this to `os.path.join(PACKAGEDIR, 'data', 'short_cadence_month_lookup.csv')`

christinahedges:
There's an issue with the current implementation of from_archive where it doesn't accept split campaigns........problematic....

lightkurve/data/short_cadence_month_lookup.csv
christinahedges: I assume that date string is a starttime, we can change it to that.

lightkurve/mast.py
christinahedges: I think this will probably work. Let's try it.

lightkurve/mast.py
christinahedges: Sure, what names do you want?

* search_products: search mast for any products that match
* search_kepler_products: search mast for only Kepler/K2 specific products
* get_kepler_products: download Kepler/K2 specific products

lightkurve/mast.py
christinahedges: I guess not

lightkurve/mast.py
christinahedges: Yeah it's always an array. A list is probably better.

lightkurve/mast.py
christinahedges: I'm still unsure what to rename these, let's come up with something together

lightkurve/mast.py
christinahedges: The sys.stdout doesn't remove astropy.log messages

barentsen:
This is the way forward!  Thanks @christinahedges!

PR #107, opened 2018-04-12
Fixed split campaign bug

### LINKED ISSUE #105: from_archive can not handle split campaigns
In the master branch version of from_archive split campaigns come up with errors:

screen shot 2018-04-06 at 5 27 49 pm

christinahedges:
from_archive now works for split campaigns.

Fixes #105

### DIFF
--- a/lightkurve/mast.py
+++ b/lightkurve/mast.py
@@ -103,8 +103,9 @@ def search_kepler_tpf_products(target, cadence='long', quarter=None,
     # Identify the campaign or quarter by the description.
     quarter_or_campaign = campaign if campaign is not None else quarter
     if quarter_or_campaign is not None:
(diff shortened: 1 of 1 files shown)


barentsen:
Ahhh, good catch.  Can you add a little comment that says
`# We remove dashes (used for e.g. "C9-1" and "C9-2") and convert to lowercase to ensure all Campaigns match`

It looks good to merge otherwise!

barentsen:
Actually, before we merge, let's add a unit test for 101/102 and 91/91 because it is not unlikely that we may accidentally re-introduce this bug in the future.

barentsen:
This looks good!  We can ignore the unrelated appveyor issue.  Thanks @christinahedges!

Issue #95, opened 2018-03-27
Allow lightkurve's psf fitting priors to be input in image coordinates and/or RA, Dec


amcody:
Currently one must input detector column and row, which is proving problematic for stars in crowded superstamp regions (e.g., M67) that go across channel boundaries where the detector coordinates are discontinuous. It would be a lot easier to use the local coordinate system of the TPF to specify star positions (though one would still use the detector coordinates to retrieve the PRF model).

gully:
Currently working on this now with AMC!

mirca:
This is really cool! Which branch are you both working on?

gully:
We are discussing the possibility of adding an option to apply `pos_corr` offsets directly inside the `phot.fit()`.  Currently the sole argument of `phot.fit()` is `tpf_flux`.  AMC and I suggest the following change:
Instead of taking `tpf.flux` as an argument, make it take in `tpf`.  This way, the fitting code will have access to `tpf.poss_corr1` (and 2).  

poss_corr would be used to *shift* the priors left-right and up-down, to maintain the desired prior window in the face of spacecraft-induced motion.  This feature can either be the default, or an opt-in keyword-arg.

barentsen:
Closing this issue because we do not currently have resources to develop PRF photometry features, unfortunately!

PR #111, opened 2018-04-20
[WIP] Make PRF fitting prior more flexible to account for spacecraft movement

gully:
Work in progress.
Relates to Issue #95


gully:
We are seeing that sometimes the results of PRF fitting can be outside the allowed prior window.  That result cannot be, so something is wrong.  

- [ ] Identify cause of result-out-of-prior-window bug
- [ ] Identify and eliminate, or simply suppress the warnings that occur during PRF fitting.  Perhaps provide a log of fails at the end of the fitting procedure.
- [ ] The `cadences=` subset kwarg does not seem to work due to a reshape error at the end of fitting.

christinahedges:
@amcody and @barentsen would like to refactor this slightly to make PRFphotometry part of SceneModel. Then the call would be

`SceneModel(...).fit()`

christinahedges:
Having talked with @amcody we're going to remove the new functions `ra_dec_to_uniformprior`, `ra_dec_to_gaussprior` and `stars_to_prior` from this PR and move them to a new PR which will fix #137. 

This PR will now only address the time varying motion priors from poscorr, and we'll have a new one for ra/dec priors.

christinahedges:
* Some campaigns that have not been reprocessed and so have no POSCORR (campaign 9) and so we will have to default to something else
* There is currently no way to give a static prior. This might not be a particularly useful case?

christinahedges:
Currently I am still having the same issue as before, where the fit is now extending outside of the bounds.

barentsen:
I'm closing this PR because it is stale (>1 yr old) and because we haven't had time to actively improve the PRF photometry sub-package.  Of course the PR will still be here for future reference re PRF photometry improvements, which would be awesome!

PR #113, opened 2018-04-24
Rolling Band Tutorial

christinahedges:
I added this rolling band tutorial. It might need tweaking to make sure that the embedded youtube video loads properly when the ipynb file is turned into rst.


christinahedges:
@barentsen can you please take a look at the video embedding? Thanks

christinahedges:
I remember now, this won't work until we make from_archive have a search radius.

barentsen:
@christinahedges Would it be hard to turn the video into an animated gif?

christinahedges:
Nope I can do that, no problem

barentsen:
This is a really good tutorial!  My only main comment is that the following sentence:

`It looks like the signal is still there in the background pixels. Unfortunately this means the signal is not astrophysical.`

Should be more something like:

`It looks like the signal is present in the background pixels with the same amplitude.  The Kepler PSF has a typical width of N pixels and so we would expect the amplitude of the signal to be much weaker in the background pixels that neighbor the star. To get a definitive answer, let's investigate whether the signal is present in the background pixels in a nearby target that is sufficiently far away from the star...`

christinahedges:
Lavly, I'll add this.

christinahedges:
rollingband

christinahedges:
@barentsen this is ready for review!

christinahedges:
@barentsen can we merge this?

christinahedges:
@barentsen this is why I merge my own pull requests.

barentsen:
Thanks @christinahedges!  This tutorial is really useful.  I made a few very minor style edits and added the tutorial to the index.  This is ready to merge now.

Note: we are going to have to update this tutorial if we implement `tpf.conesearch` as part of #138.

PR #118, opened 2018-05-04
[WIP] Added a Gaussian Process based corrector for long term trends.

### LINKED ISSUE #115: Lightcurve should have a Gaussian process based flatten routine
Savgol is quick but pretty terrible at flattening data. We should instead use celerite and a gaussian process. This is much more principled and will give us errors to propagate.

We can have it accept a mask to remove transits and burn ins due to down time. 

It should behave the same as regular flatten, either correcting the light curve or returning a light curve object with the flat trend.

There should be basic switches for the user to select.

It should be possible to give a time scale lower bound, such that the GP doesn't fit any variability on time scales of a planet transit.

christinahedges:
I've implemented a GP based corrector for long term trends. I don't like the way we call it at the moment which is 

image

I think we should perhaps have a LightCurve method called flatten_gp() which will call this.

Some key points
* The GP is much better than a savgol fit. It can interpolate data in that has been masked out and gives back errors
* The GP is much slower than a savgol fit, and that speed scales with the number of points.
* The GP tries to iteratively mask out data to find a better long term fit. This can be tweaked by the user
* There is a 'timescale' kwarg which should a) mask out data that is discrepant from the model with a boxwidth that is this time scale b) not fit a GP with a timescale shorter than this. I don't guarantee that always works...
* This won't work for transients with long time scales (e.g. supernovae) but that is probably expected.

Fixes #115


christinahedges:
It looks like appveyor is failing to install celerite in Python 2.7...@barentsen how do you want to deal with this?

barentsen:
Let's see if the appveyor issue is permanent or not!

christinahedges:
This PR has since been ingested into lightkurve in other PRs. I'm closing.

PR #119, opened 2018-05-06
makes bkjd_to_time discoverable

mirca:
Tiny PR to make the function ``bkjd_to_time`` importable from the top-level.
``bkjd_to_time`` will also show up in the sphinx docs next time we build the docs.


barentsen:
:+1: Thanks @mirca!!

PR #132, opened 2018-05-30
Revised fix to LightCurve apend method

KenMighell:
Fixed the LightCurve append method bug by adding these two lines:

if hasattr(new_lc, 'cadenceno'):  # KJM
    new_lc.cadenceno = np.append(new_lc.cadenceno, others[i].cadenceno)  # KJM

This should make the code work for Kepler **and** generic LightCurve objects.

-Ken Mighell
mighell at noao dot edu


barentsen:
Thank you @KenMighell!!!  This was a bug indeed, and the fix looks correct!  :+1:

I took the liberty to push a regression test to this PR, and added your name to the contributors. We can merge this once the tests pass.

PR #146, opened 2018-06-07
Fixed the timeobj bug

### LINKED ISSUE #116: NaN times are possible for certain quality masks.
I see NaN times when I use a quality_bitmask=None:

```python
tpf = KeplerTargetPixelFile.from_archive(201572338, quality_bitmask=None)
nan_times = tpf.time != tpf.time
nan_times.sum()

145
```

Many of these NaN times have quality values of 98304, which indicates *'No fine point', 'No data'*.
The possibility of NaN times is not necessarily a bug, and will likely only occur when quality_bitmask=None, since these quality flags are masked by default.  However, their presence does raise the question of whether we want to build in resilience to the possibility of NaN times.  They *do* have cadence numbers.

This question arose while designing `tpf.interact()`, which has so-far safely assumed times are all finite, but gets broken when NaN times are present.  I suspect it affects other places in the code that assume finite times.

christinahedges:
If the quality bitmask is "none", a file will have some points where the quality is "NoData". Where there is no data taken, there are NaNs in the time array. NaNs in the time array causes the timeobj function to break, because astropy doesn't know what date to assign to a NaN. This causes tools like interact to break.

This fixes the bug, by setting dates where there is no data to a JD of 0.

Fixes #116

@barentsen are you happy with this solution?


christinahedges:
Sorry, my autolinter has put in a load of spaces in the test file as well.

christinahedges:
Looks like travis failed at the MAST step

christinahedges:
Merging while pair coding with Geert

PR #147, opened 2018-06-07
Implemented a to_fits() method

christinahedges:
I've added a `to_fits()` method to `KeplerLightCurve`. I have made it a general method of `LightCurve` and had `KeplerLightCurve` implement it with additional header keywords for instrument, detector and mission. I've also added some unit tests and a tutorial for how to write out a fits file.


lightkurve/lightcurve.py
```diff
@@ -593,9 +598,126 @@ def to_csv(self, path_or_buf=None, **kwargs):
         """
         return self.to_pandas().to_csv(path_or_buf=path_or_buf, **kwargs)
 
+    def to_fits(self, path=None, overwrite=False, **kwargs):
+        """Writes the KeplerLightCurve to a fits fileself.
```
barentsen: `fileself` => `file`

lightkurve/lightcurve.py
```diff
@@ -593,9 +598,126 @@ def to_csv(self, path_or_buf=None, **kwargs):
...
+                                 'CREATOR': "lightkurve"}
+
+            for kw in default:
+                hdu.header['{}'.format(kw).upper()] = default[kw]
+                if default[kw] is None:
+                    warnings.warn('Value for {} is None'.format(kw))
+            if 'timeobj' in self.__dir__():
+                hdu.header['DATE-OBS'] = self.timeobj.isot[0]
+            else:
+                warnings.warn('Cannot assign DATE_OBS is None')
```
barentsen: Replace `warnings.warn` with `log.warning`.  (You may have to add `log = logging.getLogger(__name__)` at the top of the module if not already there.)

lightkurve/lightcurve.py
```diff
@@ -593,9 +598,126 @@ def to_csv(self, path_or_buf=None, **kwargs):
...
+
+            if ('quarter' in self.__dir__()) and (self.quarter is not None):
+                hdu.header['QUARTER'] = self.quarter
+            elif ('campaign' in self.__dir__()) and self.campaign is not None:
+                hdu.header['CAMPAIGN'] = self.campaign
+            else:
+                warnings.warn('Cannot find Campaign or Quarter number')
```
barentsen: `warnings` => `log`

lightkurve/lightcurve.py
```diff
@@ -592,9 +600,121 @@ def to_csv(self, path_or_buf=None, **kwargs):
...
+            for kw in default:
+                hdu.header['{}'.format(kw).upper()] = default[kw]
+                if default[kw] is None:
+                    log.warn('Value for {} is None'.format(kw))
+            if ('quarter' in dir(self)) and (self.quarter is not None):
+                hdu.header['QUARTER'] = self.quarter
+            elif ('campaign' in dir(self)) and self.campaign is not None:
+                hdu.header['CAMPAIGN'] = self.campaign
+            else:
+                log.warn('Cannot find Campaign or Quarter number')
```
barentsen: I get: `DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead`

christinahedges:
@mirca what do you think of this PR?

lightkurve/lightcurve.py
```diff
@@ -14,13 +17,17 @@
...
+from astropy.time import Time
 
 from .utils import running_mean
-
+from . import PACKAGEDIR
 
 __all__ = ['LightCurve', 'KeplerLightCurve', 'TessLightCurve',
            'iterative_box_period_search']
 
+log = logging.getLogger(__name__)
```
mirca: Should we declare ``log = logging.getLogger(__name__)`` on every module or should we import the one from ``__init__``? @christinahedges @barentsen

lightkurve/lightcurve.py
```diff
@@ -724,6 +843,43 @@ def to_pandas(self, columns=['time', 'flux', 'flux_err', 'quality',
...
+    def to_fits(self, path=None, overwrite=False, **kwargs):
+        """Export the KeplerLightCurve as an astropy.io.fits object.
+
+        Parameters
+        ----------
+        path : string, default None
+            File path, if None returns an astropy.io.fits object.
+        overwrite : bool
+            Whether or not to overwrite the file
+        **kwargs : dict
```
mirca: This is nitpick: usually we just use ``kwargs`` instead of ``**kwargs`` in the docstrings.

lightkurve/lightcurvefile.py
```diff
@@ -39,6 +39,16 @@ def time(self):
         """Time measurements"""
         return self.hdu[1].data['TIME'][self.quality_mask]
 
+    @property
+    def ra(self):
+        """Time measurements"""
```
mirca: Looks like these docstrings are incompatible?

lightkurve/lightcurvefile.py
```diff
@@ -39,6 +39,16 @@ def time(self):
...
+        return self.hdu[0].header['RA_OBJ']
+
+    @property
+    def dec(self):
+        """Time measurements"""
```
mirca: same here

mirca:
This looks great to me 👍 ! I wonder if at some point we would like to have a ``FITSFactory`` class, but let's not worry about that in this PR; food for thought.

lightkurve/lightcurve.py
barentsen: The nice thing about putting `log = logging.getLogger(__name__)` at the top of every module is that it allows you to control the stream handler and verbosity at the module-level.  It also allows you to print the origin of the log message (e.g. `lightkurve.targetpixelfile`).

lightkurve/lightcurve.py
barentsen: I'm fixing this and changing it from `kwargs` to `extra_data`, which is what it is.

lightkurve/lightcurve.py
mirca: awesome!

lightkurve/lightcurve.py
```diff
@@ -626,9 +634,118 @@ def to_csv(self, path_or_buf=None, **kwargs):
...
+            for kw in default:
+                hdu.header['{}'.format(kw).upper()] = default[kw]
+                if default[kw] is None:
+                    log.warning('Value for {} is None.'.format(kw))
+            if ('quarter' in dir(self)) and (self.quarter is not None):
+                hdu.header['QUARTER'] = self.quarter
+            elif ('campaign' in dir(self)) and self.campaign is not None:
+                hdu.header['CAMPAIGN'] = self.campaign
+            else:
+                log.warning('Cannot find Campaign or Quarter number.')
```
gully: to_fits resides in `LightCurve`, which should be unaware of Campaign or Quarter.  Recommend turning off the warning.

Issue #145, opened 2018-06-07
lightkurve should have a logger


christinahedges:
There should be a lightkurve logger that we can use to print error messages and warnings to the users.

barentsen:
Agreed!  I think it should be a very lightweight wrapper around Pythons built-in `logging` module.

christinahedges:
Yeah I recommend we use logging too, it doesn't even really need a wrapper you just need to great an object and then use it for warnings/info/debug messages

christinahedges:
Fixed in #150

PR #150, opened 2018-06-09
Introduce a default logger to lightkurve

barentsen:
This PR simplifies logging by introducing a default logger. (Addresses #145.)

### DIFF
--- a/lightkurve/__init__.py
+++ b/lightkurve/__init__.py
@@ -1,8 +1,14 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
+
 import os
 PACKAGEDIR = os.path.abspath(os.path.dirname(__file__))
+
+import logging
+log = logging.getLogger(__name__)
+log.addHandler(logging.StreamHandler())
+
 from .version import __version__
 from .prf import *
 from .lightcurve import *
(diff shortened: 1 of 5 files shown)


barentsen:
The purpose of this PR is to avoid having to have `verbose` arguments.  Instead, to make lightkurve verbose you'd execute:
```
lightkurve.log.setLevel('DEBUG')
```
and to make it quiet you'd execute:
```
lightkurve.log.setLevel('ERROR')
```

For example, this will be silent:
```
lcf = KeplerLightCurveFile.from_archive('Kepler-10', quarter=10)
```

And this will print astroquery's message:
```
lightkurve.log.setLevel('INFO')
lcf = KeplerLightCurveFile.from_archive('Kepler-10', quarter=10)
```

mirca:
This looks good to me! Thanks for this much awaited feature, @barentsen!
[This test](https://github.com/KeplerGO/lightkurve/pull/150/files#diff-3a6a2a90043504098a17134931dd25f9R147) is failing on Windows, though.

barentsen:
Actually the failing test shouldn't run on Appveyor because we should add the `@pytest.mark.remote_data` decorator, but that doesn't explain why it's failing.  I can't figure out why it's different in Windows!

mirca:
Thanks 👍

christinahedges:
This looks great, much better way to message users. Thanks @barentsen!

PR #151, opened 2018-06-10
Resolve MAST API source confusion (fixes #148)

### LINKED ISSUE #148: Downloading target from archive returns data for a different, nearby star
It's trying so hard to do a good job, but accidentally grabs a star 4 arcsec away (both have defined and downlinked apertures, which happen to touch each other)

```
from lightkurve import KeplerTargetPixelFile, LightCurve
tpf = KeplerTargetPixelFile.from_archive(6507433, quarter=8)


> Found 1 File(s)
> Downloading URL https://mast.stsci.edu/api/v0/download/file?uri=mast:Kepler/url/missions/kepler/target_pixel_files/0065/006507427/kplr006507427-2011073133259_lpd-targ.fits.gz to ./mastDownload/Kepler/kplr006507427_lc_Q011111111111111111/kplr006507427-2011073133259_lpd-targ.fits.gz ...

```
The star it gets is one pixel away from the one I want. My guess is that there's a radius search behind the scenes, and there's some bad conversations to and from RA and Dec and it's not accounting for the fact that it knows the true KIC of the target I want? But that's just a guess on my part.

Still <3 Lightkurve!

barentsen:
This PR will fix #148.  The first commit adds a failing unit test that reproduces @benmontet's problem.

### DIFF
--- a/lightkurve/tests/test_mast.py
+++ b/lightkurve/tests/test_mast.py
@@ -143,3 +143,14 @@ def test_verbosity(capfd):
     tpf = KeplerTargetPixelFile.from_archive(5728079, quarter=1, verbose=False)
     out, err = capfd.readouterr()
     assert len(out) == 0
+
+
+@pytest.mark.remote_data
+def test_source_confusion():
+    # Regression test for issue #148.
+    # When obtaining the TPF for target 6507433, @benmontet noticed that
+    # a target 4 arcsec away was returned instead.
+    # See https://github.com/KeplerGO/lightkurve/issues/148
+    desired_target = 6507433
+    tpf = KeplerTargetPixelFile.from_archive(desired_target, quarter=8)
+    assert tpf.keplerid == desired_target
--- a/lightkurve/tests/test_mast.py
+++ b/lightkurve/tests/test_mast.py
@@ -38,11 +38,9 @@ def test_search_kepler_tpf_products():
         assert(len(ca) == len(cb))
         assert(~np.any(ca['description'] == cb['description']))
         assert(~np.any(ca['dataURI'] == cb['dataURI']))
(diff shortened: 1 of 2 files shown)


barentsen:
I managed to fix #148 by passing the KIC ID to MAST as the `target_name` instead of the `objectname`, but a side-effect of this fix is that the command below now returns 1 instead of 2 tpfs, because the KIC ID is no longer resolved to coordinates:
```
KeplerTargetPixelFile.from_archive(5728079, cadence='long', radius=60, quarter=1, targetlimit=2)
```

...this gives us an extra reason to separate the conesearch into a separate function.

lightkurve/tests/test_mast.py
```diff
@@ -93,12 +91,12 @@ def test_kepler_tpf_from_archive():
...
-    # If we ask for a nearby target, it should only give back one extra with the same quarter.
+    # Ask for one extra nearby target
     tpfs = KeplerTargetPixelFile.from_archive(
-        5728079, cadence='long', radius=60, quarter=1, targetlimit=2)
+        "GJ 9827", cadence='long', radius=60, campaign=12, targetlimit=2)
     assert(isinstance(tpfs, list))
     assert(isinstance(tpfs[0], KeplerTargetPixelFile))
-    assert(tpfs[0].quarter == tpfs[1].quarter)
+    assert(tpfs[0].campaign == tpfs[1].campaign)
     assert(tpfs[0].keplerid != tpfs[1].keplerid)
```
christinahedges: Why did this unit test change? It looks like just the target has changed not purpose of the test?

christinahedges:
Looks good, I think we should work cone search in ASAP to avoid confusion.

christinahedges:
@barentsen there is a merge conflict with the test from #150 . Once you fix this please merge, we need @benmontet to be a happy customer.

PR #152, opened 2018-06-11
[WIP] Use `sphinx_automodapi` to generate the API docs

barentsen:
We should consider switching to AstroPy's `sphinx-automodapi` module to auto-generate the API docs, because it will show class properties (implemented using the `@property` decorator) in a nice table called `Attributes Summary` alongside other attributes.  I can't figure out how to achieve this using `automodule`, and this is important because we use `@property` a lot.

...however `automodapi` is a bit funny in other ways. @mirca what do you think?


mirca:
I think this is the way forward! The inheritance diagrams are also working, which is a big plus.

mirca:
It would have been the way forward if I haven't messed up your branch, 😭

mirca:
@barentsen sorry about that! I made a backup branch with the original commits (#153). One solution would be to create a new branch, ``git checkout -b "use-automodapi-take-2" mirca/fix-my-mess``, and open another PR 😢

barentsen:
@mirca Hahahahaaaa RIP this PR.  😱💀👻

👻👻👻👻👻👻👻👻👻👻👻👻

mirca:
😭😭😭😭😭😭😭

PR #153, opened 2018-06-11
[WIP] Use `sphinx_automodapi` to generate the API docs

mirca:
This is a backup for PR #152, because I messed up @barentsen's branch.

@barentsen said:

> We should consider switching to AstroPy's sphinx-automodapi module to auto-generate the API docs, because it will show class properties (implemented using the @property decorator) in a nice table called Attributes Summary alongside other attributes. I can't figure out how to achieve this using automodule, and this is important because we use @property a lot.
> 
> ...however automodapi is a bit funny in other ways. @mirca what do you think?


barentsen:
@mirca I thought you fixed the failing travis test (caused by lightkurve.__version__ not being available pre-install) but the commit seems gone?

mirca:
@barentsen what do you think of this workaround?

Issue #154, opened 2018-06-11
Print a friendly error message when `interact()` is called from a Python shell (instead of a notebook)


barentsen:
See #127 for a user experience we want to improve.

mirca:
This has been fixed in #155

PR #155, opened 2018-06-11
Fixes issue #154

ssagear:
Prints an error message when 'interact()' is called in a shell


ssagear:
@gully The error message will say interact() must be run in Jupyter Notebook, but then the python shell hangs up and is difficult to exit

lightkurve/targetpixelfile.py
```diff
@@ -772,7 +772,14 @@ def update(cadence, log_stretch):
             else:
                 vert.update(line_alpha=0)
                 fig2_dat.data_source.data['image'] = [self.flux[0, :, :] * np.NaN]
-            push_notebook()
+            try:
+                push_notebook()
+            except AttributeError:
+                print('Interact tool must be run in a Jupyter Notebook.')
```
christinahedges: Let's use the new nifty logging feature instead of print statements. You should be able to do
```
log.error('Interact tool must be run in a Jupyter Notebook.\n')
```

instead of the print statements.

christinahedges:
Still need to get rid of that notebook and then we can merge! :D

christinahedges:
@barentsen let's merge this

barentsen:
:+1:

PR #158, opened 2018-06-12
Ensure `Time` objects have the TDB time scale set

### LINKED ISSUE #149: Docstrings need to clarify that Kepler times are in TDB
The times in the Kepler data products are in the Barycentric Dynamical Time frame (TDB), which is a time system that does not include the leap seconds.  We need to make sure that lightkurve clarifies this in the docstrings that relate to times.  We can refer to Section 2.1.2 of the Kepler Archive Manual for details.

barentsen:
This PR fixes #149.

### DIFF
--- a/lightkurve/tests/test_lightcurve.py
+++ b/lightkurve/tests/test_lightcurve.py
@@ -277,13 +277,15 @@ def test_to_csv():
         pass
 
 
-def test_date():
-    '''Test the lc.date() function'''
+def test_timeobj():
+    '''Test the `timeobj` property'''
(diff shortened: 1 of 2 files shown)


barentsen:
Upon discussion with @christinahedges, we are starting to realize that `bkjd_to_time` returns the time in the reference frame of the spacecraft, as opposed to the solar system barycenter.  We want the latter, so we're going to have to fix this function!

barentsen:
I'm going to close this PR in favor of refactoring the way we deal with Time objects in PR #159.

PR #159, opened 2018-06-12
Allow astropy Time objects to be retrieved from LightCurve

barentsen:
Over in #147, @christinahedges is proposing that we allow `timeobj` (`astropy.time.Time`) to be passed on to `KeplerLightCurve`, which would make it much easier to access time in a variety of formats, but it carries the problem that `lightcurve.time` and `lightcurve.timeobj` may become inconsistent.

As an alternative, I wonder if we should pass the `time_format` argument to any `LightCurve` object.  This would allow astropy `Time` objects to be generated dynamically at any time.  (See PR attached.)

The main drawback is that `time_format = 'bkjd'` or `time_format = 'bjd'` is not supported by astropy right now.  The conversion from `bkjd` to `bjd` is trivial, but the conversion from `bjd` to `jd` is not.  The correction is pre-computed and made available in the Kepler data products, but that is not necessarily passed on to a `LightCurve` object.

Any thoughts?


christinahedges:
Ok I've changed my mind, we shouldn't have timeobj be inherited. We should just use TIME and correct it to BJD by adding 2454833.0.

I'll undo this!

christinahedges:
I've updated #147 to remove the timeobj inheritance

barentsen:
@christinahedges I am happy with this now.  How do you feel about dealing with time as proposed in this PR now?

christinahedges:
Yeah this seems like a good way to deal with it. So long as astropy_time is never required or breaks my lightcurve when I want to initialize one without a frame

`lc = lightkurve.LightCurve(x, y, e)`

Shouldn't break.

barentsen:
@christinahedges Agreed!
@mirca How do you feel about these changes?

mirca:
This looks good to me! I like the name ``astropy_time`` :D

Issue #161, opened 2018-06-13
Injection and recovery tool for lightkurve


ssagear:
Instrumental artifacts often make it difficult to distinguish real astrophysical phenomena on long timescales. Lightkurve could include a tool for injection and recovery to quantify the extent of instrumental artifacts on astrophysical signals of interest. These signals could include supernovae, transits, rotation and flares.

I propose a method in the `lightcurve` class called `.inject` that will inject a known astrophysical signal, e.g. a supernova light curve from [sncosmo](http://sncosmo.readthedocs.io/en/v1.6.x/) or a planetary transit from [ktransit](https://github.com/mrtommyb/ktransit).  One could also add a `.recover` method.

ssagear:
**Discovery of Available Parameters**

.inject_sn right now allows the user to choose the source of the supernova model (if it is built into sncosmo), but it depends on the user already knowing what parameters are taken by that source.  The method should be able to tell the user what parameters a certain source takes.

**Synthetic Light Curve Class**

It might also be good to have a "synthetic lightcurve" class that will inherit from the lightcurve class:
```python 
class SyntheticLightCurve(LightCurve):
```
This class would inherit from LightCurve and also include attributes specific to a synthetic lightcurve, e.g. planet period/radius, supernova type, etc.  This will specify that the synthetic light curve is **not** real data.

ssagear:
**Normalization of Light Curves**

All light curves will be normalized to 1.0 - supernova signals will be added, and transit signals will be multiplied. This assumes no background flux (background flux is additive) - maybe add a flag.

**Units of Supernova amplitude**

Convert the model output from sncosmo from [photons/s/cm^2] to [electrons/s]
([electrons/s] must then be converted to counts).

ssagear:
**Implementation of Inject Methods**

We might include methods like .inject.supernova() or .inject.transit() instead of different methods for each astrophysical signal we want to inject.  

@barentsen raises the possibility of supporting adding models to lightcurves:
```python
model = some_func(parameters)
synth_lc = model + data_lc
```
`model` only possesses a functional form, but is not actually a model light curve.
We have decided to revisit these abstractions after some experimentation with building directly into the `LightCurve` class.

We would do so by overriding the `__add__` method like so:

```python
def __add__(self, other):
        copy_self = copy.copy(self)
        if other.is_instance(SpecialModelClass):
             #evaluate the model at these time points
             synth_flux = other.create(self.time)
[...]

PR #162, opened 2018-06-13
[WIP] Add an injection and recovery method to lightcurve.py

ssagear:
This pull request starts a demo for a solution to #161.


lightkurve/injection.py
```diff
@@ -0,0 +1,288 @@
...
+        Returns
+        -------
+        values : array-like
+            Returns array of randomly chosen values.
+        """
+        return np.random.uniform(self.lb, self.ub)
+
+    def plot(self):
+        """Plots specified distribution."""
+        t = np.arange(self.lb, self.ub, 0.01)
```
christinahedges: So that we can see the edges of the uniform distribution, why don't you make the range something like

`np.linspace(self.lb*0.9, self.ub*1.1, 1000)`

(note `np.linspace` will allow you to fix it to 1000 points. A `np.arange(0, 1e10, 0.01)` will have 1e12 points and we probably don't need that much precision!

lightkurve/injection.py
```diff
@@ -0,0 +1,288 @@
...
+        vals = [1]*len(t)
```
christinahedges: There should be a space before and after the `*` here.

You can add a linter to your editor to to highlight these problems for you.

lightkurve/injection.py
```diff
@@ -0,0 +1,288 @@
...
+    mean : float
+        Mean of distribution
+    var : float
+        Standard deviation of distribution
+    """
+    def __init__(self, mean, var):
+        self.mean = mean
+        self.var = var
+
+    def __repr__(self):
```
christinahedges: space after ,

lightkurve/injection.py
```diff
@@ -0,0 +1,288 @@
...
+            Dictonary of planet parameters. Options are:
+                impact: an impact parameter
+                ecosw, esinw : an eccentricity vector
+                occ : a secondary eclipse depth
+                rvamp : (not sure)
+                ell : (not sure)
+                alb : (not sure)
+        """
+
+        if isinstance(period, (GaussianDistribution, UniformDistribution)):
```
christinahedges: Rather than having 3 if statements, could we make this into a loop over all the attributes? Something you have below?

lightkurve/injection.py
```diff
@@ -0,0 +1,288 @@
...
+            Flux array of lightcurve
+
+        """
+        self.model.add_data(time=time)
+        transit_flux = self.model.transitmodel
+        return transit_flux
+
+class SupernovaModel(object):
+
+    from lightkurve.injection import GaussianDistribution
```
christinahedges: We should ask @barentsen how to do this properly.

We want to be able to use `isinstance(x, GaussianDistribution)` but it won't recognize `GaussianDistribution` as a class.

gully:
Layout for recovery method:

- [x] A function that makes the initial guess
- [x] A function that returns the evaluated model given input parameters (T0, amp, z, background estimate)
- [x] A function that computes the log likelihood
- [x] A minimize function
- [ ] A function that checks if the retrieved parameters are within a threshold of the input

gully:
Ok, I'm looking over this Pull Request, here are some notes:

### 1. Improve focus and clarity of notebook tutorials
We currently have 20 Jupyter notebooks committed to the tutorials section.  We seek probably 1-3 notebook tutorials for lightkurve users to understand what the injection/recovery tool does, and see a range of its applications.  These exploratory notebooks can be moved to a different branch or workspace, or in principle can be accessed from the git revision history.

```bash
$ git diff --name-only c0d0667319ac26469f01de2d35e679dec2690841
docs/source/tutorials/2.09-inject-artifical-signals.ipynb
docs/source/tutorials/2.10-recovering-artificial-signals-old.ipynb
docs/source/tutorials/2.10-recovering-artificial-signals-python3.ipynb
docs/source/tutorials/2.10-recovering-artificial-signals.ipynb
docs/source/tutorials/2.11-modeling-injection-classes.ipynb
docs/source/tutorials/Chi-squared_tests.ipynb
docs/source/tutorials/Example_SNrecovery_Poster.ipynb
docs/source/tutorials/Exoplanet_fitting_with_batman.ipynb
docs/source/tutorials/L_dwarf_injrec.ipynb
docs/source/tutorials/MCMC_supernovae.ipynb
docs/source/tutorials/Optimization of Planetary Transits.ipynb
docs/source/tutorials/Transit_Recovery_lightkurve_integration.ipynb
docs/source/tutorials/Untitled.ipynb
docs/source/tutorials/Untitled1.ipynb
docs/source/tutorials/Untitled2.ipynb
docs/source/tutorials/Untitled3.ipynb
docs/source/tutorials/Untitled4.ipynb
docs/source/tutorials/recovery_testing_mcmc.ipynb
docs/source/tutorials/recovery_testing_mcmc_2.ipynb
docs/source/tutorials/recovery_testing_mcmc_3.ipynb
lightkurve/injection.py
lightkurve/lightcurve.py
```

Let's make 1 or 2 (or 3 if needed) high quality notebook tutorials.  The notebooks can include explanatory material, plots, and the least amount of code to show the features.  The cells should be run in order and start from [1].  We can use the rich display features of Jupyter to render equations, hyperlinks, and linked images.  We can put headings and subheadings when needed for long notebooks.  The names of the notebook should be updated to follow the section conventions `2.nn`, and `nn` should start from the end of the most recently submitted notebook (currently `2.10`).

I suspect we should separate the tutorials into science topics to appeal to targeted science communities-- supernova and planet communities.  A third tutorial could show some advanced applications of injection/recovery, such as iterating over many parameter ranges to assess detection efficiency.

gully:
## Feedback from discussion with @barentsen 

1. Is it possible to go the next step-- perform the iteration over many parameters to say the signal is routinely recovered?
2. How do you decide whether something is recovered or not?  Do we recover these parameters within *X* percent?  What is the threshold, and how do you assign it?  
3. Whether or not to have "SupernovaModel" class or do something like `SyntheticLightCurve.from_sncosmo()` or `SyntheticLightCurve.from_batman()`.

gully:
Why does the SALT2 model show a kink?

screen shot 2018-07-24 at 10 30 32 am

gully:
What attributes and methods do we want `SyntheticLightCurves` to possess?

```python
regular_attrs = np.array([ attr for attr in lc.__dir__() if attr[0:2] != '__'])\
synth_attrs = np.array([ attr for attr in lc_with_supernova.__dir__() if attr[0:2] != '__'])
diff_set = set(regular_attrs) - set(synth_attrs)
np.array(list(diff_set))

array(['quality', 'centroid_col', 'campaign', 'quarter', 'keplerid',
       'centroid_row', 'cadenceno', 'quality_bitmask', 'ra', 'correct',
       'mission', 'channel', 'dec'], dtype='<U15')
```

We probably *do* want cadenceno, but maybe not the attributes that imply the real data.

PR #163, opened 2018-06-14
Added a tpf.get_sources function which queries a catalog and returns the targets in a TPF as astropy.table

AndyTza:
This function returns a queried table of field stars that are near the tpf target. The function thus far has libraries that contain catalogues of KIC and EPIC only. 

Would be interesting to expand the library and add Gaia DR2 and other catalogues. 

You can also see a quick demo of the function [here](https://github.com/AndyTza/projects_18/blob/master/notebooks/Find%20Stars%20Function%20Demo.ipynb).


AndyTza:
This updated function will query stars (either KIC or EPIC) using a radial search (arcmins) given the tpf coordinates. 

The function is consistent with [MAST](http://archive.stsci.edu/kepler/kepler_fov/search.php) Kepler Target Search. I've made a small demo of the function [here](https://github.com/AndyTza/projects_18/blob/radius_finder/notebooks/Find%20Stars%20Function%20Demo.ipynb) for more information. 

@christinahedges

christinahedges:
@AndyTza has implemented a great find_stars function. There is a bug in vizier where the object ID's are returned as int16's which is too imprecise for Kepler IDs. I put in an issue [here](https://github.com/astropy/astroquery/issues/1173) and it looks like it will get fixed. Once it's fixed I think this looks good!

christinahedges:
@AndyTza appveyor and travis can't run the code you've added. If you click on "details" you can see the traceback for where they broke and fix it :)

C

AndyTza:
@christinahedges updates on find_stars--  appveyor & travis have passed!

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +196,122 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+                     'parameters':["KIC", "RAJ2000", "DEJ2000", "pmRA", "pmDE", "kepmag"]},
+               "EPIC":
+               {'vizier':"IV/34/epic",
+                'parameters':["ID", "RAJ2000", "DEJ2000", "pmRA", "pmDEC", "e_pmRA", "e_pmDEC", "Kpmag"]},
+               "Gaia":
+               {'vizier':"I/345/gaia2",
+               'parameters':["DR2Name", "RA_ICRS", "DE_ICRS", "pmRA", "pmDE", "e_pmRA", "e_pmDE", "Gmag"]}}
+
+        # identifies catalog
+        viz_id = (ID[catalog])['vizier']
```
christinahedges: Don't need the () here

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +196,122 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+        pars = (ID[catalog])['parameters']
```
christinahedges: Don't need ()

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +196,122 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+        #Skycoord the centre of target
+        cent = SkyCoord(ra=self.ra, dec=self.dec, frame='icrs', unit=(u.deg, u.deg))
+
+        # Choose columns from Vizier
+        v = Vizier(catalog=[viz_id], columns=pars)
+        # query around cent. with radius
+        result = v.query_region(cent, radius=radius*u.arcmin, catalog=viz_id)
+
+        # Rename names in KIC for consistency with EPIC
+        result[viz_id]['KIC'].name = 'ID'
```
christinahedges: This is going to break if you have Gaia or EPIC, there is no 'KIC' to change. You should iterate through the `'parameters'` of the catalog and change them to the ones we came up with.

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +196,122 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+
+        # Load ra & dec of all tpf pixels
+        gc_ra, gc_dec = self.get_coordinates(cadence='all')
+
+        # Reduce calculation for astroy seperation
+        c1 = np.asarray([gc_ra.ravel(), gc_dec.ravel()])
+        co = np.round(c1, decimals=5)
+
+        # Return unique pairs
+        pairs_gc = (np.unique(co, axis=1))
```
christinahedges: `gc` is a bit opaque, can we call it something that's clearer?

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +196,122 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+
+        # Make pairs_gc into SkyCoord
+        sky_pairs_gc = SkyCoord(ra=pairs_gc[0]*u.deg, dec=pairs_gc[1]*u.deg, frame='icrs', unit=(u.deg, u.deg))
+        # Make pairs in sky_data
+        sky_data = SkyCoord(ra=data['RAJ2000'], dec=data['DEJ2000'], frame='icrs', unit=(u.deg, u.deg))
+
+        new_ra, new_dec = [], []
```
christinahedges: All of this can be saved by doing a boolean array and then masking the `data` table.

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +196,122 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+            if store==True:
+                new_ra.append(data['RAJ2000'][i])
+                new_dec.append(data['DEJ2000'][i])
+                new_id.append(data['ID'][i])
+                new_pmra.append(data['pmRA'][i])
+                new_pmdec.append(data['pmDEC'][i])
+                new_kpmag.append(data['Kpmag'][i])
+
+        # Dictionary of parameters of queried statrs
+        new_table = {'ID':new_id, 'RAJ2000':new_ra, 'DEJ2000':new_dec,
```
christinahedges: You can just output the table for now, no need for a dictionary.

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +196,122 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+
+        # identifies catalog
+        viz_id = (ID[catalog])['vizier']
+        pars = (ID[catalog])['parameters']
+
+        #Skycoord the centre of target
+        cent = SkyCoord(ra=self.ra, dec=self.dec, frame='icrs', unit=(u.deg, u.deg))
+
+        # Choose columns from Vizier
+        v = Vizier(catalog=[viz_id], columns=pars)
```
christinahedges: This can be 

`data = Vizier(catalog=[viz_id], columns=pars)[viz_id]`

to reduce the lines below.

lightkurve/targetpixelfile.py
```diff
@@ -197,12 +198,10 @@ def __repr__(self):
         return('KeplerTargetPixelFile Object (ID: {})'.format(self.keplerid))
 
 
-    def find_stars(self, catalog=None, radius=0.5):
+    def query_catalog(self, catalog=None, radius=0.5, seperation_factor=12):
```
christinahedges: I'm not sure we want this separation factor, as it's redundant with radius. Why would I set radius and separation factor?

(It also needs a better name, perhaps tolerance.)

lightkurve/targetpixelfile.py
```diff
@@ -212,6 +211,11 @@ def find_stars(self, catalog=None, radius=0.5):
         radius: float
             Radius of cone search centered on the target in arcminutes.
             Default radius is 0.5 arcmin.
+        seperation_factor: float
```
christinahedges: I still don't think separation factor should be a keyword...But I could be wrong. If we use it we should change this docstring, it's confusing. Let's use something like

Tolerance for source matching. Sources that fall within this tolerance of pixels in the KeplerTargetPixelFile will be returned in the result. Default is 3 pixels (12 arcsec).

lightkurve/targetpixelfile.py
```diff
@@ -250,67 +254,66 @@ def find_stars(self, catalog=None, radius=0.5):
                      'parameters':["KIC", "RAJ2000", "DEJ2000", "pmRA", "pmDE", "kepmag"]},
                "EPIC":
                {'vizier':"IV/34/epic",
-                'parameters':["ID", "RAJ2000", "DEJ2000", "pmRA", "pmDEC", "e_pmRA", "e_pmDEC", "Kpmag"]},
+                'parameters':["ID", "RAJ2000", "DEJ2000", "pmRA", "pmDEC", "Kpmag","e_pmRA", "e_pmDEC"]},
```
christinahedges: Space after a ,

lightkurve/targetpixelfile.py
```diff
@@ -250,67 +254,66 @@ def find_stars(self, catalog=None, radius=0.5):
...
+        viz_id = ID[catalog]['vizier']
 
         #Skycoord the centre of target
         cent = SkyCoord(ra=self.ra, dec=self.dec, frame='icrs', unit=(u.deg, u.deg))
 
         # Choose columns from Vizier
-        v = Vizier(catalog=[viz_id], columns=pars)
-        # query around cent. with radius
+        v = Vizier(catalog=[viz_id], columns=ID[catalog]['parameters'])
+        # query around centre with radius
```
christinahedges: British spellings are always appreciated.

lightkurve/targetpixelfile.py
```diff
@@ -250,67 +254,66 @@ def find_stars(self, catalog=None, radius=0.5):
...
         result = v.query_region(cent, radius=radius*u.arcmin, catalog=viz_id)
 
-        # Rename names in KIC for consistency with EPIC
-        result[viz_id]['KIC'].name = 'ID'
-        result[viz_id]['pmDE'].name = 'pmDEC'
-        result[viz_id]['kepmag'].name = 'Kpmag'
+        # Rename column names
+        new_pars = ['ID', 'RA', 'DEC', 'pmRA', 'pmDEC']
+        for i in range(len(new_pars)):
+            result[viz_id].rename_column(result[viz_id].colnames[i], new_pars[i])
```
christinahedges: Much better!

lightkurve/targetpixelfile.py
```diff
@@ -250,67 +254,66 @@ def find_stars(self, catalog=None, radius=0.5):
...
 
         # Queried stats
         data = result[viz_id]
 
         # Load ra & dec of all tpf pixels
-        gc_ra, gc_dec = self.get_coordinates(cadence='all')
+        pixels_ra, pixels_dec = self.get_coordinates(cadence='all')
```
christinahedges: Much better names.

lightkurve/targetpixelfile.py
```diff
@@ -250,67 +254,66 @@ def find_stars(self, catalog=None, radius=0.5):
...
 
         # Reduce calculation for astroy seperation
-        c1 = np.asarray([gc_ra.ravel(), gc_dec.ravel()])
+        c1 = np.asarray([pixels_ra.ravel(), pixels_dec.ravel()])
```
christinahedges: These catalogs could be called something more intuitive.

lightkurve/targetpixelfile.py
```diff
@@ -250,67 +254,66 @@ def find_stars(self, catalog=None, radius=0.5):
...
-        pairs_gc = (np.unique(co, axis=1))
+        pixel_pairs = (np.unique(co, axis=1))
 
-        # Make pairs_gc into SkyCoord
-        sky_pairs_gc = SkyCoord(ra=pairs_gc[0]*u.deg, dec=pairs_gc[1]*u.deg, frame='icrs', unit=(u.deg, u.deg))
+        # Make pixel pairs into SkyCoord
+        sky_pixel_pairs = SkyCoord(ra=pixel_pairs[0]*u.deg, dec=pixel_pairs[1]*u.deg, frame='icrs', unit=(u.deg, u.deg))
         # Make pairs in sky_data
-        sky_data = SkyCoord(ra=data['RAJ2000'], dec=data['DEJ2000'], frame='icrs', unit=(u.deg, u.deg))
+        sky_data = SkyCoord(ra=data['RA'], dec=data['DEC'], frame='icrs', unit=(u.deg, u.deg))
```
christinahedges: Perhaps `sky_sources` would be more descriptive.

lightkurve/targetpixelfile.py
```diff
@@ -250,67 +254,66 @@ def find_stars(self, catalog=None, radius=0.5):
...
-        new_ra, new_dec = [], []
-        new_id = []
-        new_pmra, new_pmdec, new_e_pmra, new_e_pmdec = [], [], [], []
-        new_kpmag = []
+        sep_mask = np.array(seperation_mask, dtype=bool)
+        # Append masked list to data table
+        for i in range (len(data.colnames)):
+            data[data.colnames[i]].mask = sep_mask
+
+        # Scan and remove masked values
```
christinahedges: This should be what a mask is for...?

lightkurve/targetpixelfile.py
```diff
@@ -16,6 +16,7 @@
 from astropy import units as u
 from astropy.coordinates import SkyCoord
 from astropy.table import Table
+import numpy.ma as ma
```
christinahedges: Let's use np.ma rather than explicitly importing this.

christinahedges:
@barentsen @gully Andy and I really happy with is, could you guys take a look and review for us? Thanks!

lightkurve/targetpixelfile.py
```diff
@@ -12,11 +12,15 @@
 import matplotlib.pyplot as plt
 import numpy as np
 from tqdm import tqdm
+from astroquery.vizier import Vizier
```
barentsen: `Vizier` is only used in `utils.py`, so doesn't need to be imported here

lightkurve/targetpixelfile.py
```diff
@@ -12,11 +12,15 @@
...
+from astropy import units as u
+from astropy.coordinates import SkyCoord
+from astropy.table import Table
```
barentsen: `Table` is already being imported on line 9

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +195,88 @@ def from_archive(target, cadence='long', quarter=None, month=None,
     def __repr__(self):
         return('KeplerTargetPixelFile Object (ID: {})'.format(self.keplerid))
 
+
+    def get_sources(self, catalog=None, magnitude_limit=18, dist_tolerance=3):
+        """
+        Returns a crossmatched table of stars that are centered on the target
```
barentsen: I don't think the table the function returns is `crossmatched`?

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +195,88 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+        of the tpf file. Current catalog supported are KIC, EPIC and Gaia DR2.
+
+        Parameters
+        -----------
+        catalog: string
```
barentsen: replace `string` with `'EPIC', 'KIC', or 'Gaia'`

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +195,88 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+            Indicate catalog assigned for mission. If Kepler, catalog will be KIC
+            if K2 catalog is EPIC.
```
barentsen: I recommend re-phrasing as follows: `Catalog to query.  For Kepler target pixel files the default is 'KIC', for K2 target pixel files the default is 'EPIC'.`

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +195,88 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+        magnitude_limit : int or float
+            Limit the returned magnitudes to only stars brighter than magnitude_limit.
+            Default 18.
```
barentsen: Which band does the limit apply to?  I assume `Kp` for EPIC/KIC and `G` for Gaia -- please make this explicit in the docstring.

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +195,88 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+        dist_tolerance: float
+            Tolerance for source matching in pixels. Sources that fall within this
+            tolerance of pixels in the KeplerTargetPixelFile will be returned in the result.
+            Default is 3 pixels (12 arcsec).
```
barentsen: This description is a bit unclear.  I would rephrase as: `Maximum distance (in arcseconds) a source may be separated from the edge of the target pixel file.`

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +195,88 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+            Catalog ID from catalog.
+        RAJ200: astropy.table.column (KIC & EPIC)
+            Right ascension [degrees]
+        DEJ2000: astropy.table.column (KIC & EPIC)
+            Declination [deg]
+        pmRA: astropy.table.column (KIC & EPIC)
+            Proper motion for right ascension [mas/year]
+        pmDEC: astropy.table.column (KIC & EPIC)
+        Kpmag: astropy.table.column (KIC & EPIC)
+            Magnitude in Kepler band [mag]
```
barentsen: The per-column documentation is great, but it needs to be indented to fit inside the `result : astropy.table` indent block or the documentation-generator will be confused.

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +195,88 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+        """
+
+        if catalog is None:
+            if self.mission == 'Kepler':
+                catalog = 'KIC'
+            elif self.mission == 'K2':
+                catalog = 'EPIC'
+            else:
+                log.error('Please provide a catalog.')
```
barentsen: Because the function cannot return a valid result at this point, it must raise an exception, i.e. `raise ValueError('Please provide a catalog')`.

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +195,88 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+
+        #Skycoord the centre of target
+        cent = SkyCoord(ra=self.ra, dec=self.dec, frame='icrs', unit=(u.deg, u.deg))
+
+        #Find the size of the TPF
+        radius = np.max(self.flux.shape[1:]) * 4 * u.arcsecond
+
+        # query around centre with radius
+        data = query_catalog(cent, radius=radius, catalog=catalog)
```
barentsen: Should `dist_tolerance` be added to `radius` here?

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +195,88 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+
+        # Load ra & dec of all tpf pixels
+        pixels_ra, pixels_dec = self.get_coordinates(cadence='all')
```
barentsen: No need to compute pixel coordinates for all cadences here, instead we can simply request the pixel coordinates of the first cadence, i.e. use `self.get_coordinates(cadence=0)`

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +195,88 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+
+        # Reduce calculation for astroy seperation
+        pixel_radec = np.asarray([pixels_ra.ravel(), pixels_dec.ravel()])
+        pixel_radec = np.round(pixel_radec, decimals=5)
```
barentsen: Is rounding necessary?

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +195,88 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+
+        # Return unique pairs
+        pixel_pairs = (np.unique(pixel_radec, axis=1))
```
barentsen: This step can be dropped if we use `self.get_coordinates(cadence=0)` above.

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +195,88 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+
+        # Make pixel pairs into SkyCoord
+        sky_pixel_pairs = SkyCoord(ra=pixel_pairs[0]*u.deg, dec=pixel_pairs[1]*u.deg, frame='icrs', unit=(u.deg, u.deg))
+        # Make pairs in sky_sources
+        sky_sources = SkyCoord(ra=data['RA'], dec=data['Dec'], frame='icrs', unit=(u.deg, u.deg))
+
+        seperation_mask = []
```
barentsen: `seperation_mask` => `separation_mask`

lightkurve/tests/test_targetpixelfile.py
```diff
@@ -51,6 +55,20 @@ def test_tpf_plot():
     with pytest.raises(ValueError):
         tpf.plot(scale="blabla")
 
+def test_find():
```
barentsen: Rename to `test_get_sources()` to make it explicit which function is being tested.

lightkurve/utils.py
```diff
@@ -293,3 +300,72 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
...
+def query_catalog(coordinate, catalog="KIC", radius=0.5):
+        """
+        Returns an astropy table of the sources inside a radius query.
+        Current catalogs supported are KIC, EPIC and Gaia DR2.
+
+        Parameters
+        -----------
+        coordinate : astropy.coordinates.SkyCoord
+            Coordinate to query around.
+        catalog: string
```
barentsen: `string` => `'KIC', 'EPIC', or 'Gaia'`

lightkurve/utils.py
```diff
@@ -293,3 +300,72 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
...
+            Indicate catalog assigned for mission. If Kepler, catalog will be KIC
+            if K2 catalog is EPIC.
```
barentsen: Remove reference to Kepler or K2 from this docstring, because this function does not know about the mission it is being used for.

lightkurve/utils.py
```diff
@@ -293,3 +300,72 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
...
+        radius: float
+            Radius of cone search centered on the target in arcminutes.
+            Default radius is 0.5 arcmin.
```
barentsen: Are you sure the units expected by astroquery are `arcminutes`, not `arcseconds`?

lightkurve/utils.py
```diff
@@ -293,3 +300,72 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
...
+        result = v.query_region(coordinate, radius=radius, catalog=viz_id)
+        if len(result) == 0:
+            log.error('No sources found in queried region. Try another catalog.')
+
+        # Rename column names
+        new_pars = ['ID', 'RA', 'Dec', 'pmRA', 'pmDec', "mag"]
+        for i in range(len(new_pars)):
+            result[viz_id].rename_column(result[viz_id].colnames[i], new_pars[i])
+
+        # Queried stats
```
barentsen: The purpose of the comment `queried stats` isn't clear.

lightkurve/utils.py
```diff
@@ -293,3 +300,72 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
...
+            Catalog ID from catalog.
+        RAJ200: astropy.table.column (KIC & EPIC)
+            Right ascension [degrees]
+        DEJ2000: astropy.table.column (KIC & EPIC)
+            Declination [deg]
+        pmRA: astropy.table.column (KIC & EPIC)
+            Proper motion for right ascension [mas/year]
+        pmDEC: astropy.table.column (KIC & EPIC)
+        Kpmag: astropy.table.column (KIC & EPIC)
+            Magnitude in Kepler band [mag]
```
barentsen: Indent the column definitions to fit inside the indent block for `result`.

lightkurve/utils.py
```diff
@@ -293,3 +300,72 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
...
+        """
+
```
barentsen: The top of this function should probably contain:
```Python
supported_catalogs = ['KIC', 'EPIC', 'Gaia']
if catalog not in supported_catalogs:
    raise ValueError('catalog not one of {}'.format(supported_catalogs))
```

lightkurve/utils.py
```diff
@@ -293,3 +300,72 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
...
+        # identifies catalog
+        viz_id = ID[catalog]['vizier']
+
+        # Choose columns from Vizier
+        v = Vizier(catalog=[viz_id], columns=ID[catalog]['parameters'])
+        # query around centre with radius
+        result = v.query_region(coordinate, radius=radius, catalog=viz_id)
+        if len(result) == 0:
+            log.error('No sources found in queried region. Try another catalog.')
```
barentsen: If there are no results, the function should either return an empty Table (because it promises to return a Table) *or* it should raise an exception.

lightkurve/utils.py
```diff
@@ -293,3 +300,72 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
...
+
+        # Rename column names
+        new_pars = ['ID', 'RA', 'Dec', 'pmRA', 'pmDec', "mag"]
```
barentsen: We may want to consider using all lowercase columns names, just to simplify?

lightkurve/utils.py
```diff
@@ -3,8 +3,15 @@
 from astropy.visualization import (PercentileInterval, ImageNormalize,
                                    SqrtStretch, LogStretch, LinearStretch)
 from astropy.time import Time
+from astroquery.vizier import Vizier
+from astropy.table import Table
+import astropy.units as u
```
barentsen: The two imports above can be removed; Table and units aren't being used below.

lightkurve/tests/test_targetpixelfile.py
barentsen: Please add `@pytest.mark.remote_data` above `def test_find():` -- it tells the testing system to skip this test if no internet is available.

lightkurve/targetpixelfile.py
christinahedges: Yes, this allows us to do a `np.unique(pixel_radec)` a few lines later and find all the unique values rounded to ~0.1 arcseconds. This reduces the number of RA/Dec pairs to check from ~50,000 to ~500.

lightkurve/targetpixelfile.py
christinahedges: We are checking all the cadences for targets, including the motion. The `np.unique` above reduces the number of separations we need to calculate.

lightkurve/targetpixelfile.py
barentsen: Ah!  That means the rounding step can be dropped if we use `get_coordinates(cadence=0)` instead of `get_coordinates(cadence='all')` above.

lightkurve/targetpixelfile.py
barentsen: Ohhh I see, you are using all pixel coordinates of all cadences by design.  What is the `%timeit` i.e. speed cost?

lightkurve/targetpixelfile.py
christinahedges: We want to include the motion, incase there is a nearby source that is falling on/off the detector, which would mess up our PRF photometry. We use all the RA/Dec pairs from all cadences, reduce the precision of them to 0.1 arcseconds and then check to see if any targets are "near" to any of these pixels. Reducing the precision reduces the time needed to do the check.

lightkurve/targetpixelfile.py
christinahedges: We should only ever have to run this once for a given target, and I am guessing the cost of running this separation test is << than querying the catalog online. 

@AndyTza do you want to run a time on this? It's a good skill to practice. :)

AndyTza: 590 ms ± 13.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each) (__cadence='all'__)
208 ms ± 6.03 ms per loop (mean ± std. dev. of 7 runs, 1 loop each) (__cadence=0__)
1.5 s ± 35.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each) (__removed np.round()__)

AndyTza: @barentsen @christinahedges For the time being,  I've kept cadence='all'. Moving forward we can discuss if we want to include this step.

lightkurve/targetpixelfile.py
christinahedges: @AndyTza just pop the docstring into this format
```
result : astropy.table
        Astropy table with the following columns
        ID : Catalog ID from catalog.
        RAJ200: Right ascension [degrees]
```

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +193,87 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+        radius = (np.max(self.flux.shape[1:2]) * 4) / 60.0
+
+        # query around centre with radius
+        data = query_catalog(cent, radius=radius, catalog=catalog)
+
+        # Load ra & dec of all tpf pixels
+        pixels_ra, pixels_dec = self.get_coordinates(cadence='all')
+
+        # Reduce calculation for astroy separation
+        pixel_radec = np.asarray([pixels_ra.ravel(), pixels_dec.ravel()])
```
christinahedges: Geert has pointed out that this should only include pixels where the value is not nan. TPFs are always rectangular but have some missing nan values when the data is actually a slightly different shape.

lightkurve/targetpixelfile.py
christinahedges: I'm not sure how we feel about this. Either way, I don't think it should be the first cadence, as that one is often a little off source.

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +193,87 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+        # Make pixel pairs into SkyCoord
+        sky_pixel_pairs = SkyCoord(ra=pixel_pairs[0]*u.deg, dec=pixel_pairs[1]*u.deg, frame='icrs', unit=(u.deg, u.deg))
+        # Make pairs in sky_sources
+        sky_sources = SkyCoord(ra=data['ra'], dec=data['dec'], frame='icrs', unit=(u.deg, u.deg))
+
+        separation_mask = []
+        for i in range (len(data)):
+            s = sky_pixel_pairs.separation(sky_sources[i])  # Estimate separation
+            print (s.arcsec)
+            separation = np.any(s.arcmin <= dist_tolerance)
```
christinahedges: Let's change this so that it's not if the source is *ever* on *any* pixel, but if the source lands within a pixel at least 10% of the time.

lightkurve/utils.py
christinahedges: @AndyTza , same as above!

christinahedges:
Travis borked because of a MAST time out. It'll pass next time you run it.

lightkurve/targetpixelfile.py
AndyTza: @christinahedges do you think this is okay? Essentially what I'm doing here is that I'm finding for a given cadence where the flux is nan -- and applying that mask to the tpf.get_coordinates... That way it wouldn't look for the separation where the flux values are nan.

But I'm wondering if nans arise for different cadence numbers, this won't be an optimal solution.

lightkurve/targetpixelfile.py
gully: See my `TODO` comment about cadence below

gully:
If dist_tolerance is less than ~2 arcseconds, sources can evade selection by residing near the corners of a pixel--- such sources are not within 2 arcseconds of the center of any pixel!  You can resolve this bug by forcing a floor of sqrt(2)*2 arcseconds, or rethinking the way you define the boundary-- see "convex hull" algorithms.

lightkurve/targetpixelfile.py
```diff
@@ -190,6 +193,85 @@ def from_archive(target, cadence='long', quarter=None, month=None,
...
+
+        # Load pixel ra, dec with no nans
+        pixel_radec = np.asarray([pixels_ra[sum_nans != 0].ravel(), pixels_dec[sum_nans != 0].ravel()])
+
+        # Make pixel pairs into SkyCoord
+        sky_pixel_pairs = SkyCoord(ra=pixel_radec[0], dec=pixel_radec[1], frame='icrs', unit=(u.deg, u.deg))
+        # Make pairs in sky_sources
+        sky_sources = SkyCoord(ra=data['ra'], dec=data['dec'], frame='icrs', unit=(u.deg, u.deg))
+
+        separation_mask = []
```
christinahedges: Rather than have a list that you make into an array in line 268 you could use

```
separation_mask = np.zeros(len(data), dtype=bool)
for idx in range (len(data)):
            s = sky_pixel_pairs.separation(sky_sources[i])  # Estimate separation
            separation = np.any(s.arcsec <= dist_tolerance)
            separation_mask[idx] = separation
```

Note I've renamed `i` to `idx` because it's slightly easier to understand what `idx` is, and it's easier to read later when we have to debug.

lightkurve/utils.py
```diff
@@ -293,3 +299,67 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
...
+        result : astropy.table
+            Astropy table with the following columns
+            ID : Catalog ID from catalog.
+            RAJ200: Right ascension [degrees]
+            DEJ2000: Declination [degrees]
+            pmRA: Proper motion for right ascension [mas/year]
+            pmDEC: Proper motion for declination [mas/year]
+            Kpmag: Magnitude in Kepler band [mag]
+        """
+        supported_catalogs = ['KIC', 'EPIC', 'Gaia']
```
christinahedges: Very minor nitpick, rather than having two variables that store the names of the catalog we could have just one (`ID`) which you could print using...

`ValueError('catalog not one of {}'.format([key for key in ID.keys()]))`

You'd just need to move `ID` above this statement.

barentsen:
Feedback after Andy and Geert met:
- [ ] let's correct ra & dec to be at the epoch of the Kepler observation  (i.e. correct to `np.median(tpf.astropy_time.byear)`)
- [ ] make the epoch explicit by adding a column `epoch` to the output of `get_sources` with the median TPF epoch;
- [ ] add `column`, `row`, and `flux` as columns to the table returned by `get_sources()`;
- [ ] add `err_ra`, `err_dec`, `err_col`, `err_row` to the table returned by `get_sources()`;
- [ ] consider adding `catalog=xxx` to `tpf.plot()`, which would indicate catalog positions on the plot (note: when plotting, make sure to apply POSCORR* to the pixel coordinates);
- [ ] add a `tpf.get_star_priors(catalog=Table or string)` method;

AndyTza:
@barentsen @christinahedges I've made some final changes to the get_sources after brainstorming with @barentsen. 

The function now will be able to return to the user RA and DEC for the data epoch and also column and row positions that will be useful for the PRF.

lightkurve/targetpixelfile.py
```diff
@@ -41,6 +44,110 @@ def __init__(self, path, quality_bitmask='default', targetid=None, **kwargs):
         self.quality_mask = self._quality_mask(quality_bitmask)
         self.targetid = targetid
 
+
+    def get_sources(self, catalog=None, magnitude_limit=18, dist_tolerance=3):
```
christinahedges: dist_tolerance is not a very good name. Perhaps edge_tolerance would be more descriptive?

lightkurve/targetpixelfile.py
```diff
@@ -41,6 +44,110 @@ def __init__(self, path, quality_bitmask='default', targetid=None, **kwargs):
...
+            of the target pixel file.
+
+        Returns
+        -------
+        result : astropy.table
+            Astropy table with the following columns
+            ID : Catalog ID from catalog.
+            RA: Right ascension [degrees] (KIC & EPIC)
+            DEC: Declination [degrees]    (KIC & EPIC)
+            RA: Right ascension [degrees] (Gaia)
```
christinahedges: Why are there two kinds of RA Dec in the docstring? Just have one.

lightkurve/targetpixelfile.py
```diff
@@ -41,6 +44,110 @@ def __init__(self, path, quality_bitmask='default', targetid=None, **kwargs):
...
+        dist_tolerance: float
+            Maximum distance (in arcseconds) a source may be separated from the edge
+            of the target pixel file.
+
+        Returns
+        -------
+        result : astropy.table
+            Astropy table with the following columns
+            ID : Catalog ID from catalog.
+            RA: Right ascension [degrees] (KIC & EPIC)
```
christinahedges: If you have changed the RA/Dec from the catalog to the current epoch you MUST state that somewhere in the docstring. Perhaps here in the RA/Dec

lightkurve/targetpixelfile.py
```diff
@@ -41,6 +44,110 @@ def __init__(self, path, quality_bitmask='default', targetid=None, **kwargs):
...
+            if self.mission == 'Kepler':
+                catalog = 'KIC'
+                catalog_epoch = 2000.0
+            elif self.mission == 'K2':
+                catalog = 'EPIC'
+                catalog_epoch = 2000.0
+            else:
+                raise ValueError('Please provide a catalog. Catalog must be KIC, EPIC or Gaia.')
+        else:
+            catalog_epoch = 2015.5
```
christinahedges: This is not great if we want to add more catalogs (e.g. 2MASS) but there is pretty limited scope for doing that, so I'm happy.

lightkurve/targetpixelfile.py
```diff
@@ -41,6 +44,110 @@ def __init__(self, path, quality_bitmask='default', targetid=None, **kwargs):
...
+
+        if dist_tolerance < 3:
+            log.warning('Distance tolerance is less than 3 pixels. Sources near the edge will likely be missed.')
```
christinahedges: Can we reword this message?
'Edge tolerance is less than 3 pixels. Bright sources near the edge of the Target Pixel File may not be included.'

lightkurve/targetpixelfile.py
```diff
@@ -41,6 +44,110 @@ def __init__(self, path, quality_bitmask='default', targetid=None, **kwargs):
...
+
+        #Skycoord the centre of target
+        cent = SkyCoord(ra=self.ra, dec=self.dec, frame='icrs', unit=(u.deg, u.deg))
+
+        #Find the size of the TPF
+        radius = ((np.max(self.flux.shape[1:2]) * 4) + dist_tolerance)/ 60.0  # arcmin
+
+        # query around centre with radius and catalog
```
christinahedges: That British spelling is winning you points.

lightkurve/targetpixelfile.py
```diff
@@ -41,6 +44,110 @@ def __init__(self, path, quality_bitmask='default', targetid=None, **kwargs):
...
+
+        # Find where nans are in cadence
+        tpf_mask = np.any(np.isfinite(self.flux), axis=0)
+
+        # Coordinates of sky sources and tpf pixels
+        pixels_ra, pixels_dec = self.get_coordinates(cadence=int(len(self.cadenceno)/2))
+        pixels_radec = np.asarray([pixels_ra[tpf_mask].ravel(), pixels_dec[tpf_mask].ravel()])
+        sky_pixel_pairs = SkyCoord(ra=pixels_radec[0], dec=pixels_radec[1], frame='icrs', unit=(u.deg, u.deg))
+        sky_sources = SkyCoord(ra=tbl['ra'], dec=tbl['dec'], frame='icrs', unit=(u.deg, u.deg))
+
```
christinahedges: Add a comment here to state clearly what you're doing.

lightkurve/targetpixelfile.py
```diff
@@ -41,6 +44,110 @@ def __init__(self, path, quality_bitmask='default', targetid=None, **kwargs):
...
+            separation_mask[idx] = separation
+        sep_mask = np.array(separation_mask, dtype=bool)
+
+        # constrain with magnitude limit
+        sep_mask &= tbl['mag'] < magnitude_limit
+
+        # After applying filters we can now correct the posistions
+        source_list = tbl[sep_mask]
+
+        # Transform coordinates to median epoch of data
```
christinahedges: As above, you MUST state in the docstring that you're messing with the catalog.

lightkurve/targetpixelfile.py
```diff
@@ -41,6 +44,110 @@ def __init__(self, path, quality_bitmask='default', targetid=None, **kwargs):
...
+        tpf_epoch = np.median(self.astropy_time.byear)
+        has_pm = ~(source_list['pmra'].mask | source_list['pmdec'].mask)
+        source_list['ra'][has_pm] += (tpf_epoch - catalog_epoch) * (source_list['pmra'][has_pm] * (2.778e-7)) # mas to degree
```
christinahedges: Astropy already has a nice way of dealing with units, are you sure the units of these catalogs are always in mas/year? If yes I have no problem, but in future it might be worth using the inbuilt astropy units feature rather than hard coding a conversion.

lightkurve/targetpixelfile.py
```diff
@@ -61,6 +168,115 @@ def header(self, ext=0):
         """Returns the header for a given extension."""
         return self.hdu[ext].header
 
+    def get_prf_model(self):
```
christinahedges: `wcs` and `get_prf_model` are nothing to do with this feature. Why have these moved? Is this a problem with commit history?

lightkurve/tests/test_targetpixelfile.py
```diff
@@ -59,6 +62,18 @@ def test_tpf_plot():
         with pytest.raises(ValueError):
             tpf.plot(scale="blabla")
 
+@pytest.mark.remote_data
+def test_get_sources():
+    "Minimum sanity chech of the table it returns contains the coordinates of the tpf"
```
christinahedges: chech

christinahedges: what is 'it'?

lightkurve/tests/test_targetpixelfile.py
```diff
@@ -59,6 +62,18 @@ def test_tpf_plot():
...
+    tpf = KeplerTargetPixelFile(TABBY_TPF)
+    dat = tpf.get_sources()
+    # Smallest number the query should find is at least 1
+    assert (len(dat['ra']) >= 1)
+    assert (len(dat['id'])) >= 1
+    assert np.any([d == tpf.keplerid for d in dat['id']])
```
christinahedges: Add a comment for what this assert does

lightkurve/tests/test_targetpixelfile.py
```diff
@@ -59,6 +62,18 @@ def test_tpf_plot():
...
+    c = SkyCoord(60, 10, unit=(u.deg, u.deg))
+    result = query_catalog(c, 'Gaia', radius=1)
+    assert len(result) == 3
```
christinahedges: Add a comment for what this assert does

lightkurve/utils.py
```diff
@@ -321,3 +327,78 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
...
+            DEJ2000: Declination [degrees]    (KIC & EPIC)
+            RAJ2015.5: Right ascension [degrees] (Gaia)
+            DEJ2015.5: Declination [degrees]     (Gaia)
+            pmRA: Proper motion for right ascension [mas/year]
+            pmDEC: Proper motion for declination [mas/year]
+            Kpmag: Magnitude in Kepler band [mag]
+        """
+
+        if catalog is "Gaia":
+            log.warn('Gaia RAs and Decs are at epoch 2015.5. These RA/Decs have not been corrected.')
```
christinahedges: Ahh...but they have.

lightkurve/targetpixelfile.py
AndyTza: Yep! I merged the updated master branch with find_stars branch.

lightkurve/utils.py
AndyTza: I left query_catalogue to return the RA and DEC of the epoch of each catalogue without correcting for the proper motions, the correction to be done by get_sources!

gully:
FYI-- the tutorial breaks for me when I attempt `tpf.get_sources()`:

``` python
<ipython-input-3-ce607406b347> in <module>()
----> 1 sources = tpf.get_sources()

~/GitHub/lightkurve/lightkurve/targetpixelfile.py in get_sources(self, catalog, magnitude_limit, edge_tolerance)
    142         epoch = Column([tpf_epoch] * len(column), name='epoch', unit=(u.year))
    143         source_list.add_columns([source_flux, column, row, epoch])
--> 144         if catalog == "Gaia":
    145             source_list.rename_column('mag', 'Gmag')
    146         else:

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

PR #164, opened 2018-06-15
[WIP] Add a LightCurveCollection/TargetPixelFileCollection class

christinahedges:
We need a class that collects together many lightcurves or TPFs. These collections would be returned from 

* PSF photometry
* Cone search

Collections will have methods such as 

* Plot
* PCA: Create principle components of a group of lightcurves

I've started building this class here


christinahedges:
Closing because #173  is better

Issue #167, opened 2018-06-21
tpf.pos_corr becomes zero after calling tpf.get_coordinates()


AndyTza:
#### Problem description
When the user calls the function tpf.get_coordinates(), tpf.pos_corr is 0 for all values. After re-running the tpf 

#### Example

```python
from lightkurve import KeplerTargetPixelFile
tpf = KeplerTargetPixelFile.from_archive('kepler-10', quarter=10, quality_bitmask=2096639)
print (tpf.pos_corr1)
>> [ 0.0187123   0.01840722  0.01804379 ... -0.00236374 -0.00207365
 -0.00235928]
tpf.get_coordinates(cadence='all')
print (tpf.pos_corr1)
>> [0. 0. 0. ... 0. 0. 0.]
```

#### Expected behavior
tpf.pos_corr should return non-zero values.

christinahedges:
Fixed in #168

PR #168, opened 2018-06-21
Fixed POSCORR overwrite issue

christinahedges:
@AndyTza pointed out that the `tpf.get_coordinates()` method has been overwriting the POSCORR values with zeros in #167. This is a quick fix to correct that issue.

(poscorrs needed to be deep copied instead of pointers.)


barentsen:
This is a good fix. Thanks @christinahedges.

PR #169, opened 2018-06-21
Fixing TPFFactory to take in WCS

Emmavt:
Adding in WCS coordinates so that they are not pulled from the template file.


lightkurve/targetpixelfile.py
```diff
@@ -816,6 +818,70 @@ def to_fits(self, output_fn=None, overwrite=False):
             output_fn = "{}-targ.fits".format(self.keplerid)
         self.hdu.writeto(output_fn, overwrite=overwrite, checksum=True)
 
+    def get_hdu(image, extension):
```
christinahedges: This can be `_get_hdu()` because we don't want users to see it.

lightkurve/targetpixelfile.py
```diff
@@ -816,6 +818,70 @@ def to_fits(self, output_fn=None, overwrite=False):
...
+        """
+        if isinstance(image, fits.ImageHDU):
+            hdu = image
+        elif isinstance(image, fits.HDUList):
+            hdu = image[extension]
+        else:
+            hdu = fits.open(image)[extension]
+        return hdu
+
+    def get_WCS(mid_img,hdu,position,extension): #name of this?? add doc string
```
christinahedges: Same here, except get_WCS is quite close to get_coordinates and WCS which are two other functions. Let's think of another name.

christinahedges: Spaces after commas

christinahedges: No need for that comment

lightkurve/targetpixelfile.py
```diff
@@ -816,6 +818,70 @@ def to_fits(self, output_fn=None, overwrite=False):
...
+        WCS for an image relative to the reference image, defined as the middle
+        cadence of the FITS file.
+
+        Attributes
+        ----------
+        hdu : astropy.PrimaryHDU
+            PrimaryHDU of the image being looked at
+        mid_img : fits.ImageHDU
+            middle layer of the FITS file used as a reference image
+        Returns
```
christinahedges: Space above "Returns"

lightkurve/targetpixelfile.py
```diff
@@ -816,6 +818,70 @@ def to_fits(self, output_fn=None, overwrite=False):
...
+        """Calculates the position correction values (both in row and column) and
+        WCS for an image relative to the reference image, defined as the middle
+        cadence of the FITS file.
+
+        Attributes
+        ----------
+        hdu : astropy.PrimaryHDU
+            PrimaryHDU of the image being looked at
+        mid_img : fits.ImageHDU
+            middle layer of the FITS file used as a reference image
```
christinahedges: Let's call it 'reference image'

lightkurve/targetpixelfile.py
```diff
@@ -816,6 +818,70 @@ def to_fits(self, output_fn=None, overwrite=False):
...
+        wcs1 :  astropy.wcs.WCS
+            WCS object for this specific hdu
+        """
+        # Calculate reference WCS coords from the middle image to use later to
+        # do pos_corr calcs. Get the middle pixel column and row.
+        mid_hdu = KeplerTargetPixelFile.get_hdu(mid_img, extension)
+        wcs = WCS(mid_hdu)
+        pixelpos = skycoord_to_pixel(position, wcs=wcs)
+        column = int(np.round(pixelpos[0]))
+        row = int(np.round(pixelpos[1]))
```
christinahedges: You can probably reduce this to one line, something like:
```
col, row = int(skycoord_to_pixel(position, wcs=wcs))
```

lightkurve/targetpixelfile.py
```diff
@@ -816,6 +818,70 @@ def to_fits(self, output_fn=None, overwrite=False):
...
+
+        # Calculate the ra and dec of the center pixel of the middle image
+        skyposctr = pixel_to_skycoord(column,row,wcs=wcs)
```
christinahedges: Spaces after ,'s

lightkurve/targetpixelfile.py
```diff
@@ -816,6 +818,70 @@ def to_fits(self, output_fn=None, overwrite=False):
...
+        mid_img : fits.ImageHDU
+            middle layer of the FITS file used as a reference image
+        Returns
+        -------
+        poscorr1 :  float32
+            Position correction in the vertical direction (column)
+        poscorr2 :  float32
+            Position correction in the horizontal direction (row)
+        wcs1 :  astropy.wcs.WCS
+            WCS object for this specific hdu
```
christinahedges: You return more than these three things at the bottom.

lightkurve/targetpixelfile.py
```diff
@@ -856,21 +922,30 @@ def from_fits_images(images, position=None, size=(10, 10), extension=None,
...
+            img_data = KeplerTargetPixelFile.get_WCS(mid_img,hdu,position,extension)
+            # Add calculated pos corr shifts to the keywords dictionary
+            hdu.header['POSCORR1'] = img_data[1]
+            hdu.header['POSCORR2'] = img_data[2]
+            # Information on the center of the TPF file to be created, used to create the second header extension
+            factory.keywords['MID_COL'] = img_data[3]
+            factory.keywords['MID_ROW'] = img_data[4]
+            factory.keywords['RA_CTR'] = img_data[5]
+            factory.keywords['DEC_CTR'] = img_data[6]
+
```
christinahedges: Let's not make new header keywords if we don't have to and if people won't use them.

lightkurve/targetpixelfile.py
```diff
@@ -1021,15 +1098,38 @@ def _make_target_extension(self):
...
+        cd21 = self.keywords['CD2_1']
+        cd22 = self.keywords['CD2_2']
+
+        hdu.header['1CRVL5'] = self.keywords['RA_CTR']
+        hdu.header['2CRVL5'] = self.keywords['DEC_CTR']
+        cdelt = hdu.header['2CDLT5']
+        hdu.header['11PC5'] = -1.0*cd11/cdelt
+        hdu.header['12PC5'] = -1.0*cd12/cdelt
+        hdu.header['21PC5'] = cd21/cdelt
+        hdu.header['22PC5'] = cd22/cdelt
```
christinahedges: Let's move the calculation of the WCS into the from_fits_images() function

lightkurve/targetpixelfile.py
amcody: Actually, we need the round in there, because otherwise it just takes the integer value, which means, e.g., 5.6 --> 5 instead of 6

Emmavt:
@amcody, @christinahedges This is now finished, so if you could please review it that would be great!
 
Possible further work that Ann Marie and I discussed was to adapt the Cutout2D line of code (line949) in from_fits_images to account for the possibility of images without WCSs being passed in, as currently the code assumes the files have WCS.

amcody:
I think the channel keyword is missing when you make TPFs with this? We need it for psf photometry.

amcody:
There is a now a if wcs=None: in the code which should be wcs==None:
Not sure who did that, but can you fix it?

amcody:
This is looking good to me now and it is working locally on M35 superstamp images. Anybody else have a comment, or can we merge? @barentsen?

lightkurve/targetpixelfile.py
```diff
@@ -831,6 +833,84 @@ def get_bkg_lightcurve(self, aperture_mask=None):
                           flux=np.nansum(self.flux_bkg[:, aperture_mask], axis=1),
                           flux_err=self.flux_bkg_err)
 
+
+    def _get_hdu(image, extension):
```
barentsen: This method appears inside the `KeplerTargetPixelFile` class, but it does not actually seem to operate on the class (e.g. it does not accept or use the `self` parameter).  This suggests it should be moved outside of the class to be a standalone function instead of a class method.

The only exception to the above rule is for so-called "static methods" which return a new instance of `KeplerTargetPixelFile` (like `from_archive`).

lightkurve/targetpixelfile.py
```diff
@@ -831,6 +833,84 @@ def get_bkg_lightcurve(self, aperture_mask=None):
...
+        image : fits.ImageHDU or fits.HDUList object
+            FITS filename path or ImageHDU object to get the data from.
+        extension: integer
+            Number of the file extension containing data, depends on the file
+            type and source.
+
+        Returns
+        -------
+        hdu : astropy.PrimaryHDU
+            PrimaryHDU of the image inputted.
```
barentsen: It looks like this function can either return `PrimaryHDU` or `ImageHDU`, the docstring should clarify that.

lightkurve/targetpixelfile.py
```diff
@@ -831,6 +833,84 @@ def get_bkg_lightcurve(self, aperture_mask=None):
                           flux=np.nansum(self.flux_bkg[:, aperture_mask], axis=1),
                           flux_err=self.flux_bkg_err)
 
+
+    def _get_hdu(image, extension):
+        """Checks if an image is a single FITS image, or a FITS file,
+        and returns the appropriate PrimaryHDU.
+
```
barentsen: Please explain in the docstring when someone may actually want to use this function?  It's not 100% clear to me that we need this to be a function, as opposed to being a `if-else` statement when you retrieve the WCS further below?

We want to make sure all functions inside `KeplerTargetPixelFile` have a very clear purpose, otherwise users may be confused and our code will be harder to maintain.

lightkurve/targetpixelfile.py
```diff
@@ -831,6 +833,84 @@ def get_bkg_lightcurve(self, aperture_mask=None):
...
+        """
+        if isinstance(image, fits.ImageHDU):
+            hdu = image
+        elif isinstance(image, fits.HDUList):
+            hdu = image[extension]
+        else:
+            hdu = fits.open(image)[extension]
+        return hdu
+
+    def _get_pos_corrs(mid_img, hdu, position, extension):
```
barentsen: This method should take `self` or otherwise be turned into a standalone function.  (As a general rule, all methods in a class should operate on `self`, unless they create a new `KeplerTargetPixelFile` like `from_archive` does.

lightkurve/targetpixelfile.py
Emmavt: Hi Geert, thank you for your feedback! I agree that this function could be misleading/it is unclear what the function's purpose is, so I will move the if-else statements into the from_fits and _get_pos_corrs functions, to minimise confusion.

lightkurve/targetpixelfile.py
Emmavt: I will move this function outside of KeplerTargetPixelFile.

lightkurve/targetpixelfile.py
```diff
@@ -833,84 +893,6 @@ def get_bkg_lightcurve(self, aperture_mask=None):
...
-        ractr = skyposctr.ra.deg
-        decctr = skyposctr.dec.deg
-
-        # Get positional shift of the image compared to the reference file
-        wcs1 = WCS(hdu)
-        pixelpos1 = skycoord_to_pixel(position, wcs=wcs1)
-        poscorr1 = pixelpos1[0] - pixelpos[0]
-        poscorr2 = pixelpos1[1] - pixelpos[1]
-
-        return [poscorr1, poscorr2, column, row, ractr, decctr]
```
amcody: Two questions/issues here: 
Where in the code are ractr and decctr actually used? I am not seeing this. And if they are the coordinates of the center of the middle fits file, then there is no  need to have them in this loop, because they are being calculated over and over again.

lightkurve/targetpixelfile.py
```diff
@@ -861,33 +923,73 @@ def from_fits_images(images, position=None, size=(10, 10), extension=None,
...
+
+        # Override the defaults where necessary, and pass into the header generating functions
+        ext_info = {}
+        ext_info['TFORM4'] = '{}J'.format(size[0]*size[1])
+        ext_info['TDIM4'] = '({},{})'.format(size[0],size[1])
+
+        cd11 = factory.keywords['CD1_1']
+        cd12 = factory.keywords['CD1_2']
+        cd21 = factory.keywords['CD2_1']
+        cd22 = factory.keywords['CD2_2']
```
amcody: It looks to me here like you may be takign the cd* values from the last fits image and not the middle one? I would find a way to just read in the middle image's info- maybe another stand-alone function. Because you need to extract both the CD* keywords, as well as figure out ractr, dectr. This should only require one call.

Emmavt:
@barentsen This is now updated with all the changes you mentioned, and @amcody 's comments. These changes did mean that I restructured my code such that the two helper functions _get_hdu and _get_pos_corrs were merged with from_fits_images. This does reduce unnecessary recalculation of certain values, but it makes me concerned about how from_fits_images is maybe not as general now, which is why Christina and I first decided to introduce the helper functions. If you have any thoughts on this, I'd love to hear them!

lightkurve/targetpixelfile.py
```diff
@@ -1012,23 +1014,82 @@ def from_fits_images(images, position=None, size=(10, 10), extension=None,
...
+            mid_hdu = fits.open(mid_img)[extension]
+        # Calculate reference WCS coords from the middle image. Get the middle pixel column and row.
+        wcs = WCS(mid_hdu)
+        pixelpos = skycoord_to_pixel(position, wcs=wcs)
+        column, row = int(np.round(pixelpos[0])), int(np.round(pixelpos[1]))
+
+        # Calculate the ra and dec of the center pixel of the middle image
+        skyposctr = pixel_to_skycoord(column, row, wcs=wcs)
+        ractr = skyposctr.ra.deg
+        decctr = skyposctr.dec.deg
```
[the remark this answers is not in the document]
amcody: Ah, good point- this was originally put in by me. The reason was that when I cut out the tpf, the object of interest was not necessarily in the middle of the center pixel. We need the RA and Dec of this location, and it can differ from RA_OBJ/DEC_OBJ by a couple arcsec, due to the large Kepler pixel size. Does that make sense? Now of course we are using the astropy cutout function, and I admittedly did not check whether puts the star in the middle of the center pixel and regrids the coordinates. Unless it does that, then we still need these few lines to get the exact center of the TPF for the WCS solution.

amcody:
Also, does anyone know why one of the tests is now failing? It says something about the code taking too long to build.

barentsen:
@Emmavt Thanks Emma!  I do think it's a good idea to put the superstamp-specific aspects in a helper function, just not within the `KeplerTargetPixelFile` class, unless it is a direct operation on a specific target pixel file (which the function wasn't).  It can be a function in the module though (i.e. the same python file).

(also see comment above re the sky-to-pixel-to-sky conversion which confuses me a bit)

lightkurve/targetpixelfile.py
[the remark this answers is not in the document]
barentsen: Hmmm, does this really matter?  As long as we have the (ra, dec) for a specific pixel coordinate near the center, we can do the cutout and a wcs -- i.e. can we get rid of it?  Less code is better =)

...or did I misunderstand?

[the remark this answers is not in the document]
amcody: Well, my understanding is that we need a reference point from which to define the WCS in both world and image coordinates. I was selecting that point to be the center of the TPF, which is not the same as the object location. Thus this bit of code calculates the RA and Dec of the center, so that we can put it in the header.

lightkurve/targetpixelfile.py
```diff
@@ -1012,23 +1014,82 @@ def from_fits_images(images, position=None, size=(10, 10), extension=None,
...
+            # Add calculated pos corr shifts to the keywords dictionary
+            hdu.header['POSCORR1'] = pos_corr1
+            hdu.header['POSCORR2'] = pos_corr2
+
             if position is None:
                 cutout = hdu
             else:
                 cutout = Cutout2D(hdu.data, position, wcs=WCS(hdu.header),
                                   size=size, mode='partial')
-            factory.add_cadence(frameno=idx, flux=cutout.data, header=hdu.header)
```
amcody: I am confused here as to why we need to get a wcs for the cutout. It looks to me like we aren't using it, as the header info passed to the factory is from the orginal file, not the new one. Thoughts?

lightkurve/targetpixelfile.py
```diff
@@ -1002,6 +1003,7 @@ def from_fits_images(images, position=None, size=(10, 10), extension=None,
         tpf : KeplerTargetPixelFile
             A new Target Pixel File assembled from the images.
         """
+        from lightkurve.targetpixelfile import _get_pos_corrs
```
amcody: This needs to be removed as the function is no longer there. I think that's why the build is crashing.

lightkurve/targetpixelfile.py
```diff
@@ -1012,23 +1014,82 @@ def from_fits_images(images, position=None, size=(10, 10), extension=None,
...
-                factory.keywords = hdu.header
+
+            # Get positional shift of the image compared to the reference file
+            wcs1 = WCS(hdu)
+            pixelpos1 = skycoord_to_pixel(position, wcs=wcs1)
+            poscorr1 = pixelpos1[0] - pixelpos[0]
+            poscorr2 = pixelpos1[1] - pixelpos[1]
+            # Add calculated pos corr shifts to the keywords dictionary
+            hdu.header['POSCORR1'] = pos_corr1
+            hdu.header['POSCORR2'] = pos_corr2
```
amcody: Inconsistency here between poscorr and pos_corr.

lightkurve/targetpixelfile.py
```diff
@@ -1012,23 +1014,82 @@ def from_fits_images(images, position=None, size=(10, 10), extension=None,
...
+        ext_info['22PC5'] = cd22/0.001102672560321
+        ext_info['1CRVL5'] = ractr
+        ext_info['2CRVL5'] = decctr
+
+        for m in [4, 5, 6, 7, 8, 9]:
+            if m > 4:
+                ext_info["TFORM{}".format(m)] = '{}E'.format(size[0]*size[1])
+            ext_info['TDIM{}'.format(m)] = '({},{})'.format(size[0],size[1])
+            ext_info['1CRV{}P'.format(m)] = col - int((tpfsize-1)/2) + factory.keywords['CRVAL1P']
+            ext_info['2CRV{}P'.format(m)] = row - int((tpfsize-1)/2) + factory.keywords['CRVAL2P']
```
amcody: row, col used to be defined but are no longer.

mirca:
@amcody I wouldn't worry about the failing tests reported by AppVeyor. I'm pretty sure the failures are unrelated to this PR.

barentsen:
I think we should be able to simplify this code further.  The documentation on astropy's `Cutout2D()` says that it returns a WCS that is adjusted for the cutout, so I'm not sure I understand why we have to meddle with the WCS keywords ourselves?

Is it correct to say that we need to select one image to use as the reference WCS for the TPF, and use the WCS for all other images to compute the `poscorr` values?  (That's not quite what the code is doing right now?)

amcody:
@barentsen that's exactly what I was thinking. I believe the reason why I was originally calculating the WCS keyword values myself was that I was under the impression we should have the TPF WCS info in the PC system, not the CD system (as are the superstamp images). But if it's fine to have the spatial information stored in CD keywords, then we should be able to eliminate a lot of code by just having cutout2D transfer the WCS from the middle image. Sound good?

amcody:
Okay, we're working on these changes now. We do still need to know the detector coordinates of the corner of the TPF, and it looks like Cutout2D does not calculate this. So we have to leave a few lines of code in to do that. It's a bit complicated, as we need to know the exact center of the cutout in order to find the corners. And that depends on how Cutout2D is choosing the middle pixel (is it simply rounding the object's x,y position?).

barentsen:
Why do you need to know the corner coordinates?

amcody:
Well, we figured out the behavior of Cutout2D just now.
We need to know the corner because TPF.column,row reports that (i.e., it goes into the the 1CRV5P and 2CRV5P keywords of the TPF header).

Emmavt:
Hi all, I deleted some of the excess code to do with WCS header entries, as those should be filled by the Cutout2D function. I left in the entries to do with finding the corner coords. I also added in an if/elif statement to deal with the tpf possibly having even sides. The code still assumes a square tpf.

amcody:
Whoops, the WCS wasn't getting passed into the new TPF. Emma is making more fixes... stay tuned.

Emmavt:
@barentsen This should be finished now, I fixed a few bugs in the cutout2d line.

PR #173, opened 2018-06-26
[WIP] Created a LightCurveCollections and TargetPixelFileCollections class

johnnyzhang295:
This class collects together multiple lightcurves or tpfs and allows operations over all of them such as:
- Plot
- PCA (TODO)

The TPFCollections hasn't properly implemented
- Plot
- PCA

To create a Collections class, you need to provide the lightcurves/targetpixelfiles (targets) in a tuple object. Within the Collections class, specific targets can be referenced through their keplerids. These can be used to retrieve targets, add new targets to your collection, etc. Or you can treat the Collections class as one object (the main use) and it behaves similarly to the single lightcurve or TPF class.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -21,7 +21,7 @@ from astropy.io import fits
 from astropy.time import Time
 
 
-from .utils import running_mean, bkjd_to_astropy_time
+from .utils import running_mean, bkjd_to_astropy_time, btjd_to_astropy_time
 from . import PACKAGEDIR
 
 __all__ = ['LightCurve', 'KeplerLightCurve', 'TessLightCurve',
@@ -143,6 +143,8 @@ class LightCurve(object):
         # `TimeFormat` class for BKJD.
         if self.time_format == 'bkjd':
             return bkjd_to_astropy_time(self.time)
+        elif self.time_format == 'btjd':  # TESS
+            return btjd_to_astropy_time(self.time)
         return Time(self.time, format=self.time_format, scale=self.time_scale)
 
     def properties(self):
@@ -773,6 +775,12 @@ class KeplerLightCurve(LightCurve):
         Data flux for every time point
     flux_err : array-like
         Uncertainty on each flux data point
+    time_format : str
+        String specifying how an instant of time is represented,
+        e.g. 'bkjd' or 'jd'.
+    time_scale : str
+        String which specifies how the time is measured,
+        e.g. tdb', 'tt', 'ut1', or 'utc'.
     centroid_col : array-like
         Centroid column coordinates as a function of time
     centroid_row : array-like
@@ -798,7 +806,8 @@ class KeplerLightCurve(LightCurve):
                  centroid_col=None, centroid_row=None, quality=None, quality_bitmask=None,
                  channel=None, campaign=None, quarter=None, mission=None,
                  cadenceno=None, keplerid=None, ra=None, dec=None):
-        super(KeplerLightCurve, self).__init__(time=time, flux=flux, flux_err=flux_err, time_format=time_format, time_scale=time_scale)
+        super(KeplerLightCurve, self).__init__(time=time, flux=flux, flux_err=flux_err,
+                                               time_format=time_format, time_scale=time_scale)
         self.centroid_col = self._validate_array(centroid_col, name='centroid_col')
         self.centroid_row = self._validate_array(centroid_row, name='centroid_row')
         self.quality = self._validate_array(quality, name='quality')
@@ -932,6 +941,12 @@ class TessLightCurve(LightCurve):
         Data flux for every time point
     flux_err : array-like
         Uncertainty on each flux data point
+    time_format : str
+        String specifying how an instant of time is represented,
+        e.g. 'bkjd' or 'jd'.
+    time_scale : str
+        String which specifies how the time is measured,
+        e.g. tdb', 'tt', 'ut1', or 'utc'.
     centroid_col, centroid_row : array-like, array-like
         Centroid column and row coordinates as a function of time
     quality : array-like
@@ -943,22 +958,30 @@ class TessLightCurve(LightCurve):
     ticid : int
         Tess Input Catalog ID number
     """
-    def __init__(self, time, flux=None, flux_err=None, centroid_col=None,
-                 centroid_row=None, quality=None, quality_bitmask=None,
-                 cadenceno=None, ticid=None):
-        super(TessLightCurve, self).__init__(time, flux, flux_err)
+    def __init__(self, time, flux=None, flux_err=None, time_format=None, time_scale=None,
+                 centroid_col=None, centroid_row=None, quality=None, quality_bitmask=None,
+                 cadenceno=None, sector=None, camera=None, ccd=None,
+                 ticid=None, ra=None, dec=None):
+        super(TessLightCurve, self).__init__(time=time, flux=flux, flux_err=flux_err,
+                                             time_format=time_format, time_scale=time_scale)
         self.centroid_col = self._validate_array(centroid_col, name='centroid_col')
         self.centroid_row = self._validate_array(centroid_row, name='centroid_row')
         self.quality = self._validate_array(quality, name='quality')
         self.cadenceno = self._validate_array(cadenceno)
         self.quality_bitmask = quality_bitmask
         self.mission = "TESS"
+        self.sector = sector
+        self.camera = camera
+        self.ccd = ccd
         self.ticid = ticid
+        self.ra = ra
+        self.dec = dec
 
     def __getitem__(self, key):
         lc = super(TessLightCurve, self).__getitem__(key)
         # Compared to `LightCurve`, we need to slice a few additional arrays:
         lc.quality = self.quality[key]
+        lc.cadenceno = self.cadenceno[key]
         lc.centroid_col = self.centroid_col[key]
         lc.centroid_row = self.centroid_row[key]
         return lc
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -15,9 +15,9 @@ from tqdm import tqdm
 
 
 from . import PACKAGEDIR
-from .lightcurve import KeplerLightCurve, LightCurve
+from .lightcurve import KeplerLightCurve, TessLightCurve, LightCurve
 from .prf import SimpleKeplerPRF
-from .utils import KeplerQualityFlags, plot_image, bkjd_to_astropy_time
+from .utils import KeplerQualityFlags, plot_image, bkjd_to_astropy_time, btjd_to_astropy_time
 from .mast import download_kepler_products
 
 __all__ = ['KeplerTargetPixelFile', 'TargetPixelFile',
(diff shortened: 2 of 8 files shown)


johnnyzhang295:
This is in collaboration with #164.

lightkurve/lightcurvecollection.py
```diff
@@ -0,0 +1,87 @@
+from __future__ import division, print_function
```
christinahedges: Let's put the collections classes back in lightcurve.py and targetpixelfile.py. They are only ~50 lines long they don't need their own files.

lightkurve/lightcurvecollection.py
```diff
@@ -0,0 +1,87 @@
...
+
+log = logging.getLogger(__name__)
+
+class LightCurveCollection(Sequence):
+    """
+    Collects multiple LightCurve objects together with helpful functions.
+    """
+
+    def __init__(self, lcs):
+        lightcurves = np.asarray(lcs)
```
christinahedges: There needs to be error catching here so that the class breaks if it is not passed a `LightCurve` class

lightkurve/lightcurvecollection.py
```diff
@@ -0,0 +1,87 @@
...
+from astropy.time import Time
+
+
+__all__ = ['LightCurve', 'KeplerLightCurve', 'TessLightCurve',
+           'iterative_box_period_search']
+
+log = logging.getLogger(__name__)
+
+class LightCurveCollection(Sequence):
+    """
```
christinahedges: These docstrings need to be in the following format:
http://docs.astropy.org/en/stable/development/docrules.html#docstring-standard

lightkurve/lightcurvecollection.py
```diff
@@ -0,0 +1,87 @@
...
+        try: 
+            return self.data[kep_id]
+        except:
+            raise ValueError('No LightCurve for ' + kep_id)
+
+    def append(self, lc):
+        try:
+            self.data[lc.keplerid] = lc
+        except:
+            log.warning("Input is not a lightcurve")
```
christinahedges: This should raise an exception not a warning.

lightkurve/lightcurvecollection.py
```diff
@@ -0,0 +1,87 @@
...
+        Plot a collection of LightCurves. Random colors are assigned to each plot.
+        '''
+        for i, k_id in enumerate(self.data):
+            rand_color = np.random.rand(3)
+            if i == 0:
+                #TODO: what if longest axis is not the first item?
+                axis = self.data[k_id].plot(color=rand_color, linestyle='-',label=k_id)
+            else:
+                self.data[k_id].plot(ax=axis, color=rand_color, linestyle='-', label=k_id)
+
```
christinahedges: We can work on this together on implementing this.

lightkurve/tests/test_collections.py
```diff
@@ -0,0 +1,29 @@
...
+TESS_SIM = ("https://archive.stsci.edu/missions/tess/ete-6/tid/00/000/"
+            "004/104/tess2019128220341-0000000410458113-0016-s_lc.fits")
+
+
+def test_collections_plot():
+	lc_1 = lightcurve.LightCurve(time=np.arange(1, 5), flux=np.arange(1, 5), flux_err=np.arange(1, 5))
+	lc_2 = lightcurve.LightCurve(time=np.arange(1, 5), flux=np.arange(6, 10), flux_err=np.arange(1, 5))
+	collection = LightCurveCollection((lc_1,lc_2))
+	collection.plot()
+
```
christinahedges: We should add some more thorough tests here and have one for LCC and one for TPFC

christinahedges:
You can add a tutorial for LCC and TPFC in docs/source/tutorials. Just something very short, as we will update it when we fix from_archive.

lightkurve/lightcurvecollection.py
johnnyzhang295: Done

lightkurve/lightcurvecollection.py
johnnyzhang295: Done

lightkurve/lightcurvecollection.py
johnnyzhang295: Done

lightkurve/lightcurvecollection.py
johnnyzhang295: Done

johnnyzhang295:
Also working on a jupyter notebook.

lightkurve/lightcurve.py
```diff
@@ -1085,3 +1086,153 @@ def opt_depth(to, width, data):
...
+        for lightcurve in self.data.values():
+            result += lightcurve.__repr__() + "\n"
+        return result
+
+    def plot(self):
+        """
+        Plot a collection of LightCurves. Random colors are assigned to each plot.
+        """
+        for i, k_id in enumerate(self.data):
+            rand_color = np.random.rand(3)
```
christinahedges: I think I put in a "color_cycle" into the plot function for lightkurve so it should cycle through the black and then the maplotlib standards. This is nice because it lets you change the default plot style and still go through familiar/pleasing colours. Do we need "rand_color" for this to work?

lightkurve/lightcurve.py
```diff
@@ -1085,3 +1086,153 @@ def opt_depth(to, width, data):
...
+            if i == 0:
+                #TODO: what if longest axis is not the first item?
+                axis = self.data[k_id].plot(color=rand_color, linestyle='-',label=k_id)
```
christinahedges: If it's a dictionary can we not just loop through the lcs with a for loop?

lightkurve/lightcurve.py
```diff
@@ -1085,3 +1086,153 @@ def opt_depth(to, width, data):
...
+        except:
+            raise TypeError("Input is not a lightcurve")
+
+    def __repr__(self):
+        result = ""
+        for lightcurve in self.data.values():
+            result += lightcurve.__repr__() + "\n"
+        return result
+
+    def plot(self):
```
christinahedges: Let's have plot take in an optional `ax` and `**kwargs` so it behaves like `lc.plot()`

lightkurve/lightcurve.py
```diff
@@ -1085,3 +1086,153 @@ def opt_depth(to, width, data):
...
+        Plot a collection of LightCurves. Random colors are assigned to each plot.
+        """
+        for i, k_id in enumerate(self.data):
+            rand_color = np.random.rand(3)
+            if i == 0:
+                #TODO: what if longest axis is not the first item?
+                axis = self.data[k_id].plot(color=rand_color, linestyle='-',label=k_id)
+            else:
+                self.data[k_id].plot(ax=axis, color=rand_color, linestyle='-', label=k_id)
+
```
christinahedges: Have this return an `ax` so it behaves like `lc.plot()`.

lightkurve/lightcurve.py
```diff
@@ -1085,3 +1086,153 @@ def opt_depth(to, width, data):
...
+            if tpf.time[-1] > axis_max:
+                axis_max = tpf.time[-1]
+            if len(tpf.time) > num_samples:
+                num_samples = len(tpf.time)
+            small = np.nanmin(tpf.flux)
+            if small < flux_min:
+                flux_min = small
+            large = np.nanmax(tpf.flux)
+            if large > flux_max:
+                flux_max = large
```
christinahedges: This is a few too many lines to do this check, it is possible to make it shorter but it's a pretty good solution!

lightkurve/lightcurve.py
```diff
@@ -1085,3 +1086,153 @@ def opt_depth(to, width, data):
...
+        x = np.linspace(axis_min, axis_max, num = num_samples)
```
christinahedges: No spaces after `=` inside a function

lightkurve/lightcurve.py
```diff
@@ -24,7 +24,7 @@
 from . import PACKAGEDIR
 
 __all__ = ['LightCurve', 'KeplerLightCurve', 'TessLightCurve',
-           'iterative_box_period_search']
+           'iterative_box_period_search', 'LightCurveCollection', 'TargetPixelFileCollection']
```
christinahedges: Need an indent to put this all on one line

lightkurve/lightcurve.py
```diff
@@ -1085,3 +1086,153 @@ def opt_depth(to, width, data):
...
+        except:
+            log.warning("Input is not a TPF")
+
+    def __repr__(self):
+        result = ""
+        for TPF in self.data.values():
+            result += TPF.__repr__() + "\n"
+        return result
+
+    def plot(self):
```
christinahedges: Same as above, let's have `ax` and `**kwargs` to pass to matplotlib so that the function behaves just like `tpf.plot()`. (You'll have to be careful with vmin and vmax if you set them in this function and also pass them as kwargs.)

lightkurve/lightcurve.py
```diff
@@ -1085,3 +1086,153 @@ def opt_depth(to, width, data):
...
+            if large > flux_max:
+                flux_max = large
+        x = np.linspace(axis_min, axis_max, num = num_samples)
+        y = np.linspace(flux_min, flux_max, num = num_samples)
+        _, axis = plt.subplots()
+        axis.plot(x, y)
+
+        for i, k_id in enumerate(self.data):
+            self.data[k_id].plot(ax=axis, label=k_id)
+
```
christinahedges: As in tpf.plot(), let's return `ax`.

lightkurve/lightcurve.py
```diff
@@ -1085,3 +1086,153 @@ def opt_depth(to, width, data):
...
+            small = np.nanmin(tpf.flux)
+            if small < flux_min:
+                flux_min = small
+            large = np.nanmax(tpf.flux)
+            if large > flux_max:
+                flux_max = large
+        x = np.linspace(axis_min, axis_max, num = num_samples)
+        y = np.linspace(flux_min, flux_max, num = num_samples)
+        _, axis = plt.subplots()
+        axis.plot(x, y)
```
christinahedges: There's no need to build x and y here, just create the `axis` object, plot as below and then set the x and y limits using
`axis.set_xlim(a, b)`
`axis.set_ylim(c, d)`

lightkurve/tests/test_collections.py
```diff
@@ -0,0 +1,27 @@
...
+             "/0084/008462852/kplr008462852-2011073133259_lpd-targ.fits.gz")
+K2_C08 = ("https://archive.stsci.edu/missions/k2/lightcurves/c8/"
+          "220100000/39000/ktwo220139473-c08_llc.fits")
+KEPLER10 = ("https://archive.stsci.edu/missions/kepler/lightcurves/"
+            "0119/011904151/kplr011904151-2010009091648_llc.fits")
+TESS_SIM = ("https://archive.stsci.edu/missions/tess/ete-6/tid/00/000/"
+            "004/104/tess2019128220341-0000000410458113-0016-s_lc.fits")
+
+
+def test_collections_plot():
```
christinahedges: Need to add tests for TPFC too.

johnnyzhang295:
I also created the jupyter tutorial for LightCurveCollection.

lightkurve/lightcurve.py
```diff
@@ -1087,3 +1089,71 @@ def opt_depth(to, width, data):
...
+        return result
+
+    def plot(self, ax=None, **kwargs):
+        """
+        Plot a collection of LightCurves. Random colors are assigned to each plot.
+        """
+        for i, k_id in enumerate(self.data):
+            
+            if i == 0:
+                #TODO: what if longest axis is not the first item?
```
christinahedges: I don't understand this comment, what do you mean?

lightkurve/lightcurve.py
```diff
@@ -1087,3 +1089,71 @@ def opt_depth(to, width, data):
...
+
+    def __repr__(self):
+        result = ""
+        for lightcurve in self.data.values():
+            result += lightcurve.__repr__() + "\n"
+        return result
+
+    def plot(self, ax=None, **kwargs):
+        """
+        Plot a collection of LightCurves. Random colors are assigned to each plot.
```
christinahedges: The doc string for this needs to follow the same conventions as `lc.plot()` and `tpf.plot()`, i.e. 

```
One line description

Parameters
----------
param : type
     description

Returns
-------
param : type
     description
```

lightkurve/lightcurve.py
```diff
@@ -1087,3 +1089,71 @@ def opt_depth(to, width, data):
...
+    def __init__(self, lcs):
+        try:
+            lightcurves = np.asarray(lcs)
+        except:
+            raise TypeError("Unable to parse input")
+        self.data = {}
+        for lc in lightcurves:
+            try:
+                if lc.keplerid:
+                    self.data[lc.keplerid] = lc
```
christinahedges: It is possible to create a lightkurve with a KeplerID of `None`. It is also possible to have several *different* LightCurve objects with the same KeplerID. This will break in either case. I'm not sure what the best solution for this is...My solution from previous PR was to have the LightCurves indexed by a normal integer, and if you indexed by KeplerID it would return any LightCurves that matched.

lightkurve/lightcurve.py
```diff
@@ -1087,3 +1089,71 @@ def opt_depth(to, width, data):
...
+
+    def plot(self, ax=None, **kwargs):
+        """
+        Plot a collection of LightCurves. Random colors are assigned to each plot.
+        """
+        for i, k_id in enumerate(self.data):
+            
+            if i == 0:
+                #TODO: what if longest axis is not the first item?
+                axis = self.data[k_id].plot(linestyle='-',label=k_id)
```
christinahedges: Space after a ,

lightkurve/lightcurve.py
```diff
@@ -1087,3 +1089,71 @@ def opt_depth(to, width, data):
...
+        for lightcurve in self.data.values():
+            result += lightcurve.__repr__() + "\n"
+        return result
+
+    def plot(self, ax=None, **kwargs):
+        """
+        Plot a collection of LightCurves. Random colors are assigned to each plot.
+        """
+        for i, k_id in enumerate(self.data):
+
```
christinahedges: No need for a white space here

lightkurve/targetpixelfile.py
```diff
@@ -1050,3 +1051,113 @@ def _make_aperture_extension(self):
...
+
+class TargetPixelFileCollection(object):
+    """
+    Collects multiple TPF objects together with helpful functions.
+    """
+    def __init__(self, targetPixelFiles):
+        try:
+            tpfs = np.asarray(targetPixelFiles)
+        except:
+            raise TypeError("Unable to parse input")
```
christinahedges: I don't understand where this error will ever occur, if we always check they are TPFs then the collection should always work?

lightkurve/targetpixelfile.py
```diff
@@ -1050,3 +1051,113 @@ def _make_aperture_extension(self):
...
+
+            small = np.nanmin(tpf.flux[0])
+            if small < flux_min:
+                flux_min = small
+            large = np.nanmax(tpf.flux[0])
+            if large > flux_max:
+                flux_max = large
+        
+        _, axis = plt.subplots()
+        axis.set_xlim(col_min,col_max)
```
christinahedges: Space after a comma

lightkurve/targetpixelfile.py
```diff
@@ -1050,3 +1051,113 @@ def _make_aperture_extension(self):
...
+
+    def __repr__(self):
+        result = ""
+        for TPF in self.data.values():
+            result += TPF.__repr__() + "\n"
+        return result
+
+
+
+    def plot(self):
```
christinahedges: Same as above, ax and **kwargs needs to be in the function definition.

lightkurve/targetpixelfile.py
```diff
@@ -1050,3 +1051,113 @@ def _make_aperture_extension(self):
...
+                cax = axis.imshow(self.data[k_id].flux[0], norm=norm)
+                plt.colorbar(cax, ax=axis, norm=norm)
+                
+                self.data[k_id].plot(ax=axis, show_colorbar=False)
+
+            #    axis = self.data[k_id].plot()
+            else:
+                self.data[k_id].plot(ax=axis, show_colorbar=False)
+
+        plt.axis('equal')
```
christinahedges: What is this doing?

lightkurve/targetpixelfile.py
```diff
@@ -1050,3 +1051,113 @@ def _make_aperture_extension(self):
...
+        """
+        from astropy.visualization import (ImageNormalize, LinearStretch, LogStretch)
+        row_min = 10000
+        row_max = 0
+        num_samples = 0
+        flux_min = 10000
+        flux_max = 0
+        col_min = 10000
+        col_max = 0
+        for i, tpf in enumerate(self.data.values()):
```
christinahedges: We are enforcing here that the TPF plot is square, do you mean to do this?

lightkurve/lightcurve.py
johnnyzhang295: I didn't realize that the matplotlib function will automatically adjust the axis and scale for you, so this actually isn't an issue.

lightkurve/targetpixelfile.py
johnnyzhang295: I imagined a scenario where the input wasn't something that could be split using the np.asarray method.

lightkurve/targetpixelfile.py
johnnyzhang295: I was trying to use it to scale the plot axes so that the shape is a square because each matplotlib axes returned from the tpf.plot() method has its own (x,y) scale. Although each tpf.plot() returns a square plot, when you try to add them together the resulting matplotlib axes needs to be scaled.

lightkurve/lightcurve.py
johnnyzhang295: I can get this done by having the dictionary map a KeplerID to a LIST of lightcurves, instead of just a 1-1 mapping. I can also add a dictionary entry that holds those lightcurves without KeplerIDs. 

Doing this would add a fair bit of complexity to the data structure but it would solve both of these problems.

christinahedges:
This is really good work! Thanks for putting this together Johnny it's super impressive. I'm going to ask @barentsen to please review your code now as I think it's in a really good shape. He normally does *very* thorough reviews :)

Make sure you get those temporary files (e.g. lightkurve.py~ and setup.py~) out of the PR before we merge it.

christinahedges:
Travis is failing and I'm not sure it's just a mast error it looks like there might be something a little funky with the test?

johnnyzhang295:
Yes, the Travis build is failing to download from the mast website but I think its most likely a bug with Travis and not with lightkurve. I've removed the temporary files and I'll push it again!

lightkurve/lightcurve.py
```diff
@@ -1087,3 +1089,110 @@ def opt_depth(to, width, data):
...
+    def stitch(self, normalize=False):
+        new_lc = LightCurve(time=[])
+        for lc_array in self.data.values():
+            for lc in lc_array:
+                if normalize:
+                    lc = lc.normalize()
+                new_lc.append(lc)
+                if new_lc.keplerid != lc.keplerid:
+                    log.warning(‘CAUTION: Stitching light curves with multiple IDs’)
```
christinahedges: Don't use back ticks here, it's causing Travis/Appveyor to fail. Use normal quotes! :)

lightkurve/lightcurve.py
```diff
@@ -1087,3 +1089,110 @@ def opt_depth(to, width, data):
...
+                    self.data[lc.keplerid][0].append(lc)
+                    return   
+                self.data[lc.keplerid].append(lc)
+            else:
+                self.data[lc.keplerid] = [lc]
+        except:
+            raise TypeError("Input is not a lightcurve")
+
+    def stitch(self, normalize=False):
+        new_lc = LightCurve(time=[])
```
christinahedges: Can you initialize a `LightCurve` with no flux now? This must be new, I'm surprised this doesn't fail. `LightCurve`'s should always have flux! It's fine here so long as it doesn't break but I would use
```
 new_lc = LightCurve(time=[], flux=[])
```
just in case.

lightkurve/lightcurve.py
```diff
@@ -1087,3 +1089,110 @@ def opt_depth(to, width, data):
...
+        -------
+        ax : matplotlib.axes._subplots.AxesSubplot
+            The matplotlib axes object.
+        """
+        if ax is None:
+            _, ax = plt.subplots()
+        for lc_array in self.data.values():
+            for lightcurve in lc_array:
+                lightcurve.plot(ax=ax, label=lightcurve.keplerid)
+
```
christinahedges: No need for a space here

lightkurve/lightcurve.py
```diff
@@ -1087,3 +1089,110 @@ def opt_depth(to, width, data):
...
+        return self.data.keys()
+
+    def __getitem__(self, kep_id):
+        """
+        Returns the lightcurve associated with the kepler_id. 
+        """
+        try: 
+            return self.data[kep_id]
+        except:
+            raise ValueError('No LightCurve for ' + kep_id)
```
christinahedges: This is going to be a pain to extract just one of the `LightCurve` objects. What if I want to do

```
for lightcurve in lightcurvecollection:
    lc.normalize().fold().bin().plot()
```

It will break when I get back multiple `LightCurve`'s based on KeplerID? Will they all be in a list, or be a new collection? I think we need LCC to be iterable in this way.

I think this is what we need:
1. A way to index into the LCC using a simple integer (0...n)
2. If called in a loop, LCC should iterate through every `LightCurve` in the collection.
3. If indexed by KeplerID, it should return another `LightCurveCollection` that has only the `LightCurve`s with the specified ID.

How does that sound?

lightkurve/lightcurve.py
```diff
@@ -1087,3 +1089,110 @@ def opt_depth(to, width, data):
...
+
+    def append(self, lc, **kwargs):
```
christinahedges: I still don't love this stitch keyword. Let's keep append as one thing and rely on the stitch function for something else.

lightkurve/targetpixelfile.py
```diff
@@ -1055,3 +1056,131 @@ def _make_aperture_extension(self):
...
+    Collects multiple TPF objects together with helpful functions.
+    """
+    def __init__(self, targetPixelFiles):
+        try:
+            tpfs = np.asarray(targetPixelFiles)
+        except:
+            raise TypeError("Unable to parse input")
+        self.data = {}
+        for TPF in tpfs:
+            if isinstance(TPF, TargetPixelFile) or isinstance(TPF,KeplerTargetPixelFile):
```
christinahedges: Space after a comma

lightkurve/targetpixelfile.py
```diff
@@ -1055,3 +1056,131 @@ def _make_aperture_extension(self):
...
+
+        for tpf_array in self.data.values():
+            for tpf in tpf_array:
+                tpf.plot(ax=ax, show_colorbar=False)
+
+        plt.axis('equal')
+        
+        return ax
+
+    def pca(self):
```
christinahedges: No need for a PCA method here, it doesn't make sense for TPFs

lightkurve/tests/test_lightcurve.py
```diff
@@ -442,6 +442,60 @@ def test_flatten_robustness():
...
+    lcc = LightCurveCollection((lcf_1, lcf_2))
+    lcc.append(lcf_3)
+    assert(lcc.data[lcf_3.keplerid][0] == lcf_3)
+
+@pytest.mark.remote_data
+def test_lightcurvecollection_length():
+    lcf_1 = KeplerLightCurveFile.from_archive('Kepler-10b', quarter=1).PDCSAP_FLUX
+    lcf_2 = KeplerLightCurveFile.from_archive('Kepler-8b', quarter=3).PDCSAP_FLUX
+    lcf_3 = KeplerLightCurveFile.from_archive('Kepler-16b', quarter=7).PDCSAP_FLUX
+
```
christinahedges: I don't understand why these tests are duplicates of above? Let's keep it in one test function to reduce lines.

lightkurve/lightcurve.py
johnnyzhang295: Done

lightkurve/lightcurve.py
johnnyzhang295: Done

lightkurve/lightcurve.py
johnnyzhang295: Done

lightkurve/lightcurve.py
johnnyzhang295: Done

lightkurve/targetpixelfile.py
johnnyzhang295: Done

lightkurve/targetpixelfile.py
johnnyzhang295: Done

lightkurve/tests/test_lightcurve.py
johnnyzhang295: Done!

lightkurve/lightcurve.py
johnnyzhang295: I can achieve this with the current LCC class setup, but it will require some tweaking....
What I'm thinking is to override the __iter__() method so that we can get this kind of functionality for looping you're talking about. I can also accomplish (1, 3) by changing the __getitem__() method to return a new LCC.
Thoughts?

johnnyzhang295:
New PR in #187

PR #177, opened 2018-07-02
Periodogram class and asteroseismology

johnnyzhang295:


christinahedges:
There are a lot of commits on this branch from me, make sure that you're working from the current KeplerGO master branch of `lightkurve` rather than one of my crumby old branches!

christinahedges:
This is breaking because we need to add pandas as a dependency.

Issue #180, opened 2018-07-04
"Keyword 'MISSION' not found." for some fits files


keatonb:
#### Problem description
Some Kepler light curve fits files downloaded with the wget scripts at https://archive.stsci.edu/kepler/publiclightcurves.html do not have the MISSION keyword in their headers (especially quarter 1 data for KIC 9821777 and above?). This missing keyword renders the data inaccessible to lightkurve.  I have not been able to set the mission attribute manually either. Can this attribute be None if the keyword is not in the header as is done for "Campaign" for Kepler data?

#### Example

```python
import lightkurve
klf = lightkurve.KeplerLightCurveFile(path='kplr009821777-2009166043257_llc.fits')  
print(klf.PDCSAP_FLUX)
```

#### Expected behavior
Should print the flux array, but instead returns KeyError: "Keyword 'MISSION' not found."

#### Environment:
- platform (e.g. Linux, OSX, Windows): Linux
- lightkurve version (e.g. 1.0b6): 1.0b8
- installation method (e.g. pip, conda, source): pip

mirca:
Hi @keatonb,

thanks a lot for investigating this!

However, I can't reproduce this error when I download the lightcurve file from the archive by doing:
```
>>> klf = KeplerLightCurveFile.from_archive(9821777, quarter=1)
>>> print(klf.PDCSAP_FLUX)
KeplerLightCurve(KIC: 9821777)
>>> klf.mission
'Kepler'
```

@barentsen what does this mean with respect to lightcurve files from [here](http://archive.stsci.edu/pub/kepler/lightcurves/tarfiles/)?

texadactyl:
If you find this ever to be true in the future, Astropy has proposed a header-edit process that worked for me (I was just monkeying around to see if it would work):
http://www.astropy.org/astropy-tutorials/FITS-header.html

Obviously, you need to know what goes in the element you are adding.

keatonb:
I also found that the individual files on the MAST archive do have this header, but the files in the tarballs (downloaded ~1 month ago) don't always.

texadactyl:
Okay, I found that particular FITS at 
http://archive.stsci.edu/pub/kepler/lightcurves/0098/009821777/kplr009821777-2009166043257_llc.fits
and it seems intact:
```
import lightkurve
PATH="http://archive.stsci.edu/pub/kepler/lightcurves/0098/009821777/kplr009821777-2009166043257_llc.fits"
[...]

PR #181, opened 2018-07-04
add try/except for mission property

keatonb:
Addressing #180, I added a try/except to handle files with a missing MISSION keyword, setting the property to None where there is a KeyError.

If mission is None, the class already returns a representation with ID rather than KIC or EPIC:
```
def __repr__(self):
    if self.mission is None:
        return('KeplerLightCurveFile(ID: {})'.format(self.keplerid))
    elif self.mission.lower() == 'kepler':
        return('KeplerLightCurveFile(KIC: {})'.format(self.keplerid))
    elif self.mission.lower() == 'k2':
        return('KeplerLightCurveFile(EPIC: {})'.format(self.keplerid))
```

### DIFF
--- a/lightkurve/lightcurvefile.py
+++ b/lightkurve/lightcurvefile.py
@@ -258,7 +258,10 @@ class KeplerLightCurveFile(LightCurveFile):
     @property
     def mission(self):
         """Mission name"""
-        return self.header(ext=0)['MISSION']
+	try:
+            return self.header(ext=0)['MISSION']
+        except KeyError:
+            return None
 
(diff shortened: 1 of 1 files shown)


barentsen:
Thanks so much @keatonb!  We already had this try-except block in the `KeplerTargetPixelFile` class, and it makes 100% sense to have it in `KeplerLightCurveFile` as well. Merging! :+1:

barentsen:
(@mirca Let's go without a warning until we understand where the file in #180 came from.  It may be that files with a missing MISSION keyword are harmless.)

PR #187, opened 2018-07-09
LightCurveCollections and TargetPixelFileCollections Classes

johnnyzhang295:
Created a Collections class for LightCurves and TargetPixelFiles. This class collects together multiple lightcurves or tpfs and allows operations over all of them such as plotting, for-loops, etc.

Input to a collections class should be given as an array. Example usage would be:
```
lcc = LightCurveCollections([lc1, lc2, lc3])
for lc in lcc:
    lc.normalize().fold().bin().plot()
```

or to simply plot all of them:
```
lcc.plot()
```


lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+                result[obj.keplerid] = idx
+            except AttributeError:
+                print("Object "+ str(idx) + " has no keplerid")
+        return result
+
+    def __len__(self):
+        return len(self.data)
+
+    def __getitem__(self, index):
+        """Called when indexing into LCC or such as when looping.
```
barentsen: Let's make this docstring more clear, e.g. `Called when accessing an element in the collection by index or target id.`

lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Parameters:
+        -----------
+        Index: int
```
barentsen: `Index` => `index`  (lowercase i)

lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+                print("Object "+ str(idx) + " has no keplerid")
+        return result
+
+    def __len__(self):
+        return len(self.data)
+
+    def __getitem__(self, index):
+        """Called when indexing into LCC or such as when looping.
+
+        Parameters:
```
barentsen: `Parameters:` => `Parameters` 
(because: let's follow numpy docstring standards)

lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+        -----------
+        Index: int
+            This is either the index of the array or
+            the keplerid of the LC.
```
barentsen: Clarify:`This is either the index of the element in the collection, or the target id of the object.`

lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Returns:
```
barentsen: `Returns:` => `Returns`

lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+        --------
+        Lightcurve Object
```
barentsen: Should e.g. say `LightCurve or TargetPixelFile object`.

lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        try:
+            if index > len(self.data):
+                return self.data[self.k_id[index]]
```
barentsen: Does this work when two or more objects with the same target id are added to the collection?  Please add a unit test for this case.

lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+            else:
+                return self.data[index]
+        except KeyError:
+            print("Object has no keplerid")
```
barentsen: We can't just print a message here; the code that calls this method will expect to receive a LightCurve or TargetPixelFile and likely crash if it doesn't get one.  So, instead, let's do something like `raise KeyError('The collection does not contain an object with target ID {}.'.format(self.k_id[index]))

lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def append(self, obj):
+        """Appends lightcurve object to LCC
```
barentsen: This is a generic base class, so it should say something like `Adds a new object to the collection.`.

lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Parameters:
```
barentsen: `Parameters:` => `Parameters`

lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+        -----------
+        lc: LightCurve object
+            Lightcurve target
```
barentsen: `lc` => `obj`
`LightCurve object` => `LightCurve or TargetPixelFile object`
etc

lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Returns:
```
barentsen: `Returns:` => `Returns`

lightkurve/collection.py
```diff
@@ -0,0 +1,101 @@
...
+        --------
+        None
+        """
+        self.data.append(obj)
+        try:
+            self.k_id[obj.keplerid] = len(self.data)-1
+        except AttributeError:
+            print("Object has no keplerid")
```
barentsen: The `TessTargetPixelFile` and `TessLightCurve` objects do not have a `keplerid` attribute.  I think we have to make sure that the `TargetPixelFile` and `LightCurve` objects have a generic `targetid` attribute so we can use that here.

barentsen:
@johnnyzhang295 @johnnyzhang295 Thanks, I think this is an improvement over earlier PRs and something we could merge.

I do wonder if we want to sub-class the `Collection` class from Python's built-in `Sequence` class.  In this case we could still support indexing by targetid by overriding `__get__` to build the target=>index dictionary map when it is needed (it could be cached by making sure a `_dirty_map` boolean to True when a new item is appended).  The advantage is that we may have less code to maintain, and only need to build a mapping dictionary when needed.  Thoughts?

johnnyzhang295:
@barentsen I'm not too familiar with the Collections class in python. If we opt to use a cache for the dictionary map, I wonder if this would make the code too complex. Building the dictionary map can be done in linear time so I'm not too worried about its computational complexity.

lightkurve/collection.py
```diff
@@ -0,0 +1,102 @@
...
+        List of data objects.
+    k_id: dictionary
+        Mapping target to index in self.data
+    """
+    __metaclass__ = ABCMeta #Needs to be set for Python 2.x to work properly
+    def __init__(self, data):
+        self.data = data
+        self.k_id = self.assign_hash_values()
+
+    def assign_hash_values(self):
```
barentsen: Nit-picking: let's make this a hidden method because users shouldn't have to use it.  Python syntax for this is underscore prefix, i.e. rename the function to `_assign_hash_values`  (or even better `_update_targetid_map`?)

lightkurve/lightcurve.py
```diff
@@ -1179,3 +1179,44 @@ def opt_depth(to, width, data):
...
+
+class LightCurveCollection(Collection):
+    """Represents a set of LightCurves
+
+    Attributes
+    ----------
+    data: array
+        List of lightcurve objects.
+    k_id: dictionary
+        Mapping keplerid to index in self.data
```
barentsen: `keplerid` => `targetid`

lightkurve/lightcurve.py
```diff
@@ -1179,3 +1179,44 @@ def opt_depth(to, width, data):
...
+
+        Returns
+        -------
+        ax : matplotlib.axes._subplots.AxesSubplot
+            The matplotlib axes object.
+        """
+        if ax is None:
+            _, ax = plt.subplots()
+        for lightcurve in self.data:
+           lightcurve.plot(ax=ax, label=lightcurve.keplerid)
```
barentsen: `keplerid` => `targetid`

lightkurve/targetpixelfile.py
```diff
@@ -1342,3 +1342,35 @@ def get_bkg_lightcurve(self, aperture_mask=None):
...
+
+class TargetPixelFileCollection(Collection):
+    """Represents a set of Target Pixel Files
+
+    Attributes
+    ----------
+    data: array
+        List of TPF objects.
+    k_id: dictionary
+        Mapping keplerid to index in self.data
```
barentsen: `keplerid` => `targetid`

lightkurve/tests/test_lightcurve.py
```diff
@@ -464,6 +464,25 @@ def test_fill_gaps():
...
+def test_lc_collection():
+    lc = LightCurve(time=np.arange(1, 5), flux=np.arange(1, 5), flux_err=np.arange(1, 5))
+
+    lcf = KeplerLightCurveFile(TABBY_Q8)
+    lcc = LightCurveCollection([lcf])
+    lcf2 = lcf
+
+    assert(len(lcc) == 1)
+    assert(lcc[0] == lcf)
+    assert(type(lcc[lcf.keplerid]) == type(lcf))
```
barentsen: `keplerid` => `targetid`

lightkurve/collection.py
```diff
@@ -0,0 +1,102 @@
...
+        Assigns the targetid to indexes in self.data
+
+        Parameters
+        ----------
+        None
+
+        Returns
+        -------
+        result: Dictionary
+            With keys of keplerid (int) and
```
barentsen: `keplerid` => `targetid`

lightkurve/collection.py
```diff
@@ -0,0 +1,102 @@
...
+            values of indexes (int) in the data array.
+        """
+        result = {}
+        for idx,obj in enumerate(self.data):
+            try:
+                result[obj.keplerid] = idx
```
barentsen: `keplerid` => `targetid`

lightkurve/collection.py
```diff
@@ -0,0 +1,102 @@
...
+        ----------
+        obj: LightCurve or TargetPixelFile object
+
+        Returns
+        -------
+        None
+        """
+        self.data.append(obj)
+        try:
+            self.k_id[obj.keplerid] = len(self.data)-1
```
barentsen: `keplerid` => `targetid`

johnnyzhang295:
Collections is ready to merge now! @barentsen.
Lots of improvements were made... 
1. All collections-related code has now been moved to collection.py
2. LOTS of test coverage for LightCurveCollections and TargetPixelFileCollections :)
3. New Feature- Slicing/Indexing (Could only append before)
4. An issue with `KeplerTargetPixelFile` and `targetid` has been fixed

barentsen:
Thanks @johnnyzhang295!  Stay tuned for a final review!

barentsen:
Merging.  Thanks @johnnyzhang295!

PR #191, opened 2018-07-11
"How to make a supernova lightcurve" tutorial

gully:
Introduces a tutorial for "How to make a supernova lightcurve" using the fast evolving luminous transient KSN2015k from Campaign 6.


barentsen:
Thanks @gully!  This provides a great into to supernova newcomers.  It will be really useful to have this in place before the workshop.

I'd like to suggest three changes before merging:

1) We have a rolling band problem in this target which we can't ignore (i.e. the waves towards the end).  Let's make a simple background aperture mask and use it to perform local background subtraction, so that we demonstrate at least one approach to get rid of the rolling bands in this transient.  We can keep it easy here (i.e. do it in 3-4 lines) and link to the rolling band tutorial for additional details about the phenomenon and an additional example.

2) Please replace
`np.isnan(tpf.flux).sum(axis=0)[::-1] < 3561`
with 
`tpf.hdu[2].data > 0`.
(I am fine with adding a shorthand `tpf.observed_mask` for this.)

3) Let's make it public!  Please give the tutorial a number and add it to `docs/source/tutorials/section2.rst`. and try to `make html` to see if it compiles.  (I seem to remember that sphinx raises an error over `tpf.interact()`, so you may have to move `interact()` into the text rather than an executable cell.)

Once you get it to compile, please go ahead and merge, then go to `docs/` and type `make html` and then `make upload` to make it live.  (I'm about to board plane so may not respond for the next 10 hours.)

Thanks++

barentsen:
Great!  Thanks for the extra commits @gully!  (Indeed it'd be curious to understand why SFF doesn't contribute here!  A third option is that the Poisson noise dominates the systematics?)

Merging this now to make it available to the workshop!  Travis build failure is unrelated.

PR #176, opened 2018-07-02
Issue #172 Docs: demonstrate how to load a local target pixel file

IsFlwrs:
Issue #172 Docs: demonstrate how to load a local target pixel file

Issue #172, opened 2018-06-26
Docs: demonstrate how to load a local target pixel file


barentsen:
Both the quickstart guide and the "what are target pixel files?" tutorials in the docs only use `KeplerTargetPixelFile.from_archive()`.  We need to add an example of opening a local file from disk.

(I just received a query from a user who couldn't figure out how to open a local file.)

IsFlwrs:
I'd like to take this on. Can you provide some details on the issue please?

barentsen:
@IsFlwrs Hi!  The source of our documentation lives under the `docs` directory of this repository.  We compile it by running `make html` in that directory (using the `sphinx` documentation framework), and for every release we upload the new version to http://lightkurve.keplerscience.org

Recently a few users contacted me to ask how they can read a local target pixel file using lightkurve.  This can be done using `tpf = KeplerTargetPixelFile(path)`, however our quickstart and many of the examples in our current documentation only show the feature to download the data directly from the archive, i.e. `tpf = KeplerTargetPixelFile.from_archive(targetid)`.

We should simply add the standard version of reading a file locally, i.e. `tpf = KeplerTargetPixelFile(path)`, to the quickstart and 1 or 2 places in the basic tutorials, to make it more clear that this is an alternative way of loading data!

Does that help?

PR #194, opened 2018-07-14
Add a `from_fits` method to TargetPixelFile and LightCurveFile

### LINKED ISSUE #192: Add a from_file() function to load FITS lightcurve file
#### Problem description
Attempting to use lightkurve to read in a FITS lightcurve file (e.g., ktwo246802680-c13_llc.fits), it was not straightforward in the documentation/tutorials how to do it. I suggest adding a from_file() function to mirror the well-documented w/ example from_archive() function.

#### Example

By asking staff, I was able to use the following commands to load a lightcurve file, but I did not see it as an example in the documentation.

```python
from lightkurve import KeplerLightCurveFile
lcf = KeplerLightCurveFile("ktwo246802680-c13_llc.fits")
```

I think if there was a from_file() function, it would be more obvious that a user that could load using from_archive() or from_file():

```python
from lightkurve import KeplerLightCurveFile
lcf = KeplerLightCurveFile.from_file("ktwo246802680-c13_llc.fits")
```

#### Expected behavior
Expect intro tutorial to show how to load from a local file as well as the archive.

#### Environment:
Linux
1.0b9
pip

barentsen:
This resolves #192.

It will require changes to the docs which have already been started (#172 & PR #176).

@JeffLCoughlin Can you review?

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -12,6 +12,7 @@ Direct contributions to the code base:
 - `Andy Tzanidakis <https://github.com/AndyTza>`_
 - `Sheila Sagear <https://github.com/ssagear>`_
 - `Emma Turtelboom <https://github.com/Emmavt>`_
+- `Jeff Coughlin <https://github.com/JeffLCoughlin>`_
 
 Comments, corrections & suggestions:
 
@@ -25,3 +26,4 @@ Comments, corrections & suggestions:
 - `Elisa Quintana <https://github.com/elisaquintana>`_
 - Jeff Van Cleve
 - Ashley Villar
+- `Benjamin Pope <https://github.com/benjaminpope>`_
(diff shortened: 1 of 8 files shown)


mirca:
@barentsen I was thinking whether it would be more appropriate to add ``from_fits`` at the base class ``TargetPixelFile``? The advantage being that you wouldn't have to copy the code to ``TessTargetPixelFile`` and could also reuse part of the docstring.

barentsen:
@mirca Agreed!  ...however I couldn't figure out how to call the constructor of the subclass from a method in the parent class?  (But I was doing this on a plane so I couldn't Google.)

barentsen:
Working towards the next pip release today.  Merging...

mirca:
Very nice!

JeffLCoughlin:
Sorry I missed seeing the review until now, but this is awesome!  Thanks for tackling so quick.

PR #195, opened 2018-07-14
add notebooks directive to docs/Makefile

mirca:
I'm adding the directive ``notebooks``, so that we can run ``make notebooks`` whenever we would like to execute all the notebooks in the documentation.

I think we could discuss at some point whether this should be the default behavior. 
Having this as default is particularly interesting as the number of tutorials increases and some of them might be affect by backwards-incompatible changes. The downside is the time it takes to run all notebooks.


barentsen:
Thanks @mirca!! ❤️  This works for me.

I pushed an extra commit to expand the `Building the docs` section in the documentation.  Let's merge!

PR #198, opened 2018-07-18
Adds gradient of KeplerPRF

mirca:
This PR adds the gradient of the KeplerPRF model which will be needed for us to report uncertainties on the PRF photometry results.

TODO:

- [x] add unit tests

(feel free to add task items to this list)

### DIFF
--- a/lightkurve/prf/prfmodel.py
+++ b/lightkurve/prf/prfmodel.py
@@ -108,6 +108,66 @@ class KeplerPRF(object):
                                                  grid=False).reshape(self.shape)
         return self.prf_model
 
+    def gradient(self, center_col, center_row, flux=1., scale_col=1., scale_row=1.,
+                 rotation_angle=0.):
+        """
+        This function returns the gradient of the KeplerPRF model with
+        respect to center_col, center_row, flux, scale_col, scale_row,
+        and rotation_angle.
+
+        Parameters
+        ----------
+        center_col, center_row : float
+            Column and row coordinates of the center
+        flux : float
+            Total integrated flux of the PRF
+        scale_col, scale_row : float
+            Pixel scale in the column and row directions
+        rotation_angle : float
+            Rotation angle in radians
+
+        Returns
+        -------
+        grad_prf : list
+            Returns a list of arrays where the elements are the partial derivatives
+            of the KeplerPRF model with respect to center_col, center_row, flux, scale_col,
+            scale_row, and rotation_angle, respectively.
+        """
+        cosa = math.cos(rotation_angle)
+        sina = math.sin(rotation_angle)
+
+        delta_col = self.col_coord - center_col
+        delta_row = self.row_coord - center_row
+        delta_col, delta_row = np.meshgrid(delta_col, delta_row)
+
+        rot_row = delta_row * cosa - delta_col * sina
+        rot_col = delta_row * sina + delta_col * cosa
+
+        deriv_flux = self.interpolate(rot_row.flatten() * scale_row,
+                                      rot_col.flatten() * scale_col,
+                                      grid=False).reshape(self.shape)
+
+        interp_dy = self.interpolate(rot_row.flatten() * scale_row,
+                                     rot_col.flatten() * scale_col,
+                                     grid=False, dy=1).reshape(self.shape)
+
+        interp_dx = self.interpolate(rot_row.flatten() * scale_row,
+                                     rot_col.flatten() * scale_col,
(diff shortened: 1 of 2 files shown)


barentsen:
@mira Thank you!! Am very busy next 24 hours but can't wait to review, test, and merge this!!

mirca:
Tests for the gradients are looking good so far:
screen shot 2018-07-18 at 12 51 30 pm

There are some divergences at the edges though:
screen shot 2018-07-18 at 12 51 37 pm

mirca:
Gradients with respect to ``scale_row`` and ``scale_col`` look like to be wrong.

mirca:
This is ready to be scrutinized.

mirca:
The divergences illustrated in the example above don't appear for rotation angles less than 10.

mirca:
While working on this PR I realized we can probably use the finite differences as an approximation for the gradient to compute uncertainties, since we only need to have the gradient around the Maximum a Posteriori estimator. I would expect that computing finite differences would be faster.

lightkurve/prf/prfmodel.py
```diff
@@ -108,6 +108,66 @@ def evaluate(self, center_col, center_row, flux=1., scale_col=1., scale_row=1.,
...
+
+        scale_row_times_interp_dx = scale_row * interp_dx
+        scale_col_times_interp_dy = scale_col * interp_dy
+
+        deriv_center_col = - flux * (cosa * scale_col_times_interp_dy - sina * scale_row_times_interp_dx)
+        deriv_center_row = - flux * (sina * scale_col_times_interp_dy + cosa * scale_row_times_interp_dx)
+        deriv_scale_row = flux * interp_dx * rot_row
+        deriv_scale_col = flux * interp_dy * rot_col
+        deriv_rotation_angle = flux * (interp_dy * scale_col * (delta_row * cosa - delta_col * sina)
+                                       - interp_dx * scale_row * (delta_row * sina + delta_col * cosa))
```
barentsen: Can you add a few lines of comments to explain where these equations come from?

lightkurve/prf/prfmodel.py
```diff
@@ -108,6 +108,66 @@ def evaluate(self, center_col, center_row, flux=1., scale_col=1., scale_row=1.,
...
+        and rotation_angle.
+
+        Parameters
+        ----------
+        center_col, center_row : float
+            Column and row coordinates of the center
+        flux : float
+            Total integrated flux of the PRF
+        scale_col, scale_row : float
+            Pixel scale in the column and row directions
```
barentsen: To avoid confusion with Kepler's absolute pixel scale (~4 arcsec per pixel), the docstring here should should probably say something like `Pixel scale stretch parameter, i.e. the numbers by which the PRF model needs to be multiplied in the column and row directions to account for focus changes.`

mirca:
@barentsen Thanks for the feedback! This ``python2`` failure looks weird, investigating!

lightkurve/prf/prfmodel.py
mirca: I uploaded the proof [here](https://github.com/KeplerGO/lightkurve/files/2217633/gradients.pdf). I'm usually bad at notation, so let me know how cryptic it is.

barentsen:
@mirca Whoa the PDF document is epic!!  I don't spot any mistakes.

Related to your comment about finite differences, I guess we could implement any gradient as follows:
```Python
def gradient(self, x, y, z, dx=1e-5, dy=1e-5, dz=1e-5):
    return self.evaluate(x, y, z) - self.evaluate(x + dx, y + dy, x + dz)
````

...I guess this is often pointless because we could simply let `scipy.optimize` take care of estimating the gradients using numerical methods, though perhaps this *is* useful in the situation where you want the gradients of some likelihood functions to be analytical and that of others to be numerical?  Or can scipy or tensorflow deal with such a situation in other ways?

(^ Don't let my academic questions stop us from fixing the unit test and merging hehehe.)

barentsen:
Unhappy Travis :scream:

mirca:
Dang!! 😿

barentsen:
Travis passes! @mirca you are awesome, like the L1 norm!!!

barentsen:
This is fantastic!  Thanks @mirca!  I'm merging now!

mirca:
Thank you @barentsen! I'm so glad this got merged. Let's uncomment [#L576](https://github.com/KeplerGO/lightkurve/blob/master/lightkurve/prf/tpfmodel.py#L576) :)

PR #203, opened 2018-07-23
Ensure every `LightCurve`, `LightCurveFile`, and `TargetPixelFile` object has a generic targetid

barentsen:
This PR ensures that every `LightCurve` object, i.e. be it a `KeplerLightCurve` or `TessLightCurve`, has a generic `targetid` attribute that returns either the `keplerid` or `ticid`.

This enables tools to print an object's identifier without having to worry about the mission the data came from.

The same change is made for `LightCurveFile` and `TargetPixelFile` objects.

Assigning to @johnnyzhang295 to review!

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -869,12 +869,13 @@ class KeplerLightCurve(LightCurve):
     keplerid : int
         Kepler ID number
     """
-    def __init__(self, time, flux=None, flux_err=None, time_format=None, time_scale=None,
+    def __init__(self, time=None, flux=None, flux_err=None, time_format=None, time_scale=None,
                  centroid_col=None, centroid_row=None, quality=None, quality_bitmask=None,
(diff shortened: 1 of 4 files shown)


johnnyzhang295:
@barentsen Overall it looks very good and I like it. My only (potential) concern is that users will get confused between an object's `targetid` and its "mission_id", i.e. `keplerid, ticid, etc.` and when to use one versus the other. However, if you don't think this would be an issue then it's all good!

barentsen:
This is a good point.  I wonder if we should just get rid of `keplerid` and `ticid` and only ever use `targetid` consistently for all objects.  @christinahedges would you agree with such change?

barentsen:
I'm going to merge this now so we can #187 use `targetid` instead of `keplerid`.

I do think we should deprecate `keplerid` and `ticid` altogether in favor of a single `targetid`, but it will require changes to the tutorials etc, so let's do that in a separate PR.

PR #205, opened 2018-07-25
Make bokeh and ipywidgets optional dependencies

barentsen:
I have seen some users struggle with installing the `bokeh` and `ipywidgets` packages.  Because these are only required by `interact()`, and because it is important to keep the barrier low while we are trying to attract users, I propose we make these optional dependencies for now and print a nice message if they are not installed.

Assigning to @gully to review and merge.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
-include README.rst AUTHORS.rst LICENSE
+include README.rst AUTHORS.rst CONTRIBUTING.md LICENSE
 include lightkurve/data/*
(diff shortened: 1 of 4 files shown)


barentsen:
The AppVeyor error is unrelated (AppVeyor keeps getting stuck on the PSF photometry tests, I wonder if the AppVeyor containers don't have enough memory or compute.)

@gully please go ahead and review!

mirca:
👍

barentsen:
TY!

Issue #200, opened 2018-07-20
`tpf.get_model()` does not work when FLUX_BKG is NaN


barentsen:
The solution will be to change this line:

```
flux_prior = GaussianPrior(mean=np.nanmean(self.flux_bkg),
                                       var=np.nanstd(self.flux_bkg)**2)
```

...it needs to be robust in case self.flux_bkg is nan.

Emmavt:
@barentsen  Currently, tpf.get_model() does work when flux_bkg is NaN, it just sets the Background Prior with this NaN value for mean and var. Do we want to set it to a default value, or just not set the bkg prior if flux_bkg is NaN?

barentsen:
@Emmavt I think we'll need something like this:

```Python
if np.all(np.isnan(self.flux_bkg)):
    # estimate background prior from self.flux instead of self.flux_bkg
```

barentsen:
Fixed by #206 -- closing!

PR #206, opened 2018-07-25
Fix Issue #200 - make tpf.get_model robust for NaN background flux

Emmavt:
Makes tpf.get_model() robust for TPFs with an all NaN background flux, e.g. from Campaign 0. 
If background flux is all NaNs, sets it to the median of flux, which has been 5 sigma clipped until convergence.

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -12,7 +12,6 @@ from matplotlib import patches
 import matplotlib.pyplot as plt
 import numpy as np
 from tqdm import tqdm
-from astropy.stats import sigma_clip
 
 from . import PACKAGEDIR
 from .lightcurve import KeplerLightCurve, TessLightCurve, LightCurve
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -926,8 +926,8 @@ class KeplerTargetPixelFile(TargetPixelFile):
             if np.all(np.isnan(self.flux_bkg)):  # If TargetPixelFile has no background flux data
                 # Use the median of the lower half of flux as an estimate for flux_bkg
                 clipped_flux = np.ma.masked_where(self.flux > np.percentile(self.flux,50), self.flux)
-                flux_prior = GaussianPrior(mean=np.ma.nanmedian(clipped_flux),
-                                           var=np.ma.nanstd(clipped_flux)**2)
+                flux_prior = GaussianPrior(mean=np.ma.median(clipped_flux),
+                                           var=np.ma.std(clipped_flux)**2)
             else:
                 flux_prior = GaussianPrior(mean=np.nanmedian(self.flux_bkg),
                                            var=np.nanstd(self.flux_bkg)**2)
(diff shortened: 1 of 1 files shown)


lightkurve/targetpixelfile.py
```diff
@@ -922,6 +923,14 @@ def get_model(self, star_priors=None, **kwargs):
...
             kwargs['prfmodel'] = self.get_prf_model()
         if 'background_prior' not in kwargs:
+            if np.all(np.isnan(self.flux_bkg)):  # If TargetPixelFile has no background flux data
+                # 5 sigma clip the TargetPixelFile flux until convergence
+                clipped_flux = sigma_clip(self.flux, sigma=5, iters=None)
+                bkg = []
+                for cadence in self.flux:
+                    # Set background flux as the median of the clipped flux for each cadence
+                    cadence_bkg = bkg.append(np.full(self.shape[1:],np.ma.median(cadence))
+                self.flux_bkg = np.stack(bkg, axis=0)
```
barentsen: Ohhh, you are deriving a background estimate for each cadence?  Does this work?  The PSF photometry tool right now assumes this prior is a single value, i.e. a constant background estimate.  For simplicity, `TPFModel` has been developed to model a single cadence.  What we could do is have `get_model` take a `cadence` parameter that estimates the background for that specific cadence.

lightkurve/targetpixelfile.py
```diff
@@ -922,8 +923,14 @@ def get_model(self, star_priors=None, **kwargs):
         if 'prfmodel' not in kwargs:
             kwargs['prfmodel'] = self.get_prf_model()
         if 'background_prior' not in kwargs:
-            flux_prior = GaussianPrior(mean=np.nanmean(self.flux_bkg),
-                                       var=np.nanstd(self.flux_bkg)**2)
+            if np.all(np.isnan(self.flux_bkg)):  # If TargetPixelFile has no background flux data
+                # 5 sigma clip the flux until convergence, and use as an estimate for flux_bkg
+                clipped_flux = sigma_clip(self.flux, sigma=5, iters=None)
```
barentsen: If there are a lot of bright stars, sigma-clipping may end up rejecting the background pixels.  I think the more conventional method for background estimation is to use percentiles. How about this: 

````
if targetpixelfile has no background data:
    flux_prior = UniformPrior(lb=np.percentile(self.flux, 0), lb=np.percentile(self.flux, 50)
```

lightkurve/targetpixelfile.py
```diff
@@ -12,6 +12,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 from tqdm import tqdm
+from astropy.stats import sigma_clip
```
barentsen: If we decide to switch to percentiles, we can get rid of this import.

barentsen:
The test fails but it looks like a MAST connection problem instead of a problem related to this PR.  Restarting the test!  When they pass, we can merge!

barentsen:
Travis passed!  (Appveyor will be fixed by #208.)

Merging!

PR #207, opened 2018-07-25
New function to estimate local background flux for a TPF

Emmavt:
This function estimates the local background flux for a TPF, unlike the pipeline which gives global background flux. Estimates the local background flux by clipping the TPF flux over 5 sigma, and then takes the median of the remaining fluxes. For each cadence, sets the local bkg flux estimate as this median for the whole cadence. Returns an array with the same dimensions as the TPF containing the estimated local bkg fluxes for each cadence.

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -892,7 +892,7 @@ class KeplerTargetPixelFile(TargetPixelFile):
         """ Returns an estimate of the local background flux for a given
         TargetPixelFile as an array of the same dimensions as the TargetPixelFile.
         """
-        # 5 sigma clip the TargetPixelFile flux until convergence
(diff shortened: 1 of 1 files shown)


christinahedges:
Hi Emma! Great job on this PR. It's super helpful for PRF photometry. Unfortunately we're not makign a lot of headway on the PRF module at the moment. Also, Tom has recently added in a new function in #231 which does a similar job to this one using a threshold aperture mask that we're shortly going to merge into lightkurve. This should make it easy to estimate a local background. I'm going to close this for now, if you do find a use case and think there is a strong need for this tool in lightkurve, open a PR with a tutorial to show us how you would like to use it!

We miss you in the GO office, hope you're enjoying the new term!

PR #208, opened 2018-07-25
Have AppVeyor skip the test on which it keeps hanging

barentsen:
The PRF-model fitting unit test tends to make AppVeyor hang.  We should understand this in more detail, but for now I suggest we skip the test in AppVeyor by marking it with the `remote_data` decorator.


mirca:
@barentsen 👍, I cancelled the AppVeyor builds which were hanging, so this patch should start building soon.

mirca:
Neat trick, @barentsen! Thanks! Merging this in 3, 2, 1, ...

PR #216, opened 2018-08-04
Interactively define a mask by selecting individual pixels in a TPF plot

gully:
Very similar to `tpf.interact()`, adds the ability to select your pixel mask.


gully:
Major takeaway from the Coolstars 20 hackday: 
In order for this feature to work, we need to use Bokeh Callbacks:
https://bokeh.pydata.org/en/latest/docs/user_guide/interaction/callbacks.html

We spent most of the hack day trying to understand callbacks and how to write javascript code that could talk across data sources.

gully:
pixel_selector

svaberg:
This example creates a bokeh server inside the Jupyter notebook. Could we use this technique? https://github.com/bokeh/bokeh/blob/0.13.0/examples/howto/server_embed/notebook_embed.ipynb

gully:
Ok, so I tried the "bokeh server inside the Jupyter notebook" approach that @svaberg recommended and... it works!  The prototype achieves the desired behavior exactly.

I will work on this pixel selector tool today, with a goal of adding functionality to the `TargetPixelFile` class.  I will revisit some strategy questions and features once I have a minimum viable prototype.  One main question will be whether we want to combine `.interact()` with the pixel selector, or whether they should be two separate tools.  There are tradeoffs to both approaches, so it may be a matter of enumerating costs and benefits.

gully:
### User Interface / User Experience improvements
- [ ] Add `lasso_select` tool
- [ ] Initiate with a sane display, perhaps the pipeline mask if available
- [ ] Sane defaults for the lightcurve y-axis scaling
- [ ] Decide whether to offer an option for continuous renormalization of lightcurve 
- [ ] Decide whether to disallow selection of always-`NaN` pixels

gully:
20180827_pixel_select_2x_demo

gully:
Check it out @marusazerjal @svaberg @nattieg @rrpastro!  Our Cool Stars 20 hack day project is working!

svaberg:
That looks really cool!

lightkurve/targetpixelfile.py
```diff
@@ -684,6 +684,216 @@ def update(cadence, log_stretch):
...
+                                    cadence=lc.cadenceno, quality=lc.quality)
+                callback4('junk')
+
+            source2.on_change('selected', callback)
+
+            n_cadences = len(self.time)
+            amp_slider = Slider(start=0, end=n_cadences-1, value=0, step=1,
+                    title="TPF slice index", width = 800)
+
+            def callback2(attr, old, new):
```
barentsen: You'll want to give the various callback functions a more descriptive name, e.g. `callback2` could be `cadence_slider_callback`.

lightkurve/targetpixelfile.py
```diff
@@ -684,6 +684,216 @@ def update(cadence, log_stretch):
...
+                                     renderers=[r],
+                                     mode='mouse',
+                                     point_policy="snap_to_data"))
+            # Vertical line to indicate the cadence shown in Fig 2
+            vert = Span(location=0, dimension='height', line_color='firebrick',
+                        line_width=4, line_alpha=0.5)
+            fig1.add_layout(vert)
+
+
+            fig2 = figure(plot_width=300, plot_height=300, x_range=(min(xx), max(xx+1)),
```
barentsen: The code below can be clarified by calling this e.g. `fig_pixels` instead of `fig2`.   Likewise, I'd use e.g. `fig_lc` instead of `fig1`.

gully:
### Feedback from Kepler/K2 GO Office team meeting on Sept. 4, 2018:
- [ ] Modularize code into standalone file `interact.py`
- [ ] Add support for customization through kwargs
- [ ] Remove the "normalize lightcurve" checkbox, make that the default 
- [ ] Rename as "interact" supplanting the previous (IPyWidgets) version
- [ ] Replicate the frame-by-frame ↔️  functionality from previous version
- [ ] nice-to-have: Replicate the "play" ⏯ functionality from previous version
- [ ] Remove the save mask button for now (deploy as a pilot/demo until we build consensus on what format to save)
- [ ] Workarounds to deselect or "toggle" an already selected pixel (see https://github.com/bokeh/bokeh/issues/2326 ).  @nksaunders suggests a "lock in" / "reset" formalism.


### More UI/UX/stability enhancements
- [ ] Decide whether to disallow selection of always-`NaN` pixels-- anticipating normalization bugs
- [ ] Figure out why lasso selection does not work
- [ ] Figure out why error messages pop up if you're not on the default Jupyter notebook (*e.g.* local host 8889 vs. 8888)
- [ ] Move the `TOOLS` to the top to make more room for the color bar
- [ ] Consider kwarg for labeling of the colorbar: percentile or electrons/second
- [ ] Pad or right-align the labels of the color bar to avoid dynamic window resizing 
- [ ] Do we want to make special considerations for long-tall pixel masks (*e.g. * saturated bleed trails)
- [ ] Mimic the informative plot title labels from existing `.interact()`
- [ ] Fix tool tips rendering bug (quality code shown as `???`).
- [ ] Enable more tools in the TPF image view (box zoom), or is that confusing with box select?

christinahedges:
Hey Gully,

I get this error when I try to open your interact tool in a notebook that doesn't have the local host '8888':


image


Is this a bokeh issue? Should I open it with them?

Christina

gully:
Yeah, this behavior is known, it's actually a Jupyter limitation.  There's a link to the bokeh workaround in the comment above yours:
> Figure out why error messages pop up if you're not on the default Jupyter notebook (e.g. local host 8889 vs. 8888) ...

gully:
Ok @christinahedges, this Jupyter port bug should be fixed now.  You can now indicate the notebook url in a kwarg:

```python
tpf.interact(notebook_url='localhost:8889')
```

The bokeh issue I cited explains that there's no way to determine the port programmatically for Jupyter reasons.  A bit of googling found a potential javascript workaround, but it seemed OS-dependent, and slightly sketchy.  So a manual kwarg fix seems like the best solution.

gully:
Goal: add support for conda-installing bokeh automatically in the lightkurve recipe: https://github.com/conda-forge/lightkurve-feedstock/tree/master/recipe

lightkurve/targetpixelfile.py
```diff
@@ -480,209 +480,14 @@ def to_fits(self, output_fn=None, overwrite=False):
...
-        fig1.add_layout(vert)
-
-        # Figure 2 shows the Target Pixel File stamp with log screen stretch
-        if self.mission in ['Kepler', 'K2']:
-            title = 'Pixel data (CCD {}.{})'.format(self.module, self.output)
-        elif self.mission == 'TESS':
-            title = 'Pixel data (Camera {}.{})'.format(self.camera, self.ccd)
+    def interact(self, classic=False, lc=None, notebook_url='localhost:8888'):
+        from . import interact
+        if not classic:
```
barentsen: Is there a good reason to keep the classic version?  If not, I'd keep the code base simple and only offer the new version!

lightkurve/interact.py
```diff
@@ -0,0 +1,580 @@
...
+
+    n_pixels = tpf.flux[0, :, :].size
+    pixel_index_array = np.arange(0, n_pixels, 1, dtype=int).reshape(tpf.flux[0, :, :].shape)
+
+    # Bokeh cannot handle many data points
+    # https://github.com/bokeh/bokeh/issues/7490
+    if len(lc.cadenceno) > 30000:
+        raise RuntimeError('Interact cannot display more than 30000 cadences.')
+
+    def modify_doc(doc):
```
barentsen: Shall we call this e.g. `create_interact_ui` instead of `modify_doc`?

lightkurve/interact.py
```diff
@@ -0,0 +1,580 @@
...
+
+        # The data source includes metadata for hover-over tooltips
+        source = prepare_lightcurve_datasource(lc)
+        source2 = prepare_tpf_datasource(tpf)
```
barentsen: How about `source` => `lc_source` and `source2` => `tpf_source`?

barentsen:
I wonder if a workaround around the `localhost:8888` restriction is to have interact start its own Bokeh app server (i.e. `Server(app, port=5000).start()`) and embed it in a notebook using `IPython.display.HTML`.  An additional benefit is that it would allow `interact` to work outside of notebooks as well.

gully:
- [ ] Improve the consistency and information in docstrings

barentsen:
I wonder if a work-around for the `localhost:8888` restriction is to serve the tool using a server, i.e. the pseudo-code would look like this:

```
# Launch our app in a bokeh server
bokeh.server.server.Server(interact_tool_app).start()
if we are in a terminal:
    launch_browser('url-to-our-bokeh-app')
elif we are in a notebook:
     display(HTML("""<iframe id="iframe" sandbox="allow-same-origin allow-scripts" width=800 height=800 src="url-to-our-bokeh-app"></iframe>"""))
```

...however I would encourage you to let the bokeh folks know why we think the `localhost:8888` restriction is an annoying problem (=> people tend to run multiple notebook servers), and ask if they think this is the best work-around.  The issues I saw did not convince me that this isn't an issue that could just be resolved in bokeh.

lightkurve/interact.py
```diff
@@ -0,0 +1,394 @@
...
+    """
+
+    lc = tpf.to_lightcurve(aperture_mask=tpf.pipeline_mask)
+
+    n_pixels = tpf.flux[0, :, :].size
+    pixel_index_array = np.arange(0, n_pixels, 1, dtype=int).reshape(tpf.flux[0, :, :].shape)
+
+    # Bokeh cannot handle many data points
+    # https://github.com/bokeh/bokeh/issues/7490
+    if len(lc.cadenceno) > 30000:
```
barentsen: To be future-proof, I recommend we do not hard-code the `30000` limit and make this an optional parameter of the interact function instead?  The default of that parameter can be 30000.

gully:
By the way the tests all fail because bokeh is an optional dependency.  Can/should we add bokeh to the continuous integration?  ping @barentsen

barentsen:
I had a go at improving docstrings, object names, and test coverage.  @gully please review and merge if you are ok with the changes and if the tests pass!

I'd like the test coverage to improve a bit more for the `interact` module, it is still only 55%, but I'm ok with a future PR to improve it, i.e. we can merge now!  (Pro-tip: if you click on the `codecov` link, you can see exactly which lines aren't covered by tests yet.)

barentsen:
@gully Success!  I found a way to make the unit tests pass on both Travis (where we install bokeh) and Appveyor (where we don't).

The final step to merge this PR is to remove the files under `tutorials/under_development/widget_dev/` and update the actual tutorial instead.  Can you do this?

gully:
Tasks:
- [ ] Select a great demo object for the tutorial
- [ ] Make an animated gif for the documentation
- [ ] Update the interact tutorial notebook.

barentsen:
I'm going to go ahead and merge this, but note that we will need to update the tutorial!

barentsen:
Opened #258 so we don't forget to update the tutorial.

PR #227, opened 2018-08-13
Support `targetid` in favor of `keplerid` & `ticid`

barentsen:
I propose to store the KIC/EPIC/TIC ID of an object using the `targetid` attribute, rather than having separate `keplerid` and `ticid` attributes.  This simplifies the code and will make it easier to treat Kepler or TESS data objects in the same way.


barentsen:
We can get rid of `keplerid` right away because we are still in beta.  Once we release 1.0 non-beta proper, we will start adopting a more strict policy to avoid breaking existing code and to maintain a CHANGELOG.

PR #231, opened 2018-08-24
Adding a new option to aperture_mask

mrtommyb:
I added a new option to for the shape of an aperture_mask. In addition to the 'pipeline' and 'all' options, this adds an method that selects the mask by looking for pixels above a threshold, where the threshold is set is sigma -- although I used robust statistics.


mrtommyb:
There are a couple of things that I think need changing. The most glaring of which is that I arbitrarily hard coded that the target star has to be within +/- 2 pixels of the center of the pixel mask. This could be a parameter.

mrtommyb:
Here are a couple of screenshots showing it in action.
screenshot 2018-08-24 12 38 37
screenshot 2018-08-24 12 38 59

barentsen:
Thanks @mrtommyb!  Selecting pixels above a threshold is something I find myself doing a lot by hand, so it makes sense to automate this!  I hope we'll be able to review this asap.

JeffLCoughlin:
For K2 is there an option, or should we consider adding one, that extends the aperture in the direction of motion?

Or maybe create an aperture this way, but on the averaged flux image across the entire campaign?

mrtommyb:
Currently I use a median image to calculate the pixels above threshold, but we could probably instead look at the union of all pixels above threshold in any image.

mrtommyb:
There is a bug in this pull request with the threshold keyword not being passed to the _parse_aperture_mask function. I'm working on a fix. I'm going to need some help with how you want this variable to be passed. I'm going to implement it as part of kwargs.

lightkurve/targetpixelfile.py
```diff
@@ -375,10 +380,60 @@ def _parse_aperture_mask(self, aperture_mask):
...
+
+        #find brightest pix within 5x5 of central pix
+        #this assumes target is close to center of postage stamp
+        median_image_center = median_image[central_pixel[0] - 2:central_pixel[0] + 2,
+                                           central_pixel[1] - 2:central_pixel[1] + 2]
+        brightest_pixel = np.unravel_index(median_image_center.argmax(),
+                                        median_image_center.shape)
+        brightest_pixel_y, brightest_pixel_x = brightest_pixel
+
+        region_value = labeled_values[central_pixel[0] + brightest_pixel_y - 2,
```
christinahedges: Why are we using 5x5 pixels here? Should this contiguous region size be an additional parameter? What if I'm looking at a galaxy, a saturated star or at a faint source in a crowded region?

lightkurve/targetpixelfile.py
```diff
@@ -375,10 +380,60 @@ def _parse_aperture_mask(self, aperture_mask):
...
 
-    def centroids(self, aperture_mask='pipeline'):
+    def _threshold_mask(self, threshold):
+        """Identify region that is above the threshold and contiguous
+        with the central pixel
+
+        Parameters
+        ----------
+        threshold : float
+            A value for the number of sigma above the background to draw the
```
christinahedges: Is number of "sigma above the background" an easy to understand/common threshold?

lightkurve/targetpixelfile.py
```diff
@@ -375,10 +380,60 @@ def _parse_aperture_mask(self, aperture_mask):
...
+        aperture_mask : ndarray
+            2D boolean numpy array containing `True` for selected pixels.
+        """
+
+        # calculate a median image
+        median_image = np.nanmedian(self.flux + self.flux_bkg, axis=0)
+        vals = median_image[np.isfinite(median_image)].flatten()
+
+        # calculate what the theshold value is in counts
+        mad_cut = (1.4826 * MAD(vals) * threshold) + np.median(median_image)
```
christinahedges: Are you assuming that the background is the median of the image here? That isn't always a strong assumption.

lightkurve/targetpixelfile.py
```diff
@@ -394,7 +449,7 @@ def centroids(self, aperture_mask='pipeline'):
         col_centr, row_centr : tuple
             Arrays containing centroids for column and row at each cadence
         """
-        aperture_mask = self._parse_aperture_mask(aperture_mask)
+        aperture_mask = self._parse_aperture_mask(aperture_mask, **kwargs)
```
christinahedges: Docstring needs to be updated to include the new kwargs.

lightkurve/targetpixelfile.py
```diff
@@ -407,7 +462,8 @@ def centroids(self, aperture_mask='pipeline'):
         return col_centr, row_centr
 
     def plot(self, ax=None, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
-             show_colorbar=True, mask_color='pink', style='fast', **kwargs):
+             threshold=3, show_colorbar=True, mask_color='pink', style='fast',
```
christinahedges: Why is threshold now needed in the plot function? Shouldn't we be passing the aperture mask instead? 

Let's have a keyword `aperture_mask` with a default of `tpf.pipeline_aperture`

lightkurve/targetpixelfile.py
```diff
@@ -858,7 +917,7 @@ def mission(self):
         except KeyError:
             return None
 
-    def aperture_photometry(self, aperture_mask='pipeline'):
+    def aperture_photometry(self, aperture_mask='pipeline', **kwargs):
```
christinahedges: Are we happy that we're now passing strings ('pipeline', 'threshold') or boolean arrays to the same keyword?

lightkurve/targetpixelfile.py
```diff
@@ -1375,7 +1434,7 @@ def astropy_time(self):
         """Returns an AstroPy Time object for all good-quality cadences."""
         return btjd_to_astropy_time(btjd=self.time)
 
-    def aperture_photometry(self, aperture_mask='pipeline'):
+    def aperture_photometry(self, aperture_mask='pipeline', **kwargs):
```
christinahedges: Doc string needs updating.

lightkurve/targetpixelfile.py
christinahedges: Docstring needs updating to reflect the new kwargs and 'threshold' option.

lightkurve/targetpixelfile.py
```diff
@@ -346,21 +348,24 @@ def prf_photometry(self):
...
+        aperture_mask : array-like, 'pipeline', 'all', 'threshold', or None
             A boolean array describing the aperture such that `False` means
             that the pixel will be masked out.
             If None or 'all' are passed, a mask that is `True` everywhere will
             be returned.
             If 'pipeline' is passed, the mask suggested by the pipeline
             will be returned.
+            If 'threshold' is passed, all pixels contigious with the brightest 
+            pixel (within +/- 2 pixels of the central pixel) that are above the 
+            threshold (in sigma) are returned.
```
christinahedges: Docstring needs updating to reflect the new "threshold" keyword.

christinahedges:
@barentsen I've just gone through and made my changes on this, can you take a look?

@mrtommyb I have taken out the contiguous region part of this PR. It's harder to understand and shouldn't affect most of our sources.

lightkurve/targetpixelfile.py
```diff
@@ -345,21 +347,26 @@ def prf_photometry(self):
...
         ----------
-        aperture_mask : array-like, 'pipeline', 'all', or None
+        aperture_mask : array-like, 'pipeline', 'all', 'threshold', or None
             A boolean array describing the aperture such that `False` means
             that the pixel will be masked out.
             If None or 'all' are passed, a mask that is `True` everywhere will
             be returned.
             If 'pipeline' is passed, the mask suggested by the pipeline
             will be returned.
+            If 'threshold' is passed, all pixels contigious with the brightest
```
barentsen: typo: contigious => contiguous

lightkurve/targetpixelfile.py
```diff
@@ -345,21 +347,26 @@ def prf_photometry(self):
...
+            pixel (within +/- 2 pixels of the central pixel) that are above the
+            threshold (in sigma) are returned.
+        threshold : int
+            The threshold above which to create the mask, in sigma.
```
barentsen: Let's add: `this parameter is only used if aperture_mask='threshold'`

barentsen:
I made the following changes:
* I applied the `tpf.verb()` rule to this PR and renamed the new method `tpf.create_threshold_mask()`.
* To keep the API clean, I propose that we do not accept `threshold` as a parameter in all functions that accept the `aperture_mask` parameter.  Instead I propose that we require a user to explicitly call `create_threshold_mask()`  if they want to change the threshold, e.g. `tpf.plot(aperture_mask=tpf.create_threshold_mask(5))`.  This makes the API future-proof against other forms of aperture mask creation.
* I rebased on master.

I am happy for this to be merged, but I noticed that the feature which ensure that a mask is contiguous has disappeared.  Is this not important?

mrtommyb:
I think we might want/need to revisit this. The contiguous region part of the code may have been a little complex but I think it important/nessessary. Otherwise you get weird situations like the one below.
@barentsen 

screenshot 2018-11-18 00 15 24

mrtommyb:
The pull request included code to ensure that the theshold region was contiguous

barentsen:
Thanks @mrtommyb!  I agree we should make this smarter.  I opened #337 to discuss the possible solutions.  Would you like to share your thoughts on the best solution there?

PR #235, opened 2018-09-05
Fixed flatten issue

christinahedges:
There was a small issue in the current flatten implementation. 

We break the light curve into segments where there are breaks in time. If these segments were short we were just changing the flatten. Now if there are short segments we just take the median.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -278,6 +278,10 @@ class LightCurve(object):
         """
         if break_tolerance is None:
             break_tolerance = np.nan
+        if polyorder >= window_length:
+            polyorder = window_length - 1
+            log.warning("polyorder must be smaller than window_length, "
+                        "using polyorder={}.".format(polyorder))
         lc_clean = self.remove_nans()
         # Split the lightcurve into segments by finding large gaps in time
         dt = lc_clean.time[1:] - lc_clean.time[0:-1]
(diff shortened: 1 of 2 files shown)


lightkurve/lightcurve.py
```diff
@@ -291,16 +291,13 @@ def flatten(self, window_length=101, polyorder=2, return_trend=False,
...
-                segment_window_length = h - l
-                if (segment_window_length % 2) == 0:
-                    segment_window_length -= 1  # window_length must be odd
-            if segment_polyorder >= segment_window_length:
-                segment_polyorder = segment_window_length - 1
-            trend_signal[l:h] = signal.savgol_filter(x=lc_clean.flux[l:h],
-                                                     window_length=segment_window_length,
-                                                     polyorder=segment_polyorder, **kwargs)
+            # If the segment is too short, just take the median
+            if np.any([window_length > (h - l), (h - l) < break_tolerance]):
```
barentsen: I wonder if it is more clear to write: `if (window_length > (h - l)) or ((h - l) < break_tolerance):`

lightkurve/lightcurve.py
```diff
@@ -291,16 +291,13 @@ def flatten(self, window_length=101, polyorder=2, return_trend=False,
...
+                trend_signal[l:h] = np.nanmedian(lc_clean.flux[l:h])
```
barentsen: Instead of using `nanmedian`, do we want to enforce `polyorder=1`?

PR #238, opened 2018-09-07
[DOC]: Update recover-a-planet tutorial to use astropy bls code

### LINKED ISSUE #202: Update `how to recover a known planet` to use bls.py
The `transit-periodogram` repo that is being used in our `How to recover a planet` tutorial has been updated significantly as part of @dfm's efforts to get it merged into AstroPy (cf. https://github.com/astropy/astropy/pull/7391)

We need to update our tutorial to use the repo from which the AstroPy PR originates, which is:
https://github.com/dfm/bls.py 

We may also want to add a few weasel words to point out that the package is experimental.  We don't want @dfm to be inundated with issues and support request for an experimental package!

mirca:
This simple PR updates the tutorial on how to recover a planet by using  astropy's BLS.
Note that BLS is available only on the dev version of astropy until the next release. 

See http://mirca.github.io/lightkurve/tutorials/2.02-recover-a-planet.html for the updated tutorial.

Fixes #202


barentsen:
Fantastic!  I am going to try to run the notebook here, but it looks great!

barentsen:
This tutorial is looking great, but I have the following concern: anyone who isn't using the very latest dev version of astropy is going to get a `ImportError: cannot import name 'bls'` exception.

This is a problem because our beginner tutorials should be as accessible as possible, i.e. they shouldn't require users to install the dev version of astropy.

I think the solution is to extract the `BoxLeastSquares` from astropy and make it available in a standalone package, so that we can tell users to run `pip install boxleastsquares` until AstroPy v3.1 (?) is released and fairly mainstream.  Perhaps this is something we should discuss and work on with @dfm in NYC later this month? What do you think?

barentsen:
Update: of course `bls.py` is *already* pip-instalable (https://github.com/dfm/bls.py).  The API does look slightly different from the version that was merged into AstroPy, which we may want to update, but otherwise let's edit this tutorial so that it instructs users to use `pip install bls.py`.  The tutorial can also explain that this feature will soon be standard in astropy...

mirca:
> but otherwise let's edit this tutorial so that it instructs users to use pip install bls.py.

I like that idea! I had forgotten about ``bls.py``, thanks!

christinahedges:
I've updated @mirca's PR: the tutorial now has a little text and runs. I've added DFM's bls.py as a dependency for lightkurve, in the future we're going to create a periodogram class for it. bls.py can run in both python2 and python3 and is available now, where as the astropy implementation will only be available in a year or so and is going to work in python 3 only.

dfm:
I think that the astropy implementation should be in v3.2 which should be out in ~2-3 months shouldn't it?

christinahedges:
@dfm even if it comes out pretty quick, it is only in astropy 3.2 which will mean we can't have full lightkurve capability in python 2.7. Do you think it's better to wait for astropy?

I also want this in so I can develop a BLSPeriodogram class, we can switch it out to the astropy implementation in the future.

mirca:
@christinahedges nice! thanks for finishing this PR, chedges!

mirca:
@christinahedges I believe the AppVeyor failure will go away if you add ``pip install bls.py`` to ``appveyor.yml``.

mirca:
I'm closing this in favor of #360  💃

PR #228, opened 2018-08-21
add scatter method to light curves, for more flexible plotting

zkbt:
`lc.plot()` is an awesome quick tool. I thought it'd be nice if there were a `lc.scatter()` parallel to it, which would allow folks to use arrays to color individual markers with the `c=[some-array]` keyword. The particular case I was thinking of for this was to be able to match up actual time (in JD) to phase-folded time across multiple plots (see the [test](https://github.com/KeplerGO/lightkurve/compare/master...zkbt:scatter?expand=1#diff-446974477ccdd6a0c85fd33748597d68R154)). Is this useful?

I wrote this as an entirely new `scatter` function because I was nervous to mess with `plot`. Would you prefer me to compress this into fewer lines of code by adding a `usescatter=True/False` keyword to `plot`, which could then trigger in the (very few) lines where the two functions would differ?

PR #239, opened 2018-09-09
Modified plot() method of the lightcurve class

KenMighell:
In the file lightcurve.py I renamed the old plot method to plot_old.  I have created a new plot method  
which is identical to plot_old *except* that if the new scatter keyword is set to True (False is the  
default) then a *scatter* plot will be made [ax.scatter()] instead of a line plot [ax.plot()].  This will  
be very useful when marking rare events.


barentsen:
Thanks for opening the PR @KenMighell!  Good news: we have been refactoring the `plot` method in PR #228 to support scatter-plots in addition to a new default matplotlib style (which is much prettier than the current default).  I expect that PR will be merged in the next 1-2 days, so I'm going to close this PR for now because #228 will add the scatter-plot ability!

Thanks++

KenMighell:
Hi Geert,

> On Sep 8, 2018, at 7:08 PM, Geert Barentsen <notifications@github.com> wrote:
> 
> Thanks for opening the PR @KenMighell <https://github.com/KenMighell>! Good news: we have been refactoring the plot method in PR #228 <https://github.com/KeplerGO/lightkurve/pull/228> to support scatter-plots in addition to a new default matplotlib style (which is much prettier than the current default).
> 
I have felt for a long time that the old matplotlib style was very clunky.  

> I expect that PR will be merged in the next 1-2 days,
> 
Excellent!  I proposed my solution today because I would need it on Monday.  

> so I'm going to close this PR for now because #228 <https://github.com/KeplerGO/lightkurve/pull/228> will add the scatter-plot ability!
> 
Ok.


Cheers,

-Ken

PR #244, opened 2018-09-11
Rework search interface

nksaunders:
Pulled cone search functionality out of `from_archive()` into its own function. `cone_search()` works for both `KeplerTargetPixelFiles` and `KeplerLightCurveFiles`, and returns a collection (also added a new collection class for `LightCurveFiles`).

### DIFF
--- a/lightkurve/collections.py
+++ b/lightkurve/collections.py
@@ -151,7 +151,11 @@ class LightCurveCollection(Collection):
 
 class LightCurveFileCollection(Collection):
     """Class to hold a collection of LightCurveFile objects.
-
+    
+    Parameters
+    ----------
+    lightcurvefiles : array-like
+        List of KeplerLightCurveFile or TessLightCurveFile objects.
     """
 
     def __init__(self, lightcurvefiles):
--- a/lightkurve/collections.py
+++ b/lightkurve/collections.py
@@ -151,7 +151,7 @@ class LightCurveCollection(Collection):
 
 class LightCurveFileCollection(Collection):
     """Class to hold a collection of LightCurveFile objects.
-    
+
     Parameters
     ----------
     lightcurvefiles : array-like
@@ -174,8 +174,19 @@ class TargetPixelFileCollection(Collection):
         super(TargetPixelFileCollection, self).__init__(tpfs)
 
     def plot_all(self, ax=None):
-        """
+        """Individually plots all TargetPixelFile objects in a single 
+        matplotlib axes object.
 
+        Parameters
+        ----------
+        ax : matplotlib.axes._subplots.AxesSubplot
+            A matplotlib axes object to plot into. If no axes is provided,
+            a new one will be created.
+
+        Returns
+        -------
+        ax : matplotlib.axes._subplots.AxesSubplot
+            The matplotlib axes object.
         """
 
         if ax is None:
--- a/lightkurve/collections.py
+++ b/lightkurve/collections.py
@@ -174,7 +174,7 @@ class TargetPixelFileCollection(Collection):
         super(TargetPixelFileCollection, self).__init__(tpfs)
 
     def plot_all(self, ax=None):
-        """Individually plots all TargetPixelFile objects in a single 
+        """Individually plots all TargetPixelFile objects in a single
         matplotlib axes object.
 
         Parameters
@@ -192,8 +192,11 @@ class TargetPixelFileCollection(Collection):
         if ax is None:
             import matplotlib.pyplot as plt
             _, ax = plt.subplots(len(self.data), 1,
-                                 figsize=(7,(7*len(self.data))))
+                                 figsize=(7, (7*len(self.data))))
 
-        for i,tpf in enumerate(self.data):
-            tpf.plot(ax=ax[i])
+        if len(self.data) == 1:
+            self.data[0].plot(ax=ax)
+        else:
+            for i, tpf in enumerate(self.data):
+                tpf.plot(ax=ax[i])
         return ax
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -716,7 +716,7 @@ class LightCurve(object):
             legend_labels = ax.get_legend_handles_labels()
             if (np.sum([len(a) for a in legend_labels]) != 0):
                 ax.legend()
-                        
+
         return ax
 
     def plot(self, **kwargs):
--- a/lightkurve/search.py
+++ b/lightkurve/search.py
@@ -121,9 +121,11 @@ class SearchResult(object):
                         'cadence to limit your search.'.format(len(self.products)))
 
         # return single tpf or lcf
-        if 'lpd-targ.fits' in self.products['productFilename'][0]:
+        tpf_files = ['lpd-targ.fits', 'spd-targ.fits']
+        lcf_files = ['llc.fits', 'slc.fits']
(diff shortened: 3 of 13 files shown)


barentsen:
Our whole confusion wrt `target_name` vs `object_name` in the MAST astroquery API has appeared before.  It is partially documented in issue #148 and a regression test exists in test_mast.py.  Let's make sure we add a bunch of in-line code comments to explain why we need `target_name`.

barentsen:
We will want to think carefully about what we want the final API to look like.  For example, to keep things easy to understand, we may want to follow the rule that static methods of a class always return a new instance of that class, for example:

* **`KeplerLightCurveFileCollection.query_region(ra, dec)`** always returns a collection.
* **`KeplerLightCurveFileCollection.query_object(name)`** always returns a collection.
* **`KeplerLightCurveFileCollection.query_channel(channel)`** always returns a collection.
* **`KeplerLightCurveFile.from_archive()`** always returns a single file.
* **`KeplerLightCurve.from_archive()`** always returns a single light curve object (which may be stitched together from multiple quarters).

Alternatively, we don't *have* to use static methods.  For example: a totally different API using more simple functions could look like this:
```Python
from lightkurve import query_region
query_region(ra, dec, mission='kepler', product='lightcurvefile')
```
this might return a more generic `QueryResult` object that provides a table of filenames (!) and the optional ability to convert the search result into LightCurveFile objects etc.  This may have the advantage that the results of a query can be inspected *before* all the data products are downloaded?

Does anyone else want to doodle a dream API?  It may be good to draft a few different options (using pseudocode) before deciding.

christinahedges:
I prefer the second idea!

nksaunders:
Agreed, I think a simple function that always returns a `QueryResult` or `TargetList` object would be easier to understand as a user. The table could also include `ra` and `dec` info which would be useful at a glance.

 In the case that `.from_archive()` only returns a single file / light curve object, it should throw a clear error that prompts with available quarters (instead of returning a collection with all available as it currently does).

lightkurve/collections.py
```diff
@@ -149,6 +149,15 @@ def plot(self, ax=None, **kwargs):
         return ax
 
 
+class LightCurveFileCollection(Collection):
+    """Class to hold a collection of LightCurveFile objects.
+
```
barentsen: We'll want to add
```
    Parameters
    ----------
    lightcurvefiles : array-like
        List of KeplerLightCurveFile or TessLightCurveFile objects.
```

lightkurve/lightcurve.py
```diff
@@ -635,7 +635,7 @@ def _create_plot(self, method='plot', ax=None, normalize=True,
             legend_labels = ax.get_legend_handles_labels()
             if (np.sum([len(a) for a in legend_labels]) != 0):
                 ax.legend()
-
+
```
barentsen: Let's destroy these whitespaces =)

lightkurve/search.py
```diff
@@ -0,0 +1,460 @@
...
+    @property
+    def obsID(self):
+        return np.asarray(np.unique(self.path['obsid']), dtype='int')
+
+    @property
+    def target_name(self):
+        return np.asarray(np.unique(self.path['target_name']))
+
+    @property
+    def RA(self):
```
barentsen: We've been using lowercase `ra` elsewhere so we'll probably want to stick to that.

nksaunders:
Current functionality:
```
tpf = search_tpf(ID).download() # TargetPixelFile
lcf = search_lcf(ID).download() # LightCurveFile
tpfc = search_tpf(ID, radius=radius).download_all() # TargetPixelFileCollection
lcfc = search_lcf(ID, radius=radius).download_all() # LightCurveFileCollection
```

lightkurve/search.py
```diff
@@ -0,0 +1,530 @@
...
+import logging
+import numpy as np
+import copy
+
+from astroquery.mast import Observations
+from astroquery.exceptions import ResolverError
+from astropy.table import Table, unique
+from astropy.coordinates import SkyCoord
+from astropy.io import ascii
+from .lightcurve import KeplerLightCurve, TessLightCurve
```
barentsen: We can remove this line because `search.py` never returns or uses `LightCurve` objects

lightkurve/search.py
```diff
@@ -0,0 +1,530 @@
...
+        self.quarter = quarter
+        self.month = month
+        self.cadence = cadence
+        self.filetype = filetype
+        self.quality_bitmask = quality_bitmask
+
+    @property
+    def targets(self):
+        """Returns a table of targets and their RA & dec values produced by search"""
+        mask = ['target_name','s_ra','s_dec']
```
barentsen: PEP8 nitpick: a comma should always be followed by a white-space.

lightkurve/search.py
```diff
@@ -0,0 +1,530 @@
...
+    def download(self):
+        """
+        Downloads a single KeplerTargetPixelFile or KeplerLightCurveFile object from search result.
+        If multiple files are present in `products`, only the first will be downloaded.
+
+        Returns
+        -------
+        KeplerTargetPixelFile : `KeplerTargetPixelFile` object
+            Returns a single `KeplerTargetPixelFile` for first entry in products table
+        KeplerLightCurveFile : `KeplerLightCurve` object
```
barentsen: `KeplerLightCurve` object => `KeplerLightCurveFile` object

barentsen:
@nksaunders At your convenience, can you go to this branch and run `git rebase master` to re-play all the commits in this PR against the current master.  (This will resolve the current merge conflict and make the diff easier to understand.)

If this is your first rebase, come over and we'll do it together.  (Spoiler: git will tell you what to do on the command line after you execute the command.)

nksaunders:
> @nksaunders I believe Travis started failing because of a bug on MAST's end. I have alerted @scfleming. Don't worry about fixing it tonight!

Queries from MAST are working as expected again today!

lightkurve/collections.py
```diff
@@ -159,3 +172,17 @@ class TargetPixelFileCollection(Collection):
     """
     def __init__(self, tpfs):
         super(TargetPixelFileCollection, self).__init__(tpfs)
+
+    def plot_all(self, ax=None):
+        """
```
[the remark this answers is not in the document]
nksaunders: Ok! The collections module also needs to be reworked at some point to give collections more functionality.

barentsen:
@nksaunders Can you do `git reset master lightkurve/mast.py` to ignore the changes to `mast.py` for now?

lightkurve/search.py
```diff
@@ -0,0 +1,531 @@
...
+from astropy.coordinates import SkyCoord
+from astropy.io import ascii
+from .lightcurvefile import KeplerLightCurveFile
+from .targetpixelfile import KeplerTargetPixelFile
+from .collections import TargetPixelFileCollection, LightCurveFileCollection
+
+from . import PACKAGEDIR
+log = logging.getLogger(__name__)
+
+
```
barentsen: Let's enable `from lightkurve import search_targetpixelfile` (rather than `from lightkurve.search import ...`) by doing this:

```suggestion
__all__ = ['search_targetpixelfile', 'search_lightcurvefile']

```

barentsen:
@nksaunders Note that short cadence data has slightly different filenames (suffix `spd-targ` and `slc`, I believe).  You'll want to add those!

barentsen:
Merging in 3 ... 2 ...

PR #246, opened 2018-09-12
kepler quarters

GiantSable:
hi i've just started and worked through the quickstart and am up and running. however i'm baffled by working out which quarter to use for each star from the kepler database. or what star belongs to what quarter. is there a list?


barentsen:
@GiantSable Thank you for your feedback!  The different quarters represent different time periods in the original 4-year mission of Kepler.  You can find the dates listed here:

https://keplerscience.arc.nasa.gov/data-products.html#kepler-data-release-notes

We definitely need to explain this as part of the documentation of Lightkurve.  Many thanks for this feedback, I will open an issue to ensure we do this!

PR #249, opened 2018-09-12
Hotfix for plotting bug

christinahedges:
Plots currently don't work because there is a tiny error in the way we implemented it. Currently they plot nothing. I'm fixing it.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -617,7 +617,7 @@ class LightCurve(object):
             if ax is None:
                 fig, ax = plt.subplots(1)
             if method == 'scatter':
-                sc = ax.scatter(x=self.time, y=flux, **kwargs)
+                sc = ax.scatter(self.time, flux, **kwargs)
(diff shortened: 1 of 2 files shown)


lightkurve/lightcurve.py
```diff
@@ -626,7 +626,7 @@ def _create_plot(self, method='plot', ax=None, normalize=True,
             elif method == 'errorbar':
                 ax.errorbar(x=self.time, y=flux, yerr=flux_err, **kwargs)
             else:
-                ax.plot(x=self.time, y=flux, **kwargs)
```
barentsen: Mea culpa!

barentsen:
Thanks @christinahedges !

lightkurve/lightcurve.py
christinahedges: Actually......That was my fault........

barentsen: :open_mouth:

PR #250, opened 2018-09-13
Fix for scatter plot issue.

christinahedges:
Currently if you try to use

`lc.scatter(c='C3')`

It breaks because it tries to add a colorbar to something that isn't colorbar-able. This fixes the issue by checking if the color kwarg is a string.


lightkurve/lightcurve.py
```diff
@@ -618,7 +618,8 @@ def _create_plot(self, method='plot', ax=None, normalize=True,
                 fig, ax = plt.subplots(1)
             if method == 'scatter':
                 sc = ax.scatter(self.time, flux, **kwargs)
-                if show_colorbar and ('c' in kwargs) and hasattr(kwargs['c'], '__iter__'):
+                if show_colorbar and ('c' in kwargs) and \
+                  (not isinstance(kwargs['c'], str)) and hasattr(kwargs['c'], '__iter__'):
```
barentsen: Let's add a comment to explain that we need this condition to avoid showing a colorbar when a fixed color is set using a string.  The purpose of this if statement will not be obvious otherwise.

barentsen:
Thanks @christinahedges!  Do you think we can add `lc.scatter(c='C3')` to the existing plotting unit test?

lightkurve/lightcurve.py
christinahedges: 👍

christinahedges:
@barentsen can we merge this?

barentsen:
Looks good!  Thanks @christinahedges !

PR #253, opened 2018-09-14
[WIP] Improving SFF implementation, docs, API and tools!

christinahedges:
# What is this?
The current implementation of SFF has been really useful, but it can be improved. I'm working on improvements including

* clearer implementation, with better comments and variable names (including refactoring)
* better diagnostic plots
* better docs and tutorials to show *how* to use the SFF corrector
* a clearer api
* more tools for the user including an ✨ interact ✨ implementation

(*Note: I'll be opening a separate PR for a quick bugfix in the current implementation soon*)

# What needs doing?

- [ ] Clear refactoring
- - [ ] Sketch out the ideal structure
- - [ ] Better comments
- [ ] Better tools
- - [ ] Breakpoint splitter
- - [ ] SFF Interact
- - [ ] GPs instead of a BSpline?
- [ ] New tutorials
- - [ ] How to split a LightCurve into two chunks around the breakpoint
- - [ ] How to SFF with the new API
- [ ] Better diagnostic plots
- - [ ] Plot the spline, and the arclength
- - [ ] Plot windows and before/after fix
- [ ] Way better unit tests
- - [ ] SFF integration tests?


gully:
I like the idea of `GPs instead of a BSpline`.  It may also be worth using a GP for the mapping from (x,y) centroids to arclength, which is presently performed with a polynomial; alternatively a polynomial could be kept, but an automatic testing for polynomial model complexity (AIC or BIC), could help improve the bias/variance tradeoff there.

christinahedges:
This has been captured in the new corrector API, so I'm closing.

PR #255, opened 2018-09-19
Patch for incorrect SFF Bspline

christinahedges:
# What was wrong?

The SFF implementation was using a bspline to remove long term trends on the **windowed** light curves. This was trying to remove "long term trends" using a potentially tiny cut out of the data.

# What have you fixed?

I changed the loop structure of the implementation to quickly fix this specific problem. There are way more fixes coming in #253 !

Note that I'm going to have to change one of our tests, as I have changed the `sff.fit_bspline()` function. 

# What is the effect?
This actually in some cases makes stuff worse. We get sharp edges where the windows end now. I could really do with a little bit more time to debug and finish with #253 but this will help @barentsen in a pinch?

### Before:
image
image

### After
image
image

### Before
image

### After
image


christinahedges:
@barentsen let's get this merged!

barentsen:
Changes look good.  Merging!

Issue #259, opened 2018-09-24
Raise an exception if you try to use from_archive() on KeplerLightCurve 


ekaterinailin:
#### Problem description

... instead of ```KeplerLightCurveFile```. Add a helpful message to the exception.

#### Example

```python
import lightkurve
lc = lightkurve.KeplerLightCurve.from_archive(211117077)

AttributeError: type object 'KeplerLightCurve' has no attribute 'from_archive'
```

#### Expected behavior

Explain the user that they need to use ```KeplerLightCurveFile.from_archive()``` instead.

#### Environment:
- platform (e.g. Linux, OSX, Windows): Ubuntu 16.04
- lightkurve version (e.g. 1.0b6): 1.0b15
- installation method (e.g. pip, conda, source): pip

ekaterinailin:
See pull request #263.

ekaterinailin:
Thank you (and mostly @nksaunders I presume) for implementing the new ``search_targetpixelfile`` and ``search_lightcurvefile`` functions! I am using them in AltaiPony already. They are even easier to build upon than the old ``from_archive`` utils.

barentsen:
That's great to hear @ekaterinailin!   Yes, thank you @nksaunders!!

PR #263, opened 2018-09-25
Added a function from_archive() to KeplerLightCurve()... 

ekaterinailin:
...that throws an exception and points to KeplerLightCurveFile.from_archive(). Fixes issue #259.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1136,6 +1136,23 @@ class KeplerLightCurve(LightCurve):
                                                      overwrite=overwrite,
                                                      **extra_data)
 
+    def from_archive(target, **kwargs):
+        '''
+        A function that can be called like KeplerLightCurve.from_archive()
+        but only throws an exception to indicate that
+        KeplerLightCurveFile.from_archive() should be used instead.
+
+        Parameters:
+        -------------
+        target : str or int
+            KIC/EPIC ID or object name.
+        **kwargs : dict
+            Keywords arguments passed to `KeplerLightCurveFile.from_archive().
+        '''
+
+        raise ValueError("KeplerLightCurve.from_archive() is not available. " \
+                         "Use KeplerLightCurveFile.from_archive() instead.")
+        return
 
 class TessLightCurve(LightCurve):
     """Defines a light curve class for NASA's TESS mission.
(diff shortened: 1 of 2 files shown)


christinahedges:
Hi Ekaterina! Great job with this PR. Could you add a unit test to `test_lightcurve.py` to check this? The test should just check that `KeplerLightCurve.from_archive('Kepler-10')` throws your exception.

christinahedges:
You can write a test for exceptions in a similar way to this:

```python
def test_invalid_lightcurve():
    """Invalid LightCurves should not be allowed."""
    err_string = ("Input arrays have different lengths."
                  " len(time)=5, len(flux)=4")
    time = np.array([1, 2, 3, 4, 5])
    flux = np.array([1, 2, 3, 4])
        LightCurve(time=time, flux=flux)
    assert err_string == err.value.args[0]
```

ekaterinailin:
This is helpful, thank you!

ekaterinailin:
I tried flavors of this one:
```python
    def __getattr__(self, name):
        if name in ["from_archive"]:
            raise AttributeError("KeplerLightCurve.{0} is not available. " 
                                 "Use KeplerLightCurveFile.{0} instead."
                                 .format(name) )
        else:
            raise AttributeError("module '{}' has no attribute '{}'"
                                                .format(self.__class__,name) )
```
But 
```
lc = KeplerLightCurve.from_archive(someID)
```
does not call this overridden version of ```__getattr__```. 

Question is: How does it need to be modified to be called? I think, I ran into the problem that ```from_archive``` would be a ```@staticmethod``` and is not called on an instance of ```KeplerLightCurve```.  

Otherwise this trick should work.

barentsen:
It looks like we have to use an obscure Python trick to make this possible, using the `__metaclass__` attribute.  Details here:  https://stackoverflow.com/questions/3155436/getattr-for-static-class-variables-in-python

christinahedges:
@barentsen  only codecov is failing, can we merge this?

barentsen:
We can't merge yet -- it doesn't quite work just yet (cf. Ekaterina's comment above).

barentsen:
It looks like we are planning to remove `from_archive()` as part of #244, so we can close this for now, unless we back-track on our plans to remove it.

PR #266, opened 2018-09-26
[WIP] Enhancing lightcurve.periodogram()

ojhall94:
We are developing a number of enhancements to the periodgram class(). I have branched off the most recent status of PR #189 and am building upon that work. We have established a number of working points:

Improving from_lightcurve():
+ [x] Allow for more astropy.stats.LombScargle() keywords
+ [x] Allow for seismic modification of the LombScargle transform (i.e. oversmapling, supernyquist)
+ [x] User choice of x-axis units?
+ [x] Allow limits and grids in both frequency and period

Housekeeping:
+ [x] Add repr
+ [x] Add periodogram properties (i.e. nyquist, frequency binwidth etc.)
+ [x] Expand API on existing and new functions
+ [x] Add to_pandas() function
+ [ ] ~~Add from_csv & from_fits functions (compatibility with external pipeline output)?~~
+ [x] All lightcurve properties should be transferred, i.e. 'mission', 'quarter' etc.

Improve visualisations:
+ [ ] Multi axis (days vs microHz) periodogram
+ [ ] Bokehplot interact functionality?
+ [ ] Implement an Echelle diagram

Unit tests:
+ [ ] Expand unit test coverage for new and existing functions

Asteroseismic components:
+ [ ] Implement multiple dnu/numax options
+ [ ] Allow for user-input dnu/numax for use with scaling relations?
+ [ ] Implement uncertainty propagation through the seismic scaling relations
+ [x] Option for SNR spectrum?
+ [ ] Implement user input of corrections to the scaling relations?

Things to think about:
+ [x] Any unique functionality a planet-person would want in their periodogram?
+ [ ] How would this work for ensemble research?
+ [x] Ideally, a user would prefer to use as many quarters as possible, as longer time baseline improves frequency resolution. How do we make this compatible with lightkurves format?

I will be making improvements hopefully roughly in that order, top to bottom. Feel free to chip in!

Note: This is a WIP. Commits may contain comments of questions/working points/reminders that will be removed before final build.


gully:
I can consult on options regarding:

> Bokehplot interact functionality?

Specifically our new `interact.py` framework has been redesigned with modularity in mind, and can be extended for periodograms if a powerful visualization can be conceived.

barentsen:
Thanks for working on this @ojhall94!

ojhall94:
lightcurve.Periodogram periodogram generation and plotting options are now fully functional! Feel free to try them out. Next step is some housekeeping, then on to asteroseismic implementation.

lightkurve/lightcurve.py
```diff
@@ -798,23 +799,66 @@ def to_csv(self, path_or_buf=None, **kwargs):
...
-            The regular grid of frequencies to use.  The frequencies must be
-            in units microhertz.  Alternatively, an AstroPy Quantity object can
-            be passed with any unit of type '1/time'.
+        nterms : int
+            Default 1. Number of terms to use in the Fourier fit.
+        nyquist_factor : int
+            Default 1. The multiple of the average Nyquist frequency. Is
+            overriden by maximum_frequency (or minimum period).
+        samples_per_peak : int
+            The approximate number of desired samples across the typical peak.
```
barentsen: I'm still confused as to what a "peak" is.  Can you explain in a few words in the docs?

lightkurve/periodogram.py
```diff
@@ -2,64 +2,251 @@
 from astropy import units as u
```
barentsen: Let's add one or two lines of comments at the top of `periodogram.py` just to explain what the module is for.

lightkurve/periodogram.py
```diff
@@ -2,64 +2,251 @@
...
+            'estimate_mean_density', 'stellar_params', 'standardize_units']
+
+###Add uncertainties <- review these [astero]
+numax_s = 3090.0 # Huber et al 2013
+deltanu_s = 135.1
+teff_s = 5777.0
+
+class InputError(Exception):
+    """Raised if user inputs both frequency and period kwargs"""
+    pass
```
barentsen: A common convention in Python is to raise a `ValueError` if there is a problem with the values of the input.  I would recommend to raise a `ValueError` here too, for the sake of not requiring the user to have to import a custom exception.  (The `ValueError` is in the standard library and does not need to be imported -- easier to use!)

lightkurve/periodogram.py
```diff
@@ -2,64 +2,251 @@
...
+        The LightCurve from which the Periodogram is computed. Holds all  of
+        the `LightCurve object's properties.
     frequencies : array-like
         List of frequencies.
-    powers : array-like
-        Power measurements.
+    periods : array-like
+        List of periods (1 / frequency)
+    power : array-like
+        The power-spectral-density of the Fourier timeseries.
```
barentsen: If we use `periods` and `frequencies` as plural, should we also use the plural `powers` here?  (With docstring `List of powers of ....`)

On a related note: is `amplitudes` better than `power`, or does amplitude have a different meaning?

lightkurve/periodogram.py
```diff
@@ -2,64 +2,251 @@
...
+    nyquist : float
+        The Nyquist frequency of the lightcurve.
```
barentsen: I recommend specifying the unit expected.  (Probably the same as the unit of the frequencies used as input?  If that's the case, let's say so.)

lightkurve/lightcurve.py
ojhall94: I'm using the language samples_per_peak because this is what is used in astropy.LombScargle. I don't think its a descriptive name, so I have now changed it to 'oversample_factor', with the text:

        oversample_factor : int
            The frequency spacing, determined by the time baseline of the
            lightcurve, is divided by this factor, oversampling the frequency space.
            This parameter is identical to the samples_per_peak parameter in
            astropy.LombScargle()

lightkurve/periodogram.py
ojhall94: You mean with regards to astropy.units? I have included:

"""This module lets us attack a unit to a value or an array of values. This
allows us to keep track of what units our data are in, and easily switch
between different units."""

lightkurve/periodogram.py
ojhall94: I have replaced all instances of InputError with ValueErrors!

lightkurve/periodogram.py
ojhall94: Of the power, nyquist, or both? Either way, I'll make sure to specify their relation to the freq_unit parameter, which is defined just below.

lightkurve/periodogram.py
ojhall94: Re: plural/singular Im not sure what to do here--- I started off writing frequencies and periods but as Im going along I find myself typing 'frequency' and 'period' and having to correct myself quite a lot, which makes me think that that might be the natural thing to type or just me. Im leaning towards changing everything to singular.

Re: amplitudes/power, the units on the y-axis are technically 'power spectral density', but I will have to do a bit more reading before I can justify and resolve this comment.

barentsen:
I bet this isn't news to anyone on this PR, but some of Guy Davies' code is available for reference in this gist: https://gist.github.com/christinahedges/2de4a94fc192e4ff193b733e42d219a4

barentsen:
Also: @dfm was kind enough to post an asteroseismology example here, which will be interesting for inspiration:  https://gist.github.com/dfm/0fba6dae50bbbe110417b0a249ae78d5

lightkurve/periodogram.py
ojhall94: Update:

Re: Plural/singular, I have now changed 'frequencies' to 'frequency' and 'periods' to 'period' to match the nomenclature in lightkurve as a whole.

Re: Amplitudes/Power: while using a LombScargle periodogram with Power Spectral Density normalisation (which we are), referring to the y-axis as 'power' for short and 'Power Spectral Density' in full (as is done when plotting) is appropriate. We can revisit when we include other periodogram methods such as BLS.

lightkurve/periodogram.py
```diff
@@ -1,99 +1,947 @@
...
+        self.frequency = frequency
+        self.period = period
+        self.power = power
+        self.nyquist = nyquist
+        self.frequency_spacing = fs
+        self._format = _format
+        self.meta = meta
+
+        if self.period is None:
+            self.period = 1./self.frequency
```
barentsen: Will we also convert period to frequency if `self.frequency is None`, or should we make `self.frequency` a required argument?  (In which case it should not default to None.)

lightkurve/periodogram.py
```diff
@@ -1,99 +1,947 @@
...
+        {'frequency', 'period'}. Default 'frequency'. Determines the format
+        of the periodogram. If 'frequency', preferred units will be frequency.
+        If 'period', preferred units will be period.
+    meta : dict
+        Free-form metadata associated with the Periodogram.
     """
-    def __init__(self, frequencies=None, powers=None):
-        self.frequencies = frequencies
-        self.powers = powers
+    def __init__(self, lc=None, frequency=None, period = None, power=None,
```
barentsen: I wonder if the signature should be `def __init__(self, frequency, power, [optional attributes...])`, because I imagine the Periodogram does not make sense if we don't have at least `frequency` and `power` -- and all the rest is optional?

lightkurve/periodogram.py
ojhall94: Frequency is now a required argument (as is Power). Period is a calculated property as 1/frequency.

lightkurve/periodogram.py
ojhall94: Agreed, as per our discussion yesterday. Updated in next commit.

ojhall94:
The progress made in this PR has been prepped for merge in #266, and the seismic work is continuing in #282, so this should probably be closed. I'll keep the copy of the branch though!

christinahedges:
Hey Ollie, great job on this! We've merged in a V1.0 of your PR in #277 and so I'm going to close this one. You'll still have access to all your WIP functions and I see you've already opened a new asteroseismology branch. Good work! Keep it up! 👍

barentsen:
Thanks @ojhall94!! :tada:

ojhall94:
Thanks for the hard work tidying up and finding bugs!

PR #274, opened 2018-10-04
A few tiny TESS patches

christinahedges:
Some tiny fixes to TESS problems:

* TessTPF should use the `default` aperture mask by default
* The default mask should exclude ImpulsiveOutliers
* When using `tpf.plot` with Tess TPFs there are often lots of zeros, which wrecks out percentile cut for the automatic scaling in log space. I put in a little fix for that.


lightkurve/utils.py
```diff
@@ -399,7 +399,7 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
     """
     if ax is None:
         _, ax = plt.subplots()
-    vmin, vmax = PercentileInterval(95.).get_limits(image)
+    vmin, vmax = PercentileInterval(95.).get_limits(image[image > 0])
```
barentsen: Do TESS TPFs contain a lot of zeros?  I'm a bit concerned about this change because Kepler has lots of genuine negative flux values due to imperfect background subtraction.  However the other changes all look great, so I'm happy to merge this now and re-evaluate this `image > 0` statement in the future.

PR #275, opened 2018-10-04
Changing BJD to JD in the astropy_time function

CheerfulUser:
The astropy_time function currently makes the BJD time an astropy.time.Time object with the type JD. These JD astropy times will be offset by the delay time to the barycentre, which although small, may be important for some cases. 

I changed the bkjd_to_astropy_time function in utils.py to input JD instead of BJD into astropy.time.Time, following the the procedure in Section 2.3.2 in the Kepler Archive Manual. The change introduced two new variables to the function bkjd_to_astropy_time, being the timeslice and timecorr, which are retrieved from the header and data entry respectively.  

The astropy_time properties were also changed in the targetpixelfile and lightcurvefile classes, to incorporate the two new variables introduced in bkjd_to_astropy_time.

### DIFF
--- a/lightkurve/lightcurvefile.py
+++ b/lightkurve/lightcurvefile.py
@@ -240,7 +240,9 @@ class KeplerLightCurveFile(LightCurveFile):
     @property
     def astropy_time(self):
         """Returns an AstroPy Time object for all good-quality cadences."""
-        return bkjd_to_astropy_time(bkjd=self.time)
+        timecorr = self.hdu[1].data['TIMECORR'][self.quality_mask]
+        return bkjd_to_astropy_time(bkjd=self.time, bkjdcorr=timecorr, 
+            timeslice=self.hdu[1].header['TIMSLICE'])
 
     def get_lightcurve(self, flux_type, centroid_type='MOM_CENTR'):
         if flux_type in self._flux_types():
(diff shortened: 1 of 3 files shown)


barentsen:
@CheerfulUser Thank you for this PR!

You are entirely correct to say that the Kepler Archive Manual explains that the TIMECORR and TIMSLICE keywords are required (quote) *to revert back to non-barycentric times*.  However: I believe that if you apply these, you end up with non-barycentric times in the reference frame of Kepler rather than Earth (and of course this matters because Kepler is far away from Earth).  And I don't think you are ever interested in the time frame of Kepler.

At the bottom of `utils.bkjd_to_astropy_time()` you will see that we set `scale='tdb'` in the astropy Time object, which is short for Barycentric Dynamical Time (TDB):
`return Time(jd, format='jd', scale='tdb')`
you can convert to other timescales, like `utc` or `tt` (both of which are in the Earth frame instead of the barycenter frame), you can simply use the astropy syntax, e.g.:
`tpf.astropy_time.utc`
or
`tpf.astropy_time.tt`.

So I don't think this PR is correct.  However all of what I say above is my personal understanding which has not been validated by data or other people, so I may be very wrong.  (I wonder if we can get @dacmess to weigh in.)

Ryan: if you have specific evidence to suggest that `tpf.astropy_time.utc` does not give you Julian Days in the Earth frame, I would love to know!

CheerfulUser:
Thanks for addressing my request, the current way makes sense. I missed what the scale argument did in astropy Time!

PR #276, opened 2018-10-04
document how to remove outliers asymmetrically

anand-sundaram-zocdoc:
This is my attempt to address https://github.com/KeplerGO/lightkurve/issues/221. I updated the docstring for remove_outliers to explicitly describe usage of the relevant members of the **kwargs passed to sigma_clip and updated the tests for the remove_outliers function as a way to include an example of this kind of usage and demonstrate that it achieves the intended goal described in the issue.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -420,6 +420,15 @@ class LightCurve(object):
         are removed if they are separated from the mean flux by `sigma` times
         the standard deviation.
 
+        Examples
+        --------
+        >>> lc = LightCurve([1, 2, 3, 4, 5, 6, 7, 8], [1, 1, 1000, 1, 1, -1000, 1, 1])
+        >>> lc_clean = lc.remove_outliers(sigma=1, sigma_lower=float('inf'), sigma_upper=1)
+        >>> lc_clean.time
+        [1, 2, 4, 5, 6, 7, 8]
+        >>> lc_clean.flux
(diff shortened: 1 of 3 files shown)


lightkurve/lightcurve.py
```diff
@@ -430,6 +430,16 @@ def remove_outliers(self, sigma=5., return_mask=False, **kwargs):
             were removed. Entries marked as `True` are considered outliers.
         **kwargs : dict
             Dictionary of arguments to be passed to `astropy.stats.sigma_clip`.
+            Examples:
+            ---------
```
barentsen: Oops, it looks like you placed the `Examples` section inside the `Parameters` section.  Instead, what we'll want is something like this:
```
def remove_outliers(...):
    """One-line sentence describing the method.

    Longer explanation of the method.

    Examples
    --------
    >>> some code ...
    expected output of code

    Parameters
    ---------
    param1 : type
         description of param1
    etc...
```

Can you re-format it like this?  Thanks++!

anand-sundaram-zocdoc:
@barentsen I think I addressed your comments now and got most tests to pass. I don't really understand why code coverage is failing, since I only added to the docstring.

barentsen:
Thanks @anand-sundaram-zocdoc!  I gave the docstring an edit and added a unit test.  This is good to merge as soon as Travis and Appveyor are happy

PR #279, opened 2018-10-08
fix error in 3-line summary of commands

keatonb:
In demonstration of how a folded light curve can be produced in three lines, there was a reference to a byproduct of an intermediate step, flat_lc, that was not created in the condensed, 3-line version.


barentsen:
Thanks @keatonb!!

Issue #270, opened 2018-09-30
Can't get data from archive for short candece


AIofHuman:
#### Problem description

Good day! Thank a lot for your library! 
I try to get short cadence data from archive. But get error then I point month. I try to these code: month = [1], month = [1,2,3], month = 'all'. But always get error. Please help!
#### Example

```python
import lightkurve
from lightkurve import KeplerLightCurveFile
quartal = 7
KeplerID = 9146018
lcf1 = KeplerLightCurveFile.from_archive(KeplerID, quarter=quartal,cadence='short',month=[1])

#### Expected behavior
<!-- Describe the behavior you expected and how it differs from the behavior observed in the example. -->

#### Environment:
- platform - Windows 10:
- lightkurve version - 1.0b17
- installation method - pip:

**Comment by christinahedges** [comment]
Hi @AIofHuman! I can't seem to reproduce the error on my end, could you send a copy of the error message that you got?

Thanks!
C

**Comment by AIofHuman** [comment]
Hi Christina! Thanks a lot for you quick answer!
In attach, file copy of notebook with error message. Also I add fits file  KIC 9146018 for first month of 7 quarter.
Best regards,
Boris.

[attach.zip](https://github.com/KeplerGO/lightkurve/files/2435070/attach.zip)

**Comment by christinahedges** [comment]
Hi Boris,

Can you send a screenshot of the error message? Alternately you can upload your notebook [here](https://gist.github.com/) on gist so that we can see it as a webpage.

C

**Comment by AIofHuman** [comment]
Screenshot
![image](https://user-images.githubusercontent.com/41274676/46309222-672c0480-c5c4-11e8-82f7-27642a382428.png)

**Comment by christinahedges** [comment]
Hi Boris,

Thanks for the screenshot. This error isn't reproducible on our end, could you run the following and tell us what version number of lightkurve you are using?

``` python
import lightkurve
print(lightkurve.__version__)
```

AIofHuman:
![image](https://user-images.githubusercontent.com/41274676/46309564-562fc300-c5c5-11e8-80ad-a3715f03dea2.png)

christinahedges:
Hi Boris,

I'm running the same version and can't reproduce the error. 

![image](https://user-images.githubusercontent.com/14965634/46313061-bac63000-c57b-11e8-88c3-bbe1f783d638.png)

Sometimes MAST goes down and we can't download or query for files. Can you try to execute the same code again and verify you have the same error?

[...]

PR #280, opened 2018-10-08
Add --remote-data to Appveyor

barentsen:
A user is having issues with `from_archive` on Windows (cf. issue #270).  This PR enables the `--remote-data` flag on Appveyor to check if it's all working.  We had previously disabled it because it would fail at random times.


christinahedges:
That appveyor error does look similar to the one in #270.

lightkurve/search.py
```diff
@@ -624,6 +624,10 @@ def _filter_products(products, campaign=None, quarter=None, month=None,
         month = np.atleast_1d(month)
         # Get the short cadence date lookup table.
         table = ascii.read(os.path.join(PACKAGEDIR, 'data', 'short_cadence_month_lookup.csv'))
+        # The following line is needed for systems where the default integer type
+        # is int32 (e.g. Windows/Appveyor), the column will then be interpreted
+        # as string which makes the test fail.
+        table['StartTime'] = table['StartTime'].astype(np.int64)
```
barentsen: I borrowed this workaround from https://github.com/astropy/astroquery/pull/838

PR #277, opened 2018-10-05
Improved Periodogram v1.0

christinahedges:
Ollie has improved our periodogram code and created a wicked implementation that is thoroughly documented. I've trimmed it down to a working 'v1.0' of the code that we can merge for now, while he keeps working on more asteroseismology tools.

**TODO**

- [ ] Add unit tests
- [ ] Add tutorial

PR #282, opened 2018-10-09
[WIP] Added asteroseismic functionality to .periodogram()

ojhall94:
This PR continues #266 (closed) and #277 (merged).

The plan is to implement some basic asteroseismic stuff, so we can go to a Mass and Radius for a star, and leave opportunity for expansions later.

**New working points**
What I want to do:
- [ ] Merge master (`lightkurve 1.0.0`) into this pull request.
- [ ] Specify that this is for solar-like oscillators only
- [ ] Log a mention when using a TESS lightkurve object that this is suboptimal for TESS data with only a couple of sectors.
- [ ] Add option to calculate numax using autocorrelation
- [ ] Make sure we can recreate the contents of [this blog](https://ojhall94.github.io/blogs/autocorrelate.html) using only lightkurve (including plots!)
- [ ] Add ability for user to interact with keywords for various estimation methods
- [ ] Rewrite scaling relations (asteroseismology.py) to use the neat, compact version described in Bellinger+19.
- [ ] Write unit tests for scaling relations
- [ ] Write unit tests for seismic functions in Periodogram()
- [ ] Add a simple echelle diagram
- [ ] Provide a thorough test of this method

Stretch goals:
- [ ] Make the ACF and Echelle plots interactive

Important discussions to have:
- [ ] Do we return errors on numax & dnu?

**Old working points:**
+ [x] Implement basic smoothing search for numax
+ [ ] Implement autocorrelation search for numax
+ [x] Implement Empirical search for Dnu
+ [x] Implement autocorrelation search for Dnu (given numax)
+ [x] Implement a .find_[numax/dnu]_diagnostics() which returns key diagnostics of the method used
+ [x] Allow for user-input dnu/numax for use with scaling relations
+ [x] Implement uncertainty propagation through the seismic scaling relations
+ [x] Implement user input of corrections to the scaling relations
+ [x] Add a median smoothing filter to periodogram

Note: This is a WIP. Commits may contain comments of questions/working points/reminders that will be removed before final build.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,27 @@
-1.0b31 (unreleased)
-===================
+
+1.0.0 (2019-04-08)
+==================
+
+This is the first stable release of Lightkurve.  It was prepared with the help of 45 contributors!
+
+This release contains major changes to the ``LombScarglePeriodogram`` class:
+
+- Changed the default behavior of ``LombScarglePeriodogram.from_lightcurve()`` to use ``normalization='amplitude'`` and ``oversample_factor=5`` (the previous defaults were ``normalization='psd'`` and ``oversample_factor=1``). The docstring has been expanded to help users understand these options. [#491]
+
+- Added a ``LightkurveWarning`` to alert users of the changes to the default behavior. [#493]
+
+- Deprecated the ``min_frequency``/``max_frequency`` arguments in favor of ``minimum_frequency``/``maximum_frequency`` to be consistent with the other Periodogram classes. [#478]
+
+- Likewise, deprecated the ``min_period``/``max_period`` arguments in favor of ``minimum_period``/``maximum_period`` to be consistent with the other Periodogram classes. [#478]
+
+Other changes are:
+
+- Improved ``PLDCorrector`` to be more robust against the presence of NaNs. [#479, #488]
+
+- Improved ``search_tesscut`` to avoid crashing in the event of an empty search result, and to ensure that the files it returns carry the search string as the ``targetid`` attribute. [#475, #477]
+
+- Various minor bug fixes. [#488, #490, #494]
+
 
 
 1.0b30 (2019-03-27)
--- a/lightkurve/version.py
+++ b/lightkurve/version.py
@@ -1,3 +1,3 @@
 # It is important to store the version number in a separate file
 # so that we can read it from setup.py without importing the package
-__version__ = "1.0b31.dev"
+__version__ = "1.0.0"


ojhall94:
Note: In this build I have made the numax and delta nu calculations functions of the SNRPeriodogram() object, because they are currently not equipped to deal with non-flattened periodograms. This should be rectified in the future, but not necessarily for this PR.

ojhall94:
periodogram.flatten().estimate_stellar_parameters(`Teff`) now returns a dictionary of values and uncertainties for numax, dnu, input Teff, and stellar Mass, Radius, and log(g) (log10 of the surface gravity) in appropriate astropy units.

I will iterate over the following during the coming month or so. We're almost there!
- [ ] Add option to calculate numax using autocorrelation
- [ ] Add ability for user to interact with keywords for various estimation methods
- [ ] Add diagnostic plots for autocorrelation numax method
- [ ] Add diagnostic plots for autocorrelation dnu method
- [ ] Write unit tests for asteroseismology.py
- [ ] Write unit tests for seismic functions in Periodogram()

barentsen:
Thank you for working on this @ojhall94 !!!

lightkurve/asteroseismology.py
```diff
@@ -0,0 +1,353 @@
...
+from matplotlib import pyplot as plt
+import astropy
+from astropy import units as u
+from astropy import constants as const
+
+# from . import PACKAGEDIR, MPLSTYLE
+log = logging.getLogger(__name__)
+
+__all__ = ['dnu_mass_prior', 'estimate_radius', '_get_radius_err', 'estimate_mass',
+            '_get_mass_err', 'estimate_logg','_get_logg_err']
```
barentsen: By convention, an underscore at the start of a function name means "this is a hidden helper function which a user is never supposed to interact with", so I would remove them from `__all__`, unless you think these functions should be accessible by users, in which case we'd remove the underscore prefix.

ojhall94: Hi Geert,

Thanks for the comment, i'll fix this on the next commit!

ojhall94:
I've been thinking about how this PR fits in with the lightkurve structure; do we really want to include the ability to calculate numax and dnu in lightkurve? In my mind, a comparable ability would be a function to return a basic estimate for the transit depth, which the code doesn't do (as far as I'm aware).

Would it maybe be better to not return dnu and numax, but write a comprehensive tutorial on how to use the Periodogram() final state to compute those values yourself?

Whatever the outcome of this thought is, I think its best if we make a separate PR in which we add the ability to plot_echelle(dnu), as that should be a Periodogram function, and is not necessarily tied to lightkurve's ability to calculate dnu.

barentsen:
Thanks for reflecting on this @ojhall94!  I think there are three criteria that determine whether or not a feature belongs in Lightkurve:

1.  Is the feature a common task which is currently tedious to do?
2. Can we implement a non-controversial method to execute the task?
3. Would adding the feature help new students get started?

I think all three of these are true for estimating numax and deltanu in the case of a strong detection.  (We can care less about low signal-to-noise cases, that turns into a research project!)

ojhall94:
I tried to forcefully update to master so I could rewrite the asteroseismic methods described in this PR, because the infrastructure and my vision has changed so much it made no sense to work with the code I already had, but I guess I broke it.

...I'll be opening a new PR. Sorry everyone!

PR #284, opened 2018-10-09
update mast download directory

### LINKED ISSUE #283: `from_archive` shouldn't create a mastDownload folder every time we use it.
`from_archive()` creates a new mastDownload folder in every working directory you use it in. This means we have lots of trailing 'mastDownload' folders with Kepler-10 downloaded in them about 8 times.

We should instead use the astropy cache directory and store it there.

nksaunders:
resolve #283: `from_archive()` should download to the astropy cache in user's home directory, and not make a new mastDownload directory every time it's called

### DIFF
--- a/lightkurve/mast.py
+++ b/lightkurve/mast.py
@@ -38,7 +38,16 @@ def download_products(products):
     # Note: by default, MAST will only let us download "Minimum Recommended
     # Products" (MRPs), which do not include e.g. Target Pixel Files.
     # We need to set `mrp=False` to ensure MAST downloads whatever we want.
-    dl = Observations.download_products(products, mrp_only=False)
+
+    # check if download directory exists
+    if not os.path.isdir(os.path.expanduser('~')+'/.astropy'):
+        os.mkdir(os.path.expanduser('~')+'/.astropy')
+
+    # download to astropy cache (~/.astropy)
+    download_dir = os.path.expanduser('~')+'/.astropy'
+
+    dl = Observations.download_products(products, mrp_only=False,
+                                        download_dir=download_dir)
     return dl['Local Path']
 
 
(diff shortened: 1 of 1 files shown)


lightkurve/mast.py
```diff
@@ -38,7 +38,16 @@ def download_products(products):
     # Note: by default, MAST will only let us download "Minimum Recommended
     # Products" (MRPs), which do not include e.g. Target Pixel Files.
     # We need to set `mrp=False` to ensure MAST downloads whatever we want.
-    dl = Observations.download_products(products, mrp_only=False)
+
+    # check if download directory exists
+    if not os.path.isdir(os.path.expanduser('~')+'/.astropy'):
+        os.mkdir(os.path.expanduser('~')+'/.astropy')
```
christinahedges: If you don't have write access this will break with a funny error message. Can you add an error message?

lightkurve/mast.py
```diff
@@ -38,7 +38,24 @@ def download_products(products):
...
+    if os.path.isdir(cache_dir):
+        download_dir = cache_dir
+    else:
+        # if it doesn't exist, make a new cache directory
+        try:
+            os.mkdir(cache_dir)
+            download_dir = cache_dir
+        # downloads locally if OS error occurs
+        except OSError:
+            log.warning('Warning: unable to create astropy cache directory. '
```
barentsen: Perhaps tune this warning message to say ".lightkurve-cache" directory instead of "astropy cache" directory.  Then this should be good to be merged!

nksaunders: Fixed!

barentsen:
I confirmed this works on my Linux machine.  I am a bit concerned that Appveyor may not have tested this properly on Windows (where we skip the `--remote-data` tests), but we can look into this as part of #280.  

Merging in 10 ... 9 ... 8 .... 7 ....

PR #285, opened 2018-10-09
Hotfix for plotting style

christinahedges:
Plotting style wasn't being used by default for `lcf.plot()`. It was using the ugly old plot style instead, which was horrifying. This makes it beautiful.


barentsen:
:+1:

Issue #286, opened 2018-10-10
Lightkurve should have better names


christinahedges:
The lightkurve API should have better names. Methods should take the form either `lc.verb()` or `lc.verb_noun()`. Properties should take the form `lc.noun`. This will make it easier for users to understand the API.

**We should use as few different verbs as possible**. More verbs is confusing when you're trying to search for the functions you want. Acceptable verbs:

1. `to_` : indicates a conversion of type
2. `from_` :indicates a new object will be created
3. `estimate_`: indicates values will be built using a method that does not always produce the exact answer.
4. `extract_`: indicates values will be built using a method that **does** always produce the exact answer
5. `show_`: indicates information will be printed
6. `remove_`: indicates cadences/values/properties will be removed from the object

![image](https://user-images.githubusercontent.com/14965634/46705795-fef1aa00-cbe5-11e8-840f-3f87b588e623.png)

barentsen:
This was resolved in #287. Yay!

PR #287, opened 2018-10-10
Rename methods according to the lc.noun and lc.verb() rule

barentsen:
Over in #286 the idea was raised to give our properties and methods more consistent names, by ensuring that properties use nouns (e.g. `lc.ra`) and that methods use verbs (e.g. `lc.flatten()`).

This PR shows the most significant changes that would be needed to implement this change.  Tutorials will need to be updated as well.


lightkurve/convenience.py
```diff
@@ -9,22 +9,22 @@
 
 
 def cdpp(flux, **kwargs):
```
christinahedges: Should we have something that is a method (verb) also be allowed to be a property (noun)? I think this is more confusing? It's estimating every time.

lightkurve/targetpixelfile.py
```diff
@@ -833,7 +826,7 @@ def prf_photometry(self, cadences=None, parallel=True, **kwargs):
         return prfphot
 
     def prf_lightcurve(self, **kwargs):
```
christinahedges: What is a `prf_lightcurve`? It feels like we only need `extract_prf_photometry`? There's no equivalent `sap_lightcurve`.

lightkurve/convenience.py
barentsen: Agreed!  Will change.

lightkurve/targetpixelfile.py
barentsen: I am not going to rename the PRF methods for now and move them to a separate package instead.

barentsen:
I have now also updated the tutorial notebooks where necessary.  This should be good to merge if we agree on the new names!

barentsen:
@gully That's a very valid concern.  I feel like we can still do this now because we are only 6 months in and still in beta, and because we aren't changing names that are very popular.  Of the changes made, I expect that only the renaming of `centroids()` and `cdpp()` would be noticed by users.  We could keep those names and add a `DeprecationWarning` if we think it is important to do so?

mirca:
@gully That concern makes sense to me. But I think as long as we deliver a very nice release notes for the next version the net gain of having a more consistent API will be worth it.

PR #288, opened 2018-10-11
Add Periodogram.smooth() function

ojhall94:
At T'DA5 there as a very strong user case for this function to be included (and not doing so from the offset was a major oversight on my part). 

I have included in this pr:
Periodogram.smooth() with keyword argument `filter_width` in units of frequency. It returns a `smoothed_periodogram`, e.g. in the sense that the command:

`lc.to_periodogram().smooth().plot()`

works as expected. I've also added a unit test.

The function as it stands only works for Periodogram objects with evenly spaced grids of frequency, and making sure it can do so for evenly spaced grids of period will require a bit more thought.

image

### DIFF
--- a/lightkurve/periodogram.py
+++ b/lightkurve/periodogram.py
@@ -13,6 +13,7 @@ from astropy.stats import LombScargle
 from astropy import __version__
 from astropy import units as u
 from astropy.units import cds
+from astropy.convolution import convolve, Box1DKernel
 
 from . import MPLSTYLE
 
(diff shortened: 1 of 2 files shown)


lightkurve/periodogram.py
```diff
@@ -321,6 +322,48 @@ def bin(self, binsize=10, method='mean'):
...
+        #Check to see if we have a grid of evenly spaced periods instead.
+        if np.isclose(np.median(np.diff(self.frequency.value)), fs.value):
+            box_kernel = Box1DKernel(np.ceil(filter_width/fs))
+            smooth_power = convolve(self.power.value, box_kernel)
+            smooth_power = u.Quantity(smooth_power, self.power.unit)
+
+            smooth_pg = copy.deepcopy(self)
+            smooth_pg.power = smooth_power
+        else:
+            raise NotImplementedError("This smoothing function is not yet compatible with a grid of unevenly spaced frequencies.")
```
barentsen: Nit-picking: I'd make this error message more direct by saying: `The frequencies in this Periodogram are unevenly spaced.  The smooth() function requires a grid of evenly spaced frequencies at this time.`

lightkurve/tests/test_periodogram.py
```diff
@@ -79,6 +78,27 @@ def test_bin():
     p = lc.to_periodogram()
     assert len(p.bin(binsize=10).frequency) == len(p.frequency)//10
 
+def test_smooth():
+    '''Test if you can smooth the periodogram and check any pitfalls'''
```
barentsen: Nitpicking: to be consistent with the PEP 8 (code style) and PEP 257 (docstring) conventions, always use double quotes in triple-quoted comments.

lightkurve/periodogram.py
```diff
@@ -321,6 +322,48 @@ def bin(self, binsize=10, method='mean'):
         binned_pg.power = binned_power
         return binned_pg
 
+    def smooth(self, filter_width = 0.1/u.day):
+        """Smooths the power spectrum using a simple one dimensional smoothing
+        filter. This method requires a Periodogram class built using an evenly
+        spaced grid of periods.
```
barentsen: Nitpicking: in the docstring, it is worth saying out loud exactly what operation is being applied, i.e. `convolving with a numpy Box1DKernel`, because many users will probably understand what this means.

barentsen:
Apart from my nitpicky comments, I would be happy for @christinahedges to merge this!

lightkurve/periodogram.py
ojhall94: Done!

lightkurve/tests/test_periodogram.py
ojhall94: Done, and replaced them elsewhere in test_periodogram() also.

lightkurve/periodogram.py
ojhall94: Done!

PR #290, opened 2018-10-12
Combined p._estimate_background() & p.smooth() | p.estimate_snr() is now p.flatten()

ojhall94:
In order to allow the user to interact more with how the Signal-To-Noise spectrum is generated, p.smooth() now contains a keyword 'method', which is either `boxkernel` for smoothing through convolution with a numpy Box1D kernel, or `logmedian`, for a moving median in log10 space. 

p.estimate_snr() is functionally the same, but renamed to p.flatten() to match lightkurve nomenclature, and takes keyword arguments for p.smooth().

Added unit tests for all changes.

image
image

### DIFF
--- a/lightkurve/periodogram.py
+++ b/lightkurve/periodogram.py
@@ -322,18 +322,34 @@ class Periodogram(object):
         binned_pg.power = binned_power
         return binned_pg
 
-    def smooth(self, filter_width = 0.1/u.day):
-        """Smooths the power spectrum by convolving with a numpy Box1DKernel.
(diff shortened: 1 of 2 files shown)


barentsen:
@christinahedges I am happy for you to merge this!

barentsen:
I am going to go ahead and merge this.  We will still have the ability to revise the `Periodogram` class as part of adding the asteroseismology features in the near term.

@ojhall94 Thank you so much for working on this!!

PR #293, opened 2018-10-14
Add Time = NaN warnings to LightCurve and TargetPixelFile

### LINKED ISSUE #252: Trigger a warning if time=NaN are present
If a `TargetPixelFile` is opened with `quality_bitmask=None`, it is likely that some of time values are NaN  (usually these are cadences that have the NO DATA quality flag set).

The presence of cadences with time=NaN break several of the methods on the LightCurve class.  I do think it is valuable to allow the user to not mask out any cadences at all, but we should add a warning, likely in the constructors of LightCurve and TargetPixelFile, to warn users!

JoseALermaIII:
Resolves #252 by adding time = NaN warnings to `TargetPixelFile` and `LightCurve` initializers.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -61,6 +61,9 @@ class LightCurve(object):
             self.time = np.arange(len(flux))
         else:
             self.time = np.asarray(time)
+            # Trigger warning if time=NaN are present
+            if np.isnan(self.time).any():
+                log.warning('Warning: NaN times are present in LightCurve')
         self.flux = self._validate_array(flux, name='flux')
         self.flux_err = self._validate_array(flux_err, name='flux_err')
         self.time_format = time_format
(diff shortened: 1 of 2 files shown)


barentsen:
@JoseALermaIII Thank you for working on this!

I removed the warning from `targetpixelfile.py` for now, because it is very common to have NaN time values in TPFs.  They are usually masked out via the `self.quality_mask` mechanism, so I'm thinking we don't need a warning there for now.  (Which is opposite to what I said in #252!)

The warning in the `LightCurve` constructor looks great.  I am going to merge this as soon as the tests pass.

Again, thank you for contributing!

JoseALermaIII:
@barentsen You're welcome!

It was fun learning about all the different aspects of light curves and the package itself enough to add the warning.

I hope to be able to help more in the future!

PR #298, opened 2018-10-16
Add UserWarning for time=NaN warning in LightCurve and added tests

JoseALermaIII:
Added a `UserWarning` to the time=NaN warning portion of the LightCurve constructor in lightcurve.py so that tests could be added to test_lightcurve.py

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -63,6 +63,7 @@ class LightCurve(object):
             self.time = np.asarray(time)
             # Trigger warning if time=NaN are present
             if np.isnan(self.time).any():
+                warnings.warn('Warning: NaN times are present in LightCurve', UserWarning)
                 log.warning('Warning: NaN times are present in LightCurve')
         self.flux = self._validate_array(flux, name='flux')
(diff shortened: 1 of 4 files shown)


JoseALermaIII:
I think I can test the logging portion using pytest's decorators (if I read the docs correctly). I'll try it out tomorrow

JoseALermaIII:
You got it! I'll check out Astropy's warning class to see which portions are relevant, add it, and remove the call to `logging.warning`.

barentsen:
Thanks @JoseALermaIII!  I think the definition of `LightkurveWarning` will simply be:
```
class LightkurveWarning(Warning):
    pass
```

JoseALermaIII:
Astropy's warning system has a custom `AstropyUserWarning` class for `UserWarning`s that inherits from `AstropyWarning`. 

Since we are keeping the warning classes simple for now, I implemented the `LightkurveWarning` class in `lightcurve.py` and added a test for the `LightkurveWarning` class in `test_utils.py`

barentsen:
Thank you @JoseALermaIII !!

PR #303, opened 2018-10-18
replaced copy.deepcopy(self) with self.copy()

TheWintersKing:
replaced copy.deepcopy(self) with self.copy() except for the one in copy(self) function


barentsen:
Thank you for contributing!!  This issue has now been resolved as part of 8ab980b6fbe25ecdb4ffde3555427b2785868b98.

PR #305, opened 2018-10-19
Implement PLD corrector for target pixel files

nksaunders:
Added a simple version of 2nd order PLD to the correctors module, and made it callable by `KeplerTargetPixelFile` objects with the `correct()` function.


barentsen:
woowwwwww <3

barentsen:
Let's have a discussion in person about the API.  I worry users will be confused that there is both a `TargetPixelFile.correct()` and a `LightCurve.correct()`, even thought it is of course true that PLD needs the pixel data and SFF can run on the light curve.  Anyone has good ideas on how to resolve this potential confusion?

nksaunders:
For consistency with #319, the API of this corrector has been reworked. An usage example currently looks like this:

```python
import lightkurve
from lightkurve.correctors import PLDCorrector

tpf = lightkurve.search_targetpixelfile(247887989).download()
pld = PLDCorrector(tpf)

corrected_lc = pld.correct()
```

`.correct()` takes optional arguments `transit_mask` and `aperture_mask` to improve de-trending.

lightkurve/correctors.py
```diff
@@ -487,3 +490,95 @@ def breakpoints(self, campaign):
         independently. However, this is not implemented yet in this version.
         """
         raise NotImplementedError()
+
+class PLDCorrector(object):
+    """Implements the Pixel Level Decorrelation (PLD) method to de-trend
+    Target Pixel Files.
```
barentsen: Is there literature we should cite here, e.g. the Spitzer paper?

lightkurve/correctors.py
```diff
@@ -487,3 +490,95 @@ def breakpoints(self, campaign):
...
+            Data flux for every time point
+
+        Returns
+        -------
+        corrected_lightcurve : LightCurve object
+            Returns a corrected lightcurve object.
+        """
+
+        if aperture_mask is None:
+            aperture_mask = self.tpf.pipeline_mask
```
barentsen: I recommend replacing this with `aperture_mask = tpf._parse_aperture_mask(aperture_mask)` so that it supports exactly the same syntax as elsewhere (e.g. this will add support for `aperture_mask='threshold'` etc.)

You can still define your own default by e.g. setting `aperture_mask='threshold'` in the function signature on line 505.

lightkurve/correctors.py
```diff
@@ -17,11 +17,15 @@
 from astropy.stats import sigma_clip
 
 from .utils import channel_to_module_output
-from .lightcurve import LightCurve
-from .lightcurvefile import KeplerLightCurveFile
+from .lightcurve import LightCurve, KeplerLightCurve, TessLightCurve
+from .lightcurvefile import KeplerLightCurveFile, TessLightCurveFile
+from .targetpixelfile import KeplerTargetPixelFile, TessTargetPixelFile
 
+from sklearn.decomposition import PCA
```
christinahedges: Is this better than what we use for the CBV corrector? Should we consider updating the CBV corrector in a future PR?

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+        ----------
+        aperture_mask : array-like, 'pipeline', 'all', 'threshold', or None
+            A boolean array describing the aperture such that `True` means
+            that the pixel will be used.
+            If None or 'all' are passed, all pixels will be used.
+            If 'pipeline' is passed, the mask suggested by the official pipeline
+            will be returned.
+            If 'threshold' is passed, all pixels brighter than 3-sigma above
+            the median flux will be used.
+        transit_mask : array-like
```
christinahedges: Can you add that `True` means that the cadence will be used in the analysis.

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+            An array of indices to be included in de-trending model. This should
+            exclude cadences which occur during transit.
+        n_components_first : int
+            Number of first-order PLD components to reduce to with PCA.
```
christinahedges: Could you point to where I might learn about how to choose this number?

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+        # parse aperture
+        aperture = self.tpf._parse_aperture_mask(aperture_mask)
+
+        # find pixel bounds of aperture on tpf
+        xmin, xmax = min(np.where(aperture)[0]),  max(np.where(aperture)[0])
+        ymin, ymax = min(np.where(aperture)[1]),  max(np.where(aperture)[1])
+
+        # crop data cube to include only desired pixels
+        # this is required for superstamps to ensure matrix is invertable
+        flux_crop = self.flux[:, xmin:xmax+2, ymin:ymax+2]
```
christinahedges: Why + 2 here? If I use `aperture_mask='all'` isn't this going to break?

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+        if len(transit_mask) == 0:
+            transit_mask = np.where(self.time)
+        M = lambda x: x[transit_mask]
+
+        #  generate flux light curve from desired pixels
+        lc = M(self.tpf).to_lightcurve(aperture_mask=aperture_mask)
+
+        # set aperture values
+        aperture_vals = np.zeros(aperture_mask.shape)
+        aperture_vals[np.where(aperture_mask)] = 1
```
christinahedges: Slightly shorter
`aperture_vals = np.copy(aperture_mask).astype(int)`

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+
+        # compute raw flux light curve
+        self.aperture_flux = np.array([f*aperture_vals for f in flux_crop]).reshape(len(flux_crop), -1)
+        rawflux = np.sum(self.aperture_flux.reshape(len(self.aperture_flux), -1), axis=1)
+
```
christinahedges: Can be shortened to make this a little easier to read:
```python
     # Zero flux outside aperture
     self.aperture_flux = f * np.atleast_3d(aperture_vals).transpose([2,0,1])
     # Sum up as a function of time
     raw_flux = np.sum(self.aperture_flux, axis=(1, 2))
```

Alternatively:
```python
     aperture_vals = np.atleast_3d(np.copy(aperture_mask).astype(int)).transpose([2,0,1])
     # Zero flux outside aperture
     self.aperture_flux = f *  aperture_vals
     # Sum up as a function of time
     raw_flux = np.sum(self.aperture_flux, axis=(1, 2))
```

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+        # first order PLD
+        f1 = self.aperture_flux / rawflux.reshape(-1, 1)
```
christinahedges: I find these reshapes a little confusing, and I find `np.atleast_Xd` to be a little more explicit, but this is a personal choice.

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+        X = np.hstack([np.ones(X1.shape[0]).reshape(-1, 1), X1, X2])
+
+        # mask transits in design matrix
+        MX = M(X)
+
+        # define gaussian process parameters amp, tau
+        y = M(rawflux) - np.dot(MX, np.linalg.solve(np.dot(MX.T, MX),
+                                np.dot(MX.T, M(rawflux))))
+        amp = np.nanstd(y)
+        tau = 30.
```
christinahedges: Why is the time scale length hard coded here? Is this something that should be a knob we could turn?

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+        X2 = pca.fit_transform(f2)
+
+        # combine them and add a column vector of 1s for stability
+        X = np.hstack([np.ones(X1.shape[0]).reshape(-1, 1), X1, X2])
+
+        # mask transits in design matrix
+        MX = M(X)
+
+        # define gaussian process parameters amp, tau
+        y = M(rawflux) - np.dot(MX, np.linalg.solve(np.dot(MX.T, MX),
```
christinahedges: Please define what `y` is here. Is this the preliminary correction of the motion?

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+                                np.dot(MX.T, M(rawflux))))
+        amp = np.nanstd(y)
+        tau = 30.
+
+        # set up gaussian process
+        kernel = celerite.terms.Matern32Term(np.log(amp), np.log(tau))
```
christinahedges: Is this kernel the standard one to use? Do we need a jitter term for the white noise?

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+                        -1), axis=1)**2)
+
+        # compute
+        A = np.dot(MX.T, np.linalg.solve(sigma, MX))
+        B = np.dot(MX.T, np.linalg.solve(sigma, M(rawflux)))
+        C = np.linalg.solve(A, B)
+
+        # compute detrended light curve
+        model = np.dot(X, C)
+        self.detrended_flux = rawflux - model + np.nanmean(rawflux)
```
christinahedges:
```suggestion
        self.detrended_flux = rawflux - (model - np.nanmean(model))
```

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+
+        # calculate errors (ignore warnings related to zero or negative errors)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", RuntimeWarning)
+            flux_err = np.nansum(self.flux_err[:, aperture_mask]**2, axis=1)**0.5
+
```
christinahedges: If I understand this correctly, at line 634 ish you're calculating the maximum likelihood estimation of `tau` and `amp`. But not computing the errors on those parameters. If we were to calculate the errors on those parameters, we could propagate the error on the model, and then propagate them through to `flux_err`.

This might be a functionality that would make things rather slow, so it might be worth having a switch to turn something like this off.

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+                                  flux_err=flux_err, **keys)
+
+        else:
+            warnings.warn("Input TargetPixelFile not recognized as a Kepler "
+                          "or TESS observation.", LightkurveWarning)
+            return LightCurve(time=self.time, flux=self.detrended_flux,
+                              flux_err=flux_err)
+
+    def diagnose(self, ax=None):
+        """Plot a figure to diagnose performance of PLD correction."""
```
christinahedges: I feel like this should have some more interesting diagnostics than just plotting the results, but I don't have a good intuition for what they would be.

lightkurve/targetpixelfile.py
```diff
@@ -1077,6 +1077,28 @@ def _open_image(img, extension):
             factory.add_cadence(frameno=idx, flux=cutout.data, header=hdu.header)
         return factory.get_tpf(hdu0_keywords=allkeys, ext_info=ext_info, **kwargs)
 
+    def correct(self, method='pld', aperture_mask=None):
```
christinahedges: I know we've had a discussion about this before, but we will now have
```
lc.correct()    # lightkurve.LightCurve
tpf.correct()  # lightkurve.LightCurve
```

@barentsen how do you feel about this? We could be more explicit and do something like this:
```
lc.SFFcorrect()    # lightkurve.LightCurve
tpf.PLDcorrect()  # lightkurve.LightCurve
```

And then calling e.g. `lc.PLDcorrect()` would raise an exception and direct you to use the TPF. I don't feel super strongly about this, but it might make it easier for people to understand.

lightkurve/tests/test_correctors.py
```diff
@@ -71,3 +72,15 @@ def test_sff_corrector():
...
+def test_pld_corrector():
+    # download tpf data for a target
+    target = 247887989
+    tpf = search_targetpixelfile(target).download()
+
+    # instantiate PLD corrector object
+    pld = PLDCorrector(tpf)
+
+    # produce a PLD-corrected light curve
+    corrected_lc = pld.correct()
```
christinahedges: I feel there should be a few more tests, given the amount of code we're adding. We need tests for 

`pld.diagnose()`
`pld.correct(mask=X)`
`pld.correct(difference number of components)`
`tpf = search_targetpixelfile('EPIC NUMBER').download()`
`tpf = search_targetpixelfile('TIC NUMBER').download()`

This would be a nice place to add some scope lightkurves where we know the ground truth and do some test to see we retrieve the correct answer!

christinahedges: `pld.correct(aperture_mask='all')`
`pld.correct(aperture_mask=threshold')`
`pld.correct(aperture_mask=tiny)`

lightkurve/correctors.py
```diff
@@ -487,3 +491,212 @@ def breakpoints(self, campaign):
...
+        # define gaussian process parameters amp, tau
+        y = M(rawflux) - np.dot(MX, np.linalg.solve(np.dot(MX.T, MX),
+                                np.dot(MX.T, M(rawflux))))
+        amp = np.nanstd(y)
+        tau = 30.
+
+        # set up gaussian process
+        kernel = celerite.terms.Matern32Term(np.log(amp), np.log(tau))
+        gp = celerite.GP(kernel)
+        sigma = gp.get_matrix(M(self.time)) + \
```
christinahedges: Add comments here please

nksaunders:
To do: @christinahedges and I discussed opening a new PR to implement a separate Gaussian Process function to the `correctors` module. Simple options include: periodic / non-periodic kernel, timescale (tau).

barentsen:
@nksaunders Note that Travis/Appveyor failed for reasons that are unrelated to your PR.  This has been resolved in meanwhile (#397), so you'll want to do a `git fetch` followed by a `git rebase keplergo/master` to get the tests to pass.

lightkurve/correctors.py
```diff
@@ -487,3 +491,210 @@ def breakpoints(self, campaign):
...
+                n_components_first=25, n_components_second=20, use_gp=True):
+        """Returns a PLD systematics-corrected LightCurve.
+
+        Pixel Level De-Correlation (PLD) was developed by Deming et. al (2015)
+        to remove systematic noise caused by spacecraft jitter for the Spitzer
+        space telescope. It was adapted to K2 data by Luger et. al (2016, 2018)
+        for the Everest pipeline. For a detailed description of PLD, please see
+        the papers by Deming and Luger.
+
+        Our simple implimentation of PLD is performed by first calculating the
```
barentsen:
```suggestion
        Our simple implementation of PLD is performed by first calculating the
```

lightkurve/correctors.py
```diff
@@ -487,3 +491,210 @@ def breakpoints(self, campaign):
...
+            m_i = \sum_l a_l \frac{f_{il}}{\sum_k f_{ik}} + \sum_l \sum_m b_{lm} \frac{f_{il}f_{im}}{\left( \sum_k f_{ik} \right)^2} + \alpha + \beta t_i + \gamma t_i^2
+
+        where $m_i$ is the noise model at time $t_i$, $f_{il}$ is the flux in
+        the $l^\text{th}$ pixel at time $t_i$, $a_l$ is the first-order PLD
+        coefficient on the linear term, and $b_{lm}$ is the second-order PLD
+        coefficient on the $l^\text{th}$, $m^\text{th}$ pixel pair. $\alpha$,
+        $\beta$, and $\gamma$ are the Gaussian Process terms applied to capture
+        long-period variability.
+
+        We perform Principle Component Analysis (PCA) to reduce the number of
```
barentsen:
```suggestion
        We perform Principal Component Analysis (PCA) to reduce the number of
```

barentsen:
@nksaunders The Appveyor (Windows) Python 3 test failures seem to indicate a legitimate problem with PLD on Windows.  Any idea what's going on?  (I recommend Googling the error message.)

The Appveyor Python 2 failure seems to be related to the celerite installation.  Hmmm...  this would potentially break installing Lightkurve on Windows with Python 2, so we'll want to have a closer look there too.

lightkurve/correctors.py
nksaunders: This is a good suggestion, and is likely more within the scope of a standalone GP fitter in `lightkurve`. I'll include it when we add the new GP function.

lightkurve/targetpixelfile.py
nksaunders: I retained the option to call `tpf.correct()`, but all tutorials and examples should use the new corrector API (#319). Let's chat about whether it's worth removing this function altogether.

lightkurve/correctors.py
nksaunders: PCA from `scikit-learn` is what I first learned, and I'm not sure how it compares to what is already used in the `CBVCorrector`. If you have thoughts about how to do a head-to-head comparison we could give it a try!

lightkurve/correctors.py
nksaunders: The available range for these knobs will depend on how many pixels are included in the aperture. Before this gets merged, I want to give some consideration to intelligent defaults based on number of pixels. With that in mind, it should require very little user intervention.

lightkurve/correctors.py
barentsen: `CBVCorrector` doesn't actually carry out PCA, it merely reads in basis vectors from the pipeline-produced CBV files.  Adding an extra dependency (sklearn) comes at some cost, but it doesn't look like numpy or scipy have an easy-to-use PCA tool, so I think I'm fine with this!

barentsen:
@nksaunders See 14c74a7144efc82ff2530fef4685338c8a4af2e1 for a few changes needed to enable math rendering (you can check using `cd docs; make html`).

The docstring of `PLDCorrector.correct()` ends a bit abruptly.  Can you add a few sentences to explain what happens after the PLD model is solved, i.e. how is the noise model applied to the light curve.

barentsen:
The example added to the docstring of `PLDCorrector` broke the tests (due to doctest).  I'm fixing this now...

barentsen:
Tests pass!  Fantastic work @nksaunders.  Thanks++++.  :+1: :+1: :+1:

PR #307, opened 2018-10-24
Add five cadences of a real TESS TPF to `tests/data`

barentsen:
I added a unit test for a real TESS Sector 1 TPF.  To keep the file size small, I edited the original TPF to throw away all cadences except the first five.  I did this as follows:

```Python
from astropy.io import fits
f = fits.open("https://archive.stsci.edu/hlsps/tess-data-alerts/hlsp_tess-data-alerts_tess_phot_00025155310-s01_tess_v1_tp.fits")
f2 = f.copy() 
f2[1].data = f[1].data[0:5]
HDUList(f2).writeto('tess25155310-s01-first-cadences.fits')
```

Assigning to @mrtommyb to review and merge.

### DIFF
--- a/lightkurve/tests/test_targetpixelfile.py
+++ b/lightkurve/tests/test_targetpixelfile.py
@@ -385,7 +385,7 @@ def test_threshold_aperture_mask():
     lc = tpf.to_lightcurve(aperture_mask=tpf.create_threshold_mask(threshold=1))
     assert (lc.flux == 1).all()
 
-
+    
 def test_tpf_tess():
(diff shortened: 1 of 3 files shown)


lightkurve/tests/test_targetpixelfile.py
```diff
@@ -378,18 +378,28 @@ def test_tess_simulation():
     assert np.isnan(tpf.time).sum() == 0
```
christinahedges: If we have this, do we need the "TESS_SIM" data set and `test_tess_simulation()` function?  Can you delete these two things?

PR #309, opened 2018-10-25
Add useful collections functions

nksaunders:
The collections class has the opportunity for useful plotting and analysis functions. This PR aims to expand on its functionality and allows users to work with collections in similar ways to their `TargetPixelFile`, `LightCurve`, and `LightCurveFile` constituent parts.


christinahedges:
We should add some sort of PCA function so I can get all the nearby targets and do a local PCA

christinahedges:
I'm closing because the new collections API has "stitch" and new plotting functions that give the same functionality.

PR #311, opened 2018-10-30
Made Collection much more simple

christinahedges:
Collection was broken, I'd like to use it for an email blaster. This has made it temporarily more simple before we can merge in #309


barentsen:
:+1:  :trolleybus: :poultry_leg: :cow: :pen: :cactus:

PR #312, opened 2018-10-30
Better windows default

christinahedges:
The default for windows in `lc.correct()` is useless, i've set it to 10 which is a reasonable default.


barentsen:
Thank you for this Peel Request @christinahedges!

PR #314, opened 2018-10-30
Suppressed the scipy warning on lc.flatten()

christinahedges:
`lc.flatten()` had a gross error message from scipy.savgol_filter which will apparently be fixed in scipy v1.2.0, until then I've just suppressed it because it is gross.

image

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -318,7 +318,7 @@ class LightCurve(object):
             else:
                 # Scipy outputs a warning here that is not useful, will be fixed in version 1.2
                 with warnings.catch_warnings():
-                    warnings.simplefilter('ignore')
+                    warnings.simplefilter('ignore', FutureWarning)
                     trend_signal[l:h] = signal.savgol_filter(x=lc_clean.flux[l:h],
                                                              window_length=window_length,
                                                              polyorder=polyorder,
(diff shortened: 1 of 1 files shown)


lightkurve/lightcurve.py
```diff
@@ -312,13 +312,18 @@ def flatten(self, window_length=101, polyorder=2, return_trend=False,
...
             if np.any([window_length > (h - l), (h - l) < break_tolerance]):
                 trend_signal[l:h] = np.nanmedian(lc_clean.flux[l:h])
             else:
-                trend_signal[l:h] = signal.savgol_filter(x=lc_clean.flux[l:h],
-                                                         window_length=window_length,
-                                                         polyorder=polyorder,
-                                                         **kwargs)
+                # Scipy outputs a warning here that is not useful, will be fixed in version 1.2
+                with warnings.catch_warnings():
+                    warnings.simplefilter('ignore')
```
barentsen: Let's specifically ignore `FutureWarning`, in case other warnings happen some day which we do want to know about:
```suggestion
                    warnings.simplefilter('ignore', FutureWarning)
```

barentsen:
@christinahedges Please "accept" the suggested change using the fancy new GitHub feature, then please go ahead and merge!

christinahedges:
Suggestion was accepted, merging is occurring.

Issue #165, opened 2018-06-15
from_archive() outputs a message that cannot be silenced.


christinahedges:
Since PR #150 the verbose keyword has been removed from from_archive. However it still prints an ugly message that cannot be silenced. We could capture this information and output it to the logger as debug information if we thought this was important? Or we could just always silence it.

![image](https://user-images.githubusercontent.com/14965634/41491883-912a2154-70b0-11e8-8130-7409c53c907d.png)

barentsen:
But you can silence this with `log.setLevel('WARNING')` right?  So the issue is that astroquery outputs it as `info` rather than `debug`?

I quite like that fact that, when a Python package downloads something, it is verbose about it by default (because downloads often hang your code on a plane etc...)  If setting `log.setLevel()` is too cumbersome, we could re-introduce a `verbose` argument that does it for the user.

christinahedges:
But there's no log level that will get rid of the message above.

barentsen:
You're right.  The issue is that astroquery's `_download_file` produces this message using a progressbar, not the logging module.  Hmmmmm.

gully:
Maybe make an issue / pull request in Astroquery to silence progressbar output upon request?

barentsen:
Right, I think a PR to astroquery would be a solution.  (Though it wouldn't be until the next astroquery release for the fix to become available.)

texadactyl:
When already in the cache ("Found cached file..."), I cannot see this behavior with WARN or ERROR but I can see the reported behavior with INFO or DEBUG.  So, Astroquery obeys its own log level.

If not yet cached locally ("Downloading ..."), then I see this behavior for any level.  The error seems to be that in {Astroquery treetop}/astroquery/query.py `_download_file`, the progress bar is being emitted without being conditioned on the log level.  See line 277 in https://github.com/astropy/astroquery/blob/master/astroquery/query.py.

I filed issue #1179 on the Astroquery github site.

texadactyl:
Astroquery issue #1179 was closed/rejected by Astroquery today saying it was a an Astropy issue.

After looking at the Astropy source code, it is apparent that the ProgressBarOrSpinner() API referenced at line 277 in query.py has no way of silencing it.  It would take a design & documentation change to get an enhancement (new quieter function that obeyed the log levels) through Astropy and then it would take an Astroquery change to use it in query.py, assuming either group had an interest.  I have no history/leverage with them.

[...]

PR #315, opened 2018-10-30
Suppress printed output statements

nksaunders:
In response to #165: added a decorator to `utils` module which redirects printed outputs from a function to `null`, and then restores to default.

### DIFF
--- a/lightkurve/utils.py
+++ b/lightkurve/utils.py
@@ -2,6 +2,7 @@
 from __future__ import division, print_function
 import logging
 import sys
+import os
 import warnings
 
 from astropy.visualization import (PercentileInterval, ImageNormalize,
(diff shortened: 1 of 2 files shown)


lightkurve/utils.py
```diff
@@ -436,3 +438,19 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
 class LightkurveWarning(Warning):
     """Class for all Lightkurve warnings."""
     pass
+
+
+@contextlib.contextmanager
+def suppress():
```
barentsen: Sweet!  Can we call it `suppress_stdout()` to make its purpose more explicit?

lightkurve/utils.py
```diff
@@ -436,3 +438,19 @@ def plot_image(image, ax=None, scale='linear', origin='lower',
...
+    '''
```
barentsen: {nitpicking}
Per [PEP257](https://www.python.org/dev/peps/pep-0257/), a common Python convention is to use double quotes instead of single quotes for docstrings, and to put one-line docstring on one line, i.e.
```suggestion
    """A simple decorator to suppress function print outputs."""
```

barentsen:
@nksaunders Thanks for this!!  Decorators!!  If you rename `suppress` to `suppress_stdout` then I'm happy for you to merge this!

lightkurve/utils.py
nksaunders: Good to know 👍

nksaunders:
@barentsen Take a look at the changes I just made! It's much more decorator-ish now. If you're happy with it I think it's ready to merge

Issue #316, opened 2018-10-31
Bug Report: KeplerTargetPixelFile opens TESS Target Pixel Files 


KenMighell:
#### Problem description

KeplerTargetPixelFile() opens TESS Target Pixel Files instead of throwing an exception.

#### Example

```python
import lightkurve as lk
# Assume tp.fits is a TESS Target Pixel File
tpf = lk.KeplerTargetPixelFile('tp.fits'). # works without thowing an exception  :-(
```
While you can look at the headers without problem, using the to_lightcurve() function
will fail because channel information is missing.

 ```python
import lightkurve as lk
# Assume tp.fits is a TESS Target Pixel File
tpf = lk.KeplerTargetPixelFile('tp.fits')
tpf.to_lightcurve(). # <--- crashes because tp.fits is not a Kepler Target Pixel File
```
See the attached PNG file to see the gory details.

#### Expected behavior

Using KeplerTargetPixelFile to open a TESS Target Pixel File should fail and throw an exception.

#### Environment:
- platform : Max OS X (High Sierra)
- lightkurve version:  1.0b19
- installation method: GitHub source

![bug_report](https://user-images.githubusercontent.com/17578185/47802468-8fa33d80-dced-11e8-8cb3-a34e1e81b2be.png)

barentsen:
Thanks @KenMighell!  This looks like a bug we should fix!

barentsen:
@nksaunders Do you want to have a look?  

My initial thinking is that we may want to issue a `LightkurveWarning` (using `warnings.warn`, cf.  https://github.com/KeplerGO/lightkurve/pull/298#issuecomment-430282055) if the TPF does not have MISSION=TESS in the header?

KenMighell:
I think that the fact that to_lightcurve() crashes with a TESS TPF warrants an exception and not just a warning.

-Ken

barentsen:
My worry about raising an exception is that it then becomes hard for a home-made target pixel files to be loaded (or perhaps even future versions of official TESS products).

Two additional mitigation actions we can consider:
* We could make `to_lightcurve` more robust to not crash if the CHANNEL keyword is missing from the TPF (which is why it crashed).
* Similar to `astropy.io.fits.open()`, we could add a more simple `lightkurve.targetpixelfile.open()` function which detects the file type and returns the appropriate object. This will help users to avoid making the mistake!

KenMighell:
Ok. Warning it is.

-Ken

[...]

PR #317, opened 2018-10-31
Intelligently read in fits files

nksaunders:
#316: Make sure that fits files are read in as the correct type. Add a `targetpixelfile.open()` function to check Kepler or TESS specific keywords and correspondingly open either a `KeplerTargetPixelFile` or `TessTargetPixelFile`.

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -583,13 +583,6 @@ class TargetPixelFile(object):
         return show_interact_widget(self, lc=lc, notebook_url=notebook_url,
                                     max_cadences=max_cadences)
 
-    def open(self, path):
-        """
-        Opens a fits file, detects its type, and returns the appopriate
-        `KeplerTargetPixelFile` or `TessTargetPixelFile`.
-        """
-        pass
-
 
 class KeplerTargetPixelFile(TargetPixelFile):
     """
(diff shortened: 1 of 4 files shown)


nksaunders:
```python
from lightkurve.targetpixelfile import open_fits

open_fits('/path/to/kepler/or/k2.fits') # returns a `KeplerTargetPixelFile`
open_fits('/path/to/tess.fits') # returns a `TessTargetPixelFile`
```

lightkurve/targetpixelfile.py
```diff
@@ -1389,3 +1413,19 @@ def get_bkg_lightcurve(self, aperture_mask=None):
                               flux=np.nansum(self.flux_bkg[:, aperture_mask], axis=1),
                               flux_err=flux_bkg_err,
                               **keys)
+
+def open_fits(path):
```
barentsen: I think we can simply call it `open()`!  Can anyone think of a reason to call it `open_fits()`?

nksaunders: I opted for `open_fits()` over `open()` to preserve the built-in python function `open()`. But if you think this won't be an issue I'm happy to simplify it!

In that case, it should probably throw an exception if `path` doesn't point to a fits file.

barentsen: I think I prefer `lightkurve.targetpixelfile.open()` for consistency with `astropy.io.fits.open()`.

barentsen:
@nksaunders Do you think we should extend `open()` to also detect and read in `LightCurveFile` files?

nksaunders:
@barentsen `open()` now detects `TargetPixelFiles` and `LightCurveFiles` based on header information and returns the corresponding object.

I wasn't sure where to put this function, so I moved it to the `search` module for now. I initially thought it would make sense in `utils`, but as both `lightcurvefile` and `targetpixelfile` depend on `utils`, it created circular imports when I tried to return the correct object. Any thoughts on where it should end up?

lightkurve/tests/test_search.py
```diff
@@ -125,3 +127,17 @@ def test_source_confusion():
     desired_target = 6507433
     tpf = search_targetpixelfile(desired_target, quarter=8).download()
     assert tpf.targetid == desired_target
+
+def test_open():
+    k2_path = os.path.dirname(os.path.abspath(__file__)) + '/data/test-tpf-star.fits'
```
barentsen: I suggest replacing `os.path.dirname(os.path.abspath(__file__))` with `PACKAGEDIR`  (which is defined in `__init__.py`, i.e. you can do `from . import PACKAGEDIR`)

nksaunders:
This looks good to me! @barentsen if you're happy with it, go ahead and merge.

barentsen:
Unit tests are green!  Merging in 3 ... 2 ...

PR #320, opened 2018-11-02
Quick fix for test_search 

christinahedges:
I was getting a broken test in test_search when I ran the tests locally, this fixes that broken test.


barentsen:
Looks like a good fix.  Two questions:
* do we need this function?
* is it tested?

christinahedges:
Yeah, it's useful. It reduces a potentially large table to just the key information.

nksaunders:
👍 this looks good to me!

PR #323, opened 2018-11-02
Increase the speed of lc.bin()

christinahedges:
`lc.bin()` was pretty slow for me so I've made some minor changes to increase speed.

1. When there were no errors we were calculating the standard deviation in each bin. I think this is bad form. We shouldn't be able to magic errors out of thin air. If there are no errors, there should continue to be no errors.
2. We kept calling `np.array_split()` which was becoming a little costly to do each time. 

This is a really minor speed increase of  ~ a factor 2.


barentsen:
These changes look sensible!  I plan to merge this before lunch.

PR #325, opened 2018-11-05
[Bugfix] Allow quality_bitmask to be passed into download

### LINKED ISSUE #321: Cannot pass a quality bitmask using the new search_targetpixelfile API
This breaks with an obscure error:

image

nksaunders:
fix #321: `wrapper` now takes any number of arguments


barentsen:
Thanks @nksaunders !

PR #326, opened 2018-11-05
[Docs] New demo gif 

nksaunders:
updated gif with new search API


nksaunders:
@barentsen if this is merged, will the gif automatically update with the correct web address? I want to make sure it won't break the docs

barentsen:
@nksaunders This won't break anything!  You can verify by building a local version of the docs first, using:
```
$ cd docs
$ make clean
$ make html
```
...if the output of that looks good, you can update the online docs using:
```
$ make upload
```

To update the GitHub README, you will want to use the direct GitHub link (the one you get when clicking the "Raw" button) and copy/paste that URL into README.md.  (If that makes sense?)

Issue #265, opened 2018-09-25
Add a way to have interact export the mask


gully:
The new `.interact()` is great!  It would be even better to have an option that allows reuse of the mask that was generated in the process.  Issue #264 addresses conceivable ways that the mask could be saved.

gully:
Currently thinking about default names for saving the mask:

```
import os

fn = tpf.hdu.filename()
base = os.path.basename(fn)
outname = base.rsplit('_')[0] + '_custom_mask.fits'
```

results in: `ktwo210690913-c13_custom_mask.fits`

Demerit: makes strong assumption for TPF file names (there is exactly one underscore).

JeffLCoughlin:
Maybe replace the .fits on the end with _custom_mask.fits ?  Then there's no assumption on delimeters in the filename.

outname = base.rsplit('.fits')[0] + '_custom_mask.fits'

barentsen:
At the risk of being verbose, I like `base.rsplit('.fits')[0] + '-custom-aperture-mask.fits'`

Real-world example:
```Python
> from lightkurve import search_targetpixelfile
> tpf = search_targetpixelfile("TRAPPIST-1").download()
> print(os.path.basename(tpf.path).rsplit('.fits')[0] + '-custom-aperture-mask.fits')
ktwo246199087-c12_lpd-targ-custom-aperture-mask.fits
```

gully:
Do you think it's misleading to include `targ`, since the output file is no longer a target pixel file?

barentsen:
I'm not worried about that, personally!

barentsen:
The `interact()` tool now has a `Save lightcurve` button which saves both the custom light curve and the aperture mask as a FITS file (cf. #329).  Closing!

Issue #289, opened 2018-10-12
Better support for reading in non-mission produced lightcurve fits files, e.g. to_fits() or HLSPs


gully:
A common use case is: "I want to read in a Vanderburg & Johnson K2SFF lightcurve for a given target-of-interest".  This High Level Science Product (HLSP) has several special features that sets it apart from a common Kepler-pipeline produced KeplerLightCurve or a custom user-produced LightCurve:

1. Existence of many (N_lc ~ 20) lightcurves of the same target with slightly different aperture choices
2. Existence of custom aperture mask that delivered each of the N_lc lightcurves
3. Existence of non-standard derived time-series information, e.g. arclength, adopted masks
4. Existence of additional metadata-- time-setpoints of window boundaries, model complexity of detrending, etc.

Upon discussion with @christinahedges, we think the best path forward would be:
1. Extend the `.from_fits()` reader to support K2SFF data.  *Alternatively* make a custom/standalone K2SFF reader/converter, perhaps in `utils`.
2. Add a `meta` keyword to the LightCurveCollection class as a catch-all destination for any type of metadata you wish to associate with your LightCurveCollection; this `meta` could store the custom aperture masks.  *Alternatively*, store everything in an HDU extension, as was suggested in #264 .

I'm not sure about what to do with point `3` above, but in the very least it could be (awkwardly) crammed into the `meta`--not great, but something.

The mere discussion of support for a specific HLSP presupposes that interacting with HLSPs is a common and desirable operation that lightkurve should support.  If we support this HLSP, why not support others?

gully:
For some clarity:

```python
from astropy.io import fits
demo_path = 'http://archive.stsci.edu/hlsps/k2sff/c16/211300000/94805/'\
            'hlsp_k2sff_k2_lightcurve_211394805-c16_kepler_v1_llc.fits'

hdus = fits.open(demo_path)
```

yields:

```
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU      25   ()      
  1  BESTAPER      1 BinTableHDU    247   3333R x 6C   [D, D, D, D, I, J]   
  2  CIRC_APER0    1 BinTableHDU    247   3333R x 6C   [D, D, D, D, I, J]   
  3  CIRC_APER1    1 BinTableHDU    247   3333R x 6C   [D, D, D, D, I, J]   
[...]

Issue #264, opened 2018-09-25
How to save custom aperture masks?


gully:
Imagine you painstakingly create a custom aperture mask for aperture photometry.  Perhaps you've done so with our slick new `.interact()` method.  Or maybe you've done so programmatically by some other means that may be difficult to reproduce.  What are your options for saving your work?

Chiefly, the options come down to:  

0. Do nothing--- leave it up to the user to save masks with their desired format (status quo)
1. Save just the mask (in some format-to-be-named-later)
2. Save the lightcurve with *e.g.* the `.to_fits()` method with an HDU extension for the mask
3. Save a new target pixel file with a new HDU extension for the mask

JeffLCoughlin:
I think option 2 would be great. Right now for our TPF files HDU extension #2 (the third and final extension) is the pipeline aperture. Adding additional apertures as extension #3, #4, #5, etc. seems natural and convenient. Their format could mimic the existing #2 format.

I believe this would allow the file to be used normally by any software, as the first 3 extensions remain untouched.

lacalaca85:
I second Jeff, No 2 sounds great. Light curve is readily reproducible with the stored custom mask. We struggled with this in pyke where overwriting the mask files was rather easy... 
And we will be able read back the mask from the light curve file onto the original TPF, right?

barentsen:
I'd be fine with #2!  Who would like to volunteer to try and implement it?

gully:
It looks like option 2 wins based on unanimous feedback so far: 3/3 votes.  So then the question becomes implementation, what needs to change in order to "save the lightcurve with e.g. the .to_fits() method with an HDU extension for the mask":

The `.to_fits()` method currently works on LightCurve objects, which presently have no way to know what mask generated them.  It seems to me that there are a few design and practical considerations for moving forward:

1. Modify `to_fits()` to support HDU mask extensions, and/or verify that it's possible and robust.  Do we want to mimic extant formats from, e.g. K2SFF/EVEREST/Kepler Pipeline?
2. Modify LightCurve or its subclasses to support a `.mask_that_generated_this_lightcurve` property (or a more-concise-attribute-name-to-be-decided-upon-later), which could get seamlessly passed to `to_fits()`).  This attribute may not be desirable or may not be possible to support.
3. Decide on the `interact()` user experience: click a button to save? Pass a kwarg-- `interact(export_lightcurve=True)` ? Do we want invisible/automatic triggering upon a Kwarg?
[...]

PR #329, opened 2018-11-08
Add a "Save lightcurve" button to tpf.interact()

gully:
Will address issue #265.


gully:
The export feature works now, but lacks tests.  I added two new kwargs to allow users to save the mask with a custom path name, and to allow startup from an input mask (rather than the pipeline mask, which remains the default if not startup mask is given).

Here is a demo of an "advanced patterns with interact" tutorial.  I propose two tutorials.  A simple interact tutorial similar to the existing one, and an advanced patterns one.  We should redo the advanced patterns for a TPF with two targets in the aperture, and show how to get lightcurves from both targets with the next export feature.

https://gist.github.com/gully/0b4cb33613305df12f273aa6a4e5f0e4

lightkurve/targetpixelfile.py
```diff
@@ -550,7 +550,8 @@ def to_fits(self, output_fn=None, overwrite=False):
             output_fn = "{}-targ.fits".format(self.targetid)
         self.hdu.writeto(output_fn, overwrite=overwrite, checksum=True)
 
-    def interact(self, lc=None, notebook_url='localhost:8888', max_cadences=30000):
+    def interact(self, notebook_url='localhost:8888', max_cadences=30000,
+                startup_aperture=None, exported_filename=None):
```
barentsen: I think we'll want to name the parameter `aperture_mask` instead of `startup_aperture`,  to be consistent with other methods that take the same parameter?

lightkurve/lightcurve.py
```diff
@@ -1227,7 +1240,7 @@ def to_pandas(self, columns=['time', 'flux', 'flux_err', 'quality',
         """
         return super(KeplerLightCurve, self).to_pandas(columns=columns)
 
-    def to_fits(self, path=None, overwrite=False, **extra_data):
+    def to_fits(self, path=None, overwrite=False, aperture_mask=None, **extra_data):
```
barentsen: All the data written by `to_fits` is held as class attributes.  I think we'll want the same with `aperture_mask`, i.e. I think we'll want `aperture_mask` to become an optional attribute that can be passed to the constructor of `LightCurve`?  If set, `to_fits` would then write `self.aperture_mask` as an optional extension.

lightkurve/interact.py
```diff
@@ -340,6 +350,7 @@ def create_interact_ui(doc):
                                 width=490)
         r_button = Button(label=">", button_type="default", width=30)
         l_button = Button(label="<", button_type="default", width=30)
+        export_button = Button(label="Export Mask", button_type="success", width=120)
```
barentsen: Should this button be named `Save Lightcurve`?  (Given that this button is really creating a new light curve file with aperture mask extension, which I think is the right thing to do!)

lightkurve/interact.py
```diff
@@ -263,8 +264,18 @@ def stretch_change_callback(attr, old, new):
 
     return fig, stretch_slider
 
+def make_default_export_name(tpf, suffix='custom-aperture-mask'):
```
barentsen: Should this be `suffix='custom-lc'`, given that the button is creating a light curve file with aperture mask extension?

lightkurve/targetpixelfile.py
gully: ✅

lightkurve/lightcurve.py
gully: `aperture_mask` has the distinction of being the only metadata that is ndarray, and requires a special ImageHDU extension.  The `extra_data` keywords assume inputs are vectors or scalar/text, so `aperture_mask` will still require some special handling, but you're right that it need not be a standalone kwarg.

lightkurve/interact.py
gully: Ideally, yes.  As it presently stands, lightkurve offers no convenient way to read in the product of a `to_fits()` operation (see #289), at least not while preserving metadata.  The word `mask` was chosen over `lightcurve` to encourage a pattern of mask reuse, rather than lightcurve reuse, for now.  I had considered zeroing out the time and flux hdu extension as a safeguard, too, but decided a nudge was adequate.

lightkurve/interact.py
gully: see comment above.

gully:
FYI, Travis is failing for known, good reasons:  we removed the non-functional `lc` kwarg, a holdover from the previous interact version, before pixel selection was possible.

lightkurve/interact.py
barentsen: We have a unit test which tests that the output of `LightCurve.to_fits()` can be read in by `KeplerLightCurveFile`.  It's annoying that `.PDCSAP_FLUX` won't necessarily work for such object (not written by default), but otherwise we have a well-functioning way to read in data written using `to_fits()`, I think!  I'd go ahead and make this a full `Save Lightcurve` feature, and work on improving the generic `LightCurveFile` class in a separate PR?

barentsen:
@gully Let's try to get this feature merged this week.  I'm making a release plan for v1.0 and would like to include this!

lightkurve/lightcurve.py
gully: I am going to leave `aperture_mask` as a new keyword to `to_fits`, to accelerate the current atomic pull request, and open issue #341 to add an optional `aperture_mask` class attribute to `LightCurve`.

barentsen: 100% agreed!

gully:
The pixel selection is no longer working, and the jump-to-cadence-upon-click yields unusual replotting behavior.  I think something changed in bokeh version 1.0.1, or the introduction of the aperture_mask kwarg...

gully:
The bokeh version 1.0 change to selection callbacks has eliminated our ability to apply clever deselection recursion tricks: it is now impossible with our current strategy to deselect the **most-recently selected pixel**.  The gif animation below shows a demo of the inability to deselect the most-recently selected pixel.
20180920_callback_fail

gully:
There are certainly ways to overcome this pixel-deselection-limitation, but all the ones I can think of require either hacks with strange side effects, or an overhaul of how we structure interact.  I'll keep thinking, and other ideas are welcomed.

gully:
Ready for review @barentsen

gully:
More todo:
- [ ] Get codecov to be 💚  instead of 🔻 
- [ ] run `tpf._parse_aperture_mask(aperture_mask)` on input masks to support `aperture_mask='threshold'`
- [ ] Rebase onto the the quick-fix for bokeh v1.0.0 to get the nice warnings.
- [ ] Update the conda-forge version to demand bokeh v1.0 or greater.
- [ ] Ask folks whether the no-deselection-of-last-selected-pixel is a deal-breaker.

lightkurve/interact.py
```diff
@@ -299,14 +312,13 @@ def show_interact_widget(tpf, lc=None, notebook_url='localhost:8888', max_cadenc
...
-    if lc is None:
-        lc = tpf.to_lightcurve(aperture_mask=tpf.pipeline_mask)
-    else:
-        if len(lc.time) != len(tpf.time):
-            log.error("The custom lightcurve provided to interact() does not contain "
-                      "the same number of cadences ({}) as the target pixel file ({})."
-                      "".format(len(lc.time), len(tpf.time)))
-            return None
+    if aperture_mask is None:
+        aperture_mask=tpf.pipeline_mask
```
christinahedges:
```suggestion
        aperture_mask = tpf.pipeline_mask
```

lightkurve/interact.py
```diff
@@ -388,22 +396,39 @@ def go_left_by_one():
             if existing_value > np.min(tpf.cadenceno):
                 cadence_slider.value = existing_value - 1
 
+        def export_mask():
+
```
christinahedges: Docstring? Or no extra lines.

lightkurve/interact.py
```diff
@@ -388,22 +396,39 @@ def go_left_by_one():
...
+            if tpf_source.selected.indices != []:
+                selected_indices = np.array(tpf_source.selected.indices)
+                selected_mask = np.isin(pixel_index_array, selected_indices)
+                lc_new = tpf.to_lightcurve(aperture_mask=selected_mask)
+                lc_new.to_fits(exported_filename, overwrite=True,
+                               aperture_mask=selected_mask.astype(np.int),
```
christinahedges: You're passing an aperture into a light curve fits file. This is not standard behavior. Is this going to confuse users between `LightCurveFiles` and custom light curve fits files? Can you read in a custom light curve fits file with light curve? Then how do you access the mask?

Maybe we should chat about this as a group?

christinahedges:
Also this needs more unit tests, and we need more tests for the `to_fits()` method that test the aperture mask function.

lightkurve/interact.py
gully: @christinahedges Some related discussions at #264 and to-a-lesser-extent at #265

gully:
It's not clear if the `aperture_mask` kwarg in `to_fits` will stay-- Geert recommends making it an optional attribute #341

lightkurve/interact.py
```diff
@@ -263,8 +266,18 @@ def stretch_change_callback(attr, old, new):
 
     return fig, stretch_slider
 
+def make_default_export_name(tpf, suffix='custom-lc'):
+    '''makes the default name to save a custom intetract mask'''
```
barentsen: Nitpick: pep8 includes a convention to always use double quote characters for triple-quoted strings/docstrings.

lightkurve/interact.py
barentsen: Following discussions, I think we decided to go ahead and accept `aperture_mask` as an argument to `to_fits`.

Note: it would be fair for `lc.to_fits` to accept both a boolean or an integer mask.  However in the case of the integer mask we should follow the standard Kepler/TESS bit convention.  In this current PR we pass `selected_mask.astype(np.int)` which will be incorrect (i.e. selected pixels will show value `1` instead of `3`).

lightkurve/lightcurve.py
```diff
@@ -949,7 +949,7 @@ def to_periodogram(self, nterms=1, nyquist_factor=1, oversample_factor=1,
                                            freq_unit=freq_unit,
                                            **kwargs)
 
-    def to_fits(self, path=None, overwrite=False, **extra_data):
+    def to_fits(self, path=None, overwrite=False, aperture_mask=None, **extra_data):
         """Writes the LightCurve to a FITS file.
 
         Parameters
```
barentsen: Please add an explanation of `aperture_mask` to the docstring as part of the `Parameters` section, clarifying that we accept both a boolean mask or an integer-mask, that an integer mask must always follow the Kepler/TESS convention, and that a boolean mask will be converted to an integer-mask when being stored in the FITS file.

lightkurve/tests/test_targetpixelfile.py
```diff
@@ -369,7 +369,6 @@ def test_interact():
     for tpf in [KeplerTargetPixelFile(filename_tpf_one_center),
                 TessTargetPixelFile(filename_tess)]:
         tpf.interact()
-        tpf.interact(lc=tpf.to_lightcurve(aperture_mask='all'))
```
barentsen: Do we expect this to break?  If not, let's keep it in the unit test?

lightkurve/interact.py
```diff
@@ -299,14 +312,13 @@ def show_interact_widget(tpf, lc=None, notebook_url='localhost:8888', max_cadenc
...
-    if lc is None:
-        lc = tpf.to_lightcurve(aperture_mask=tpf.pipeline_mask)
-    else:
-        if len(lc.time) != len(tpf.time):
-            log.error("The custom lightcurve provided to interact() does not contain "
-                      "the same number of cadences ({}) as the target pixel file ({})."
-                      "".format(len(lc.time), len(tpf.time)))
-            return None
+    if aperture_mask is None:
+        aperture_mask = tpf.pipeline_mask
```
barentsen: I recommend removing the above two lines in favor of:
```aperture_mask = tpf._parse_aperture_mask(aperture_mask)```

This will enable all the traditional masks to be supported.  To ensure that the pipeline mask is the default, change `aperture_mask=None` into `aperture_mask='pipeline'` in the function signature.

File: lightkurve/targetpixelfile.py [hunk 12]
```diff
@@ -569,10 +568,20 @@ def interact(self, lc=None, notebook_url='localhost:8888', max_cadences=30000):
         max_cadences : int
             Raise a RuntimeError if the number of cadences shown is larger than
             this value. This limit helps keep browsers from becoming unresponsive.
+        aperture_mask: np.ndarray
+            An aperture mask to be used as the starting mask at the beginning of
+            a new .interact() visualization. Defaults to the Kepler Pipeline Mask.
```
barentsen: I recommend copying the docstring for `aperture_mask` from `tpf.extract_aperture_photometry()` or `tpf._parse_aperture_mask()` and pasting it here.  I think it will benefit clarity if every function that accepts an `aperture_mask` as a parameter accepts exactly the same options and has exactly the same docstring.

barentsen:
I left a number of fairly small comments above.  @gully Do you want to have a go at reviewing them?  Also please rebase on master. Thx+++.

lightkurve/lightcurve.py
gully: OK, I added the docstring info and shifted aperture mask in preparation for custom Tess/Kepler behavior in commit 2bf8e91.  Still TODO: add the function that parses a mask into the different TESS and Kepler conventions.

lightkurve/tests/test_targetpixelfile.py
gully: Yes, we've deprecated the input custom lightcurve "lc" kwarg with the new pixel selection.

lightkurve/lightcurve.py
gully: OK, we can now accept a boolean mask, and convert it to a bitmask.

lightkurve/utils.py
```diff
@@ -511,3 +511,40 @@ def detect_filetype(header):
     # if one of them is Undefined we expect `.lower()` to yield an AttributeError.
     except (KeyError, AttributeError):
         return None
+
+
+def boolean_mask_to_bitmask(aperture_mask):
```
barentsen: This function is only used in `lightcurvefile.py` and there are no use cases that require end users to call this function, at least none that I can think of right away?
For these reasons, I wonder if we want to move this function to `lightcurvefile.py` and make it a hidden function.  End users would still be able to call this function if they *really* wanted to, but I think our public API will be easier for users to navigate if we hide functions like this from the visible API.  Thoughts?

lightkurve/utils.py
```diff
@@ -511,3 +511,40 @@ def detect_filetype(header):
...
+    ----------
+    aperture_mask : array-like
+        2D aperture mask. The mask can be either a boolean mask or an integer
+        mask mimicking the Kepler/TESS convention; boolean or boolean-like masks
+        are converted to the Kepler/TESS conventions.
+
+    Returns
+    -------
+    bitmask : numpy uint8 array
+        A bitmask mimicking the Kepler/TESS convention
```
barentsen: We will want the docstring to describe exactly what is being set in the returned mask, i.e. `3` where `aperture_mask == 1` and 1 otherwise?  We don't implement the full convention etc.

lightkurve/interact.py
```diff
@@ -265,7 +275,18 @@ def stretch_change_callback(attr, old, new):
     return fig, stretch_slider
 
 
-def show_interact_widget(tpf, lc=None, notebook_url='localhost:8888', max_cadences=30000):
+def make_default_export_name(tpf, suffix='custom-lc'):
+    """makes the default name to save a custom intetract mask"""
+    fn = tpf.hdu.filename()
```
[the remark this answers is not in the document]
gully: Fixed in caef960

lightkurve/utils.py
gully: fixed in 980c8d0

gully:
We're currently testing some `to_fits()` behavior by saving some fits files locally during `pytest`.  Folks probably don't want extraneous fits files appearing whenever they run `py.test`, so we should probably delete these files automatically.  We should also either check for existing files, or use esoteric names so that we don't overwrite and then delete a bona fide fits file with the same name.  

None of this matters for Travis, which gets a fresh environment every single cycle.

lightkurve/utils.py
barentsen: I guess I meant `lightcurve.py`. It should probably just be a hidden function there?

barentsen:
* Can we figure out a way to add a `bokeh.Div` object next to the button, which would have its text set to "Saved {filename}` after the button is clicked?

* We may want to add the following to `TargetPixelFile._parse_aperture_mask()`:
```
if mask & 2 > 0:
    return lcf.hdu[2].data & 2 > 0
```

gully:
Updates to interact button user interaction: clicking the button prints a message that includes the file name.  Re-clicking the button with no change in aperture grays out the message.  Selecting a new set of pixels removes the message until saved again.  Attempting to save no pixels warns the user of saving nothing, changing the color to orange until another click of the mask.  Not shown: deselection works and has the same behavior for all but the *last-selected-pixel*.

20181113_interact

barentsen:
Thanks @gully. I gave it another test and I think this is ready to be merged.  Great work!

As always, there are plenty of opportunities for future improvements, most notably improving our documentation and support for custom light curve files, and increasing test coverage (which is tricky for bokeh apps).  We will no doubt be making progress on those things in the future, one step at a time!

Merging in 3 ... 2 ... 1 ....

PR #334, opened 2018-11-14
[Hotfix] Fix `pos_corr2` property name

nksaunders:


barentsen:
Thanks, that makes sense!

PR #335, opened 2018-11-16
Ensure default `column` and `row` parameters are defined

nksaunders:
When a `KeplerTargetPixelFile` is generated by the `KeplerTargetPixelFileFactory`, the default header has no defined value for column and row (it's set to `astropy.io.fits.card.Undefined`). 

This PR ensures that the output of the factory has the required header keywords to perform basic tasks like plotting.


barentsen:
LGTM!  @nksaunders please merge once Travis and Appveyor are green!

PR #338, opened 2018-11-19
Mitigate 60 warnings triggered by the unit tests

barentsen:
This PR catches or avoids approximately 60 warnings of various types (mostly `LightkurveWarning`) that were being raised by various unit tests.

To remain sensitive to any unexpected warnings appearing, it is important that we explicitly deal with or silence all harmless warnings.

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -90,6 +90,20 @@ class TargetPixelFile(object):
         else:
             self._hdu = value
 
+    def get_keyword(self, keyword, hdu=0, default=None):
+        """Returns a header keyword value.
+
+        If the keyword is Undefined or does not exist,
+        then return ``default`` instead.
+        """
+        try:
+            kw = self.hdu[hdu].header[keyword]
+        except KeyError:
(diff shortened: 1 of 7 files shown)


barentsen:
@christinahedges Can you look over the changes (as a sanity check) and merge?

lightkurve/targetpixelfile.py
```diff
@@ -98,42 +112,22 @@ def header(self):
...
-            out = self.hdu[1].header['1CRV5P']
-        except KeyError:
-            out = 0
-        # ensure output has a value
-        if isinstance(out, fits.card.Undefined):
-            return 0
-        else:
-            return out
+        """CCD pixel column number ('1CRV5P' header keyword)."""
+        return self.get_keyword('1CRV5P', hdu=1, default=0)
```
nksaunders: This is a much more elegant way of handling missing header values. Nice!

barentsen: I should have made a separate PR for this, but yes, let's use the same logic for all keywords by having a single `get_keyword` method!

Issue #125, opened 2018-05-15
Add a `plot_kepler_ffi()` convenience function


barentsen:
@christinahedges prototyped the implementation as follows:

```Python
from astropy.io import fits
import numpy as np
import matplotlib.pyplot as plt

def plot_ffi(filename, **kwargs):
    ffi = fits.open(filename)
    n=10
    fig, ax = plt.subplots(n, n, figsize=(30, 30), sharey=True, sharex=True)
    for i in range(n):
        for j in range(n):
            ax[i,j].axis('off')
            ax.set_facecolor('black')
    plt.xticks([])
    plt.yticks([])
    plt.subplots_adjust(hspace=0.02, wspace=0.02)


    #4, 3, 1, 2
    ax[0, 2].imshow(ffi[4].data[::-1, :], origin='top left', **kwargs)
    ax[0, 3].imshow(ffi[3].data[::-1, ::-1], origin='top right', **kwargs)
    ax[1, 2].imshow(ffi[1].data, origin='bottom left',**kwargs)
    ax[1, 3].imshow(ffi[2].data[:, ::-1], origin='bottom right', **kwargs)

    ax[2, 2].imshow(ffi[20].data[::-1, :], origin='top left', **kwargs)
    ax[2, 3].imshow(ffi[19].data[::-1, ::-1], origin='top right', **kwargs)
    ax[3, 2].imshow(ffi[17].data, origin='bottom left',**kwargs)
    ax[3, 3].imshow(ffi[18].data[:, ::-1], origin='bottom right', **kwargs)

    ax[0, 4].imshow(ffi[8].data[::-1, :], origin='top left', **kwargs)
    ax[0, 5].imshow(ffi[7].data[::-1, ::-1], origin='top right', **kwargs)
    ax[1, 4].imshow(ffi[5].data, origin='bottom left',**kwargs)
    ax[1, 5].imshow(ffi[6].data[:, ::-1], origin='bottom right', **kwargs)

    ax[2, 4].imshow(ffi[24].data[::-1, :], origin='top left', **kwargs)
    ax[2, 5].imshow(ffi[23].data[::-1, ::-1], origin='top right', **kwargs)
    ax[3, 4].imshow(ffi[21].data, origin='bottom left',**kwargs)
    ax[3, 5].imshow(ffi[22].data[:, ::-1], origin='bottom right', **kwargs)

    ax[0, 6].imshow(ffi[12].data[::-1, :], origin='top left', **kwargs)
[...]

PR #342, opened 2018-11-19
[WIP] Add a simple `FullFrameImage` class

barentsen:
While diagnosing a target with a funny lightcurve, I experienced the need to inspect the area surrounding the target in the FFI data.  So I built a simple `FullFrameImage` class that makes it easy to display an FFI cutout given a sky coordinate.

We will want to reflect carefully about if and how such a class might belong in Lightkurve, so I'm just opening this PR as a demo.  Code from #125 could find a home in this PR too.

Demo:

```Python
from lightkurve import FullFrameImage
ffi = FullFrameImage("https://archive.stsci.edu/pub/kepler/ffi/kplr2012121122500_ffi-cal.fits")
ffi.plot_cutout(skycoord)
```
screenshot from 2018-11-19 15-46-28

### DIFF
--- a/lightkurve/__init__.py
+++ b/lightkurve/__init__.py
@@ -21,3 +21,4 @@ from .convenience import *
 from .periodogram import *
 from .collections import *
 from .search import *
+from .ffi import *
--- a/lightkurve/ffi.py
+++ b/lightkurve/ffi.py
@@ -1,0 +1,34 @@
+"""Tools to interact more easily with Full Frame Image files."""
+from __future__ import division, print_function
+
+from astropy.io import fits
+from astropy.wcs import WCS
+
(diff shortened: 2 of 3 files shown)


christinahedges:
@barentsen the FFI class should probably be able to plot the whole FFI as well, I remember hard coding a way to do that.

christinahedges:
The name of this module would easily be confused with TESS FFIs, unless you anticipate this module also being able to deal with TESS FFIs

barentsen:
This PR has been open for a year, but at no point have we actively missed having a ``FullFrameImage`` class.  There also haven't been any user requests for this feature, so I'm going to close this PR.

If you read this message and feel strongly that Lightkurve should offer this feature, then please do open a new issue!

Issue #188, opened 2018-07-09
Sometimes lightkurve returns inconsistent endianness


gully:
#### Problem description
We found an example where lightcurve time and flux arrays have different *endianness*, which foils certain model fitting.

#### Example
```python
from lightkurve import KeplerTargetPixelFile

tpf_star = KeplerTargetPixelFile.from_archive(1867630, quarter=1)
lc_star = tpf_star.to_lightcurve(aperture_mask='all')

lc_star.time.dtype, lc_star.flux.dtype

(dtype('>f8'), dtype('float32'))
```

#### Expected behavior
We expect both datatypes to be the same, ideally float32.

#### Environment:
We're on the `injrec` branch of lightkurve on @ssagear 's fork.

barentsen:
Target pixel files store `TIME` as `>f8` (i.e. 64-bit float) and `FLUX` as `>f4` (i.e. 32-bit float):

![screenshot from 2018-11-19 11-54-03](https://user-images.githubusercontent.com/817669/48731422-d51ea080-ebf1-11e8-8e44-bd635ba078ce.png)

This is by design because TIME is recorded at a much higher precision than FLUX.  Along the way FLUX does get converted from `>f4` to `float32` because numpy operations are applied to extract aperture photometry, but to my knowledge there is no important difference between those two alternative notations.

I think it makes sense for the `LightCurve` class to retain whatever numeric data types were passed along.  @gully Do you remember if there was a specific reason why you opened the issue?  If not, I think we should close it!

barentsen:
That's interesting.  I find that most operations on a dataframe obtained using `lc.to_pandas()` work perfectly fine, but somehow `df.describe()` does crash?!

Note that `LightCurve.fill_gaps()` includes a workaround for this issue:
https://github.com/KeplerGO/lightkurve/blob/a9dc2dd065d74c7c6536ec4cc38b464281f674c0/lightkurve/lightcurve.py#L418-L421

It looks like we'll either want to move this workaround to the constructor of `LightCurve`, or otherwise duplicate it in the `to_pandas()` method. Thoughts?  I don't understand the nature of this issue particularly well.

barentsen:
I believe I fixed this in #343.

PR #343, opened 2018-11-20
Fix the pandas byteorder bug

barentsen:
Pandas only supports data stored in the native byteorder format.  The most common platform (x86) uses little endian, however the FITS file format stores data in big endian, and so it is common for `lightkurve` to pass data in the wrong byteorder to pandas (e.g. see #188).

This PR adds a fix in `to_pandas()` to ensure we always pass little endian to pandas.  I also added unit tests to verify the correct byteorder behavior.

@gully Please review and merge!

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -862,6 +862,12 @@ class LightCurve(object):
         for col in columns:
             if hasattr(self, col):
                 data[col] = vars(self)[col]
+                # We need to ensure pandas gets the native byteorder.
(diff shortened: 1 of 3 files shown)


dhomeier:
> Since those systems are presumably uncommon, I'm fine with this solution as it presently stands

They are probably mostly restricted to some POWER and SPARC systems, which you are most likely to encounter in supercomputing environments these days, if at all.
It should be fairly straightforward to amend the patch for those, even if you don't wish to check `sys.byteorder`, by changing to
`data[col] = data[col].byteswap().newbyteorder('=')`

PR #345, opened 2018-11-21
Have `tpf.create_threshold_mask()` return only one contiguous mask

### LINKED ISSUE #337: Enable `tpf.create_threshold_mask()` to return a contiguous mask
### Problem
The `aperture_mask='threshold'` feature of Lightkurve is a bit simplistic in that it returns all pixels above a median flux threshold.  This means that the mask returned is often not contiguous and may include neighbor stars, see e.g. https://github.com/KeplerGO/lightkurve/pull/231#issuecomment-439681783.  Although we could keep this behavior as an option, most of the time users will want a smarter mask.

### Possible solutions
I am aware of four possible ways to make the threshold feature smarter.
It could return the contiguous mask that...
1. contains the center pixel;
2. contains the brightest pixel (in terms of median flux);
3. contains the largest number of pixels; or
4. contains the most flux.

Thoughts?  I wonder if option 4 makes the most sense.

### Example implementation
Here's an example implementation of scenario 2 written by @mrtommyb as part of his original PR [#231]:
```Python
# identify pixel above the threshold and label contiguous regions
threshold_region = np.where(median_image > mad_cut, 1, 0)
labeled_values = label(threshold_region)[0]

# find brightest pixel
brightest_pixel_y, brightest_pixel_x = np.unravel_index(median_image.argmax(), median_image.shape)
region_value = labeled_values[brightest_pixel_y, brightest_pixel_x]
aperture_mask = np.where(labeled_values == region_value, 1, 0)
return aperture_mask.astype(bool)
```

barentsen:
This PR attempts to resolve #337, i.e. `aperture_mask='threshold'` should return a mask that marks only one contiguous region of the TPF.

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -395,7 +395,7 @@ class TargetPixelFile(object):
         self._last_aperture_mask = aperture_mask
         return aperture_mask
 
-    def create_threshold_mask(self, threshold=3):
+    def create_threshold_mask(self, threshold=3, reference_pixel=None):
         """Returns an aperture mask creating using the thresholding method.
 
         This method will identify the pixels in the TargetPixelFile which show
@@ -405,13 +405,19 @@ class TargetPixelFile(object):
         with 1.4826.
 
         If the thresholding method yields multiple contiguous regions, then
-        only the region closest to the central pixel is returned.
+        only the region closest to the ``reference_pixel`` is returned.
+        If ``reference_pixel`` is None, then the region closest to the center
+        of the mask will be returned.
 
         Parameters
         ----------
         threshold : float
             A value for the number of sigma by which a pixel needs to be
             brighter than the median flux to be included in the aperture mask.
+        reference_pixel: (int, int) tuple
+            If multiple contiguous regions of pixels fall above the threshold,
+            the region closest to this pixel coordinate will be returned.
+            If `None` (default), the center pixel of the mask will be used.
 
         Returns
         -------
@@ -419,6 +425,8 @@ class TargetPixelFile(object):
             2D boolean numpy array containing `True` for pixels above the
             threshold.
         """
+        if reference_pixel is None:
+            reference_pixel = (self.shape[0] / 2, self.shape[1] / 2)
         # Calculate the median image
         with warnings.catch_warnings():
             warnings.simplefilter('ignore')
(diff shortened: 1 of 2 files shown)


barentsen:
This PR is ready for review!  By default, the `threshold` feature will now select the contiguous region that is closest to the central pixel:

screenshot from 2018-11-21 17-04-40

Optionally, you can override this behavior by calling `tpf.create_threshold_mask()` by hand and specifying a `reference_pixel`.  The region of pixels that are above the threshold and closest to the `reference_pixel` will be returned.  Example:

screenshot from 2018-11-21 17-04-54

Alternatively, passing `reference_pixel='all'` will return all regions above the threshold:

screenshot from 2018-11-21 17-05-12

@mrtommyb Would you be willing to review and, if happy, merge?

lightkurve/targetpixelfile.py
```diff
@@ -404,18 +405,33 @@ def create_threshold_mask(self, threshold=3):
         in a robust way by multiplying the Median Absolute Deviation (MAD)
         with 1.4826.
 
+        If the thresholding method yields multiple contiguous regions, then
+        only the region closest to the (col, row) coordinate specified by
+        `reference_pixel` is returned, e.g. `reference_pixel=(0, 0)` will pick
+        the region closest to the bottom left corner.
+        By default, the region closest to the center of the mask will be
+        returned. If `reference_pixel='all'` then all regions will be returned.
```
christinahedges: Perhaps all contiguous regions should be returned when `reference_pixel=None`.

'Multiple' reference pixels doesn't seem to make a tone of sense to me. No reference pixel it seems intuitive that it would pass back all regions.

This is a nitpick.

lightkurve/targetpixelfile.py
```diff
@@ -404,18 +405,33 @@ def create_threshold_mask(self, threshold=3):
...
         Parameters
         ----------
         threshold : float
             A value for the number of sigma by which a pixel needs to be
             brighter than the median flux to be included in the aperture mask.
+        reference_pixel: (int, int) tuple, None, or 'all'
+            (col, row) pixel coordinate closest to the desired region.
+            For example, use `reference_pixel=(0,0)` to select the region
+            closest to the bottom left corner of the target pixel file.
+            If `None` (default) then the region closest ot the center pixel
```
christinahedges:
```suggestion
            If `None` (default) then the region closest to the center pixel
```

barentsen:
Thank you for the review @christinahedges!  I completely agree with your comment.  I have edited the method as follows:
* `reference_pixel="center"` returns the region closest to the center (this is the default);
* `reference_pixel=(col, row)` returns the region nearest to (col, row);
* `reference_pixel=None` returns all regions.

This interface makes much more sense that what it was before.  Thank you!!

Issue #237, opened 2018-09-07
a way to know the un-phase-folded time in a folded light curve


zkbt:
#### Enhancement description
Thanks to @christinahedges's updates in #228, we'll be able to color individual data points in a `lc.scatter()`. One extra feature that would be nice, for pedagogical (and maybe scientific) purposes, would be to be able to say something like
```
folded = lc.fold(period)
folded.scatter(c=lc.time)
```
so that the colors indicate the un-phase-folded time. 

#### Proposed solution 
Right now, `lc.fold` sorts the data by their phase-folded time. This is necessary to make line plots look sensible, but it means you don't have an easy way to figure out what the original time was for each phase-folded point. I suggest modifying the `FoldedLightCurve` class so that it can keep track of an `original_time` attribute, as well as the phase-folded `time` attribute. Then, we'd be able to do
```
folded = lc.fold(period)
folded.scatter(c=folded.original_time)
```
with lots of delightful colors.

I'm happy to write this change + test and submit it as a PR, if y'all think this sounds OK. Any proposals for a different attribute name than `original_time`?

barentsen:
Thanks @zkbt!  I agree that adding an `original_time` attribute is a good solution.  I don't particularly like the name `original_time` but I can't think of a better suggestion!

You'd be very welcome to submit a PR if you wanted to! :heart:

barentsen:
I wonder if *every* LightCurve object should have a `parent` attribute which contains the original LightCurve object from which the current object was created (or `None` otherwise).

The `parent` attribute would be set by all methods that create a new LightCurve from an existing one, including `fold`, `bin`, `flatten`, `correct`, etc.

The main drawback I see is that it will increase the memory cost of a LightCurve object significantly upon each operation. This may not stop us, but it will be useful to check how expensive this change would be.

christinahedges:
I like the idea of keeping history, but I'm not sold about keeping a lightcurve after every new method. Can we keep the "original" once and then a record of every command applied to it in the `lc.history` string?

barentsen:
But if we only keep `original` then the color-plotting issue won't be resolved, e.g. this won't work:
```
folded = lc.bin().fold()
folded.scatter(c=folded.original.time)
````
(of course binning a lightcurve before folding is a silly thing to do, but other operations may alter the `time` attribute as well)

I do quite like `parent`; it can exist in addition to a `history` attribute of course.

barentsen:
[...]

PR #348, opened 2018-11-22
Pass extra data from LightCurve to FoldedLightCurve

barentsen:
This PR changes `LightCurve.fold()` to pass along `time_original`, `targetid`, `label`, and `meta` to the `FoldedLightCurve` object it returns.

This addresses #237.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -362,9 +362,13 @@ class LightCurve(object):
         # fold time domain from -.5 to .5
         fold_time[fold_time > 0.5] -= 1
         sorted_args = np.argsort(fold_time)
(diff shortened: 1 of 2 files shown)


christinahedges:
@barentsen this looks great to me. 

Only one question: Do we want it to be called "time_original"? 

I'm guessing yes, because then I can tab complete time and see that there are two. 

If you're happy with that, please go ahead and merge.

(It does kind of beg the question why `FoldedLightCurve` hands back a phase in the `time` attribute, not a time in days. But that's a question for another time.)

barentsen:
Thank you @christinahedges!  I very slightly favored `time_original` over `original_time` because of tab-completion.  If anyone comes with a counter-argument in the next few days, we can revise!  For now, I'm going to go ahead and merge!

PR #349, opened 2018-11-22
Transit masking module

### LINKED ISSUE #327: Implement a transit masking feature
Various trend and systematics removal methods will benefit from the ability to mask out known planet transits before "fitting the trend" to be removed.  To make this easy, we should consider:
1. Providing an easy method to generate an appropriate transit mask, e.g. `mask = LightCurve.create_transit_mask(phase, period, duration)`?
2. Have our detrending/correction methods accept such mask, e.g. `LightCurve.flatten(transit_mask=mask)`?  Those methods will then require some interpolation of the trend over the transit points.

Any thoughts on this feature or its ideal design?

For inspiration, Everest implements transit masking as follows: https://rodluger.github.io/everest/using_everest.html#masking-transits

christinahedges:
Added a module that adds the ability to mask transits. Fixes #327.


lightkurve/tests/test_nexsci.py
```diff
@@ -0,0 +1,35 @@
...
+    ''' Test the dataframe can be read '''
+    df = get_nexsci_data()
+    assert(len(df) >= 0)
+
+def test_trappist():
+    ''' Test the planet mask works '''
+    t = np.arange(0, 100, 0.001)
+    m = find_planet_mask('TRAPPIST-1', t)
+    # Trappist 1 should have a planet transiting 7.9% of the time
+    assert np.isclose((1 - (m.sum()/len(t))), 0.079, atol=0.001)
```
nksaunders: This assertion fails in python 2.7 because `sum()` is returning an int. I believe a quick fix is to replace `m.sum()/len(t)` with `float(m.sum())/len(t)`.

lightkurve/tests/test_nexsci.py
```diff
@@ -0,0 +1,35 @@
...
+
+def test_unknown():
+    ''' Test the unknown planet mask works '''
+    t = np.arange(0, 100, 0.001)
+    m = create_planet_mask(t, 10, 0, 5)
+    # This should have a planet transiting 75% of the time
+    assert np.isclose((1 - (m.sum()/len(t))), 0.75, atol=0.01)
```
nksaunders: The same issue will likely arise here

christinahedges:
@barentsen this is probably not necessary for v1.0, is there a way you can avoid adding it to the v1 release? 

Also please let me know what you think of the module name.

christinahedges:
@barentsen 
Be aware that the `find_planet_mask()` only works if time is in JD not BKJD. 

How do you want to deal with that?

christinahedges:
I added a very similar extension to `starry`, I can lift the same, updated code from that extension and put it in lightkurve to make this a bit more streamlined.

https://rodluger.github.io/starry/v1.0.0.dev5/extensions.html

christinahedges:
@barentsen this is a bit of a dead PR, can you tell me whether you think something like this is worth merging in? If you think so, I will dust it off and update it.

barentsen:
My thinking:
* We should absolutely add the `create_planet_mask()` feature to Lightkurve, perhaps as a method in `LightCurve`?
* I'm worried that adding features to access the Nexsci API and interact with exoplanet parameters will add a lot of maintenance work down the line, in particular because Nexsci is changing its API right now. Would this be better suited for a separate Python package focused solely on accessing the Nexsci planet parameters table?

barentsen:
Good news: a `LightCurve.create_transit_mask()` method has now been added via #808!

The second part of this PR entailed the addition of an API to query exoplanet data from Nexsci.  I prefer not to add this to Lightkurve core right now because third-party data querying tends to require a lot of future maintenance work (e.g. the Nexsci API is changing as we speak).  I think we should either make that functionality available as a separate lightweight package, or use the existing `astroquery.nasa_exoplanet_archive` sub-package for this.

Closing! Thanks everyone! :+1:

PR #352, opened 2018-11-26
Fixed errorbar xlabel bug (Fixes #328)

### LINKED ISSUE #328: `errorbar()` provides incorrect x label for folded light curves
Similar to #6: `errorbar()` returns `Time` instead of `Phase` for folded light curves.

image

I believe adding a `FoldedLightCurve.errorbar()` wrapper function similar to the existing `.plot()` and `.scatter()` would resolve this issue.

christinahedges:
@nksaunders pointed out that `lc.fold().errorbar()` has the wrong xlabel. This fixes that, and adds a test it will never happen again.

Fixes #328

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1115,6 +1115,25 @@ class FoldedLightCurve(LightCurve):
             ax.set_xlabel("Phase")
         return ax
 
+    def errorbar(self, **kwargs):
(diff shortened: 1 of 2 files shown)


barentsen:
Thank you @christinahedges.  This looks good!!  Extra bonus points for the extra unit tests! Merging in 3 ... 2 ...

PR #355, opened 2018-11-27
Quick fix for Bokeh v1.0.0

barentsen:
@gully I think it would be good to merge a quick workaround fix for Bokeh v1.0.0 (cf. https://bokeh.pydata.org/en/latest/docs/releases.html#handling-of-selections), separately from the ongoing work to add a `Save lightcurve` button (#329).

Does this PR look right?  I am aware that this fix causes the deselection of pixels to no longer work, but it does seem to restore all other functionality, so I'd like to suggest we go ahead and merge this.

(If we can restore the deselection feature, all the better of course, but I didn't find an obvious workaround?)

### DIFF
--- a/lightkurve/interact.py
+++ b/lightkurve/interact.py
@@ -344,8 +344,12 @@ def show_interact_widget(tpf, lc=None, notebook_url='localhost:8888', max_cadenc
         # Callbacks
         def update_upon_pixel_selection(attr, old, new):
             """Callback to take action when pixels are selected."""
-            if tpf_source.selected.indices != []:
-                selected_indices = np.array(tpf_source.selected.indices)
+            if ((sorted(old) == sorted(new)) & (new != [])):
+                # Trigger recursion
+                tpf_source.selected.indices = new[1:]
+
+            if new != []:
+                selected_indices = np.array(new)
                 selected_mask = np.isin(pixel_index_array, selected_indices)
                 lc_new = tpf.to_lightcurve(aperture_mask=selected_mask)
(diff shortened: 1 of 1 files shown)


gully:
Let's merge #329, yeah?  There's nothing wrong with it, and it will likely not change much-- we're stuck with the new behavior.

barentsen:
Ah, I'll have a careful look at #329 asap!

gully:
I fixed this patch, so it's ok to ship 🚢 , in case we want more time ⏳ to spot-check #329 😃 

:shipit:

gully:
Note ⚠️ however, that this patch will break pixel selection for `.interact()` for all users with bokeh<`1.0` 😢 

I just demo'ed with bokeh version 0.13 and got unresponsive pixel selection. 👎

gully:
The above ⬆️ comment implies that this patch must be accompanied by an update 🆕 to the bokeh version by the user.  We could detect and compare the bokeh version, and notify 🔔the user that interact requires v1.0 or greater of bokeh.

barentsen:
Ah! Understood.  Let's add something like this:
```
if bokeh.__version__[0] == '0':
    warnings.warn("interact() requires Bokeh version 1.0 or later", LightkurveWarning)
```
Can you do that?

barentsen:
I'm going to go ahead and merge this so we can push a new version of Lightkurve to pypi later today!

gully:
Cool!  After you push to pypi, I'll modify the conda-forge to require bokeh 1.0 or greater for the new release.

PR #359, opened 2018-11-30
Change argument name from sigma_clip to sigma in the method estimate_cdpp

mirca:
Hi y'all,
I propose to change the argument named ``sigma_clip`` to ``sigma`` from the method ``estimate_cdpp``. My rationale is the following: 

``sigma_clip`` is already the name of a function that is being imported from ``astropy.stats``, so we are doing something like:
```
from astropy.stats import sigma_clip

def estimate_cdpp(sigma_clip=5.):
     ...
```
which is fine to some extent. But basically, because of Python's scope resolution, we can't use the function ``sigma_clip`` inside ``estimate_cdpp`` if we ever need it.

Also, the method ``remove_outliers``, which is used inside ``estimate_cdpp``, already expects an argument named ``sigma``, so just for consistency I think it would be a little clearer to have ``sigma`` instead of ``sigma_clip``.

Let me know y'all's thoughts.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -598,7 +598,7 @@ class LightCurve(object):
         return self.estimate_cdpp(**kwargs)
 
     def estimate_cdpp(self, transit_duration=13, savgol_window=101,
-                      savgol_polyorder=2, sigma_clip=5.):
+                      savgol_polyorder=2, sigma=5.):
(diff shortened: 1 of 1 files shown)


barentsen:
Thanks @mirca!  I think I agree that we could just call this `sigma`.  I don't think it will break a lot of code  because this is quite an obscure option.

Unless someone else speaks up, I'd be happy to merge this over the weekend!

PR #362, opened 2018-12-04
Added an iterative feature to flatten to iteratively clip outliers when flattening. 

christinahedges:
When flattening light curves with flares or transits, the flatten always makes the transits look funny, because it doesn't remove those outliers before flattening.

I've added in an iterative sigma-clipping which flattens, removes outliers, and then re-flattens the data ignoring those outliers. 

### Problem (before)
image
image

### Solution (after)

image

image

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -11,6 +11,7 @@ import warnings
 
 import numpy as np
 from scipy import signal
+from scipy.interpolate import interp1d
 from matplotlib import pyplot as plt
 
 from astropy.stats import sigma_clip
@@ -257,7 +258,7 @@ class LightCurve(object):
         return copy.deepcopy(self)
 
     def flatten(self, window_length=101, polyorder=2, return_trend=False,
-                break_tolerance=5, **kwargs):
+                break_tolerance=5, niters=2, sigma=3, **kwargs):
         """Removes the low frequency trend using scipy's Savitzky-Golay filter.
 
         This method wraps `scipy.signal.savgol_filter`.
@@ -279,6 +280,11 @@ class LightCurve(object):
             individually. A gap is defined as a period in time larger than
             `break_tolerance` times the median gap.  To disable this feature,
             set `break_tolerance` to None.
+        niters : int
+            Number of iterations to iteratively sigma clip and flatten. If more than one, will
+            perform the flatten several times, removing outliers each time.
+        sigma : int
+            Number of sigma above which to remove outliers from the flatten
         **kwargs : dict
             Dictionary of arguments to be passed to `scipy.signal.savgol_filter`.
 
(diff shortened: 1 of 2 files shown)


mirca:
This is very cool! ``flatten`` is clearly more robust this way, thanks!

One minor suggestion: I would suggest perhaps swapping the convention adopted in ``mask`` (ie, flux values for which ``mask == True`` should be indeed masked) in order to follow the trend of [astropy](http://docs.astropy.org/en/stable/api/astropy.table.MaskedColumn.html) and [numpy](https://docs.scipy.org/doc/numpy/reference/maskedarray.generic.html#accessing-only-the-valid-entries). I imagine that being done like: rename every instance of ``mask`` to ``not_mask`` (except in the argument list of ``flatten``) and then in the beginning of ``flatten`` take ``not_mask = ~mask``.

lightkurve/lightcurve.py
```diff
@@ -290,39 +301,54 @@ def flatten(self, window_length=101, polyorder=2, return_trend=False,
...
-
-        trend_signal = np.interp(self.time, lc_clean.time, trend_signal)
+
+        if mask is None:
+            mask = np.ones(len(self.time), dtype=bool)
+        # No NaNs
+        mask &= np.isfinite(self.flux)
+        # No outliers
+        mask &= np.nan_to_num(np.abs(self.flux - np.nanmedian(self.flux))) <= (np.nanstd(self.flux) * sigma)
+#        ax = self.plot(normalize=False)
```
barentsen: :point_up: Let's exterminate this line!

lightkurve/lightcurve.py
```diff
@@ -290,39 +301,54 @@ def flatten(self, window_length=101, polyorder=2, return_trend=False,
...
+        for iter in np.arange(0, niters):
+#            ax.scatter(self.time[mask], self.flux[mask])
```
barentsen: :point_up: Exterminate!

barentsen:
@christinahedges Do you want to address the concerns related to the meaning of `mask`?  Then, this should be good to go!

JeffLCoughlin:
Very nice!

If it proves worth trying in the future, below is some C code I wrote a while back for sigma clipping on Kepler light curves. It uses MAD and adjust sigma based on the number of data points, so in theory only requires a single pass to remove outliers.

```
  detrenddur = 0.02043981;  // Set detrending window to one cadence duration. Or can be a fraction of a transit duration

  for(l=0;l<1;l++)  // Number of outlier passes - ONLY NEED ONE WHEN USING MAD - JUST LEAVING LOOP IN CASE I EVER WANT TO CHANGE MY MIND
    {
    for(i=0;i<ndat;i++)
      data[i].resid = data[i].flux - data[i].model;  // Calculate residuals

  // std = STDEV(data.resid,ndat);  // Saving this line in case I go back to std instead of mad
    for(i=0;i<ndat;i++)
      tmpdob3[i] = data[i].resid;
    mad = MAD(tmpdob3,ndat);  // MAD is Median Absolute Devitation - better than STD - only need one pass.
    std = 1.4826*mad;  // https://www.mathworks.com/help/stats/mad.html  AND https://en.wikipedia.org/wiki/Median_absolute_deviation

    sigreject = sqrt(2)*INVERFC(1.0/ndat);  // Calculate sigma threshold based on number of data points

    for(i=0;i<ndat;i++)  // Keep track of which data points to remove. 0 is good, if flagged as 1, remove later
      tmpintarry[i]=0;

    for(i=0;i<ndat;i++)
      {
      k=0;
      sw1=0;
      for(j=0;j<ndat;j++)
        if(fabs(data[i].phase-data[j].phase) < detrenddur)  // look at points within the detrend duration
          {
          tmpdobarry1[k]=data[j].resid;
          k++;
          }
      med = MEDIAN(tmpdobarry1,k);

      if(fabs(data[i].resid - med) > sigreject*std)  // Keep data points within sigreject standard deviations of median
        tmpintarry[i] = 1;
      }

    m=0;
    for(i=0;i<ndat;i++)
      if(tmpintarry[i]==0)  // Keep good data points
        {
        data[m].time = data[i].time;
        data[m].phase = data[i].phase;
        data[m].flux = data[i].flux;
        data[m].model = data[i].model;
        data[m].resid = data[i].resid;
        m++;
        }
      else
        if(m<midti)
          midti--;  // If I remove a data point, I have to update midti to keep track of it
    ndat=m;  // Update number of data points to those not thrown out
    }



double MAD(double x[], int n)  // Median Absolute Deviation
  {
  double xmed,w[N];
  int z;

  xmed = MEDIAN(x,n);
  for(z=0;z<n;z++)
    w[z] = fabs(x[z] - xmed);

  return MEDIAN(w,n);
  }



```

mirca:
Hey @JeffLCoughlin, that's really cool! It seems then that ``flatten`` can be sped up by orders of magnitude! We could probably make use of ``pybind11`` to directly integrate your C++ code into ``lightkurve``.

JeffLCoughlin:
One note I forgot is I developed that for use on phased data, but it could be adapted to non-phased data (would obviously want to increase the detrendung window to much bigger than one cadence duration). I think the key point is just the use of MAD, and calculating the number of sigma to clip via the number of data points and the complementary inverse error function (INVERFC), and that's probably quick enough to just write in python from scratch.

But only likely worth revisiting if there is a need for a CPU savings, or if 3 sigma is over or under agressive. For Kepler-prime long-cadence lightcuves w/ ~70,000 data points the statistically correct sigma came out to around 4.3, and 5.0 for short-cadence w/ 30 times more. For K2 I get something like 3.6 for long-cadence and 4.5 for short if I assume ~4,000 long-cadence data points per campaign.

PR #361, opened 2018-11-30
Changed lc.fold() phase argument to transit midpoint.

christinahedges:
I've changed lc.fold() phase argument to transit midpoint to fix issue #357. I've also fixed #353 and fixed the unit test. 

### Fold now takes `transit_midpoint` as an argument, not phase.
image

Note that if `transit_midpoint` is in JD it will warn the user that it doesn't look right. It just raises a warning, not an exception, because it's conceivable that the user knows what they're doing. @barentsen do you want to chime in on this?

### Folded light curve attribute `time_original` is now in the correct order.
image

Fixes #357 
Fixes #353

PR #363, opened 2018-12-06
[DOC] Replace phase with transit_midpoint

brennv:
So tutorial docs conform with #361


barentsen:
@brennv Thank you so much for spotting and fixing this bug.  We really appreciate it!  Merging in 3 ... 2 ... 1 ...

Issue #365, opened 2018-12-06
Implement from_archive methods for TESS


mrtommyb:
Implement methods to get data directly from the MAST into lightkurve.

mrtommyb:
I am working on this.

mrtommyb:
I will implement this for the 2-min cadence data. However, we also want to interface with TESScut to access FFI data.

barentsen:
Thanks @mrtommyb, that sounds great! Please note: `from_archive()` has been deprecated in favor of the more flexible `lightkurve.search_targetpixelfile()` and `lightkurve.search_lightcurvefile()` functions. They are the ones that will need updating. 

@nksaunders was also standby to help edit these functions for TESS. Do report any progress here and feel free to open [WIP] PRs early and often for early feedback!

mrtommyb:
So, do you want me to implement things or are we doing it on #367

barentsen:
#367 should be ready or almost ready to be merged -- it would be really useful to test

PR #367, opened 2018-12-06
Expand search to TESS data products

nksaunders:
#365: This PR adds quick checks to determine the observing spacecraft for data products on MAST. 

This works to find target pixel files:
```python
import lightkurve as lk

tic_id = 273985862
lk.search_targetpixelfile(tic_id, spacecraft='Tess')
```
```
   obsID    target_name                     productFilename                        description     distance
----------- ----------- ------------------------------------------------------- ------------------ --------
17000015500          -- tess2018206045859-s0001-0000000273985862-0120-s_tp.fits Target pixel files       --
```

...and light curve files:
```python
lk.search_lightcurvefile(tic_id, spacecraft='Tess')
```
```
   obsID    target_name                     productFilename                     description  distance
----------- ----------- ------------------------------------------------------- ------------ --------
17000015500          -- tess2018206045859-s0001-0000000273985862-0120-s_lc.fits Light curves       --
```
I added the spacecraft keyword to distinguish between K2 and TESS observations, because the current API considered tic_ids within the numerical range of K2.

- [x] Rework Kepler-specific keywords in download functions
- [x] Add ability to return `TessTargetPixelFile` or `TessLightCurveFile` objects from download


nksaunders:
FFIs are also available on MAST and queryable with astropy. @barentsen & @mrtommyb, would it be worth introducing the ability to search and download FFIs? Maybe a `search_ffi()` or  `search_fullframeimage()` function?

barentsen:
Thanks @nksaunders!

First thought: the MAST API ([docs here](https://mast.stsci.edu/api/v0/_productsfields.html)) already has fields called `obs_collection`  (= mission identifier) and `project` (= processing project).  I wonder if we want to follow the same nomenclature and pass that keyword on the astroquery.  Of course, by default it would be e.g. `project=['Kepler', 'K2', 'TESS']`.

barentsen:
Regarding FFIs: we should think about adding a `search_fullframeimage()`, but I think it'll need a lot more discussion and thinking than supporting light curve and target pixel files, so I would split FFI support off in separate issues/PRs.

lightkurve/search.py
```diff
@@ -147,12 +147,18 @@ def download(self, quality_bitmask='default', download_dir=None):
...
+        tess_lcf_file = '_lc.fits'
+        if any(file in self.table['productFilename'][0] for file in kepler_tpf_files):
             return KeplerTargetPixelFile(path[0], quality_bitmask=quality_bitmask)
-        elif any(file in self.table['productFilename'][0] for file in lcf_files):
+        elif any(file in self.table['productFilename'][0] for file in kepler_lcf_files):
             return KeplerLightCurveFile(path[0], quality_bitmask=quality_bitmask)
+        elif tess_tpf_file in self.table['productFilename'][0]:
+            return TessTargetPixelFile(path[0], quality_bitmask)
+        elif tess_lcf_file in self.table['productFilename'][0]:
+            return TessLightCurveFile(path[0], quality_bitmask=quality_bitmask)
```
barentsen: Can we replace all of the above with `open(self.table['productFilename'][0])` ?

lightkurve/search.py
```diff
@@ -240,7 +256,7 @@ def _default_download_dir(self):
         return download_dir
 
 
-def search_targetpixelfile(target, radius=None, cadence='long', quarter=None,
+def search_targetpixelfile(target, radius=None, cadence='long', project='Kepler', quarter=None,
```
barentsen: I think we may want the default to be `project=['Kepler', 'K2', 'Tess']`?

lightkurve/search.py
nksaunders: Yes, good point! I didn't consider that `open()` takes care of a lot of this functionality.

nksaunders:
Updated usage:

```python
tic = 273985862

tpf = lk.search_targetpixelfile(tic, project='TESS').download()
lcf = lk.search_lightcurvefile(tic, project='TESS').download()
```
These will return a `TessTargetPixelFile` or `TessLightCurveFile`.

You can pass a `radius = <arcsec>` keyword into both search functions to perform a cone search.

lightkurve/search.py
```diff
@@ -240,7 +232,8 @@ def _default_download_dir(self):
         return download_dir
 
 
-def search_targetpixelfile(target, radius=None, cadence='long', quarter=None,
+def search_targetpixelfile(target, radius=None, cadence='long',
+                           project=['Kepler', 'K2', 'TESS'], quarter=None,
```
barentsen: Let's make sure we add a good docstring for `project` here and in `search_lightcurvefile`!

lightkurve/search.py
```diff
@@ -438,20 +433,26 @@ def _search_products(target, radius=None, filetype="Lightcurve", cadence='long',
...
+    observations = _query_mast(target, project=project, cadence='long', radius=radius)
+
     if len(observations) == 0:
         raise SearchError('No data found for target "{}."'.format(target))
-    products = Observations.get_product_list(observations)
-    result = join(products, observations, join_type='left')  # will join on obs_id
+
+    # mask out FFIs from observations    
+    mask = np.array(['FFI' not in obs['target_name'] and
+                     'FFI' not in obs['obs_collection'] for obs in observations])
```
barentsen: Do you think this hard-coded mask is necessary?  I imagine the call to `_filter_products` on line 449 below will take care of retaining only the `filetype` requested?

barentsen:
Thanks @nksaunders! This is getting close! We'll want to add a couple of unit tests for TESS data though before merging.

lightkurve/search.py
nksaunders: This was added because the `join` on line 446 fails when FFIs are included due to missing keywords in their table entries. I struggled with the simplest way to exclude them for both Kepler and TESS, but if you see a clearer fix let me know!

barentsen:
Note: the TESScut service is currently supported as follows:

screenshot from 2018-12-07 14-43-23

@gully is investigating if we need a nicer API!

mrtommyb:
We should add a sector keyword

barentsen:
To support queries by sector number, it looks like we'll have to parse the `sxxxx` sector number from the filename (e.g. `tess2018206045859-s0001-0000000261136641-0120-s_tp.fits`)

barentsen:
Before we can merge this, we need to revise some of the logic in `_filter_products` -- e.g. `search_targetpixelfile("pi Mensae")` does not return anything right now, whereas I would expect it to return a TESS product by default.

barentsen:
I reduced the complexity of `search._filter_products()` by adding three helper functions: `_mask_kepler_products()`, `_mask_k2_products()`, and `_mask_tess_products()`.  

This PR needs another careful review and additional testing, but I think we're close to supporting TESS without breaking Kepler/K2, e.g.:

screenshot from 2018-12-09 21-38-43

nksaunders:
Thanks @barentsen! I was struggling with how to handle arbitrary combinations of `['Kepler', 'K2', 'TESS']` project keywords being passed in, and I think you've found the best approach with the mask functions.

While experimenting, I found that astropy gives a `NoResultsFound` warning for certain TESS targets. For example: 

```python
lk.search_targetpixelfile(38847248)
```
gives the following error:
```
WARNING: NoResultsWarning: Query returned no results. [astroquery.mast.core]
```

But it also returns the desired search result. My solution is to catch and filter the `AstropyWarning` raised by `Observations/query_criteria()`.

barentsen:
@nksaunders I was surprised to see all those new warnings raises, I don't quite understand why we didn't see them in the past.  We can suppress them, because we'll throw our own!

barentsen:
Great!  I'm going to go ahead and merge this.  Thanks @nksaunders!

amcody:
FYI I just tried to install this version of lightkurve, and it says it cannot find the sklearn module? I can install myself, but thought maybe it should be required or something?

amcody:
Same with the george package. And in this case, i am unable to pip install it, due to a problem with gcc being missing. I think if we are going to require these other packages, we want to make sure they're easy for people to get installed.

amcody:
Okay, IGNORE my above comments! Nick switched his branch recently from tess-search to search-tess, and the latter does not produce these problems..

mrtommyb:
@barentsen is this merged? I just cloned the master repo and python setup.py install I get 
```
import lightkurve
lightkurve.search.search_targetpixelfile('pi Mensae')

WARNING: NoResultsWarning: Query returned no results. [astroquery.mast.core]
No data found for target "pi Mensae."

```

PR #372, opened 2018-12-07
Solves issue in astropy.convolution.convolve affecting lightkurve's periodogram box method

mirca:
As explained by @barentsen, `astropy 3.1` introduced some behavior that resulted in an error when computing periodograms using `lightkurve`. I didn't completely understand what's going on on the astropy side, but this PR should bring back the expected behavior. I also noted that with `astropy 3.2dev` we didn't need this fix, but it works with it regardless.

### DIFF
--- a/lightkurve/periodogram.py
+++ b/lightkurve/periodogram.py
@@ -3,6 +3,7 @@ from __future__ import division, print_function
 
 import copy
 import logging
+import math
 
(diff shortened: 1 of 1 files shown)


barentsen:
Thanks @mirca!  This looks good!

barentsen:
PS: You are the best!

PR #374, opened 2018-12-08
remove impulsive outlier as a default flag

mrtommyb:
I don't think that this should be a default flag as it flags flares intrinsic to the star.


barentsen:
I'm happy to merge this because`ImpulsiveOutlier` isn't in any of the default bitmasks for Kepler either.  I think the risk is that we'll see more light curves with annoying artifact spikes which ruin plots.

@christinahedges Do you remember why you added this to the default mask back in #274?

christinahedges:
@barentsen no recollection of why this is in there... please go ahead.

barentsen:
Great, merging!

@mrtommyb Do keep an eye out for additional changes to the default mask for TESS.  We certainly didn't think this through carefully yet!

Issue #373, opened 2018-12-07
interact() fails on TESScut TPFs


barentsen:
`interact()` appears to fail on TESScut TPFs because the CADENCENO column is set to [0, 0, 0, ... 0].

![screenshot from 2018-12-07 14-34-17](https://user-images.githubusercontent.com/817669/49676108-3edddd80-fa2d-11e8-916c-9b54f4f0fd86.png)

Also see https://twitter.com/GeertHub/status/1071172334982094848

gully:
Will investigate.  I suspect a fix to this will require an overhaul of interact, since so much of its indexing relies on the existence of cadences.

gully:
Update: Geert recommends a quick workaround-- arbitrarily assign indices as fake cadence numbers for TESSCut.  Avoids overhauling interact and achieves close to the desired behavior.

barentsen:
This has been resolved as part of #375.  Closing!

PR #375, opened 2018-12-10
Prevent `tpf.cadenceno` from returning zeros for TESScut TPFs

barentsen:
`interact()` does not work on TESScut TPFs because the `CADENCENO` column is all zeros (cf. #373).  This will likely be resolved in a few months (cf. https://twitter.com/GeertHub/status/1071172334982094848 ), but for now I propose we merge this workaround.

@gully Can you review, test, and merge?

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -168,7 +168,12 @@ class TargetPixelFile(object):
     @property
     def cadenceno(self):
         """Return the cadence number for all good-quality cadences."""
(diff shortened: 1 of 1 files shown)


gully:
The current implementation tests if the first cadenceno is 0, are we convinced cadences starting with 0 could not be valid?  What do you think of a check for unique cadences? `len(set(cadences)) == len(cadences)`  would fail for nan cadences, so would have to be cleaned, which adds unnecessary complexity.

gully:
I tried interact on a tess-cut TPF and it worked fine with this change.

PR #377, opened 2018-12-12
[bugfix] LightCurve.bin() did not bin QUALITY and CADENCENO correctly

barentsen:
Christina reported verbally that `KeplerLightCurve.bin(15).estimate_cdpp()` failed.  I believe this PR addresses the bug, and adds a regression test.

@christinahedges Please review, test, and merge!  Thx++.

### DIFF
--- a/lightkurve/tests/test_lightcurve.py
+++ b/lightkurve/tests/test_lightcurve.py
@@ -346,6 +346,7 @@ def test_bin():
     lc = KeplerLightCurve(time=np.arange(10),
                           flux=2*np.ones(10))
     lc.bin(5).remove_outliers()
+    # Second regression test for #377
     lc = KeplerLightCurve(time=np.arange(1000) * 0.02,
                           flux=1*np.ones(1000) + np.random.normal(0, 1e-6, 1000),
                           cadenceno=np.arange(1000))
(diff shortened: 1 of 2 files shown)


lightkurve/tests/test_lightcurve.py
```diff
@@ -342,6 +342,10 @@ def test_bin():
                     flux=2*np.ones(10))
     binned_lc = lc.bin(binsize=2)
     assert_allclose(binned_lc.flux_err, np.zeros(5))
+    # Regression test for #377
+    lc = KeplerLightCurve(time=np.arange(10),
+                          flux=2*np.ones(10))
+    lc.bin(5).remove_outliers()
```
christinahedges: Can I suggest we add a test that confirms we can estimate the cdpp when there is a cadence number set? This was the issue we were having.

```suggestion
lc = lk.KeplerLightCurve(time=np.arange(1000) * 0.02,
                          flux=1*np.ones(1000) + np.random.normal(0, 0.00002, 1000), cadenceno=np.arange(1000)+3000)
np.isclose(lc.estimate_cdpp(), 5, rtol=2)
```

christinahedges:
Great job, this fixes the issue I was having! 

image

PR #382, opened 2018-12-17
Added errorbar and scatter methods to LCF

### LINKED ISSUE #376: Add a scatter method to lightcurvefile objects
Currently, lightcurvefile objects can only be plotted with continuous lines, if one wishes to show both SAP_FLUX and PDC_SAPFLUX together. 

When one does lcf.scatter(), it returns the error message
AttributeError: 'KeplerLightCurveFile' object has no attribute 'scatter'

The current way around this is to do, e.g., lcf.SAP_FLUX.scatter(), using LightCurveFile. Let's add a few lines to get Kepler/TessLightCurveFile to do scatterplots too.

christinahedges:
I added errorbar and scatter methods to `lightcurvefile` in the same way as we have them for `lightcurve`. I haven't added tests because it actually looks like we were testing these methods without them ever breaking in `test_lightcurve.py`. Gonna see what code cov says.

Fixes #376 for @amcody 

```python
@pytest.mark.remote_data
def test_lightcurve_plots():
    """Sanity check to verify that lightcurve plotting works"""
    for lcf in [KeplerLightCurveFile(TABBY_Q8), TessLightCurveFile(TESS_SIM)]:
        lcf.plot()
        lcf.plot(flux_types=['SAP_FLUX', 'PDCSAP_FLUX'])
        lcf.SAP_FLUX.plot()
        lcf.SAP_FLUX.plot(normalize=False, title="Not the default")
        lcf.SAP_FLUX.scatter()
        lcf.SAP_FLUX.scatter(c='C3')
        lcf.SAP_FLUX.scatter(c=lcf.SAP_FLUX.time, show_colorbar=True, colorbar_label='Time')
        lcf.SAP_FLUX.errorbar()
        plt.close('all')
```

### DIFF
--- a/lightkurve/lightcurvefile.py
+++ b/lightkurve/lightcurvefile.py
@@ -153,7 +153,8 @@ class LightCurveFile(object):
         ax : matplotlib.axes._subplots.AxesSubplot
             The matplotlib axes object.
         """
-        return self._create_plot(method='plot', **kwargs)
+        return self._create_plot(method='plot', flux_types=flux_types,
+                                 style=style, **kwargs)
 
 
     def scatter(self, flux_types=None, style='lightkurve', **kwargs):
@@ -188,7 +189,8 @@ class LightCurveFile(object):
         ax : matplotlib.axes._subplots.AxesSubplot
             The matplotlib axes object.
         """
-        return self._create_plot(method='scatter', **kwargs)
+        return self._create_plot(method='scatter', flux_types=flux_types,
+                                 style=style, **kwargs)
 
     def errorbar(self, flux_types=None, style='lightkurve', **kwargs):
         """Plot the light curve file using matplotlib's `errorbar` method.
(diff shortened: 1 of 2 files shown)


amcody:
@christinahedges this works fine for me now! Maybe the coverage would increase if you added a line with lcf.scatter()?

lightkurve/lightcurvefile.py
```diff
@@ -118,7 +114,116 @@ def plot(self, flux_types=None, style='lightkurve', **kwargs):
...
-                lc.plot(label=ft, **kwargs)
+                if method == 'plot':
+                    lc.plot(label=ft, **kwargs)
+                elif method == 'scatter':
+                    lc.scatter(label=ft, **kwargs)
+                elif method == 'errorbar':
+                    lc.errorbar(label=ft, **kwargs)
+
+
+    def plot(self, flux_types=None, style='lightkurve', **kwargs):
```
barentsen: Caution: it looks like the `flux_types` and `style` arguments are not passed to `self._create_plot()` in this function below.  In the `LightCurve` class we accepted these as part of `**kwargs`, we'll want to do the same here.

barentsen:
Thanks for this PR @christinahedges!!  I'd be happy for this to be merged if the keyword bug above ☝️ is fixed and if we add the three methods to the tests as a sanity check.

PR #390, opened 2019-01-18
[WIP] Add iterative sine fitting to lightkurve (Fixes #310)

### LINKED ISSUE #310: Allow for fitting with Astropy.LombScargle through lightkurve.Periodogram()?
#### Problem description
Ben Pope (@benjaminpope) has mentioned how Periodogram() could be made more useful by allowing use of the astropy.LombScargle fitting functionality (http://docs.astropy.org/en/stable/api/astropy.stats.LombScargle.html#astropy.stats.LombScargle.model). 
This would allow for users to evaluate the model at the frequency of highest power and be more useful for modelling coherent signals. 

I initially didn't include in this functionality as I felt it fell outside of the scope of what Periodogram() should do, but given @benjaminpope's work below I thought it best to reconsider.

#### Example
An example of @benjaminpope using this can be found here: (https://github.com/benjaminpope/hot/blob/master/notebooks/test.ipynb).

ojhall94:
Please refer to issue #310 for the initial discussion.

I'm attempting to include a simple function that allows for iterative sine fitting for removal of periodic noise. I have included this as a function in Periodogram, although I feel like it would be better suited to be a LightCurve function that potentially leverages over Periodogram.

I am keeping up the notebook below with an example of its functionality in practice, labeling which features are now new in lightkurve in this PR.

https://github.com/ojhall94/keplergohacks/blob/master/%23310.ipynb

I'll write unit tests once we've reached a consensus on the format!


christinahedges:
Hi Ollie! This is really useful. I wanted exactly this in light curve not that long ago. Quick question: how do we do this when we increase the nterms? (I think nterms will become very important for people fitting star spots)

Thanks for contributing this! I'll take an in-depth look at it soon!

ojhall94:
Hi!

> I wanted exactly this in light curve not that long ago.
So does this mean you think the functionality should remain as in the current example? Should we do a separate PR for iterative sine fitting to the LightCurve class that uses this functionality?

>Quick question: how do we do this when we increase the nterms?
So since Periodogram.model() generates a new periodogram from which it returns the model, this might cause issues as you suggest. Maybe we can make the nterms value a property of a Periodogram object that is used when calling model() for consistency?

ojhall94:
Alright, I've made some updates. I saw you included a get_transit_model(), so I've modelled (pun intended) the get_lombscargle_model() function after that.

- I changed the name to get_lombscargle_model()
- LombScarglePeriodogram now has the attribute _LS_object, analogous to _BLS_object. This is important for consistency.
- get_lombscargle_model() takes an input time array. This is different to get_transit_model(), where `time` is stored as an attribute. I feel like it gives more freedom if a user can model on to any time array of choosing, for whatever reason.

I have yet to check the issue you raised with nterms, I will check that next, and write unit tests after. Otherwise, should be good to go.

ojhall94:
Note to self: Make sure SNRperiodogram inherits this function.

ojhall94:
I've added unit tests, and Pytest runs locally. I think this is ready for review now :)

barentsen:
Awesome :+1:   Will review ASAP!  (Probably when we return to Ames on Monday!)

christinahedges:
@ojhall94 i believe this functionality now exists in `lightkurve`. I'm including a screenshot of the functionality below.

image

I'm going to close this PR as I believe it is outdated. Please feel free to discuss here and reopen if you disagree.

PR #397, opened 2019-01-26
Fix Travis error (`Can't open /etc/init.d/xvfb`)

barentsen:
Travis has been failing in the past 1-2 weeks with the following error message:
```
$ sh -e /etc/init.d/xvfb start
sh: 0: Can't open /etc/init.d/xvfb
The command "sh -e /etc/init.d/xvfb start" failed and exited with 127 during .
```

This is because [Travis changed the way](https://hk.saowen.com/a/3f9c720bb360f9daa8ad29bdd3fbc48f34ce2aad47215b5802dd4fcc9e781730) in which the `xvfb` service (X Virtual FrameBuffer) ought to be started.  We use `xvfb` to enable our plots to be tested in a headless environment.

This PR should fix the issue!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ python:
   - 3.6
   - 3.7
 dist: xenial
+services:
+  - xvfb  # X Virtual FrameBuffer (enables plotting in a headless environment)
 before_install:  # optional dependencies
   - pip install codecov
   - pip install pytest-cov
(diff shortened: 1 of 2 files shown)


barentsen:
This PR revealed two other platform-related unit test issues in `test_lightcurve.py` so I pushed two additional commits to fix these.  All tests pass now, hence merging!

PR #403, opened 2019-01-29
Add distinct `LombScarglePeriodogram` and `BoxLeastSquaresPeriodogram` classes

barentsen:
The existing `Periodogram` class is sufficiently generic to hold the results of e.g. both a `BoxLeastSquares` and a `LombScargle` period search, however both methods have a few specific attributes and a different initializations.  To support these, I propose we add two light-weight subclasses,  `LombScarglePeriodogram` and `BoxLeastSquaresPeriodogram`, to capture what is specific about both methods.

One new feature this will enable is to find a planet in one line, e.g.:
```python
planet_period = lc.to_periodogram("bls").period_at_max_power
```


lightkurve/periodogram.py
```diff
@@ -695,3 +502,225 @@ def plot(self, **kwargs):
...
+                                      targetid=lc.targetid, label=lc.label)
+
+
+class BoxLeastSquaresPeriodogram(Periodogram):
+
+    def __init__(self, *args, **kwargs):
+        super(BoxLeastSquaresPeriodogram, self).__init__(*args, **kwargs)
+
+    @staticmethod
+    def from_lightcurve(lc, **kwargs):
```
barentsen: :point_up:  This should take all the relevant arguments to initialize a BLS.

barentsen: e.g. we need a min_period, max_period, frequency_factor ...  and a docstring

christinahedges:
@barentsen I don't know what appveyors problem is could you take a look?

barentsen:
The Appveyor stack trace shows that this test raises a `MemoryError`:
```python
# Ridiculous BLS spectra should break.
>           lc.to_periodogram(method='bls', frequency_factor=0.00001)
```

You can probably fix this by making the `frequency_factor` slightly larger to keep Appveyor's precious little brain from over-flowing, or does it have to be that small?

christinahedges:
@barentsen I can't do that, the test is to show that if users require a very high resolution BLS, the initialization breaks with a value error and tells them to calm down.

barentsen:
Fantastic work @christinahedges!  :+1: :+1: :+1: 

This is ready to merge.  I might give Appveyor a few minutes to pass.

In future work we will want to add these awesome new features to our docs.

Issue #407, opened 2019-02-01
Search provides two TPFs for Kepler-76


christinahedges:
#### Problem description
Kepler-76b is in two tpfs. When we search for it, we get both back

#### Example
```Python
lk.search_targetpixelfile('kepler-76', quarter=3)
```
![image](https://user-images.githubusercontent.com/14965634/52144244-9d30d200-2612-11e9-824f-ad0d7471eac8.png)

#### Expected behavior
It should only provide one file.

nksaunders:
After investigating, it seems that the resolution for the radius argument of astroquery's `Observations.query_criteria()` is not precise enough to distinguish these two targets. I'll open a PR to add a warning for cases like this, but I don't see an obvious way to fix it yet.

barentsen:
There are two distinct Target Pixel Files at this location, separated by 11 arcseconds (going by the RA and Dec coordinates in the TPF headers).  The question is why the `distance` column suggests they are at distance 0.

barentsen:
The screenshot of the MAST portal below demonstrates that the two targets have distinct footprints (cf. two white rectangles on the sky view), but their distance from the input search coordinate is incorrectly said to be zero (cf. distance column in the table).

This is problematic because it means that programmatic access to a target is not possible in a predictable way if the target happens to have a close neighbor.

@scfleming @mustaric Is this a known Portal API issue?

![screenshot from 2019-02-01 15-39-26](https://user-images.githubusercontent.com/817669/52155426-98324980-2637-11e9-850c-57d5d7466f31.png)

scfleming:
I'm on it, stay tuned.

barentsen:
Thanks @scfleming!! 🎉

scfleming:
OK so the cause is that these two targets are so close together in the sky that their footprints (which are 4 or 8 arcsecond squares in the Portal, cant remember which) on the sky overlap each other when searching for the exact position of the one star.  See attached screenshot doing a very tight search of 1 arcsecond directly on the coordinate of the other target.

The distance column is the separation between the search coordinate and the closest edge of an observation's footprint.  So, because this coordinate is within the footprints of BOTH Kepler targets, that column is listed as "0.0" for both targets, even though on-sky they are separated by 7+ arcseconds.  Note the mouseover description text for Distance in the Portal UI is not correct, I will file a ticket here to fix that.

[...]

PR #410, opened 2019-02-01
Give warning when Search provides multiple TPFs for a single target

nksaunders:
#407: Astroquery's `Obervations.query_criteria()` function occasionally returns multiple TPFs or LCFs for a single target. This PR adds a warning that the `SearchResult` contains more targets than expected.


barentsen:
Isn't this warning somewhat redundant with the warning that gets triggered when `download()` is called on a `SearchResult` with more than one entry?

I think I'd like to resolve #407 by figuring out why the accurate distance to the input target coordinates isn't reported.

barentsen:
After discussion, we've decided to close this for now until we hear from MAST folks over in #407, because this really looks like a bug on their end.

Issue #299, opened 2018-10-17
Periodogram() should have power normalisation options


ojhall94:
#### Problem description
I've received some feedback from Prof. Tim Bedding that units of Power Spectral Density are useful for solar-type oscillators, but not necessarily for other types. Power Spectral Density should be the default, but other types could be possible.

I'll seek out some more community feedback before building a functional example.
#### Example
```python
pg = lightcurve.to_periodogram(normalization="psd")
```

ojhall94:
Hi Daniel!

Thanks for your great comment--- This is a use case I haven't personally experienced so I'm glad you raise such a clear example. I've had Tim fill me in on the details while he was still here, so I'll try and write a pull request for this soon!

barentsen:
@danielhey Thank you for weighing in!!  Do you have a suggestion as to what Lightkurve's ideal interface might be to enable you to avoid the extra conversions?

danielhey:
Hi all, sorry for the delay in my response!
  
I've been thinking about it and I think a great way for accessing amplitudes instead of fluxes would be an `@property`, something like:
```python
@property
def amplitude(self):
    return self.flux -2.5 * np.log10(self.flux)
```

That way flux is still the default (as it should be!), and the rel. amplitude won't be calculated unless it's needed. Likewise in the `from_lightcurve()` method, maybe a keyword arg could specify the chosen scaling for the spectrum, i.e. normalise=['psd', 'amp', 'power']. 

Attached below are the appropriate rescaling factors for the amplitude and power spectrum:
```python

ndata = len(time)
sc = LombScargle(time, sig).power(fr, normalization='psd')

# 1. if you want to plot the amplitude spectrum
fct = np.sqrt(4./ndata)
amplitude = np.sqrt(sc) * fct

# 2. if you want to have the power spectrum
fct = np.sqrt(4./ndata)
power = fct**2. * sc
```

If you're not working on it already Oliver, I can write up a quick PR.

ojhall94:
Hi Daniel,

If you wanted to do this one that would be great! You've got more experience on this topic so it would almost certainly turn out better than if I did it. Having it as a property is a neat idea, and a kwarg in from_lightcurve() was what I was thinking as well.

Cheers!

barentsen:
I struggle to think of anything more exciting than @danielhey preparing a PR and @ojhall94 testing and reviewing it. :heart: 

[...]

PR #416, opened 2019-02-04
Power spectrum normalisation options

danielhey:
G'day all,

I've added in a power normalization option for the periodogram class following #299 without hopefully changing too much. The spectrum can now be normalized to either `psd` (default) or `amplitude`. i.e.,

```python
pg = lc.to_periodogram(normalization='amplitude')
```

The `amplitude` option should give identical results to Period04 (albeit, in ppm - so multiplied by 1e6). 
During comparison with Period04, I've also noticed that the periodogram (both psd and amplitude) can become somewhat diminished as a consequence of the `oversample_factor` of 1 (see below plot). I've changed the default value to 2 to fix this, which gives the proper amplitudes.

oversample_example


lightkurve/periodogram.py
```diff
@@ -530,8 +533,8 @@ def __repr__(self):
     def from_lightcurve(lc, min_frequency=None, max_frequency=None,
                         min_period=None, max_period=None,
                         frequency=None, period=None,
-                        nterms=1, nyquist_factor=1, oversample_factor=1,
-                        freq_unit=1/u.day, **kwargs):
+                        nterms=1, nyquist_factor=1, oversample_factor=2,
```
barentsen: Question: did you intentionally change the default `oversample_factor` value from `1` to `2`?  Can you explain in a few words why this is a better default?

lightkurve/periodogram.py
```diff
@@ -603,6 +606,10 @@ def from_lightcurve(lc, min_frequency=None, max_frequency=None,
         freq_unit : `astropy.units.core.CompositeUnit`
             Default: 1/u.day. The desired frequency units for the Lomb Scargle
             periodogram. This implies that 1/freq_unit is the units for period.
+        normalization : str
```
barentsen: I suggest making the two valid options explicit in the docstring:
```suggestion
        normalization : 'psd' or 'amplitude'
```

lightkurve/periodogram.py
```diff
@@ -603,6 +606,10 @@ def from_lightcurve(lc, min_frequency=None, max_frequency=None,
...
+            Default: `psd`. The desired normalization of the power spectrum. 
+            Can be either power spectral density (`psd`) or amplitude 
+            (`amplitude)
```
barentsen:
```suggestion
            (`'amplitude'`)
```

lightkurve/periodogram.py
```diff
@@ -603,6 +606,10 @@ def from_lightcurve(lc, min_frequency=None, max_frequency=None,
         freq_unit : `astropy.units.core.CompositeUnit`
             Default: 1/u.day. The desired frequency units for the Lomb Scargle
             periodogram. This implies that 1/freq_unit is the units for period.
+        normalization : str
+            Default: `psd`. The desired normalization of the power spectrum.
```
barentsen:
```suggestion
            Default: `'psd'`. The desired normalization of the power spectrum. 
```

lightkurve/periodogram.py
```diff
@@ -603,6 +606,10 @@ def from_lightcurve(lc, min_frequency=None, max_frequency=None,
         freq_unit : `astropy.units.core.CompositeUnit`
             Default: 1/u.day. The desired frequency units for the Lomb Scargle
             periodogram. This implies that 1/freq_unit is the units for period.
+        normalization : str
+            Default: `psd`. The desired normalization of the power spectrum. 
+            Can be either power spectral density (`psd`) or amplitude
```
barentsen:
```suggestion
            Can be either power spectral density (`'psd'`) or amplitude 
```

lightkurve/periodogram.py
```diff
@@ -697,23 +710,34 @@ def from_lightcurve(lc, min_frequency=None, max_frequency=None,
...
+            power = power * (cds.ppm**2)
+
+            # Rescale power to units of ppm^2 / [frequency unit]
+            power = power / fs
 
-        power = power * (cds.ppm**2)
+        # Amplitude spectrum
+        elif normalization == 'amplitude':
+            factor = np.sqrt(4./len(lc.time))
+            power = np.sqrt(power) * factor * 1e6
```
christinahedges: Above you create `flux_scaling` and then set it to `1`. However, here you multiply by 1e6. Why is this not the `flux_scaling` from above, set to `1e6`?

lightkurve/periodogram.py
danielhey: Hi guys, I did indeed change the `oversample_factor` on purpose. Although I can not speak for exoplanets, in asteroseismology, an oversample factor of 1 corresponds to critical sampling, which I believe is usually only done with solar-like oscillators.

In cases where the mode lifetime is much shorter than the duration observations (i.e. they can be resolved into Lorentzians) it doesn't do any harm to use a critically sampled spectrum. This was a pretty common scenario with _Kepler_ data. But now with _TESS_ we've seen that even some of the solar-like oscillators are not being resolved fully, such as nu Indi in the 1 month data. 

The image I attached in my original comment is a 𝛿 Scuti with coherent modes lasting longer than the 4 yr time-base of observations (common for 𝛿 Scts). Because the spectrum is critically sampled, many of the peaks are not sampled at their exact maximum and their amplitudes (and frequencies) are slightly off. For cases like these, to obtain accurate readings of the frequency and amplitude you'd need to do either some kind of interpolation, or simply oversample the spectrum instead.

The only loss incurred by oversampling is that the frequency bins are no longer statistically independent, and the obvious extra computing budget required. I've spoken to Tim Bedding regarding this and he has also suggested oversampling the spectrum by default. He has actually suggested that it should go up to 5 as a trade-off between proper sampling of frequencies & computing time.

To answer your comment below, @christinahedges,
> Please can you tell me how you determined what the true amplitudes are for your target? If this is just to match another code, I'll need a concrete reason before I change the default.

The true (or at least very close to true) amplitudes were obtained by correct sampling of the frequency grid (oversampling :) ). In the below plot I've zoomed in on the central frequency from my original comment. You can see that with `oversample_factor=1` the frequency grid is insufficient to correctly pick out the right frequency (and consequently, the amplitude is underestimated). In fact, even  `oversample_factor=2` does not fully capture the mode.
oversample_example 1

It is straightforward to show that this is the case by using some artificial data (with known amplitudes) and running it through different oversample factors

Sorry for the very long response to this one point, please let me know if you have any more questions! 

Finally, there's a nice (if not mathematically dense) discussion of oversampling in [this paper](https://academic.oup.com/mnras/article/388/4/1693/981666).

ojhall94:
Hi everyone,
Apologies for being late to the party--- this addition is awesome and I'm super happy @danielhey is making this addition. Thanks for the extensive explanations and links, I'm learning a lot and will definitely be reading your recommended papers.

>Also @ojhall94, do you have any comments?
I think I like the p.normalize() option. This will give more freedom to users who input data in one format and want to express it in another, when not using the LightCurve.to_periodogram() method.

>Period04 is the standard code we use (which is also well tested)
This isn't explicitly relevant to this PR, but it should be noted that Period04 is (to my understanding) the go-to tool for this stuff among large sections of the asteroseismic community. Anecdotally (@keatonb) I've heard people get lightkurve lightcurves, save them, and then put them into Period04. Every step we take to get closer to Period04 functionality is more people using lightcurve :)

keatonb:
> This isn't explicitly relevant to this PR, but it should be noted that Period04 is (to my understanding) the go-to tool for this stuff among large sections of the asteroseismic community. Anecdotally (@keatonb) I've heard people get lightkurve lightcurves, save them, and then put them into Period04. Every step we take to get closer to Period04 functionality is more people using lightcurve :)

See #427

lightkurve/periodogram.py
keatonb: Hi, I just noticed this discussion and thought I might chime in on the issue of oversampling. Within asteroseismology there are definitely going to be two different schools of thought here. Those interested in solar-like oscillators typically fit features in the frequency domain, where oversampling really complicates the distribution of power and associated statistics. For classical oscillators (also for binaries), where signals are often assumed to be coherent throughout the time series, oversampling (usually by a factor of 5-10) and amplitude normalization allows one to quickly read off precise signal properties directly from the periodogram (which are probably then refined by fits to the time series).  This decision should be explicitly chosen by the user based on their specific science case.  I think a good default is oversampling=1 since this requires the least computation time, it yields an easy-to-interpret Chi^2 2 degrees-of-freedom noise distribution in the power spectrum, and it matches the default normalization of 'psd'. 
I caution that the heights of oversampled peaks in amplitude spectra risk misinterpretation when the assumption that the underlying signal is coherent in phase, frequency, and amplitude throughout the time series is not valid, making it a risky default.

Since amplitude normalization and oversampling often go hand-in-hand, perhaps it would be a good idea to issue a warning when a periodogram is computed with normalization='amplitude' and oversample_factor=1?

Would this be a useful thing to add a tutorial about?  I'd be happy to work on that when I find some spare time...

Thanks @danielhey for adding this normalization option.  I've had it on my own to-do list for a while, and now it's one less thing... :smile:

ojhall94: I agree with Keaton on all counts, especially

>Since amplitude normalization and oversampling often go hand-in-hand, perhaps it would be a good idea to issue a warning when a periodogram is computed with normalization='amplitude' and oversample_factor=1?

I also personally feel like lightkurve should be the path of least resistance for people learning to work with these data or having a quick look at them. The less we complicate things the better.

Thanks for adding your voice to the discussion @keatonb, this back and forth is proving very insightful!

timbedding: Hi all, I hope you don't mind me butting in :-) 
I agree completely that lightkurve should be aimed primarily at those with less experience.  With this in mind, I would suggest a default that gives (i) a properly sampled spectrum, and (ii) units that are easy to interpret.  

(i) for my money, "properly sampled" means oversample_factor>~5.  Note that over- vs critically- is unfortunate choice of names that we are stuck with.  The sampling referred to as "critical" gives a ragged ugly curve (see Dan's plots) that not only makes it hard to extract the true frequencies and amplitudes, it actually throws away some information (happy to explain further).  It is true that some experts like to use critical for solar-like modes that are well-resolved.  But that is a specialist application (and less relevant to TESS than Kepler because of the shorter segments).  Most applications (rotation, pulsations, transits) are much better served by providing a well-sampled spectrum that is a closer approximation to the true spectrum.

(ii) units are a subtle issue (see Appendix A1 of https://ui.adsabs.harvard.edu//#abs/1995A&A...293...87K/abstract ).  Power density is used for solar-like oscillators, especially when measuring noise backgrounds from granulation.  Again, this is a fairly specialised and minority area.  I would think the vast majority of users just want to know the amplitude of the sinusoidal variations.  

There are good reasons why Period04 adopted defaults that give fully-sampled spectra with easy-to-interpret amplitudes, and I strongly encourage lightkurve to do the same.

amcody:
I have a question- is the discrete fourier transform (DFT) that Period04 outputs *exactly* the same as the Lomb Scargle when you change the normalization? I had the [potentially] mistaken impression that the two periodograms were slightly different. If they are indeed the same, then when you compute a false alarm probability for a particular power level in the Lomb Scargle, can that be translated into an amplitude in the DFT? Note that I'm aware in finding significant signals, Period04 does a Monte Carlo to take red noise into account. But just curious as to why people frequently draw a line across  their Lomb Scargle periodograms (e.g., to find "significant" peaks above it), whereas no one seems to do this with the DFT/Period04.

timbedding:
Yes, DFT and Lomb-Scargle should be same apart from a normalisation factor.  I have not explicitly tested the python LS routine, but I think @danielhey might have?

Yes, if you want to define a false-alarm threshold you can measure the noise in the DFT simply by taking the average value in regions where there is no signal.  You can then set a threshold of 3 or 4 times the noise as your threshold  There is no difference between this and using the LS statistics, both must be done with care because the noise is usually not flat.

There is a good reason not to draw a line and take all peaks above it, and that is the spectral window.  If the data have gaps, then every peak has sidelobes that will sit above your line, but are not real.  The correct approach is to "clean" away the peaks and their sidelobes one at a time (often called prewhitening), and this is what people do with Period04.  This issue is less important with Kepler and TESS since the data usually don't have gaps, but it is still a good idea.  Even completely continuous data has sidelobes from the sinc function, and sidelobes from a strong mode can still be well above the background noise.

keatonb:
Despite my previous comment, I must agree with @timbedding regarding default normalization (see comments on a post above). Amplitude normalization and oversampling = 5 are very sensible defaults for most science cases. These are easiest to interpret for the uninitiated, and best enable lightkurve to develop into an end-to-end Python alternative to Period04 (see my proposed #427).

I recommend this solution: .to_periodogram() defaults to amplitude normalization and oversampling_factor=5, and a convenience function .to_psd() calls the same routines with psd normalization and oversampling_factor=1 defaults for those who prefer it.

Side note: I'm not convinced that the DFT and L-S results are identical in all cases, but I expect any differences to be negligible for anything other than very sparse, irregularly sampled data. Certainly for Kepler/K2/TESS, I would not hesitate to treat them as the same.

lightkurve/periodogram.py
ojhall94: Hi Tim, thanks for adding to the discussion! This is definitely healthy for lightkurve (and educational for me), given that periodogram() was written by somebody who has only studied solar-like oscillators 😬. 

I agree, and I also think a tutorial based around the different use cases discussed in this thread would be an excellent learning tool (although for a future PR, probably)

amcody:
Thanks for the input. I was confused regarding classical vs. Lomb Scargle periodogram. If you look at the formulae for them, they are different, and so I don't see how they can be identical (after all, if they were, why would Lomb Scargle have its own name?). Anyway, I read up a bit more to clear my confusion and found this article instructive: https://arxiv.org/pdf/1703.09824.pdf  For the LS vs. DFT differences, see equations 31-34 as well as Figure 16 for a comparison plot.

Why do I bring this up? Well, @keatonb asked that we reproduce the behavior of Period04, and that  uses a DFT, NOT a Lomb Scargle. Yes, it's true that these have very small differences for the evenly spaced data of Kepler/K2. But what if someone wants to throw out a bunch of compromised datapoints? I feel like if we are going to emulate Period04, then we need a new class for the DFT, which can be computed by looping over a simple equation. My proposal is that if you use the DFT, you would get an oversampled periodogram in amplitude units, and if you use LS, then you get a periodogram in power units with an oversample factor of 1 or max 2.
Thoughts??

keatonb:
> Why do I bring this up? Well, @keatonb asked that we reproduce the behavior of Period04, and that uses a DFT, NOT a Lomb Scargle. Yes, it's true that these have very small differences for the evenly spaced data of Kepler/K2. But what if someone wants to throw out a bunch of compromised datapoints? I feel like if we are going to emulate Period04, then we need a new class for the DFT, which can be computed by looping over a simple equation. My proposal is that if you use the DFT, you would get an oversampled periodogram in amplitude units, and if you use LS, then you get a periodogram in power units with an oversample factor of 1 or max 2.
> Thoughts??

@amcody  I see no reason that the L-S would not be equally valid (if not superior) for the typical Period04 workflow of selecting peaks from the periodogram, refining the fits of those frequencies to the time series, prewhitening and repeating on the residuals. L-S and DFT are both periodograms with the only difference being a frequency-dependent normalization. In most use cases of lightkurve (Kepler, K2, TESS), the results are practically identical.  The L-S periodogram is actually formulated to be an improvement for the case of data with gaps (this is what you mean by compromised data?), so I believe this to be the preferred choice (not to mention much, much faster!).  Regardless, this is outside the current issue of normalization and oversampling, and perhaps we should discuss this further at #427.

I strongly discourage oversampling by a factor of 2, which produces neither the simple statistics of a "critically" sampled spectrum, nor the amplitude and frequency precision of a well oversampled spectrum. I think that oversampling of ~5 provides the most direct interpretability in the broadest astrophysical contexts as an easily overridden default. Adding a .to_psd() helper function with different defaults would be an easy way to make both preferences equally accessible.

danielhey:
Hi all, I'm pleased (and surprised!) with how much discussion this PR has generated. 

I've experimented with @christinahedges suggestion of implementing a `pg.normalize()` method in the Periodogram class but haven't found an elegant way to do it. The issue is that if we want to re-normalize an already normalized periodogram (say, from `psd` to `amplitude`), then we'd need to keep track of the original periodogram (perhaps a `self.unnormalized` variable?). I've tried to make the PR as unobtrusive as possible, but can continue implementing this if everyone thinks it's worth it. 

For now, I have changed the default to `normalization='amplitude'` and `oversample_factor=5` following the comments above. I've also included a convenience function in the `LightCurve` class (`to_psd`) which will return a critically sampled PSD, for all you solar-like oscillator fans out there.

It seems like some things have changed with the logmedian smoothing test. Any tips on how to cleanly integrate the test would be appreciated!

keatonb:
Hey @danielhey,

> I've experimented with @christinahedges suggestion of implementing a `pg.normalize()` method in the Periodogram class but haven't found an elegant way to do it. The issue is that if we want to re-normalize an already normalized periodogram (say, from `psd` to `amplitude`), then we'd need to keep track of the original periodogram (perhaps a `self.unnormalized` variable?). I've tried to make the PR as unobtrusive as possible, but can continue implementing this if everyone thinks it's worth it.
> 

Instead of storing the entire unnormalized spectrum that is only different by some multiplicative constant, can't you instead just store the current normalization factor?  Then you can un-normalize before re-normalizing.

> For now, I have changed the default to `normalization='amplitude'` and `oversample_factor=5` following the comments above. I've also included a convenience function in the `LightCurve` class (`to_psd`) which will return a critically sampled PSD, for all you solar-like oscillator fans out there.
> 

Great!  Thanks for adding that!

keatonb:
This still waiting on any further changes before merging?  I'm looking forward to implementing the new normalization into some code!

barentsen:
@keatonb thank you for reminding us of this PR!  We lost sight of it in the midst of the Kepler conference and all the other open issues.

It looks like there is general consensus that this PR is very valuable and that it should be merged!

I think we are close to merging, but I think there are five small issues that need to be resolved:
1. I'd love to see the papers mentioned by @danielhey above captured in the docstring and/or in-line comments somewhere, for the sake of documenting the code as well as possible with references etc.
2. Likewise, I'd like to see some text added to the docstring of `LombScarglePeriodogram.from_lightcurve()` which explains the recommended choice of `oversample_factor` (to capture the discussion that took place on this thread about that parameter, and to guide new users).
3. I suggest removing the `to_psd()` method for now.  The reason is that we want all `to_something()` methods in Lightkurve to return an object of type `something` (e.g. `to_lightcurve` returns a `LightCurve`, `to_periodogram` returns a `Periodogram`).  There is no `psd` object, so we can't call the method `to_psd()`.  For the sake of simplicity, I wonder if we can do without the method for now, and instead require users to specify the desired normalization when the `Periodogram` is first created.
4. The unit tests are failing; we'll have to fix those.
5. Finally, we need to execute a `git rebase master` and resolve conflicts with other recent changes in `periodogram.py`.

Does that make sense?  Please do raise your hand if this doesn't make sense, or if I'm forgetting something!

@danielhey I will definitely need your help with items 1-2 (adding a few sentences may be enough).  I'm happy to help take care of items 3-4-5.  Let us know!  And thanks everyone! :+1: :+1: :+1:

danielhey:
Well, I force pushed to my repo and ruined everything. Is it worth trying to recover this @barentsen? Perhaps it might be easier if I just branch from the latest commit and work there, since my PR adds very little code.

grd349:
Hey folks - love this addition.

I love thew idea of a to_psd method but as @barentsen suggests this breaks convention.

How about a `normalisation='psd'` which then sets sampling to critical, frequency units to microHz, and normalises to power.  This would then be my standard config.

ojhall94:
I agree. A lot of people studying solar-like oscillators are and will be using lightkurve, so we will want to make this transition a painless as possible and also try to avoid giving the impression that lightkurve favours one subfield of asteroseismology over the other. Since the periodogram default should be 1/days for planet purposes anyway, `normalisation='amplitude'` (or maybe, `type='classical'`?) could be the default, with an easy switch to `normalisation='psd'` (or `='solarlike'`), as Guy suggests.

At the moment we have to import astropy.units and set the frequency units to microhertz when using lightkurve for anything, so I would actually welcome a simple kwarg in the to_periodogram() call that would me set the common preferences for solar-likes in one go!

keatonb:
I think that the `normalization` should be independent of the oversample_factor and frequency units, since these can be individually configured.  But I like the idea of including a few preset named configurations through a different keyword rather than including a separate .to_psd(). `type` is already a python function, so maybe something like a `preset` or `convention` keyword could equal one of "classical", "solarlike" or maybe even "exoplanet" to set the normalization, oversampling, and units to some convenient defaults? These should each be able to be overwritten if the corresponding keywords are specified (I guess default values of `None` for the oversample_factor, normalization, and units?  Preferring `preset = classical`).

But if this discussion of new defaults is slowing this merge, it would be nice to at least get the new amplitude normalization option implemented through the keyword, possible before we make changes to default behavior. I don't mind setting my preferences manually 😃

barentsen:
In relation to the defaults: my preferred solution is to go with a solution that minimizes the number of keywords and methods involved (because simple APIs tend to be easier to learn and maintain).

### Proposal

Based on the feedback I read above, I suggest we set the defaults to:
```normalization="amplitude", oversample_factor=None, freq_unit=None```

Then, in the docstring, we could explain that:
```If normalization="amplitude", the default value for `oversample_factor` is set to 5 and `freq_unit` is 1/day because [insert explanation]. If `normalization="psd"`, the default value for `oversample_factor` is set to 1 and `freq_unit` is set to microHz because [insert explanation].  These defaults can be overridden by the user. Users doing [insert science] typically prefer `normalization="amplitude"` because [insert explanation], users doing [insert science] tend to prefer  `normalization="psd"` because [insert explanation].```
```

### Questions
* Did I understand y'all correctly? Does my proposal make sense? Please criticize!
* Can we use `normalization="power"` instead of `normalization="psd"` to avoid the use of an acronym, or is this an incorrect synonym?  (Note: it is easy to treat `power` and `psd` as synonyms in the code if that makes snese.)

(PS: I am *loving* this thread. :heart:  Discussions like these are key to building good software!  APIs are worth thinking carefully about.)

ojhall94:
This is perfect @barentsen, imo!

`If `normalization="psd"`, the default value for `oversample_factor` is set to 1 to avoid correlations between frequency bins, and `freq_unit` is set to microHz to match conventions for the study of solar-like oscillators. These defaults can be overridden by the user. 

... users studying solar-like oscillators tend to prefer `normalization="psd"` because relative power is more important than amplitude, and `psd` rescales the power on a scale that is consistent for all lengths of observation.`

Tell me off if any of that is wrong, @grd349, @keatonb ! :)

timbedding:
Hi all, as discussed above, I really think a default of oversample_factor=1 is the wrong choice.  That value should only be used by a subset of experts (whose peakbagging code requires it) looking at a particular subset of stars (solar-like oscillators) in cases where the observations are long enough to resolve the modes fully (true for Kepler, but often not true for TESS).  For non-experts, and those looking at one of the numerous types classical pulsators (as I am doing right now, as it happens), oversample_factor=5 is the sensible default.  Otherwise, I fear that many non-experts will end up measuring incorrect frequencies and amplitudes from their data.

grd349:
Hi Tim,

I think the suggestion is that the default is normalisation='amplitude'
which would make the oversampling default 5 under the suggested paradigm.

Then the oversampling would be unity if you select 'psd' or 'power'.

I could be wrong,

G

timbedding:
Hi Guy

Ok, great.  Sorry, my mistake, I thought "oversample_factor=None" meant "oversample_factor=1".

I am really enjoying lightkurve, by the way, for both TESS and Kepler data.  Thanks to all of you!

    Tim

barentsen:
@timbedding To be explicit: would you be fine with `oversample_factor=5` as the default for amplitude normalization, and `oversample_factor=1` as the default for psd normalization?

Just so I understand: is there a reason why `oversample_factor=1` is less tricky/dangerous in psd normalization mode than it is in amplitude mode?  Or is this merely a matter of different preferences/habits in the amplitude vs psd "communities"?  (i.e. can someone complete my docstring from 3 posts up to explain this?  That would be awesome! 🤓)

timbedding:
@barentsen Yes Geert, that would be fine with me.
Oversample=1 means the spectrum is critically sampled, with every point in the spectrum independent of the others.  Some people who fit Lorentzians to modes in the power spectrum like this because the statistics of their fit are well-behaved.  (actually, it is not hard to make a correction if an oversampled spectrum is used, but they don't seem to like it for reasons I don't quite understand :-) )
The reason oversample=1 it is less likely to cause problems for people using the psd is that they are usually looking at resolved modes, that is, the power from each mode is spread over a range of frequencies (due to dampling).  Hence, any small error from measuring mode frequencies by taking the maximum of the peak is negligible compared with the intrinsic linewidth of the modes.  

(Hmm, that's probably too much information for a docstring, sorry!)

   Tim

barentsen:
> (Hmm, that's probably too much information for a docstring, sorry!)

Not at all!  This is excellent.  I'd like to include a version of your paragraph in the docstring to help beginners understand these choices! Thanks+++! 👍

danielhey:
Thanks to all for the insightful comments! This thread has become a great resource on the more subtle aspects of LS periodograms and normalization methods in asteroseismology, and I hope I've captured that in the updated docstrings.

The default `normalization` is now "amplitude", which will automatically set `freq_units=1/u.day` and `oversample_factor=5`. If "psd" normalization is chosen, the frequency units will be in `u.microhertz` and the spectrum will be critically sampled. The default frequency units and oversample factor can be overridden by specifically calling them in `to_periodogram()`.

@barentsen, can you please check my commits? In particular, whether I've done anything crazy to the git history by force pushing. If I have, then I have a separate branch with all of these changes ready to PR.

lightkurve/lightcurve.py
```diff
@@ -1048,6 +1048,12 @@ def to_periodogram(self, method="lombscargle", **kwargs):
             Keyword arguments passed to either
             `~lightkurve.periodogram.LombScarglePeriodogram` or
             `~lightkurve.periodogram.BoxLeastSquaresPeriodogram`.
+            Keywords accepted by `~lightkurve.periodogram.LombScarglePeriodogram` are:
+            ``min_frequency``, ``max_frequency``, ``min_period``, ``max_period``, ``frequency``,
+            ``normalization``.
+            ``period``, ``nterms``, ``nyquist_factor``, ``oversample_factor``, ``freq_unit``.
+            Keywords accepted by `~lightkurve.periodogram.BoxLeastSquaresPeriodogram` are:
+            ``min_period``, ``max_period``, ``period``, ``frequency_factor``.
```
barentsen: Minor change request: we moved the listing of these options to lines 1031-1038 in this same docstring (this changed just a few weeks ago, i.e. after this PR started).  To avoid repetition, would you be willing to add `normalization` on line 1034 in this docstring and removing lines 1051-1056?

AUTHORS.rst
```diff
@@ -46,3 +46,4 @@ Comments, corrections & suggestions:
 - `Molnár László <https://github.com/lacalaca85>`_
 - `Tom Wilson <https://github.com/Onoddil>`_
 - `Tara Fetherolf <https://github.com/tfetherolf>`_
+- `Timothy Bedding <https://github.com/timbedding>`_
```
barentsen: :+1:

lightkurve/periodogram.py
```diff
@@ -566,23 +570,56 @@ def from_lightcurve(lc, minimum_frequency=None, maximum_frequency=None,
...
+        (http://adsabs.harvard.edu/abs/1995A%26A...293...87K).
+
+        The parameter nterms controls how many Fourier terms are used in the
+        model. Setting the Nyquist_factor to be greater than 1 will sample the
+        space beyond the Nyquist frequency, which may introduce aliasing.
+
+        The `freq_unit` parameter allows a request for alternative units in frequency
+        space. By default frequency is in (1/day) and power in (amplitude
+        (ppm)). Asteroseismologists for example may want frequency in (microHz)
+        in which case they would pass `freq_unit=u.microhertz`.
```
barentsen: I love the detail captured in this docstring now.  :heart:  In the future we should think about walking the users through this information as part of a tutorial.

timbedding: Geert @barentsen I am happy to read through the doc but I am new to github and can't figure out how to see it!
Can you please help (or email tim,bedding@sydney.edu.au)?

barentsen: @timbedding Sure thing!  You can read the newly-proposed docstring at this url: https://github.com/KeplerGO/lightkurve/blob/466548846e0b818bd70738f48c52d5f40adbbd6b/lightkurve/periodogram.py#L561

barentsen: Oops, GitHub renders the url into a fancy widget.  The full url is: `https://github.com/KeplerGO/lightkurve/blob/466548846e0b818bd70738f48c52d5f40adbbd6b/lightkurve/periodogram.py#L561-L682`

lightkurve/lightcurve.py
danielhey: Done! I've also removed the line regarding how LombScargle(normalization) is fixed to PSD (technically unnormalized) to avoid confusion with the normalization parameter in to_periodogram().

lightkurve/periodogram.py
danielhey: I've added some of Tim's suggestions to the docstring

keatonb:
Since this will change the periodogram defaults, I think it's important to issue a warning for a while when this function is called with the new defaults.  Otherwise it's bound to break existing code and leave users confused upon the next update.

barentsen:
@danielhey Thank you for the additional commits!  Are you happy for this version to be merged?

@keatonb I guess the change in defaults you refer to is that `oversample_factor` will now be 5 instead of 1 by default.  We must certainly highlight this in the release notes.  The drawback of adding a warning is that users tend to become insensitive to warnings if you have too many, so I like to put the bar pretty high.  Do you think it is true that this change will break or impact existing code in a very significant way (in which case a warning is justified)?

keatonb:
@barentsen In addition to the new default oversampling, the biggest change is that the new default normalization will give amplitudes instead of powers.  I do expect that to break **all** code that uses the periodogram class, since the only output previously available was in power units.

barentsen:
@keatonb Aren't amplitudes the existing default? I don't think that has changed, fortunately!

keatonb:
@barentsen Take a look at the changes to the code.  The original implementation only allowed for output in units of power.  This PR provides the option for amplitude output, and sets it as the default.  I think this is a sensible default, but the change is significant and will likely be unexpected.

barentsen:
@keatonb You're absolutely right!  I had it the wrong way around.  Ok, let's add a warning.

barentsen:
I am going to go ahead and merge this!  (We are still on the fence over the introduction of a warning here, but may add one based on feedback!)

Thanks everyone for working on this!  :+1:  Any further discussions/corrections/changes continue to be very welcome as new issues or PRs, of course!

ojhall94:
I'd like to throw my voice in and say a warning is definitely necessary here. For people studying solar like oscillators, being alerted of the change in units and especially the oversample rate will be crucial if they've built this into any existing pipeline that fits anything to the frequency or amplitude data.

barentsen:
Understood!  I opened a PR to add a warning; see #493.

Issue #408, opened 2019-02-01
Bring `SFFCorrector` in line with the new `Corrector` API design


barentsen:
We need to change the API of `SFFCorrector` slightly to bring it in line with the API described in #319.  In particular:
* the constructor should take a `LightCurve` object;
* we need to deprecate `LightCurve.correct()` in favor of a new `LightCurve.to_corrector()` method;
* stretch goal: add a `diagnose()` method.

barentsen:
Resolved by #417.  Closing.

PR #417, opened 2019-02-04
Refactor `SFFCorrector` to use new `Corrector` API design

barentsen:
This PR addresses #408.

This PR also deprecates `LightCurve.correct()` in favor of a new `LightCurve.to_corrector()` method.

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -1094,8 +1094,9 @@ class KeplerTargetPixelFile(TargetPixelFile):
         """
         allowed_methods = ["pld"]
         if method == "sff":
-            raise ValueError("The 'sff' method can only be used on "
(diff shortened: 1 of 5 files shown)


lightkurve/targetpixelfile.py
```diff
@@ -1093,6 +1093,9 @@ def to_corrector(self, method="pld"):
             and `diagnose()` methods.
         """
         allowed_methods = ["pld"]
+        if method == "sff":
+            raise ValueError("The 'sff' method can only be used on "
+                             "`LightCurve` objects, not `TargetPixelFile` objects.")
```
christinahedges: Can you add something along the lines of

```
"Use `tpf.to_lightcurve()` to convert your `TargetPixelFile` to a `LightCurve` using default aperture."
```

PR #418, opened 2019-02-04
Return FFI cutouts with `search_tesscut`

### LINKED ISSUE #402: `search_targetpixelfile` does not support FFI cutouts
#### Problem description

TESS Search isn't finding my target.
#### Example

```python
import lightkurve
lk.search_targetpixelfile('TIC 206669860')
```

image

It is available on MAST

image

#### Expected behavior

It should find the target. 

#### Environment:
- platform (e.g. Linux, OSX, Windows): OSX
- lightkurve version (e.g. 1.0b6): 1.0b26.dev
- installation method (e.g. pip, conda, source): source

nksaunders:
Fixes #402: This PR extends the functionality of `search_targetpixelfile` and allows users to access FFI cutouts. 

- [x] Check if FFIs were returned by query
- [x] Append row with FFI observation
- [x] Aquire RA & dec of target
- [x] Download with `astroquery.mast.TesscutClass`
- [x] Read in data as a `TessTargetPixelFile`
- [x] Allow custom cutout size
- [x] Allow specific sectors


nksaunders:
Still to do:
- [ ] Allow `SkyCoord` objects to be queried on FFIs
- [ ] Fix `download_all()` to parse FFI cutouts and include them in downloaded object
- [ ] Discuss API: dedicated `search_TESScutout`? `search_TESS_FFIcutout`? Or should it be included in general `SearchResult`?
- [ ] Add tests

lightkurve/search.py
```diff
@@ -143,14 +143,24 @@ def download(self, quality_bitmask='default', download_dir=None):
...
+            path = Observations.download_products(self.table[:1], mrp_only=False,
+                                                  download_dir=download_dir)['Local Path'][0]
 
         # open() will determine filetype and return
-        return open(path[0], quality_bitmask=quality_bitmask)
+        return open(path, quality_bitmask=quality_bitmask)
 
     @suppress_stdout
-    def download_all(self, quality_bitmask='default', download_dir=None):
+    def download_all(self, quality_bitmask='default', download_dir=None, cutout_size=None):
```
barentsen: Same here: please add a docstring and default value for `cutout_size`.

lightkurve/search.py
```diff
@@ -231,6 +250,53 @@ def _default_download_dir(self):
...
+        sector = int(self.table[0]['description'][-2])
+        if cutout_size is None:
+            cutout_size = 5
+        cutout_path = tc.download_cutouts(coords, size=cutout_size,
+                                          sector=sector, path=download_dir)
+
+        path = os.path.join(download_dir, cutout_path[0][0])
+        return path
+
+    def _resolve_coords(self, target):
```
barentsen: I propose removing this function because it really only wraps one line, i.e. `MastClass()._resolve_object(target)`.  It'll keep our code shorter if we call that line directly where needed.

lightkurve/search.py
```diff
@@ -407,6 +473,34 @@ def search_lightcurvefile(target, radius=None, cadence='long',
...
+    target : str, int, or `astropy.coordinates.SkyCoord` object
+        Target around which to search. Valid inputs include:
+
+            * The name of the object as a string, e.g. "Kepler-10".
+            * The KIC or EPIC identifier as an integer, e.g. 11904151.
+            * A coordinate string in decimal format, e.g. "285.67942179 +50.24130576".
+            * A coordinate string in sexagesimal format, e.g. "19:02:43.1 +50:14:28.7".
+            * An `astropy.coordinates.SkyCoord` object.
+    sector : int
+        TESS Sector number.
```
barentsen: Please explain the default behavior (i.e. what does `sector=None` do).  It looks like this argument also accepts a `list of int`, so we'll want to explain that here too.

lightkurve/search.py
```diff
@@ -441,24 +535,39 @@ def _search_products(target, radius=None, filetype="Lightcurve", cadence='long',
...
+        # pull sector numbers
+        s = observations['sequence_number'][i]
+        # convert to pandas data frame to add FFI cutout row
+        products_df = masked_result.to_pandas()
+        # if the desired sector is available, add a row to the data frame
+        if s in np.atleast_1d(sector) or sector is None:
+            products_df = products_df.append({'description': 'Target pixel file (s{:02})'.format(s),
+                                              'target_name': str(target),
+                                              'productFilename': 'TESScut Full Frame Image Cutout'},
+                                              ignore_index=True)
```
barentsen: I think `products_df` has a `sequence_number` column.  I recommend setting it to `observations['sequence_number'][i]` so that the sector number will easily be available in `SearchResult.table` for cutouts.

lightkurve/search.py
```diff
@@ -231,6 +250,53 @@ def _default_download_dir(self):
...
+        -------
+        path : str
+            Path to locally downloaded cutout file
+        """
+        from astroquery.mast import TesscutClass
+        tc = TesscutClass()
+
+        # Resolve SkyCoord of given target
+        coords = self._resolve_coords(target)
+        sector = int(self.table[0]['description'][-2])
```
barentsen: This looks like it will break from sector 10 onwards.  I recommend using
```
sector = self.table[0]['sequence_number']
```
See my comment at line 568 below where we can set the sequence_number to enable this.

lightkurve/search.py
```diff
@@ -407,6 +473,34 @@ def search_lightcurvefile(target, radius=None, cadence='long',
         return SearchResult(None)
 
 
+def search_TESScutout(target, sector=None):
```
barentsen: Let's make this a lowercase function name (because all other function names in Lightkurve are lowercase).  How about `search_cutout()`?

gully:
TESS Cutouts are different than traditional TPFs in a few subtle ways that will break some lightkurve expectations.  Here are some demos:

*demo planet host sources* 
```python
import lightkurve as lk

result_cutout = lk.search_cutout(2758565)
result_classic = lk.search_targetpixelfile(9006668)
tpf_cutout = result_cutout.download()
tpf_classic = result_classic.download()
```

1. The result objects RA and DEC attributes are not populated
```python
result_cutout.ra, result_cutout.dec
(array([nan]), array([nan]))

result_classic.ra, result_classic.dec
(array([349.231362]), array([-18.606628]))
```
*impact:* low-- result objects are scarcely used directly.


2. The HDU `APERTURE` extension defaults to *floats* rather than *int* dtypes.
Float dtypes presently breaks expectations in `background_mask()`

```python
tpf_classic.background_mask.shape
(11, 11)

tpf_cutout.background_mask.shape
TypeError: ufunc 'bitwise_and' not supported for the input types...
```
*impact:* low-- we should update mask handing to gain resilience to float masks.  
*guidance for MAST*: update cuout fits files to use the same datatype as mission-provided products.

3. `cadenceno` appears populated with running index numbers, rather than absolute values

```python
tpf_classic.cadenceno
array([ 91191,  91192,  91194, ..., 110920, 110921, 110922], dtype=int32)

tpf_cutout.cadenceno
array([   0,    1,    2, ..., 1225, 1226, 1227])
```
*impact:* low.  cadenceno is scarcely relied upon in lightkurve.
*guidance for MAST:* Are these values MAST-provided or filled in?  MAST could update to real cadence numbers if available.

4. lightkurve cache location deviates from traditional directory structure
```python
tpf_classic.path
'~/.lightkurve-cache/mastDownload/TESS/tess2018234235059-s0002-0000000009006668-0121-s/tess2018234235059-s0002-0000000009006668-0121-s_tp.fits'

tpf_cutout.path
'~/.lightkurve-cache/tess-s0002-1-3_356.283124_-17.449174_5x5_astrocut.fits'
```
*impact:* low. cached data access is invisible and seamless for most users.
*guidance for MAST:* Are there prospects for standardizing cutout filenames?

5. No pos_corr available
```python
tpf_classic.pos_corr1
array([-0.02342336, -0.00911357, -0.02193538, ..., -0.00247887,
        0.00463906,  0.00013731], dtype=float32)

tpf_cutout.pos_corr1
array([0., 0., 0., ..., 0., 0., 0.], dtype=float32)
```
*impact:* medium?. pos_corr is expected by some detrending algorithms.
*guidance for MAST:* none?

gully:
Addendum: `tpf_cutout.targetid` is an empty string for cutouts.  If the TIC ID is passed into search, could/should it be passed on to the tpf.targetid?

lightkurve/search.py
```diff
@@ -20,7 +20,7 @@
 
 log = logging.getLogger(__name__)
 
-__all__ = ['open', 'search_targetpixelfile', 'search_lightcurvefile']
+__all__ = ['search_targetpixelfile', 'search_lightcurvefile', 'search_cutout', 'open']
```
christinahedges: I'm gonna suggest again that this should be called `search_TESScutout`. My concern is that the majority of TESS users won't notice that this feature exists, and anyone who tab completes and sees `search_cutout` will have no idea what it really does. 

I think we need a more informative name.

lightkurve/search.py
```diff
@@ -143,14 +145,35 @@ def download(self, quality_bitmask='default', download_dir=None):
         if download_dir is None:
             download_dir = self._default_download_dir()
 
-        path = Observations.download_products(self.table[:1], mrp_only=False,
-                                              download_dir=download_dir)['Local Path']
+        # if table contains TESScut search results, download cutout
+        if 'TESScut' in self.table[0]['productFilename']:
+            if cutout_size is None:
+                cutout_size = 5
+            elif cutout_size < 0:
```
christinahedges:
```suggestion
            elif cutout_size <= 0:
```

lightkurve/search.py
```diff
@@ -191,12 +216,27 @@ def download_all(self, quality_bitmask='default', download_dir=None):
         if download_dir is None:
             download_dir = self._default_download_dir()
 
-        path = Observations.download_products(self.table, mrp_only=False,
-                                              download_dir=download_dir)['Local Path']
+        # if table contains TESScut search results, download cutouts
+        if 'TESScut' in self.table[0]['productFilename']:
+            if cutout_size is None:
+                cutout_size = 5
+            elif cutout_size < 0:
```
christinahedges:
```suggestion
            elif cutout_size <= 0:
```

lightkurve/search.py
```diff
@@ -191,12 +216,27 @@ def download_all(self, quality_bitmask='default', download_dir=None):
...
+                raise ValueError('`cutout_size` must be positive.')
+            elif cutout_size > 100:
+                warnings.warn('Cutout size is large and may take a few minutes to download.')
+            path = [self._fetch_tesscut_path(t, s, download_dir, cutout_size)
+                    for t,s in zip(self.table['target_name'], self.table['sequence_number'])]
+        else:
+            if cutout_size is not None:
+                warnings.warn('`cutout_size` can only be specified for TESS '
+                              'Full Frame Image cutouts.', LightkurveWarning)
+
```
christinahedges: There's a bit of code duplication here between `download` and `download_all`. You could make a helper function to do this, but you don't have to do it for this PR.

christinahedges: For example, you could move this `cutout_size` test to the `_fetch_tesscut_path` in order to have it only called once.

lightkurve/search.py
```diff
@@ -409,6 +478,35 @@ def search_lightcurvefile(target, radius=None, cadence='long',
...
+        list of desired sectors can also be provided.
+
+    Returns
+    -------
+    result : :class:`SearchResult` object
+        Object detailing the data products found.
+    """
+    try:
+        return _search_products(target, filetype="ffi", mission='TESS', sector=sector)
+    except SearchError as exc:
```
christinahedges: Why are we not letting the exception be raised here? Why log it instead?

lightkurve/tests/test_search.py
```diff
@@ -85,6 +85,31 @@ def test_search_lightcurvefile(caplog):
...
+    tpf = search_string.download()
+    # Ensure the correct object has been read in
+    assert(isinstance(tpf, TessTargetPixelFile))
+    # Ensure default size is 5x5
+    assert(tpf.flux[0].shape == (5,5))
+    # Download with different dimensions
+    tpf = search_coords.download(cutout_size=4)
+    # Ensure correct dimensions
+    assert(tpf.flux[0].shape == (4,4))
+
```
christinahedges: Only other test I can suggest is trying to download the "edge" of an FFI and checking that an exception is raised.

lightkurve/search.py
barentsen: In a GO office meeting, we decided on the name `search_tesscut()`.

barentsen:
Woohoo! The tests pass! Thank you for your hard work on this @nksaunders! I'd be happy to see this merged now.

Also, thank you @gully for highlighting the differences between 'real' and 'cutout' TPFs.  We've reached out to MAST/TESS/Tesscut to resolve some of these issues, and we should continue to do so, though really on the Lightkurve side the only thing we can do is make sure our tools are resilient enough to work with cutout TPFs as well as possible.

Issue #384, opened 2018-12-27
Bring back the phase argument for lightcurve.fold() [Redux of Issues #357 and #361]


KenMighell:
#### Problem description

The lightcurve.fold() method used to have an optional phase argument.  Recently following disccusions in Issues #357 & #361, the phase argument was removed and an optional transit_midpoint argument was added.  

The removal of the phase argument broke one of my demos for the upcoming American Astronomical Society meeting in Seattle.  I partially fixed the demo by commenting out the use of the phase argument; the fix was only partial as the resulting plots do not look as good as before.

This raises the issue that **not all periodic variable objects are transiting exoplanets**.  

Phase was useful in modifying (hacking) lightkurve phase plots of RR Lyrae variables (or eclipsing binaries) produced by the lightcurve.fold() method so that the minimum flux had a phase value of zero -- this is a current convention among astronomers doing research with variable stars.  This was also useful during the production of quick-look (non-publication grade) phase plots used during data exploration.

Suggestions:

[1]  Adding phase=None to the definition of the fold() method in lightcurve.py combined with the detection of the None value doing *nothing* would enable current applications which use transit_midpoint to continue working as expected.  

[2]  If a phase argument *is* used with a float value between -0.5 and +0.5, then the phase plot X-axis data could be shifted by the requested phase value (as before) and older lightkurve apps using phase would continue working as expected.  

[3]  If *both* phase and transit_midpoint were to be defined, then the phase value would be ignored.

[4]  If phase *is* added back,  the old docstring describing the use of phase would need to be improved (clarified).

#### Environment:
- platform (e.g. Linux, OSX, Windows): Mac OS X (Sierra and El Capitan)
- lightkurve version: 1.0b26.dev
- installation method: git

barentsen:
Thanks for reporting this @KenMighell.

I'm sorry to hear the change broke your demo.  The meaning of the `phase` parameter was indeed very unclear in the past, which led to change.

For the use case you describe, perhaps a new parameter called `phase_shift` could be introduced, which would allow the phase to be shifted by some amount.  It would default to zero and be applied in addition to `transit_midpoint`.

However, we would want to be certain that `phase_shift` adds functionality that is necessary and important.  Right now one can also hack the `transit_midpoint` parameter to plot a folded RR Lyrae star with minimum flux at phase zero roughly as follows:
```Python
lc.fold(period=xxx, transit_midpoint=lc.time[np.argmin(lc.flux)])
```

[...]

PR #419, opened 2019-02-04
Rename transit_midpoint into t0

barentsen:
This PR proposes to change `transit_midpoint` into `t0` as the recommended way to specify the reference time in the `LightCurve.fold()` method, i.e.
```python
lc.fold(period=2.34, transit_midpoint=3456.7)
```
becomes
```python
lc.fold(period=2.34, t0=3456.7)
```

The reason is that `transit_midpoint` is too specific to exoplanets (cf. #384).  However this PR retains `transit_midpoint` as a valid synonym for `t0`, i.e. we won't break people's code (for now).

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -376,21 +376,23 @@ class LightCurve(object):
             return flatten_lc, trend_lc
         return flatten_lc
 
-    def fold(self, period, transit_midpoint=0.):
-        """Folds the lightcurve at a specified ``period`` and ``transit_midpoint``.
+    def fold(self, period, t0=None, transit_midpoint=0.):
+        """Folds the lightcurve at a specified ``period`` and reference time ``t0``.
 
         This method returns a `FoldedLightCurve` object in which the time
         values range between -0.5 to +0.5 (i.e. the phase).
-        Data points which occur exactly at ``transit_midpoint`` or an integer
-        multiple of ``transit_midpoint + n*period`` will have time value 0.0.
+        Data points which occur exactly at ``t0`` or an integer multiple of
+        ``t0 + n*period`` will have phase value 0.0.
 
         Parameters
         ----------
         period : float
             The period upon which to fold.
+        t0 : float, optional
+            Time corresponding to zero phase. In the same units as the
+            LightCurve's ``time`` attribute.  Defaults to 0.
         transit_midpoint : float, optional
-            Time reference point in the same units as the LightCurve's ``time``
-            attribute.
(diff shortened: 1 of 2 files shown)


christinahedges:
`t0` is not very descriptive, what's wrong with `time_midpoint`?

barentsen:
Isn't `time_midpoint` exoplanet-specific?  I'm thinking an RR Lyra star does not have a mid-point but a transit does.

christinahedges:
I'm clearly a little biased towards planets and binaries (I said it) so I'm happy to go with your decision on this one.

lightkurve/lightcurve.py
```diff
@@ -376,39 +376,42 @@ def flatten(self, window_length=101, polyorder=2, return_trend=False,
...
-        """Folds the lightcurve at a specified ``period`` and ``transit_midpoint``.
+    def fold(self, period, t0=None, transit_midpoint=0.):
+        """Folds the lightcurve at a specified ``period`` and reference time ``t0``.
 
         This method returns a `FoldedLightCurve` object in which the time
         values range between -0.5 to +0.5 (i.e. the phase).
-        Data points which occur exactly at ``transit_midpoint`` or an integer
-        multiple of ``transit_midpoint + n*period`` will have time value 0.0.
+        Data points which occur exactly at ``t0`` or an integer multiple of
+        ``t0 + n*period`` will have phase value 0.0.
```
christinahedges: Good catch

lightkurve/lightcurve.py
```diff
@@ -376,39 +376,42 @@ def flatten(self, window_length=101, polyorder=2, return_trend=False,
...
         ----------
         period : float
             The period upon which to fold.
+        t0 : float, optional
+            Time corresponding to zero phase. In the same units as the
+            LightCurve's ``time`` attribute.  Defaults to 0.
         transit_midpoint : float, optional
-            Time reference point in the same units as the LightCurve's ``time``
-            attribute.
+            Synonym for `t0`. This parameter will be ignored if ``t0`` is set.
```
christinahedges: Why are we letting people set two things? This doesn't seem pythonic...

barentsen: I was thinking that, in this specific case, breaking people's code is worse than being non-Pythonic.  But I suppose I could add a deprecation warning when people use `transit_midpoint`, so we can remove it in a few months!

barentsen: I have gone ahead and added a deprecation warning that will be issued whenever someone uses `transit_midpoint`, and updated the docstring accordingly, so it's more Pythonic know (though won't break anything right away).

PR #423, opened 2019-02-05
Discussion: make LightCurve _repr_() more fancy in Jupyter notebooks?

barentsen:
For discussion: I wonder if there is merit in showing a snippet of the data as part of the `LightCurve` repr in Jupyter notebooks, to help users understand the data columns contained in a `LightCurve`.

Here is what it could look like. Thoughts?

screenshot from 2019-02-05 08-11-52

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1362,7 +1362,10 @@ class KeplerLightCurve(LightCurve):
     def _repr_html_(self):
         """HTML representation which will display in notebooks by default."""
         html = "<p>{}</p>".format(self.__repr__())
-        html += self.to_pandas().reset_index(drop=True).to_html(max_rows=6, notebook=True)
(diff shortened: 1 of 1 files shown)


christinahedges:
This is pretty wordy. And if you use the pandas repr, I would expect to be able to interact with it exactly like a pandas table.

barentsen:
That's a fair comment!  Anyone want to vouch in favor?  If not, I'll just close this for now!

gully:
I sort of like it!  Maybe a hybrid solution is to allow `lc.head()`, or other suitably-disambiguated method name, which facilitates quicklooks.  It's nice to now what columns are available, what order-of-magnitude the values take, and what the shape is.

We could also consider adding [numpy_html](https://pypi.org/project/numpy-html/) as an optional dependency, and then displaying beautiful arrays and matrices.

PR #425, opened 2019-02-06
Added Odd and Even mask arrays to FoldedLightCurve

christinahedges:
With folded light curves, it's often useful to see odd and even transits. This can be used to see if the signal is a planet or an eclipsing binary. It's also useful to compare the shape of odd and even transits.

I've added two attributes to `FoldedLightCurve`: `odd` and `even` which returns a mask of booleans. You use it to index the folded light curve.

```
f = lc.fold(...)
f[f.odd].scatter()
f[f.even].scatter()
```

Here's a demo:

image

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -27,7 +27,7 @@ from .utils import (
     LightkurveWarning, validate_method
 )
 
-__all__ = ['LightCurve', 'KeplerLightCurve', 'TessLightCurve']
+__all__ = ['LightCurve', 'KeplerLightCurve', 'TessLightCurve', 'FoldedLightCurve']
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


lightkurve/lightcurve.py
```diff
@@ -1233,12 +1235,26 @@ class FoldedLightCurve(LightCurve):
...
         super(FoldedLightCurve, self).__init__(*args, **kwargs)
 
     @property
     def phase(self):
         return self.time
 
+    @property
+    def odd(self):
+        ''' Mask to show only odd numbered transits/oscillations. '''
+        phase = (self.t0 % self.period) / self.period
```
barentsen: I think we can just use `self.time` here instead of re-computing `phase`.

lightkurve/lightcurve.py
```diff
@@ -1233,12 +1235,26 @@ class FoldedLightCurve(LightCurve):
...
+        self.period = kwargs.pop("period", None)
+        self.t0 = kwargs.pop("t0", None)
         super(FoldedLightCurve, self).__init__(*args, **kwargs)
 
     @property
     def phase(self):
         return self.time
 
+    @property
+    def odd(self):
```
barentsen: Elsewhere in Lightkurve we have often used the suffix `_mask` to mark boolean masks.  Should we name this `odd_mask` or `odd_cadence_mask`?

christinahedges: I'm happy with that suggestion, I kept it short because it is a little easier to read when we have a long "sentence". I'll make this change.

christinahedges:
This PR is now ready to be merged. It implements a useful function for verifying planets that was requested a long time ago by Jeff Coughlin

barentsen:
I pushed an extra commit to improve the docstrings:

Screenshot from 2020-02-04 11-54-01

This should be good to merge if the tests pass!

Issue #429, opened 2019-02-06
Installation hysteresis when updating to v1.0b26


gully:
#### Problem description
Celerite became a dependency between ~v1.0b25 to v1.0b26.  When pulling from master, the new dependencies do not automatically appear, so a reinstall is needed.

#### Example

```python
In [1]: import lightkurve as lk
l---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-23246d7cf2d6> in <module>
----> 1 import lightkurve as lk

~/GitHub/lightkurve/lightkurve/__init__.py in <module>
     15 from .lightcurve import *
     16 from .lightcurvefile import *
---> 17 from .correctors import *
     18 from .targetpixelfile import *
     19 from .utils import *

~/GitHub/lightkurve/lightkurve/correctors.py in <module>
     26 from sklearn.decomposition import PCA
     27 from itertools import combinations_with_replacement as multichoose
---> 28 import celerite
     29
     30 __all__ = ['SFFCorrector', 'PLDCorrector', 'KeplerCBVCorrector']

ModuleNotFoundError: No module named 'celerite'
```

Attempting a `python setup.py develop` yields another missing dependency:

```python
ModuleNotFoundError: No module named 'pybind11'
```

The successful installation requires:
`pip install -e .`

Then everything works again!

#### Environment:
- OSX Mojave
- lightkurve 1.0b27dev
- pip, conda, & source

gully:
FYI: conda install works fine.

```python
In [1]: import lightkurve as lk

In [2]: lk.__version__
Out[2]: '1.0b26'

In [3]: import celerite
```

barentsen:
Ah!  It makes sense that a mere `git pull` from master wouldn't have installed the new `celerite` dependency (needed by `PLDCorrector`)..

PR #430, opened 2019-02-06
Move the scope of the `celerite` and `scikit-learn` imports

barentsen:
Right now `PLDCorrector.correct()` is the only method in Lightkurve that requires `celerite` and `scikit-learn`.

This PR changes `correctors.py` to import these two dependencies at the method-level rather than the module-level, which will prevent `import lightkuve` from crashing if any issue with these dependencies happen to exist.

This PR is inspired by #429.  @nksaunders or @gully please review.

### DIFF
--- a/lightkurve/correctors.py
+++ b/lightkurve/correctors.py
@@ -10,6 +10,7 @@ import warnings
 
 from bs4 import BeautifulSoup
 from tqdm import tqdm
+from itertools import combinations_with_replacement as multichoose
 
 import oktopus
 import numpy as np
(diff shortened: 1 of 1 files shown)


mirca:
@barentsen @nksaunders @gully should we remove `celerite` and `scikit-learn` from the `requirements.txt` file given that they are optional dependencies?

barentsen:
Thanks for weighing in @mirca!  My idea here is for `celerite` and `scikit-learn` to be required dependencies, because the `PLDCorrector` will hopefully become a very standard tool & we'll no doubt use GPs elsewhere too.

However, because only one function uses these packages so far, I wanted to avoid `import lightkurve` breaking just in case something is wrong.  As a general rule, I think it's OK to import a package within a function if it's the only place it is ever used.

mirca:
@barentsen totally agree with you

Issue #428, opened 2019-02-06
`BLSPeriodogram` does not work out of the box


barentsen:
I feel like the example below should work out of the box, but it doesn't for reasons I don't understand yet. Simply adding a `normalize()`, `flatten()`, or `remove_nans()` to the chain did not appear to resolve it.

Anyone willing to investigate?

```python
import lightkurve as lk
lcf = lk.search_lightcurvefile("Kepler-10", quarter=6).download()
lcf.PDCSAP_FLUX.to_periodogram("bls").plot()
```

Screenshot:
![screenshot from 2019-02-06 10-04-41](https://user-images.githubusercontent.com/817669/52363018-aea12200-29f6-11e9-964f-6ae9c6fd114e.png)

barentsen:
This has been resolved by #432!  Closing.

PR #432, opened 2019-02-07
Make `BoxLeastSquaresPeriodogram` robust against NaNs

barentsen:
This PR ensures we avoid passing NaN `flux` or `flux_err` values to AstroPy's `BoxLeastSquare`.  It explains the funny behavior reported in #428.


barentsen:
Thanks @mirca!

PR #434, opened 2019-02-09
Bugfix: `tpf.wcs` is invalid for TPF's produced using TessCut

barentsen:
In #124 we discovered that the `TargetPixelFile.wcs` property is invalid for TPFs produced using the TessCut service, which presently only populates a valid WCS in the second (aperture) extension but not the first (data table) extension.

This PR makes `TargetPixelFile` resilient against this issue and adds a regression test.

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -213,27 +213,32 @@ class TargetPixelFile(object):
         w : astropy.wcs.WCS object
             WCS solution
         """
-        # Use WCS keywords of the 5th column (FLUX)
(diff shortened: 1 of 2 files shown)


barentsen:
@gully Can you review and merge as soon as the tests pass?  This bugfix will enable `interact_sky` to work with TESSCut-produced TPFs!

gully:
Going to merge.  The appveyor is failing for known reasons (py 3.7 on windows with tesscut)

gully:
I think the tests might be failing due the an astropy version problem, some version have a particular WCS method and some don't?

barentsen:
Tests pass, merging!

PR #441, opened 2019-02-11
Support rectangular TESS FFI cutouts

nksaunders:
**By popular demand!** `search_tesscut()` now support rectangular cutouts with the following syntax:

```python
import lightkurve as lk
search_tesscut('pi Men').download(cutout_size=(4,10))
```


barentsen:
Thank you @nksaunders! :+1:

One change request: can you also update the docstring of both `download()` and `download_all()` to clarify that the `cutout_size` argument now accepts both a float and a tuple of two floats? Thanks++.

barentsen:
LGTM! :+1:

PR #450, opened 2019-02-19
Quick fix to get the Gaia Source to show all digits, overcoming Bokeh default

gully:
Thank you to @keatonb and others for pointing out that `interact_sky` Gaia Source IDs were truncated in the final two digits.  It was ostensibly due to a Bokeh default that truncates the digits for long numbers.


gully:
Note-- There may be some residual problems for Python 2.

PR #453, opened 2019-02-20
Fixed scaling issue in periodogram logmedian smoothing

grd349:
The logmedian smoothing of data with noise properties of chi^2 2 dof was producing an offset wrt the boxkernel.  The problem was that the median of the chi^2_2 distribution is 2(8/9)**3 where as the mean is 2.  I added the correction factor in periodogram.smooth and this seemed to fix the scaling issue with the flattened periodogram at the same time,

I added a suggestion for two tests to check the scaling.

### DIFF
--- a/lightkurve/periodogram.py
+++ b/lightkurve/periodogram.py
@@ -159,7 +159,7 @@ class Periodogram(object):
 
         Periodograms that are unsmoothed have multiplicative noise that is
         distributed as chi squared 2 degrees of freedom.  This noise
-        distirbution has a well defined mean and median but the two are not
+        distribution has a well defined mean and median but the two are not
         equivalent.  The mean of a chi squared 2 dof distribution is 2, but the
         median is 2(8/9)**3.
         (see https://en.wikipedia.org/wiki/Chi-squared_distribution)
--- a/lightkurve/periodogram.py
+++ b/lightkurve/periodogram.py
@@ -157,6 +157,25 @@ class Periodogram(object):
         where `filter width` is in log10(frequency) space. This is best for
         estimating the noise background, as it filters over the seismic peaks.
 
+        Periodograms that are unsmoothed have multiplicative noise that is
+        distributed as chi squared 2 degrees of freedom.  This noise
+        distirbution has a well defined mean and median but the two are not
+        equivalent.  The mean of a chi squared 2 dof distribution is 2, but the
+        median is 2(8/9)**3.
+        (see https://en.wikipedia.org/wiki/Chi-squared_distribution)
+        In order to maintain consistency between 'boxkernel' and 'logmedian' a
+        correction factor of (8/9)**3 is applied to (i.e., the median is divided
+        by the factor) to the median values.
+
+        In addition to consistency with the 'boxkernel' method, the correction
+        of the median values is useful when applying the periodogram flatten
+        method.  The flatten method divides the periodgram by the smoothed
+        periodogram using the 'logmedian' method.  By appyling the correction
+        factor we follow asteroseismic convention that the signal-to-noise
+        power has a mean value of unity.  (note the signal-to-noise power is
+        really the signal plus noise divided by the noise and hence should be
+        unity in the absence of any signal)
+
         Parameters
         ----------
         method : str, one of 'boxkernel' or 'logmedian'
@@ -207,8 +226,6 @@ class Periodogram(object):
             count = np.zeros(len(self.frequency.value), dtype=int)
             bkg = np.zeros_like(self.frequency.value)
             x0 = np.log10(self.frequency[0].value)
-            # The median of chi squared 2 dof noise is 2(8/9)**3 where as the
-            # mean is 2.  To correct for this we divide the median by (8/9)**3
             corr_factor = (8.0/9.0)**3
             while x0 < np.log10(self.frequency[-1].value):
                 m = np.abs(np.log10(self.frequency.value) - x0) < filter_width
--- a/lightkurve/periodogram.py
+++ b/lightkurve/periodogram.py
@@ -207,10 +207,13 @@ class Periodogram(object):
             count = np.zeros(len(self.frequency.value), dtype=int)
             bkg = np.zeros_like(self.frequency.value)
             x0 = np.log10(self.frequency[0].value)
+            # The median of chi squared 2 dof noise is 2(8/9)**3 where as the
+            # mean is 2.  To correct for this we divide the median by (8/9)**3
+            corr_factor = (8.0/9.0)**3
             while x0 < np.log10(self.frequency[-1].value):
                 m = np.abs(np.log10(self.frequency.value) - x0) < filter_width
                 if len(bkg[m] > 0):
-                    bkg[m] += np.nanmedian(self.power[m].value)
+                    bkg[m] += np.nanmedian(self.power[m].value) / corr_factor
                     count[m] += 1
                 x0 += 0.5 * filter_width
             bkg /= count
--- a/lightkurve/periodogram.py
+++ b/lightkurve/periodogram.py
@@ -226,7 +226,7 @@ class Periodogram(object):
             count = np.zeros(len(self.frequency.value), dtype=int)
             bkg = np.zeros_like(self.frequency.value)
             x0 = np.log10(self.frequency[0].value)
-            corr_factor = (8.0/9.0)**3
+            corr_factor = (8.0 / 9.0)**3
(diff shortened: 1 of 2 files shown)


christinahedges:
Hi @grd349, thanks for the PR! I'm jazzed you are using the new periodogram features. I have a couple of quick comments for you. 

1. Can you please explain your correction factor with a more thorough comment, perhaps pointing to a reference on the median/means of the chi square distribution, and explaining why the correction factor is needed to scale the two different flattening/smoothing functions

2. I believe your tests will fail, I suggest that you change it to something like:
```python
assert np.isclose(np.abs(p.smooth().power - 1.0), atol=0.05)
```

3. Could you explain why your test should pass? I'm not sure I follow why those should be 1 within some tolerance?

Thanks again for putting this in, these fixes are really important for our consistency!

grd349:
Hey @christinahedges,

I'm jazzed too! lightkurve is really really good!

1. & 2. Done and done.

> 3. Could you explain why your test should pass? I'm not sure I follow why those should be 1 within some tolerance?

They shouldn't!  I don't know quite what I was thinking.  A little more thought gives ...

'assert np.isclose(np.mean(p.smooth(method='logmedian').power), np.mean(p.power), atol=0.05)'

i.e., that the mean of the smoothed (logmedian) power should be consistent with the mean of the power.  

And

'assert np.isclose(np.mean(p.flatten(method='logmedian').power), 1.0, atol=0.05)'

i.e., the flattened power is the power divided by the smoothed power and we are saying the smoothed power should have ~the same mean as the power.  So the mean value of white noise that is flattened should be unity.

christinahedges:
Awesome! Thanks for adding the notes, I'll read through those and review your PR shortly. Just quickly, your tests are currently failing (it's a simple fix that I'll show you in a moment). Below I've put some quick instructions on how to run the tests locally, and check that they will pass before pushing to the PR. Running tests locally is much (>10x) faster than waiting for travis and appveyor to pass, and so running tests locally means your PR will be merged faster.

# Running `lightkurve` tests locally

`lightkurve` has many [unit tests](https://en.wikipedia.org/wiki/Unit_testing) which check that our basic functionality works as we expect. Before we can make any changes to `lightkurve` we have to check that these tests all pass. If you open a PR to `lightkurve`, these tests will be run automatically. If they pass, your code can be reviewed and potentially merged into the `lightkurve` master branch. However, running these tests online can take a long time. Running the tests locally on your machine is much faster, and will let you check that your work still maintains the expected `lightkurve` behavior as you develop your code.

## How do I run tests?

First off, you need to find the directory that your `lightkurve` installation is in. You can check this by looking at the `lightkurve` path.

```python 
import lightkurve as lk
lk.__path__
```

In a terminal, `cd` into the `lightkurve` directory. Yours should look something like this:

image
Note that here we're using the master branch. If you want to run the tests on a branch you are developing, switch to that branch using `git checkout branchname`. Once you're in the `lightkurve` directory, go to the tests directory in `lightkurve/tests`.

image

In this directory you should find several tests labeled `test_xxx.py`. You can run a test using `pytest`. For example, to test the `lightkurve.targetpixelfile` module you would execute

```
pytest test_targetpixelfile.py 
```

If the tests are successful, you should see a green success message such as
image

### Why are some of the tests marked "s"/ Why are some tests skipped?

Running some of our tests requires external data, e.g. some require data to be downloaded from MAST. These tests take a little longer, and so we skip them by default. In order to run all the tests simply use

```
pytest test_targetpixelfile.py --remote-data
```

### My tests passed, but I got warning messages
Sometimes you will get warnings in your tests, causing your result to be yellow instead of green. For example, you may get an output that looks like this:

image

While this is not ideal, some `lightkurve` tests do raise warnings currently. This will become less and less likely as `lightkurve` improves.

### My tests failed

If your test fails, don't worry, this is what tests are for. Take a look at the traceback that pytest provides for you. If your test has failed then you will see an F next to the test you've run, for example:

image

Underneath, you will then see the traceback of the test that failed. For example, the traceback below shows that there is an `AssertionError`. 

image

In the test, we have made an assertion
`assert_array_equal(lc_add.flux, lc.flux + 2)`.

However in the traceback we can see that these two arrays are not actually equal, and so the test is breaking.

image

Use this information to correct the code you're developing until the tests pass. In rare cases (such as the case above) it is the test itself that is incorrect, not the lightkurve code. If you believe there is an error in one of the tests, point it out in your PR for everyone to comment and discuss. 

## When should I run tests?

Before you open a PR to lightkurve, ideally you should run these tests locally and check that they are all passing. If they aren't passing, and you are confused as to why they are not, you can open a PR and ask for help.

## Can I write my own test?

Ideally, any PR opened to `lightkurve` with new functionality should include some tests. These tests check that the basic functionality of your PR works. That way, if in future people create new features that break your PR, we will be alerted. Read through the `pytest` documentation and take a look at our existing tests to get an idea of how to write your own.

## I can't run any tests. 

We run our unit tests using `pytest`. This should have been installed when you installed `lightkurve`. However, if your tests don't run, you may want to check all the test dependencies are installed by running

```
pip install pytest pytest-cov pytest-remotedata
```

lightkurve/periodogram.py
```diff
@@ -157,6 +157,25 @@ def smooth(self, method='boxkernel', filter_width=0.1):
         where `filter width` is in log10(frequency) space. This is best for
         estimating the noise background, as it filters over the seismic peaks.
 
+        Periodograms that are unsmoothed have multiplicative noise that is
+        distributed as chi squared 2 degrees of freedom.  This noise
+        distirbution has a well defined mean and median but the two are not
```
christinahedges:
```suggestion
        distribution has a well defined mean and median but the two are not
```

lightkurve/periodogram.py
```diff
@@ -157,6 +157,25 @@ def smooth(self, method='boxkernel', filter_width=0.1):
...
+        equivalent.  The mean of a chi squared 2 dof distribution is 2, but the
+        median is 2(8/9)**3.
+        (see https://en.wikipedia.org/wiki/Chi-squared_distribution)
+        In order to maintain consistency between 'boxkernel' and 'logmedian' a
```
christinahedges: Am I understanding correctly that `boxkernel` is then a 'mean' smoothing, not a 'median' filter?

lightkurve/periodogram.py
```diff
@@ -207,10 +226,11 @@ def smooth(self, method='boxkernel', filter_width=0.1):
             count = np.zeros(len(self.frequency.value), dtype=int)
             bkg = np.zeros_like(self.frequency.value)
             x0 = np.log10(self.frequency[0].value)
+            corr_factor = (8.0/9.0)**3
```
christinahedges:
```suggestion
            corr_factor = (8.0 / 9.0)**3
```

lightkurve/tests/test_periodogram.py
```diff
@@ -134,6 +134,9 @@ def test_smooth():
     with pytest.raises(ValueError) as err:
         p.smooth(method='logmedian',  filter_width=5.*u.day)
 
+    # Check logmedian smooth that the mean of the smoothed power should be consistent with the mean of the power
+    assert np.isclose(np.mean(p.smooth(method='logmedian').power.value), np.mean(p.power), atol=0.05)
```
christinahedges:
```suggestion
    assert np.isclose(np.mean(p.smooth(method='logmedian').power.value), np.mean(p.power.value), atol=0.05)
```

lightkurve/periodogram.py
grd349: Yes - the boxkernel is like a running mean (in fact I think it is a running mean).

grd349:
Thanks!  I had tried to run locally but I have a dependency issue with pytest and pytest-cov that I can't yet fix :) or rather :(

I'm very happy with your comments!

christinahedges:
If you still can't run them locally, try installing the following with pip:
```pip install pytest pytest-cov pytest-remotedata```

grd349:
Got it - Ta.  I had to uninstall pytest first and then run the above command.  Now it's working locally!

But now I get a fail.  

'np.mean(p.smooth(method='logmedian').power.value' 

evaluates as nan but I don't get this behaviour from the command line.

'>>> import numpy as np
>>> from lightkurve import LightCurve
>>> lc = LightCurve(time=np.arange(1000),
...                     flux=np.random.normal(1, 0.1, 1000),
...                     flux_err=np.zeros(1000)+0.1)
>>> p = lc.to_periodogram()
>>> np.mean(p.smooth(method='logmedian').power.value)
15942682403.255383
'

PR #458, opened 2019-02-22
TPF factory now calculates and outputs pos_corr values to generated file.

amcody:
I found that somehow a few of Emma's important changes from the summer did not make it through to the current version of targetpixelfile.py. In particular, the from_fits_images routine was not calculating pos_corr from the object position and WCS. It now does that, and stores pos_corr1, 2 in the header.  IT also passes the wcs of a reference image to factory.add_cadence, so that that loop does not think there is no wcs (as is the default). Having pos_corr values in the TPF header is important for use in, e.g., detrending of light curves.


lightkurve/targetpixelfile.py
```diff
@@ -1126,12 +1126,24 @@ def _open_image(img, extension):
...
+            hdu.header['POS_CORR1'] = column_current - column_ref
+            hdu.header['POS_CORR2'] = row_current - row_ref
+
             if position is None:
                 cutout = hdu
             else:
                 cutout = Cutout2D(hdu.data, position, wcs=WCS(hdu.header),
                                   size=size, mode='partial')
-            factory.add_cadence(frameno=idx, flux=cutout.data, header=hdu.header)
+            factory.add_cadence(frameno=idx, wcs=wcs_ref, flux=cutout.data, header=hdu.header)
```
barentsen: I don't understand why the `add_cadence` method has a `wcs` argument.  All it seems to be used for inside `add_cadence` is this snippet:
```Python
        if wcs is None:
            self.pos_corr1[frameno], self.pos_corr2[frameno] = None, None
```

I suggest we clean this up and remove the `wcs` argument from the `add_cadence` method.  Inside that method, we can just do e.g. this:
```Python
        if 'POS_CORR1' in header:
            self.pos_corr1[frameno] = header['POS_CORR1']
        else:
            self.pos_corr1[frameno] = None
```

lightkurve/targetpixelfile.py
```diff
@@ -1126,12 +1126,24 @@ def _open_image(img, extension):
 
             if idx == 0:  # Get default keyword values from the first image
                 factory.keywords = hdu.header
+
+
+            # Get positional shift of the image compared to the reference WCS
+            wcs_current = WCS(hdu.header)
+            column_current, row_current = wcs_current.wcs_world2pix(
+                          np.asarray([[position.ra.deg], [position.dec.deg]]).T,0)[0]
```
barentsen: I'm worried this will break if `position` is equal to `None`.  A more robust solution may be to retrieve the ra/dec of the center pixel, and use that to compute the POS_CORRs?

lightkurve/targetpixelfile.py
amcody: @barentsen yes I was struggling over this same thing myself- why is the wcs fed in the places where it is, and why doesn't it go through to the TPF in the end? Maybe when you are free we can take a look together when you get back...

amcody:
@barentsen I made the requested changes and put in a new test. However, the new test is failing because CRVAL1P and CRVAL2P are not in the test header (a new requirement- needed to output 1CRV5P, 2CRV5P in output TPF). Can we talk on Monday about how to fix this?

amcody:
@barentsen never mind the above comment- I just made a further change that should (I hope!) fix the problem. Let's see if the tests pass...

barentsen:
@amcody Sounds good, let's discuss Monday!  Looks like we are really close to getting this fix merged! :tada:

amcody:
Doh, it failed. Will have to investigate why on Mon...

amcody:
Turns out I had correctly put in CRVAL1P and CRVAL2P, but I forgot to do it twice-- once for each tpf_from_images test.  So great, no failures there anymore, BUT, the tests are still failing, on what appears to be test_search_lightcurvefile! Which is really weird, because I didn't touch that... help @barentsen ?? Is there a funny MAST timeout thing getting in the way here?

barentsen:
Thanks @amcody!  I like the new unit test!  I do suggest we add just a few more assert statements below line 411 in `test_targetpixelfile.py` to check that the TPF created in the new test really has a useful WCS.

For example, it should be easy to verify the following using a few extra asserts:
```
1CTYP5 = 'RA---TAN'
2CTYP5  = 'DEC--TAN'
1CRPX5, 2CRPX5  = some small, sensible numbers
1CRVL5, 2CRVL5  = something close to the crval you chose above 
1CUNI5  = 'deg     '
2CUNI5  = 'deg     ' 
```

amcody:
@barentsen Okay, it turns out this test was failing because it is not correct to specify the ctype as TAN but to leave in things like set_pv. We need to decide if we should leave our fake WCS in 'AIR' (airy projection) format or switch to TAN and it's relevant keywords.

amcody:
Update: I switched to TAN since that seems to be a more typical format.

amcody:
@barentsen could you press the "restart build" button, pretty please? :)

amcody:
@barentsen  Something with our MAST connection is really flaky today- now python 3.6 is passing (previously failed) but 3.7 is failing. Can we rebuild individual versions??

amcody:
@barentsen One thing we need to check is whether the tpf.wcs property exists (or needs to exist) when we do the test to create one. My preliminary check indicated that this produces an error ("Unmatched celestial axes"), but I don't know if it's a problem. I believe it does work on a real wcs- there is just some issue with the way we defined the fake one.

barentsen:
When running the new `test_tpf_wcs_from_images` test, I see the following warnings:
```
Keyword name 'POS_CORR1' is greater than 8 characters or contains characters
not allowed by the FITS standard; a HIERARCH card will be created.
```

This is because we temporarily add `POS_CORR1` as a header keyword to pass it on to the factory.  I suggest we remove the underscore when we do this to avoid this warning.  (If we get too many warnings when running tests we may not notice important ones.)  Of course the underscore does have to be there in the name of the column table!

barentsen:
The tests pass!  It wouldn't be a bad idea to add something to verify that `tpf.wcs` works, e.g. you could add this line to your new test, which works for me:
```python
assert tpf.wcs.to_header()['CDELT1'] == w.wcs.cdelt[0]
```

If you add this extra assert, and fix the warning (see message above), I'm happy for this to be merged!

amcody:
@barentsen it is repeatedly failing on various tess_search functions... I've rerun it 3 times and not sure how to fix other than to give it a day to see if MAST is having issues?

barentsen:
MAST works again.  The test failure is now due to an unrelated issue in `test_interact_sky_functions` (see below, cc/ @gully).  I verified that the tests do pass both on the Travis build that does not have the `--remote-data` flag, and on my local machine, so I am going to go ahead and merge this PR while we continue to look into the unrelated test failures.

```python
_________________________ test_interact_sky_functions __________________________
    @pytest.mark.skipif('bokeh' not in sys.modules, reason="requires bokeh")
    def test_interact_sky_functions():
        """Do the helper functions in the interact module run without syntax error?"""
        import bokeh
        from ..interact import (prepare_tpf_datasource, make_tpf_figure_elements,
                                add_gaia_figure_elements)
        tpf = TessTargetPixelFile(example_tpf)
        mask = tpf.flux[0, :, :] == tpf.flux[0, :, :]
        tpf_source = prepare_tpf_datasource(tpf, aperture_mask=mask)
        fig1, slider1 = make_tpf_figure_elements(tpf, tpf_source)
>       add_gaia_figure_elements(tpf, fig1)
lightkurve/tests/test_interact.py:130: 
tpf = TessTargetPixelFile(TICID: 25155310), fig = Figure(id='1127', ...)
magnitude_limit = 18
    def add_gaia_figure_elements(tpf, fig, magnitude_limit=18):
        """Make the Gaia Figure Elements"""
        # Get the positions of the Gaia sources
        c1 = SkyCoord(tpf.ra, tpf.dec, frame='icrs', unit='deg')
        # Use pixel scale for query size
        pix_scale = 4.0  # arcseconds / pixel for Kepler, default
        if tpf.mission == 'TESS':
            pix_scale = 21.0
        # We are querying with a diameter as the radius, overfilling by 2x.
        from astroquery.vizier import Vizier
        Vizier.ROW_LIMIT = -1
        result = Vizier.query_region(c1, catalog=["I/345/gaia2"],
                                     radius=Angle(np.max(tpf.shape[1:]) * pix_scale, "arcsec"))
        no_targets_found_message = ValueError('Either no sources were found in the query region '
                                              'or Vizier is unavailable')
        too_few_found_message = ValueError('No sources found brighter than {:0.1f}'.format(magnitude_limit))
        if result is None:
            raise no_targets_found_message
        elif len(result) == 0:
>           raise too_few_found_message
E           ValueError: No sources found brighter than 18.0
lightkurve/interact.py:227: ValueError
```

barentsen:
Well, Travis is passing now!  Merging in 3 ... 2 ... 1 ...

PR #463, opened 2019-02-27
fixing axis label in interact plot

mrtommyb:
screenshot 2019-02-27 11 45 45

### DIFF
--- a/lightkurve/interact.py
+++ b/lightkurve/interact.py
@@ -165,7 +165,12 @@ def make_lightcurve_figure_elements(lc, lc_source):
                  border_fill_color="whitesmoke")
     fig.title.offset = -10
     fig.yaxis.axis_label = 'Flux (e/s)'
-    fig.xaxis.axis_label = 'Time - 2454833 (days)'
+    if (lc.mission == 'K2') or (lc.mission == 'Kepler'):
+        fig.xaxis.axis_label = 'Time - 2454833 (days)'
+    elif lc.mission == 'TESS':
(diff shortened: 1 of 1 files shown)


barentsen:
Thanks @mrtommyb, this looks good!

lightkurve/interact.py
```diff
@@ -165,7 +165,12 @@ def make_lightcurve_figure_elements(lc, lc_source):
                  border_fill_color="whitesmoke")
     fig.title.offset = -10
     fig.yaxis.axis_label = 'Flux (e/s)'
-    fig.xaxis.axis_label = 'Time - 2454833 (days)'
+    if (lc.mission == 'K2') or (lc.mission == 'Kepler'):
+        fig.xaxis.axis_label = 'Time - 2454833 (days)'
+    elif lc.mission == 'TESS':
+        fig.xaxis.axis_label = 'Time - 2457000 (days)'
+    else:
+        fig.xaxis.axis_label = 'Time (days)'
```
barentsen: I'm a bit worried about the following: `mission` is not a property of the base `LightCurve` class, so I'm worried that this is going to crash on non-standard TPFs for which `to_lightcurve` returns a generic `LightCurve` object without mission property.  I think it'll be safer to do this instead:
```Python
fig.xaxis.axis_label = 'Time (days)'
try:
   if lc.mission ...:
        fig.xaxis.axis_label = ...
except AttributeError:  # no mission keyword available
  pass
```

mrtommyb:
ok

barentsen:
Thanks @mrtommyb!

PR #464, opened 2019-03-01
[WIP] Refactoring correctors and adding a GPCorrector for long term trends

christinahedges:
Work in progress, @nksaunders and I discussed putting something like this in lightkurve, this is an update of an old WIP I had going with the new API.


barentsen:
This is looking awesome!!  For the `GPCorrector`, we may want to considering seeking inspiration from this cell of the [tess.world](tess.world) project:

```Python
# Extract the data and convert to parts per thousand
x = np.ascontiguousarray(lc.time, dtype=np.float64)
y = np.ascontiguousarray((lc.flux - 1.0) * 1e3, dtype=np.float64)
yerr = np.ascontiguousarray(lc.flux_err * 1e3, dtype=np.float64)

# Compute the transit mask
mask = np.zeros_like(x, dtype=bool)
for i in range(len(periods)):
    mask |= get_transit_mask(x, t0s[i], periods[i], 5*durations[i])

# Temporarily increase the in transit error bars substantially
diag = np.array(yerr**2)
diag[mask] += 10000.0

# Build a GP model
with pm.Model() as model:
    logs2 = pm.Normal("logs2", mu=np.log(1e-4*np.var(y)), sd=10)
    logsigma = pm.Normal("logsigma", mu=np.log(np.std(y)), sd=10)
    logrho = pm.Normal("logrho", mu=np.log(10.0), sd=10.0)
    
    kernel = xo.gp.terms.Matern32Term(log_sigma=logsigma, log_rho=logrho)
    gp = xo.gp.GP(kernel, x, diag + tt.exp(logs2), J=2)
    pm.Potential("loglike", gp.log_likelihood(y))
    
    map_soln = xo.optimize()
    pred = xo.utils.eval_in_model(gp.predict(), map_soln)

# Flatten the light curve
y -= pred
```

I believe the main difference in the above snippet is the use of PyMC, rather than scipy.minimize, for the fitting.

barentsen:
@nksaunders and I did some pair-experimentation with GP and Exoplanet fitting today, benefiting from the awesome notebooks now available at [tess.world](tess.world).  This is a (very coarse) write-up of three key lessons we learned, which a future `GPCorrector` class in Lightkurve should take into account:

1. GP's can be strongly affected by outliers or the presence of transits.  However, rather than masking those out, Dan FM recommends we temporarily inflate the `flux_err` values for those cadences to encourage the model to ignore them without being penalized.  (This is being done in the tess.world snippet posted above.)  This is both faster and easier than having to predict the GP across gaps (though we have to ask Christina or DFM again why).

2. Based on looking at examples, we found that it is extremely important to use the `flux_err` inflation trick (or masking) on outlier points.  Even a single extreme outlier can cause the GP to be wildly incorrect across many days.  We believe that the tess.world snippet posted above is not currently robust against a single extreme outlier.

3. It is important to set the `rho` parameter of the `Matern32Kernel` to a value that is significantly larger than the typical duration of an exoplanet transit.  This seems obvious, but it often goes wrong because we tend to parametrize `rho` in terms of cadences instead of time units, and so applying the same code on 30-min Kepler vs 2-min TESS vs 1-min Kepler data can fail for this reason.  We should think about this in the design of GPCorrector.

barentsen:
A number of improvements to the Corrector classes have been merged into master since this PR was opened.  As a result, it wouldn't be smart to try to rebase and merge this PR as-is.  Instead, I have opened a new PR to execute the refactor proposed here (#519).

We can keep using this PR to get the proposed `GPCorrector` merged in.

barentsen:
I'm going to close this PR because we'll want to revise `GPCorrector` to use the new `RegressionCorrector` framework (cf. #613).  The code attached to this PR will continue to remain immensely useful for that purpose! :+1::+1:

Issue #452, opened 2019-02-20
light curve created using a TESS target pixel file with .interact fails to open


mrtommyb:
#### Example

```python
import lightkurve as lk
tpfs = lk.search_targetpixelfile('TIC 206544316').download_all()
tpf = tpfs[1]
tpf.interact('http://localhost:8889')
% Saved file tess2018234235059-s0002-0000000206544316-0121-s_tp-custom-lc.fits 
lc = lk.TessLightCurveFile('tess2018234235059-s0002-0000000206544316-0121-s_tp-custom-lc.fits', )
```
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/anaconda3/envs/exoplanet/lib/python3.7/site-packages/astropy/io/fits/column.py in _get_index(names, key)
   2007         try:
-> 2008             indx = names.index(key.rstrip())
   2009         except ValueError:

ValueError: 'QUALITY' is not in list

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
<ipython-input-14-80ea932783c4> in <module>
----> 1 lc = lk.TessLightCurveFile('tess2018234235059-s0002-0000000206544316-0121-s_tp-custom-lc.fits', )

~/anaconda3/envs/exoplanet/lib/python3.7/site-packages/lightkurve-1.0b30.dev0-py3.7.egg/lightkurve/lightcurvefile.py in __init__(self, path, quality_bitmask, **kwargs)
    382         self.quality_bitmask = quality_bitmask
    383         self.quality_mask = TessQualityFlags.create_quality_mask(
--> 384                                 quality_array=self.hdu[1].data['QUALITY'],
    385                                 bitmask=quality_bitmask)
    386         # Early TESS releases had cadences with time=NaN (i.e. missing data)

~/anaconda3/envs/exoplanet/lib/python3.7/site-packages/astropy/io/fits/fitsrec.py in __getitem__(self, key)
    485 
    486         if isinstance(key, str):
--> 487             return self.field(key)
    488 
    489         # Have to view as a recarray then back as a FITS_rec, otherwise the

~/anaconda3/envs/exoplanet/lib/python3.7/site-packages/astropy/io/fits/fitsrec.py in field(self, key)
    675         # NOTE: The *column* index may not be the same as the field index in
    676         # the recarray, if the column is a phantom column
--> 677         column = self.columns[key]
    678         name = column.name
    679         format = column.format

[...]

PR #466, opened 2019-03-18
Support reading-in interact-generated custom lightcurves

gully:
See discussion in Issue #452.  This PR attempts to support reading-in custom lightcurves.


gully:
OK, this now works for TESS files only:
```python

import lightkurve as lk
tpfs = lk.search_targetpixelfile('TIC 206544316').download_all()
tpf = tpfs[1]

lcf = lk.TessLightCurveFile('tess2018234235059-s0002-0000000206544316-0121-s_tp-custom-lc.fits', )
lcf.plot()
lc = lcf.get_lightcurve('FLUX')
lc.plot()
```
Note that `SAP_FLUX` and `PDCSAP_FLUX` appear as attributes, but are not valid flux types:
```python
lcf.PDCSAP_FLUX
```
yields:
```python
KeyError: 'PDCSAP_FLUX is not a valid flux type. Available types are: <bound method LightCurveFile._flux_types of TessLightCurveFile(TICID: None)>'
```

gully:
Todo:
- [ ] Add support for `KeplerLightCurveFile`s
- [ ] Add unit tests
- [ ] Spot check conceivable ways other parts of the code expect metadata that will be missing

gully:
Seeking early feedback to gauge the merits and conceivable unintended consequences of this approach @barentsen @christinahedges

barentsen:
@gully One issue here is that `TessLightCurve.to_fits()` creates a file with a `SAP_QUALITY` column (which is the behavior of the base LightCurve class), but in the TESS data format this column appears to be called `QUALITY` instead of `SAP_QUALITY`.  I think we'll want to correct `TessLightCurve.to_fits()` on this front.

lightkurve/lightcurvefile.py
```diff
@@ -380,8 +380,16 @@ def __init__(self, path, quality_bitmask='default', **kwargs):
...
 
         self.quality_bitmask = quality_bitmask
+        # Custom lightcurves may not have quality flags; assign zeros
+
+        if 'QUALITY' in self.hdu[1].data.columns.names:
+            quality_vector = self.hdu[1].data['QUALITY']
+        elif 'SAP_QUALITY' in self.hdu[1].data.columns.names:
+            quality_vector = self.hdu[1].data['SAP_QUALITY']
+        else:
+            quality_vector = self.hdu[1].data['TIME']*0.0
```
barentsen: I recommend using `np.zeros` here instead of a multiplication (because it'll be easier to read the code, and because this will break if TIME equals `nan` or `inf`).

lightkurve/lightcurvefile.py
```diff
@@ -397,6 +405,20 @@ def __repr__(self):
...
+        else:
+            centroid_col = self.hdu[1].data["TIME"][self.quality_mask]*0.0
+            centroid_row = self.hdu[1].data["TIME"][self.quality_mask]*0.0
+
+        if 'QUALITY' in self.hdu[1].data.columns.names:
+            quality_vector = self.hdu[1].data['QUALITY']
+        elif 'SAP_QUALITY' in self.hdu[1].data.columns.names:
+            quality_vector = self.hdu[1].data['SAP_QUALITY']
+        else:
+            quality_vector = self.hdu[1].data['TIME']*0.0
```
barentsen: I recommend assigning `self._quality_vector` in the constructor and re-using it here, so we avoid code duplication.

lightkurve/lightcurvefile.py
barentsen: Also note: we may want to move this code block to a `_get_quality()` method in the base class, so that `KeplerLightCurveFile` can call it and behave in exactly the same way.

lightkurve/lightcurvefile.py
```diff
@@ -397,6 +405,20 @@ def __repr__(self):
         return('TessLightCurveFile(TICID: {})'.format(self.targetid))
 
     def get_lightcurve(self, flux_type, centroid_type='MOM_CENTR'):
+        if centroid_type+"1" in self.hdu[1].data.columns.names:
+            centroid_col = self.hdu[1].data[centroid_type + "1"][self.quality_mask]
+            centroid_row = self.hdu[1].data[centroid_type + "2"][self.quality_mask]
+        else:
+            centroid_col = self.hdu[1].data["TIME"][self.quality_mask]*0.0
+            centroid_row = self.hdu[1].data["TIME"][self.quality_mask]*0.0
```
barentsen: We'll want to use `np.zeros()` here too (or `np.zeros() * np.nan`?  we probably want nans here?)

barentsen:
Ping @gully, does this :point_up: help you move forward?

gully:
I made the requested changes and I've updated with ability to read in `.interact()`-produced lightcurves for both Kepler and TESS.  Still some subtasks to do above.

gully:
The tests seem to be failing due to a `search` / MAST / http issue.

barentsen:
@gully I fixed this test failure in #471.  Would you like to `git rebase master` and see if the tests pass now?

lightkurve/lightcurvefile.py
```diff
@@ -159,6 +171,9 @@ def __init__(self, path, quality_bitmask='default', **kwargs):
         elif filetype is None:
             warnings.warn("Given fits file not recognized as Kepler or TESS "
                           "observation.", LightkurveWarning)
+        elif filetype == 'KeplerLightCurveFile-interact':
+            warnings.warn("Custom-made Kepler fits files have limited "\
+                          "functionality.", LightkurveWarning)
```
barentsen: I'm not sure this warning is particularly useful (because it is not actionable).  Is there a specific feature that will not work?  If so, I think we should only have a warning when a non-functioning feature is called?

lightkurve/lightcurvefile.py
```diff
@@ -373,6 +394,9 @@ def __init__(self, path, quality_bitmask='default', **kwargs):
         elif filetype is None:
             warnings.warn("Given fits file not recognized as Kepler or TESS "
                           "observation.", LightkurveWarning)
+        elif filetype == 'TessLightCurveFile-interact':
+            warnings.warn("Custom-made TESS fits files have limited "\
+                          "functionality.", LightkurveWarning)
```
barentsen: I think we'll want to get rid of this warning as well to avoid warning-fatigue (=> we don't want people to think that ignoring warnings is OK, so we should use them sparingly).

lightkurve/tests/test_lightcurve.py
```diff
@@ -277,6 +279,42 @@ def test_lightcurve_copy():
...
+    assert_array_equal(lc.time, hdu[1].data['TIME'])
+    assert_array_equal(lc.flux, hdu[1].data['FLUX'])
+
+    # TESS has QUALITY while Kepler/K2 has SAP_QUALITY:
+    if mission == "TESS":
+        assert "QUALITY" in lcf_custom.hdu[1].columns.names
+        assert_array_equal(lc.quality, hdu[1].data['QUALITY'])
+    if mission in ["K2", "Kepler"]:
+        assert "SAP_QUALITY" in lcf_custom.hdu[1].columns.names
+        assert_array_equal(lc.quality, hdu[1].data['SAP_QUALITY'])
```
barentsen: I love these tests! :+1:

lightkurve/utils.py
```diff
@@ -484,6 +486,16 @@ def detect_filetype(header):
...
         # use `telescop` keyword to determine mission
         # and `creator` to determine tpf or lc
+
+        if "SOURCE" in header.keys():
+            # Could be a custom lightcurve
+            if header['SOURCE'] == 'lightkurve interact':
+                if header['mission'].lower() == 'tess':
+                    return 'TessLightCurveFile-interact'
+                elif header['mission'].lower()[0] == 'k':
+                    return 'KeplerLightCurveFile-interact'
```
barentsen: I'm hesitant to define interact-generated files as a different file type just because the flux column has a different name (or are there other key differences?).  I think we should think of them as standard files -- perhaps the only difference being that if you call `lcf.SAP_FLUX` you get a nice error message that says: `This file does not have a SAP_FLUX column. The available flux columns are: ...`

lightkurve/lightcurve.py
```diff
@@ -1609,6 +1609,7 @@ def _make_aperture_extension(hdu_list, aperture_mask):
         hdu = super(TessLightCurve, self).to_fits(path=None,
                                                     overwrite=overwrite,
                                                     **extra_data)
+        hdu[1].columns.change_name('SAP_QUALITY', 'QUALITY')
```
barentsen: Maybe add a comment here to make it explicit that we do this because the TESS file format is subtly different in the name of this column.

gully:
OK, all changes made.  :)  Probably ready for user testing and/or merging.

lightkurve/utils.py
```diff
@@ -484,6 +484,17 @@ def detect_filetype(header):
...
         # and `creator` to determine tpf or lc
+
+        # Non-pipeline file types from e.g. lightkurve .interact()
+        if "SOURCE" in header.keys():
+            # Could be a custom lightcurve
+            if header['SOURCE'] == 'lightkurve interact':
+                if header['mission'].lower() == 'tess':
+                    return 'TessLightCurveFile'
+                elif header['mission'].lower()[0] == 'k':
+                    return 'KeplerLightCurveFile'
```
barentsen: Do we need lines 488-495 here?  I think the existing lines that follow (497-517) will already identify an interact-generated file correctly (based on the TELESCOP and CREATOR keywords).  If that's true, then let's get rid of lines 488-495, because less code is better!

gully: The interact-generated fits files do not have the `TELESCOP` keyword, unfortunately.  We could add this header in the `to_fits()` method, but any existing `interact()`-generated fits files would not be readable.

barentsen: I don't think that's true, e.g. `tests/data/test_K2_interact_generated_custom-lc.fits` shows `TELESCOP= 'KEPLER  '` in the header.  Do you want to have a go to see if it all works without this extra code?

lightkurve/lightcurvefile.py
```diff
@@ -265,9 +283,15 @@ def get_lightcurve(self, flux_type, centroid_type='MOM_CENTR'):
...
+        elif flux_type in ['SAP_FLUX', 'PDCSAP_FLUX']:
+            # Must be a custom FITS file:
+            warnings.warn("This custom file does not have a {} column. "
+                        "The available flux columns are {}".
+                        format(flux_type, self._flux_types()), LightkurveWarning)
+            return None # Could possibly return NaN array.
         else:
             raise KeyError("{} is not a valid flux type. Available types are: {}".
-                           format(flux_type, self._flux_types))
+                           format(flux_type, self._flux_types()))
```
barentsen: I think we can remove lines 286-291, because lines 292-294 already take care of producing an informative error message if the requested `flux_type` is not available.  (And I think it's better to raise a clear error here rather than returning `None`.)

gully: Currently `SAP_FLUX` appears as an attribute (`lcf.SAP_FLUX` exists), even when it is not available and would raise an error.  Is that OK?

barentsen: That doesn't worry me!  It would of course be better if the interact-files populated `SAP_FLUX` instead of `FLUX`, because the former is the conventional way of storing uncorrected aperture photometry flux. However that doesn't have to be resolved as part of this PR, and probably the real answer here is to start supporting/advertising `lcf.get_lightcurve()` as the preferred method of obtaining the default-choice light curve from a file.

lightkurve/utils.py
gully: TESS interact-generated fits files do not populate the `TELESCOP` keyword:

`TELESCOP=  / telescope`

I will populate this metadata in `to_fits()` to eliminate the extra code.

gully: Thought: rather than lots of if/then logic, we could define a nested dictionary here and use keys to get the filetype string: 

`lookup_table = {'tess':{'to_fits':'TESSLightCurveFile', 'targetpixel':'TESSLightCurveFile'}, 'kepler':[...]}`

lightkurve/utils.py
```diff
@@ -484,22 +484,27 @@ def detect_filetype(header):
...
         creator = header['creator'].lower()
         origin = header['origin'].lower()
         if telescop == 'kepler':
             # Kepler TPFs will contain "TargetPixelExporterPipelineModule"
             if 'targetpixel' in creator:
                 return 'KeplerTargetPixelFile'
             # Kepler LCFs will contain "FluxExporter2PipelineModule"
-            elif 'fluxexporter' in creator or 'lightcurve' in creator:
+            elif ('fluxexporter' in creator or 'lightcurve' in creator
+                or 'to_fits()' in creator):
```
barentsen: should we make this `Curve.to_fits()`?

lightkurve/utils.py
```diff
@@ -484,22 +484,27 @@ def detect_filetype(header):
...
                 return 'KeplerLightCurveFile'
         elif telescop == 'tess':
             # TESS TPFs will contain "TargetPixelExporterPipelineModule"
             if 'targetpixel' in creator:
                 return 'TessTargetPixelFile'
             # TESS LCFs will contain "LightCurveExporterPipelineModule"
-            elif 'lightcurve' in creator:
+            elif 'lightcurve' in creator or 'to_fits()' in creator:
```
barentsen: Can we simplify to:
```elif 'lightcurve' in creator.lower()```

barentsen: I'm worried that some day we'll have a TPF.to_fits(), we can prolly just remove to_fits here and it;ll still work.

Issue #465, opened 2019-03-15
Feature request: interactive slicing of short-cadence TPF with interact()


ColinLittlefield:
This is a follow-up to Issue #201, which addressed the difficulty of using interact() on TPFs with more than 30,000 cadences. Although the workaround of slicing the TPF is helpful, I would be interested in the more in-depth solution discussed by @gully in #201: namely, having a range slider that allows the user to update in real-time the window used to slice the TPF.

I've found interact() to be an excellent tool for determining the optimal extraction aperture and for identifying data artifacts. Thanks for all of your work on it!

#### Environment:
- Windows
- lightkurve version 1.0b29
- installation method: pip

barentsen:
Thank you for the feedback @ColinLittlefield!  We will discuss this a bit more and report back!

PR #468, opened 2019-03-22
Short Cadence range slider for .interact()

gully:
Here is a working prototype of a short cadence range slider for interact aimed at Issue #465 .

This Pull Request realizes the dream of @nattieg, @ColinLittlefield, and @rrpastro, and may be of interest to @keatonb and others.

### DIFF
--- a/lightkurve/interact.py
+++ b/lightkurve/interact.py
@@ -170,7 +170,7 @@ def make_lightcurve_figure_elements(lc, lc_source):
             fig.xaxis.axis_label = 'Time - 2457000 (days)'
     except AttributeError:  # no mission keyword available
       pass
-        
+
 
     ylims = get_lightcurve_y_limits(lc_source)
     fig.y_range = Range1d(start=ylims[0], end=ylims[1])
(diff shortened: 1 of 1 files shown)


gully:
20190321_lightkurve_sc_interact

gully:
Neat feature of this mode:
If you pass in a long cadence lightcurve, it works the same as before-- no Display Window slider.
If you pass in a short cadence (or have more than `max_cadences=30000` kwarg), you automagically get a range slider.  You can exceed max_cadences by increasing the window size manually, but you will pay the price in latency.

gully:
Update-- @barentsen recommends a data-decimation scheme to handle the performance issues of short cadence in `.interact()`.  The number of rendered data points would be reduced by showing 1-of-every-N-points, similar to the `.interact_bls()` approach from @christinahedges.  Zooming in would increase the fraction of points displayed.  There could also be synergies with datashader/holoviews, but these are likely beyond the scope of this PR, and may add more complexity.

ColinLittlefield:
Thank you all for your work on this! The data-decimation scheme strikes me as an excellent solution. Do you have an estimate as to when this might become available? I have a student who's about to begin work on a short-cadence K2 light curve; she's become adept at using .interact(), so if possible, I'd like for her to be able to use the new feature.

gully:
Hi @ColinLittlefield,
Thank you for staying engaged with the Issue, and for sharing lightkurve with your student!

Technically the window slider is available now if you follow the Step 1 guidelines in GitHub `command line instructions` link at the bottom of this pull request, and then follow the `development version installation` [instructions](http://docs.lightkurve.org/about/install.html#installing-the-development-version) in our documentation.  

In practice, this multi-step workaround may be counter productive if you're seeking a long-term stable solution.  I suspect it would be another 1-2 months for other developers to weigh-in on this feature, and they may decide not to merge the feature in the end depending on how much complexity it adds to the long-term maintenance of the code base.

Feel free to email our [helpdesk](https://keplerscience.arc.nasa.gov/helpdesk.html) with details of your use-case to help raise the visibility of the issue with a wider group of decision makers.

gully:
Update: @barentsen and I discussed an option to seamlessly update the data with the bokeh-native zoom and scroll tools (the wheel zoom and box zoom, for example).

gully:
Update here: I noticed that bokeh offers support for "level of detail" scaling:

https://bokeh.pydata.org/en/latest/docs/user_guide/tools.html#controlling-level-of-detail

It may be worth looking into setting a level-of-detail threshold to make the performance better for short cadence lightcurves.

barentsen:
I'm going to close this for now because we are running out of person power to finish and support new features, but I hope someone will be able to pick this up in the future!!  It looks great!

Issue #444, opened 2019-02-12
BoxLeastSquaresPeriodogram input parameters do not match those suggested by LightCurve.to_periodogram


Onoddil:
#### Problem description
``lc.to_periodogram`` has in its docstring:
```
Keywords accepted by `~lightkurve.periodogram.BoxLeastSquaresPeriodogram` are:
            ``min_period``, ``max_period``, ``period``, ``frequency_factor``.
```
however, ``BoxLeastSquaresPeriodogram.from_lightcurve`` looks for ``maximum_period = kwargs.pop("maximum_period", None)`` (and the same for ``minimum_period``). This then leads to the case where recommendations from one method fail when being passed to the second.

#### Example

```python
import lightkurve as lk
lc = lk.search_lightcurvefile('wasp-6').download()
p1 = lc.remove_nans().bin(5).to_periodogram(method='BLS', min_period=1, max_period=6)
```

leading to ``TypeError: power() got an unexpected keyword argument 'min_period'``

#### Expected behavior
Consistency between docstring for ``to_periodogram`` and ``BoxLeastSquaresPeriodogram`` ``kwargs.pop`` lines

barentsen:
Thanks for reporting this @Onoddil. We must improve the docstring!

barentsen:
Clearly it's not great that LombScargle expects `min_period` and BLS expects `minimum_period`.

I suggest we change both `LombScarglePeriodogram` and `BoxLeastSquaresPeriodogram` to deprecate and rename the following arguments:
* `min_period` => `minimum_period`
* `max_period` => `maximum_period`
* `min_frequency` => `minimum_frequency`
* `max_frequency` => `maximum_frequency`

I suggest this because both `astropy.stats.LombScargle` and `astropy.stats.BoxLeastSquares` use `minimum_period/frequency` and `maximum_period/frequency` and it would be good to be consistent with astropy.

@ojhall94 @christinahedges Would you have any objections to this change?  We can keep the old argument names around for a few months to ease the pain.

ojhall94:
No objections from me! Thanks for finding this bug @Onoddil!

barentsen:
This has now been fixed as part of #478.  Closing!

PR #478, opened 2019-03-28
Rename min_period/max_period into minimum_period/maximum_period in `LombScarglePeriodogram`

barentsen:
This PR addresses #444.

TODO: Restore support for the old arguments with a deprecation warning?


ojhall94:
Hi Geert! Thanks for doing this (tedious) job. test_periodogram.test_error_messages() will need updating to match the updated syntax also!

barentsen:
Thanks @ojhall94!  Yep, we'll need to update the tests too!

barentsen:
Awesome.  Thanks @ojhall94, you're the best!  I added support for the old arguments (with a deprecation warning) so I don't think this will break anything.  I plan to merge this shortly unless anyone objects!

PR #485, opened 2019-04-02
Add a badge to run tutorials in Google Colab

barentsen:
This PR brings three changes:

1. It changes the Sphinx prologue of each tutorial to show badges instead of text (cf. [current behavior](https://docs.lightkurve.org/quickstart.html)).
2. It adds a Google Colab badge which takes the user straight to a Colab-hosted version of the notebook.
2. It adds a few cells at the top of the `quickstart.ipynb` notebook with installation instructions (i.e. a cell with `!pip install lightkurve` and some explainer text).  These extra cells are intended to make it easy for a user to install lightkurve when opening the notebook in Colab.  These extra cells have the metadata `"nbsphinx": "hidden"` so they are not shown in sphinx.  However these extra cells may confuse users who already have lightkurve installed, and they are inconsistent with our recommendation to install lightkurve using conda, so I'm not yet convinced this is the right solution.

Preview:
Screenshot from 2019-04-02 08-08-17

/cc @gully

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -81,19 +81,14 @@ nbsphinx_prolog = r"""
 
 .. only:: html
 
-    .. role:: raw-html(raw)
-        :format: html
+    .. raw:: html
 
-    .. nbinfo::
-        **This page is a static version of an interactive Jupyter notebook**
-
-        - Try the interactive version: :raw-html:`<a href="https://mybinder.org/v2/gh/KeplerGO/lightkurve/master?filepath=docs/source/{{ docname }}"><img alt="Binder badge" src="https://mybinder.org/badge_logo.svg" style="vertical-align:text-bottom"></a>`
-
-        - Download the source file: `{{ docname }}`__
-
-    __ https://github.com/KeplerGO/lightkurve/blob/master/docs/source/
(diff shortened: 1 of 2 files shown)


barentsen:
I removed the changes to `quickstart.ipynb` from this PR for now, so we can just merge the new badges and worry about dependency management in the cloud separately.

@gully Can you have a look and merge if happy?

gully:
Indeed, dependency management in the cloud is an issue.  When I tried naively to run one of the notebooks, I got:


```python
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-73a736ec1f11> in <module>()
----> 1 from lightkurve import search_targetpixelfile
      2 # First we open a Target Pixel File from MAST, this one is already cached from our previous tutorial!
      3 tpf = search_targetpixelfile(6922244, quarter=4).download()
      4 
      5 # Then we convert the target pixel file into a light curve using the pipeline-defined aperture mask.

ModuleNotFoundError: No module named 'lightkurve'

NOTE: If your import is failing due to a missing package, you can
manually install dependencies using either !pip or !apt.

To view examples of installing some common dependencies, click the
"Open Examples" button below.
```

Issue #490, opened 2019-04-03
Periodogram normalization doesn't work for custom frequency ranges


keatonb:
#### Problem description
The `.to_periodogram()` function allows the periodogram to be calculated across a custom frequency range, but the way the psd normalization is implemented, the units returned are only correct if the periodogram is computed between 0 and the Nyquist.  This is a problem when passing any of the min_frequency, max_frequency, min_period, max_period, or nyquist_factor kwargs through to the astropy periodogram function.  

#### Example

```python
#Min example
import numpy as np
import matplotlib.pyplot as plt
import lightkurve as lk

time = np.arange(1,100,0.1)
flux = 1. + 0.1 * np.sin(2.*np.pi*time) + 0.2*np.random.randn(time.size)

lc = lk.LightCurve(time=time,flux=flux)
fullper = lc.to_periodogram(oversample_factor = 5)
partper = lc.to_periodogram(min_frequency= 0.5, max_frequency= 1.5, oversample_factor = 5)

#these should be the same in the range of frequency overlap
plt.plot(fullper.frequency,fullper.power)
plt.plot(partper.frequency,partper.power,zorder=-1)

plt.title('psd normalization breaks for specified freq ranges')
plt.xlabel('frequency')
plt.ylabel('psd')
plt.tight_layout()
plt.savefig('normprobs.png')
plt.show() 
```
![normprobs](https://user-images.githubusercontent.com/6413923/55485607-1062be80-562b-11e9-9414-22c9db10c03c.png)

#### Expected behavior

These should be the same thing, just computed over different frequency ranges.  However, the power is completely different due to the way the power spectrum is normalized.  Rather than computing the spectrum directly in the desired units such that Parseval's theorem is naturally obeyed, the spectrum is rescaled from arbitrary units to try to enforce Parseval's theorem post hoc.  This is a hack that works only when the power spectrum is computed from 0 to the Nyquist frequency, and importantly only in the case where an actual Nyquist frequency can be determined and is used (rare in practice unless samples are strictly evenly spaced).

PR #416 provides a physical normalization for the amplitude spectrum.  I will replace the power spectral density normalization implementation to follow suit.

#### Environment:
- platform (e.g. Linux, OSX, Windows): OSX
- lightkurve version (e.g. 1.0b6): 1.0b30
- installation method (e.g. pip, conda, source): pip

barentsen:
Thank you for this extremely clear example and explanation, @keatonb!

[...]

PR #491, opened 2019-04-03
Correct normalization of power spectral density

keatonb:
Implements a physical normalization of the power spectrum based on the actual power computed at each frequency rather than scaling the relative power through the periodogram to force agreement with Parseval's theorem.  Fixes issue #490, as the current normalization causes unexpected output when computing the periodogram over custom frequency ranges.

### DIFF
--- a/lightkurve/periodogram.py
+++ b/lightkurve/periodogram.py
@@ -287,7 +287,10 @@ class Periodogram(object):
         if style is None or style == 'lightkurve':
             style = MPLSTYLE
         if ylabel is None:
-            ylabel = "Power Spectral Density [{}]".format(self.power.unit.to_string('latex'))
+            if self.power.unit == cds.ppm:
+                ylabel = "Amplitude [{}]".format(self.power.unit.to_string('latex'))
+            else:
+                ylabel = "Power Spectral Density [{}]".format(self.power.unit.to_string('latex'))
 
         # This will need to be fixed with housekeeping. Self.label currently doesnt exist.
         if ('label' not in kwargs) and ('label' in dir(self)):
@@ -552,8 +555,9 @@ class LombScarglePeriodogram(Periodogram):
     def from_lightcurve(lc, minimum_frequency=None, maximum_frequency=None,
                         minimum_period=None, maximum_period=None,
                         frequency=None, period=None,
-                        nterms=1, nyquist_factor=1, oversample_factor=1,
-                        freq_unit=1/u.day, **kwargs):
+                        nterms=1, nyquist_factor=1, oversample_factor=None,
+                        freq_unit=None, normalization="amplitude",
+                        **kwargs):
         """Creates a Periodogram from a LightCurve using the Lomb-Scargle method.
 
         By default, the periodogram will be created for a regular grid of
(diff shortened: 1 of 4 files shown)


lightkurve/periodogram.py
```diff
@@ -749,14 +749,8 @@ def from_lightcurve(lc, minimum_frequency=None, maximum_frequency=None,
...
 
-        # Normalise the according to Parseval's theorem
-        norm = np.std(lc.flux * 1e6)**2 / np.sum(power)
-        power *= norm
-
-        power = power * (cds.ppm**2)
-
         # Rescale power to units of ppm^2 / [frequency unit]
-        power = power / fs
+        power *=  2./(len(time)*oversample_factor*fs) * (cds.ppm**2)
```
ojhall94: Could you add a comment above this line explaining how the normalisation is done (just a single line, as before)? Also possibly include a 'why', which I neglected to do in the original code.

lightkurve/periodogram.py
```diff
@@ -676,7 +676,7 @@ def from_lightcurve(lc, minimum_frequency=None, maximum_frequency=None,
         # Hard coding that time is in days.
         time = lc.time.copy() * u.day
 
-        # Calculate Nyquist Frequency and frequency bin width in terms of days
+        # Approximate Nyquist Frequency and frequency bin width in terms of days
```
ojhall94: Hah!

ojhall94:
Thanks for this PR @keatonb! I have provided some minor minor comments. Does the new method return the same power as the old method for a periodogram evaluated from 0 to Nyquist?

barentsen:
I'll wait for @keatonb to respond to @ojhall94's questions, but this looks good to be merged otherwise!

lightkurve/periodogram.py
keatonb: Added something slightly more descriptive

lightkurve/periodogram.py
keatonb: A bit of a pet peeve of mine, haha.

lightkurve/periodogram.py
ojhall94: Awesome, thanks Keaton!

keatonb:
> Thanks for this PR @keatonb! I have provided some minor minor comments. Does the new method return the same power as the old method for a periodogram evaluated from 0 to Nyquist?

One would expect it to, but curiously it does not...

newnorm

It gives a result that is very close, but scaled off by a small random factor near unity.  I haven't been able to track down the source of this difference, or to decide which is the desired behavior...

ojhall94:
Paging @grd349  😬

grd349:
Hello! 

This isn't immediately clear to me. I'm a bit short on time to get into this now. 

1st ideas ... gaps in the data or flawed estimates of the variance in the time domain. 

Can we try this on evenly spaced white noise?

keatonb:
This example was for even sampling, and when I remove the sinusoid so it's just white noise, I see the same differences.  Here's a histogram of the ratio of the new to old normalization factors for 10000 realizations of white noise (1000 samples with 1% standard deviation).

normfactordist

This difference has a larger scatter when there are fewer points or the noise level is increased. My guess is that this has something to do with how the light curve is normalized or how the variance is estimated.  I have tried normalizing by the mean vs. the median, subtracting the measured vs. the known underlying mean, penalizing one degree of freedom on the variance measurement, including/excluding `fit_mean` and `center_data` in the Lomb-Scargle calculation.  These all affect the results a little, but nothing has brought the two normalization factors into perfect agreement.

grd349:
Thanks @keatonb - that's a good set of tests!  I agree that looks like it could be noise on the estimate of variance.  Not so sure about the offset.

At the sort of levels we are seeing here, is this a problem?  ~<0.5% differences in the normalisation of the periodogram are expected to be much smaller than some other contributions to the power levels.  I would say that this has no impact on my science.   Can we cope with a warning in the docs?

jsk389:
After a lot of discussing with @keatonb we have come across the reason for the difference in the normalisation factors. Surprisingly this has little to do with the estimation of the variance and more to do with the application of Parseval’s theorem.

We took 1000 realisations of a white noise signal (evenly-spaced in time) and for each one computed the power spectrum using: 

1. Full FFT (positive and negative frequencies)
2. Astropy Lomb-Scargle slow implementation
3. Astropy Lomb-Scargle  fast implementation

All of which are normalised according to the Astropy implementation.

 In each case we computed the ratio of the variance to the mean squared power in the power spectrum, i.e. the ratio of the lhs and rhs of Parseval’s theorem. 

For every case this should equal one since Parseval’s theorem is defined by the properties of the Fourier transform. These ratios are shown in the plot and labelled accordingly. 

Parseval


As expected the ratios computed using the power spectrum from the FFT are equal to one every time. However for the periodogram (both the fast and slow implementation) this is not the case and Parseval’s theorem is appear to hold. We are not sure of the reason for this happening, but it could be due to the conditions behind Parseval’s theorem being violated in some way (i.e. must be a unitary transform etc.). 

This certainly shows that there may be a problem calibrating according to Parseval’s theorem, since even in the ideal case there are issues doing it this way.

I should also note that Parseval’s theorem does not hold for unevenly-spaced data (due to the Nyquist not being well-defined) and so it should not be used as a normalisation for real data. The normalisation factor added in by @keatonb gives the expected result in all cases.

grd349:
Hi @jsk389 ( how you getting on James :) )

This looks really good - agreed, use the method @keatonb has made.

Is it easy for you to check that this method holds (or not) for data with gaps (I expect it does but I've been wrong before) but also if someone uses the gap_fill() method (I expect it does not but I've been wrong before).  I suspect the 'Parsevals'  method will also fall over on gap_filled data because that certainly biases the estimate of the variance if the gap is substantial.

barentsen:
Thank you @keatonb et al.  This has been merged into master!  (I did a manual merge to resolve the merge conflict.)

Thanks++ everyone! :+1:  Do open new issues or PRs if new findings pop up, or we improve our understanding of the normalization issues!

PR #494, opened 2019-04-08
[Bugfix] Use `quality_bitmask` keyword in download_all

nksaunders:
The `download_all()` function accepted a `quality_bitmask` keyword, but never used it when opening the downloaded files. This PR actually applies the keyword when calling the `open()` function.


barentsen:
Thanks @nksaunders! Good catch!! :+1:  It looks like we accidentally introduced this bug in #367.

I don't quite understand what's up with the test failures yet.  (Other PRs are failing too.)

Issue #489, opened 2019-04-03
KeplerLightCurveFile does not return a warning about filtered cadences


JeffLCoughlin:
By default, lightkurve.KeplerLightCurveFile has the attribute quality_bitmask default to "default", which is a very sane choice that removes cadences with particular flags.

However, it is easy to quickly open a file, e.g.,

`lc = lk.KeplerLightCurveFile("ktwo206467205-c19_llc.fits")`

and start work, without a user realizing that such cadences have been filtered, unless they dig into the documentation. This could lead to losing a bet with a co-worker about whether there are true gaps in the data or not, resulting in owing that co-worker a beverage.

Ideally, KeplerLightCurveFile and similar functions could return a warning message when the file is opened, such as:

`In[1]: lc = lk.KeplerLightCurveFile("ktwo206467205-c19_llc.fits")`
`Out[1]: Warning: Using quality_bitmask = default. 1,283 cadences have been filtered.`

barentsen:
Thanks for opening this issue @JeffLCoughlin. I understand this concern. I feel uneasy about discarding data without warning the user.

Some background: we discussed the trade-off between being "explicit" and "noisy" on a few occasions.  I believe we ended up choosing not to be verbose because the default bitmask is designed to strictly only reject cadences which 99% of users should never have to worry about.  The definition of the default bitmask is shown here:

https://github.com/KeplerGO/lightkurve/blob/a6169d4e125ee0039ff7074277de290f3d96139d/lightkurve/utils.py#L125-L127

Of course this default mask makes less sense for C19 (and perhaps C0) due to the large number of coarse point cadences in those Campaigns (I'm guessing).  I wonder if the solution is to print a warning message if the default bitmask rejects more than 5% of the cadences, so that at least users of C19 will see a warning message about the silly number of cadences being discarded.  Could that be a good solution / trade-off? Happy to discuss and revise our strategy here!

As an aside, Lightkurve does print a message about the masking if the user raises the verbosity, but that is an obscure feature.  Example:

![Screenshot from 2019-04-02 22-12-12](https://user-images.githubusercontent.com/817669/55454378-94b63280-5594-11e9-9e26-9a860280a273.png)

JeffLCoughlin:
Yea perhaps I hit more of an edge case working on C19 data. I like that middle-ground of 5% or some such number --- that might be the right tradeoff. Maybe even 1% just to play it safe for folks, but would have to look at often that comes up.

[...]

PR #495, opened 2019-04-08
Log a warning if a large number of cadences are masked out by `quality_bitmask`

barentsen:
I propose to issue a warning when more than 20% of the cadences are masked out by `quality_bitmask`, which is the case for K2 C0 and C19. (In contrast, e.g. C12 has 12% bad cadences using the default mask).  I hope this won't make Lightkurve too noisy.

This PR addresses #489.  

### Example:

```python
>>> import lightkurve as lk
>>> tpf = lk.search_targetpixelfile("K2-20", campaign=0).download(quality_bitmask="default")
Warning: 56% (2112/3753) of the cadences will be ignored due to the quality mask (quality_bitmask=1130799).
```

### DIFF
--- a/lightkurve/utils.py
+++ b/lightkurve/utils.py
@@ -85,8 +85,17 @@ class QualityFlags(object):
                                  "".format(bitmask, valid_options))
         # The bitmask is applied using the bitwise AND operator
         quality_mask = (quality_array & bitmask) == 0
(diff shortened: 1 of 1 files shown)


barentsen:
The test is failing due to a known issue (#507).  I intend to merge this later today!

PR #496, opened 2019-04-10
Add asteroseismic parameter estimator for solar-like oscillators

ojhall94:
This PR continues #266 (closed) and #277 (merged) and #282 (sloppily closed, my bad).

Between opening #282 and now, the infrastructure of .periodogram changed sufficiently to warrant me re-writing the contents of that PR.

The plan is to implement some basic asteroseismic stuff, so we can go to a Mass and Radius for a star, and leave opportunity for expansions later.

**New working points**
What I want to do:
- [x] Merge master (`lightkurve 1.0.0`) into this pull request.
- [x] Specify that this is for solar-like oscillators only
-~ [x] Log a mention when using a TESS lightkurve object that this is suboptimal for TESS data with only a couple of sectors.~ < mentioned in the docstring that its suboptimal for low SNR
- [x] Add option to calculate numax using autocorrelation
- [x] Make sure we can recreate the contents of [this blog](https://ojhall94.github.io/blogs/autocorrelate.html) using only lightkurve (including plots!)
- [x] Rewrite scaling relations (asteroseismology.py) to use the neat, compact version described in Bellinger+19.
- [x] Write unit tests for scaling relations
- [x] Write unit tests for seismic functions in Periodogram()
- [x] Add a simple echelle diagram
- [x] Provide a thorough test of this method
- [x] Implement a .find_[numax/dnu]_diagnostics() which returns key diagnostics of the method used
- [x] Add sufficient literature references

Implementation tests to clear on at least one MS and RGB star
- [x] Estimate Numax 
- [x] Estimate Dnu
- [x] Diagnostics Numax
- [x] Diagnostics Dnu
- [x] Echelle diagram
- [x] Scaling relations
- [x] Recover parameters for baseline MS star
- [x] Recover parameters for baseline RGB star

Important discussions to have:
- [x] Do we return errors on numax & dnu? [We decided that we would not]


ojhall94:
I've moved the goalposts around a little on this latest commit. Instead of return diagnostic plots, you can show a plot using the `show_plots` keyword. I know this isn't what we want in the final build, but we can shuffle around once the whole thing is up and running.

image

christinahedges:
Hi @ojhall94, `show_plots()` is great! We've also been working with some ideas along with lines of `diagnose()`.

Keep the ideas coming, let me know when you need a review!
C

ojhall94:
Hi @christinahedges! I remember when we discussed this back at Ames you suggested a diagnose() function, which I think still may be the best way to go about things. 

After discussion with @grd349 I think I will refrain from fitting a Gaussian to the ACF in favour of just returning a $\nu_{\rm max}$ without an uncertainty. I think this will allow people to use this method for a quicklook at stellar properties of their target, while at the same time discouraging from using the output in published work without further scrutiny. I will leave the infrastructure in here though in case we change our minds!

I've got tests on this PR ongoing on this Jupyter Notebook: https://github.com/ojhall94/keplergohacks/blob/master/seismo_stuff/autocorr_test.ipynb

ojhall94:
I have added a estimator for the large separation, and stopped the method from returning an uncertainty as I suggested before.

I'm basing the first guess for the Dnu estimation on the empirical relation presented in [Stello et al. 2009](https://ui.adsabs.harvard.edu/abs/2009MNRAS.400L..80S/abstract). Adding an interactive slider for Dnu to hit the right peak might be a nice addition.

Lots of failsafe testing (particularly re: correct units) is next on the list. @grd349, I think it's ready for you to play with.

image

image

grd349:
Excellent work @ojhall94!  As a first test I ran some of the stars that belong to NGC6819.  See the .ipynb that could make the basis of a tutorial perhaps? (after a lot of work)

An issue we'll need to look at is when there are large noise spikes in the data (not very often and perhaps astrophysical not true noise but noise to us).  Here the estimate_numax option is getting fooled and picking the wrong numax - once this happens estimate_dnu fails too (of course).  

We might need a better prior on the dnu/numax relationship - this would be fairly easy to do ... after all, dnu & numax are well described by a power law :) .  To make things statistical, a 2D Gaussian with full covariance would be easy to implement and we could use the Yu 2017 sample to fit the Gaussian properties.  We would need to add more stars (Serenelli sample) for the main sequence/subgiant stars but this is easily done.

Happy to take advice on how to fit the .ipynb into/close to the tutorial style - I avoided list comprehension in favour of for loops but I also used dictionaries with appends which might be a little cryptic.

Great job so far - I think I could starting using this after only a few small mods!

ojhall94:
I'm trying to come to terms with how to properly normalise the ACF. My previous approach to estimating numax involved taking the maximum value of the ACF calculated at each numax, but this approach only holds when the modes provide the highest amplitude signal in the spectrum.

We could keep it this way, but its not very robust (see Guy's example). Instead, I tried to normalise so that the ACF at zero lag is equal to a probability of 1 (as opposed to some metric of signal power, which is what np.correlate usually returns). I then save the total power of the ACF at each numax. In principle, the repeated modes should provide the highest total power for an ACF evaluated around the correct numax.

The problem with this in practice is that because the window of data I self-correlate to get the ACF increases with frequency, higher values of numax find more power in the ACF. I could normalise this by the length of the data that is being self-correlated, but this then unfairly weights up the total ACF power at very low frequencies.

I'm a bit stumped. @grd349, any thoughts?

ojhall94:
I've made the decision to revert back to my original method (using a moving hanning window), and finding numax by seeing where the power is highest. In the exception of large spikes (like in Guy's example) this works, and I think the ability to output a diagnostic plot and input a custom range of numax's to search makes it easy enough to find the right answer with a bit of human intervention.

barentsen:
:+1:

ojhall94:
Echelle diagram is implementation tested: https://github.com/ojhall94/keplergohacks/blob/master/seismo_stuff/echelle_implementation_test.ipynb

ojhall94:
Alright, @grd349, this is ready for testing now.

I've changed the way that the numax is estimated. The method I previously used was taking the total autocorrelation power in a Hanning filter, as:

image

This biases towards large filters (larger _N_) and anomalously large peaks. I've now changed it to:

image

Which reduces both those biases somewhat.

Pros: More consistent recovery of proper numax in integration tests.

Cons: Messier, harder to interpret, not really physically justified.

Normalisation as shown in Mosser & Appourchaux (2009) doesn't really apply in this scenario, as it does not treat the increased autocorrelation power due to filter width.

barentsen:
I noticed there is a new/recent paper on arXiv discussing how best to derive delta nu: http://adsabs.harvard.edu/abs/2019arXiv190508333V

ojhall94:
Thanks for bringing this paper to my attention Geert! I'll give it a careful read tomorrow!

ojhall94:
Besides tweaks to the methods (sorry @barentsen, I didn't get round to reading that paper yet) and updates to the unit tests, the main body of this PR is done. @grd349 has been working on some tests/tutorials which are included in the PR (but can move elsewhere before merging).

I'm off on holiday now until the 20th of June, after which I'll wrap up the methods asap making this hopefully ready for review!

barentsen:
@ojhall94 @grd349 Awesome!! THANK YOU!! :heart:

I insist that you enjoy your vacation. ✈️ 🛄  Vacation is super important to stay sane!

When you return from vacation, here are a few random bits of feedback that may be helpful:

1. It definitely won't be necessary to digest the new Viani paper before KASC.  I'm 100% happy to merge a best-effort algorithm now, and let it trigger a separate discussion if the community thinks it is important to use or offer alternative approaches in the future. We'll just want to make sure that methods such as `estimate_dnu` and `estimate_numax` are defined in a sufficiently generic way to potentially support different underlying algorithms should the need arise.  (I don't expect a problem there, it already seems pretty generic in this PR.)
2. I would recommend removing the `show_plots` feature from `estimate_dnu` and `estimate_numax` in favor of dedicated `plot_deltanu_diagostics` and `plot_deltamax_diagostics` methods (or some other name). I say this because it is easier to test, debug, and use smaller functions which always return the same number of parameters.  (Under the hood they can both share a helper function, of course.)
 3. Make sure we include literature references to explain where the scaling relations come from.  (But I see you already have this on your TODO list anyway!)
 4. I'd recommend against including interactive ACF/Echelle plots as part of this PR.  Interactive plots are complicated and require a lot of code, so let's develop/discuss those as a separate PR to keep this one easier to review/discuss/merge!

Again, thanks for the awesome work on this!!!

ojhall94:
Thanks for the support Geert!

1. Agreed :)
2. Can do!
4. Agreed, should be separate, more of a fun project, but also want to avoid feature creep. Can worry about this another time :)

ojhall94:
Updated the goalposts.

ojhall94:
Hey @barentsen, @christinahedges,

> 2. I would recommend removing the `show_plots` feature from `estimate_dnu` and `estimate_numax` in favor of dedicated `plot_deltanu_diagostics` and `plot_deltamax_diagostics` methods (or some other name). I say this because it is easier to test, debug, and use smaller functions which always return the same number of parameters.  (Under the hood they can both share a helper function, of course.)

I was going to start trying out this reshuffle today. Do you have any particular requests for how you'd like it to look? The main difficulty I think is that some thing I would want to show on the plot (such as the acf function and the smoothed acf function) won't be returned from the `estimate_numax()` function, and custom input such as a custom list of numaxs to evaluate the ACF at won't carry over.

A solution would be to store this stuff as properties of the SNRPeriodogram object. Storing the numax and dnu as properties would make some stuff smoother (like using the scaling relations), but I personally prefer having numax and dnu be output and fed back in to functions such as `estimate_dnu()` and (soon) `estimate_stellar_parameters()` etc.

barentsen:
@ojhall94 The typical solution is to create an internal helper function, e.g. `_estimate_numax()`, which would be called by `estimate_numax()` and `plot_numax_diagnostics()` alike.  I see you already have an internal `_autocorrelate()` method which allows code to be shared between `estimate_dnu()` and `estimate_numax()`.  This is exactly the right pattern you want to be using, however it may take a bit of trial and error to figure out what the internal method(s) need to accept/return so that they are useful for all functions they need to support!  (Does that help?)

ojhall94:
It does, thanks @barentsen! I'll have a crack at it today.

ojhall94:
Just got to add the TESS warning and implementation test on some stars--- should be done tomorrow, and then we'll ready for review, and certainly ready for TASC!

ojhall94:
Hi all,

This is done and ready for review now!

Implementation test found here: https://github.com/ojhall94/keplergohacks/blob/master/seismo_stuff/implementation_test.ipynb

barentsen:
Awesome, thanks @ojhall94!!  It's a long weekend in the US due to 4th of July, but we're going to try this out ASAP!!

ojhall94:
Managed to break some stuff, I'll tick them off when I fix things
- [ ] Can't use default estimate_numax() when using a sliced spectrum (as it will start below the zeroth bin)

lightkurve/periodogram.py
```diff
@@ -371,6 +508,40 @@ def flatten(self, method='logmedian', filter_width=0.01, return_trend=False):
             return snr, bkg
         return snr
 
+    def _get_fwhm(self, numax):
+        """For a given numax, the method calculates the expected Full Width Half
+        Maximum of the seismic mode envelope as (Mosser et al 2010) for RGB
+        stars:
```
barentsen: As a seismology-n00b, it's not clear to me exactly what a `seismic mode envelope` is. and why a specific `numax` has a predictable FWHM of the envelope.  Is this something that is easy to elucidate by adding one or two extra sentences?  If so, it would help @barentsen understand what's going on =)

lightkurve/periodogram.py
```diff
@@ -371,6 +508,40 @@ def flatten(self, method='logmedian', filter_width=0.01, return_trend=False):
...
+
+        fwhm = 0.66 * numax^0.88,
+
+        if the data is generated for a long-cadence timeseries. If the
+        timeseries is short-cadence, the relation will be:
+
+        fwhm = 0.25 * numax.
+
+        This will be based on the highest frequency in the periodogram.
```
barentsen: This sentence confuses me a bit.  The equations above have fwhm depend on numax, but this sentence suggests it is based on max(self.frequency).  Can you clarify and/or edit this sentence?

lightkurve/periodogram.py
```diff
@@ -371,6 +508,40 @@ def flatten(self, method='logmedian', filter_width=0.01, return_trend=False):
...
+            numax : float
+                The estimated position of the numax of the power spectrum. This
+                is used to calculated the region autocorrelated with itself.
+
+        Returns:
+        --------
+            fwhm: float
+                The estimate full-width-half-maximum of the seismic mode envelope
+        """
+        fs = np.median(np.diff(self.frequency.value))
```
barentsen: `fs` doesn't seem to be used in the remainder of this method, so this line can probably be removed!

lightkurve/periodogram.py
```diff
@@ -544,6 +715,354 @@ def plot(self, **kwargs):
...
+        deviation of 5 to smooth it. The frequency that results in the highest
+        value of M is the detected numax.
+
+        NOTE: This method is not robust against large peaks in the spectrum (due
+        to e.g. spacecraft rotation), nor is it robust in the case of low signal
+        to noise (such as for single sector TESS data). Exercise caution when
+        using this module!
+
+        NOTE: This function is intended for use with solar like Main Sequence
+        and Red Giant Branch oscillators only.
```
barentsen: This is a fabulous docstring :heart_eyes:

lightkurve/seismology.py
```diff
@@ -0,0 +1,206 @@
...
+
+log = logging.getLogger(__name__)
+
+__all__ = ['estimate_radius','estimate_mass','estimate_logg']
+
+"""Global parameters for the sun"""
+numaxsol = ufloat(3090, 30) # microhertz | Huber et al. 2011
+dnusol = ufloat(135.1, 0.1) # microhertz | Huber et al. 2011
+teffsol = ufloat(5772., 0.8) # Kelvin    | Prša et al. 2016
+gsol = ((const.G * const.M_sun)/(const.R_sun)**2).to(u.cm/u.second**2) #cms^2
```
barentsen: Nitpicking: [Python PEP8 convention](https://www.python.org/dev/peps/pep-0008/#constants) is to use UPPERCASE_VARIABLE_NAMES for constant values.  While not strictly required for functionality, it is good to use uppercase here for code readability!

lightkurve/seismology.py
```diff
@@ -0,0 +1,206 @@
...
+    R_e = uR.s * u.solRad
+
+    if all(b is not None for b in [numax_err, dnu_err, Teff_err]):
+        return R, R_e
+    else:
+        return R
+
+def estimate_mass(numax, dnu, Teff, numax_err=None, dnu_err=None, Teff_err=None):
+    """Calculates mass using the asteroseismic scaling relations. This method
+    borrows heavily from work done in Bellinger et al. (2019).
```
barentsen: When you say `borrows heavily`, does it mean I can go into Bellinger+ 2019 to look up what the scaling relations are?  If so, can we cite e.g. a specific equation number or table?

barentsen:
Thank you so much for working on this @ojhall94!  I'm going to be playing with this today and will be leaving additional comments as I make progress.

I can't mention it in each nit-picky comment so I'll mention it here: **this is already looking really really awesome**!

lightkurve/seismology.py
ojhall94: Thats an easy fix :)

lightkurve/seismology.py
ojhall94: I've updated the docstrings to explain the scaling relations, with appropriate references to the historical literature and Bellinger+19 as a source of the code I borrow from & as a review on the topic.

lightkurve/periodogram.py
ojhall94: Added background + references

ojhall94:
- [ ] `plot_echelle()` is not (yet) robust against periodograms that don't run from 0 to nyquist. I can (and will) fix this next :)

ojhall94:
Done!

lightkurve/seismology.py
```diff
@@ -0,0 +1,258 @@
...
+from uncertainties import umath
+
+log = logging.getLogger(__name__)
+
+__all__ = ['estimate_radius','estimate_mass','estimate_logg']
+
+"""Global parameters for the sun"""
+NUMAX_SOL = ufloat(3090, 30) # microhertz | Huber et al. 2011
+DNU_SOL = ufloat(135.1, 0.1) # microhertz | Huber et al. 2011
+TEFF_SOL = ufloat(5772., 0.8) # Kelvin    | Prša et al. 2016
```
barentsen: Oops, unfortunately the unicode `š` character is making the Lightkurve tests fail on Python 2.  Better replace it with an `s`.  Sorry Andrej!!
```suggestion
TEFF_SOL = ufloat(5772., 0.8) # Kelvin    | Prsa et al. 2016
```

lightkurve/periodogram.py
```diff
@@ -371,6 +507,47 @@ def flatten(self, method='logmedian', filter_width=0.01, return_trend=False):
...
+        approximately Gaussian, for all basic purposes, also referred to as the
+        'mode envelope'. For a given numax (the central frequency of the mode
+        envelope), the expected Full Width Half Maximum of the envelope is known
+        as a function of numax for evolved Red Giant Branch stars as follows
+        (see Mosser et al 2010):
+
+        fwhm = 0.66 * numax^0.88 .
+
+        If the maximum frequency in the periodogram is less than 500 microhertz,
+        this functio will default to the above equation under the assumption it
```
barentsen:
```suggestion
        this function will default to the above equation under the assumption it
```

lightkurve/periodogram.py
```diff
@@ -544,6 +721,422 @@ def plot(self, **kwargs):
...
+            if window < fs:
+                raise ValueError("You can't have a window smaller than the"
+                                "frequency separation!")
+            if window > (self.frequency[-1].value - self.frequency[0].value):
+                raise ValueError("You can't have a window wider than the entire"
+                                "power spectrum!")
+            if window < 0:
+                raise ValueError("Please pass a positive window.")
+
+        # Calcualte the window size
```
barentsen:
```suggestion
        # Calculate the window size
```

barentsen:
Hi @ojhall94!!  A few additional thoughts:

* Is there a reason you added the `estimate_numax()` and `estimate_dnu()` methods to `SNRPeriodogram` instead of `Periodogram`?  Doesn't every `Periodogram` have a numax and deltanu? I worry people will try calling `lc.to_periodogram().estimate_numax()` instead of `lc.to_periodogram().flatten().estimate_numax()` and not understand why the method is not found in the former case.  Is it always true that the result is useless if you don't flatten first?  If so, perhaps `estimate_numax()` could call `flatten()` behind the scenes by default if `lc.to_periodogram().estimate_numax()` is called?

* Would it name the method `estimate_deltanu()` instead of `estimate_dnu()`, because `deltanu` is the common term in "daily speech"? Thoughts?

* It is best for functions to *always* return the same number of parameters (it reduces syntax mistakes).  I wonder if we want to change `estimate_radius()`, `estimate_mass()`, and `estimate_logg()` to return an `uncertainties.ufloat` if the input parameters are `uncertainties.ufloat` objects, and a normal `float` otherwise.  That way, only one object is returned at all times.  (I'm happy to help implement that this weekend if you want me to!)

ojhall94:
Hey @barentsen! Thanks for your excellent comments on this!

> * Is there a reason you added the `estimate_numax()` and `estimate_dnu()` methods to `SNRPeriodogram` instead of `Periodogram`?  Doesn't every `Periodogram` have a numax and deltanu? I worry people will try calling `lc.to_periodogram().estimate_numax()` instead of `lc.to_periodogram().flatten().estimate_numax()` and not understand why the method is not found in the former case.  Is it always true that the result is useless if you don't flatten first?  If so, perhaps `estimate_numax()` could call `flatten()` behind the scenes by default if `lc.to_periodogram().estimate_numax()` is called?

The reason for this is that `estimate_numax()` relies on a detection of power excess. If the periodogram is not flattened, granulation noise in the star will have lots of power at low frequencies, and the method will return a wrong estimate. `estimate_dnu()` works fine on an unflattened periodogram, but only if you pass it a correct `numax` value. So really, the two of them should be paired.

Calling `flatten()` behind the scenes would be good solution. My reasoning for not doing so initially is that I think the user should be an active participant in getting the data to a place where these tools can be used (in the same way that the user will remove nans or trim outliers from a lightcurve object). I'm happy to do it either way though. Your call, I'll make the change if needed :)

> * Would it name the method `estimate_deltanu()` instead of `estimate_dnu()`, because `deltanu` is the common term in "daily speech"? Thoughts?

This is a good suggestion. I've changed all instances of `dnu` with `deltanu` in docstrings, function titles and variable names through `periodogram`, `seismology` and associated unit tests.

> * It is best for functions to _always_ return the same number of parameters (it reduces syntax mistakes).  I wonder if we want to change `estimate_radius()`, `estimate_mass()`, and `estimate_logg()` to return an `uncertainties.ufloat` if the input parameters are `uncertainties.ufloat` objects, and a normal `float` otherwise.  That way, only one object is returned at all times.  (I'm happy to help implement that this weekend if you want me to!)

The only issue I have with this suggestion is that a `ufloat` object can't have an astropy unit. At the moment, all the output and input of the seismology tools allow for (and is robust against errors in) astropy units, so it would break consistency a little if we don't return values with units here. 

However if theres anywhere to drop the units, it would be in this module--- everything comes out in solar units, and its easily included in the docstring. What do you think?

ojhall94:
`estimate_numax()` takes a very long time to run on short-cadence data as it takes 1 microhertz steps at a time. I should scale this up to 10 micohertz at least.

christinahedges:
Hey @ojhall94! This is looking really awesome, I've spent today looking through your PR and I'm blown away by the progress and new utilities here. 

The rest of the team and I have been chatting and we're going to take a crack at a quick refactor of the code you've got here. There are two benefits here; 1) refactoring will make sure that there is an "asteroseismology" module that's easy to chop and change and a "periodogram" module that is light weight and air tight. 2) I will thoroughly understand your code!

We'd like to change the API to be something like:
```python
seis = lc.to_periodogram(kwargs).to_seismology(kwargs)

seis.plot()
seis.estimate_numax()
seis.estimate_mass(Teff)
```
And ideally we'd like to leave some "room" for new functions to be added to e.g. estimate numax.

I took a shot at it today and I'm 10% of the way through so far, watch this space and comment if you think there is anything that I'm changing that is really important. Otherwise I'll take another stab tomorrow morning! Some changes that I've made so far:

1. Defaults for method from fast/slow to fastchi2/chi2 which will let us use nterms and otherwise provides the same results. @ojhall94 do you have any problems with this?
2. Added a time modeling method, this is useful for pulsators and star spots. @barentsen do you have any thoughts on the name which is currently `flux_model`
3. Now we can use nterms > 1
4. Refactored all asteroseismology tools out of `periodogram` and into `seismology`
5. Made a new class, temporarily called `SeismologyButler` which can be renamed...

Some things that I've noticed while refactoring:

* In the ACF docstring you say you're multiplying by some hanning window, but you're not multiplying by anything?
* I'm not quite following why there is an `_estimate_numax` and `estimate_numax` function, but I think in future we might have multiple estimation methods so I've left this roughly as is for now
* I think your 'helper' functions should return dictionaries, because return 6 variables from a function can get confusing, and wires easily get crossed.
* When we're estimating numax, I'm not thoroughly understanding why those defaults are chosen, and the kwargs for `_estimate_numax` are not easy for me to understand currently (what is spacing and windows? they sound so similar). Is this something that can be shown in the diagnostic plot, so I can understand the window size?

ojhall94:
Hey @christinahedges! This is going to be a long reply so bear with me... thanks for putting so much care into looking over this code!

> Hey @ojhall94! This is looking really awesome, I've spent today looking through your PR and I'm blown away by the progress and new utilities here.

Thank you! :)

> The rest of the team and I have been chatting and we're going to take a crack at a quick refactor of the code you've got here. There are two benefits here; 1) refactoring will make sure that there is an "asteroseismology" module that's easy to chop and change and a "periodogram" module that is light weight and air tight. 2) I will thoroughly understand your code!

Cool, I trust your judgement! Once the refactor is over I'll pass through and make sure it makes sense with the use cases for the modules.

> 1. Defaults for method from fast/slow to fastchi2/chi2 which will let us use nterms and otherwise provides the same results. @ojhall94 do you have any problems with this?

Part of me feels like this should be a separate PR, but if it provides the same results in power then sounds like thats probably fine! 

> 2. Added a time modeling method, this is useful for pulsators and star spots. @barentsen do you have any thoughts on the name which is currently `flux_model`

I might have already done this in #390, which was finished but never merged, and now also built on an outdated periodogram class. Maybe theres some useful code in there.

> 3. Now we can use nterms > 1

We'll probably want to add a little paragraph in the docstring about what this does.

> 4. Refactored all asteroseismology tools out of `periodogram` and into `seismology`

I just had a look at this, looks sick!

> 
> Some things that I've noticed while refactoring:
> 
> * In the ACF docstring you say you're multiplying by some hanning window, but you're not multiplying by anything?

Good catch, the ACF no longer uses a hanning window. I'll take it out.

> * I'm not quite following why there is an `_estimate_numax` and `estimate_numax` function, but I think in future we might have multiple estimation methods so I've left this roughly as is for now

This was a suggestion by @barentsen so that we can have both the `estimate_numax()` and `plot_numax_diagnostics()` functions. The plotting needs all kinds of data from the estimation process (as you've addressed in your next point), but `estimate_numax` only needs the numax output.

> * I think your 'helper' functions should return dictionaries, because return 6 variables from a function can get confusing, and wires easily get crossed.

This is a much better solution to what I was doing!

> * When we're estimating numax, I'm not thoroughly understanding why those defaults are chosen, and the kwargs for `_estimate_numax` are not easy for me to understand currently (what is spacing and windows? they sound so similar). Is this something that can be shown in the diagnostic plot, so I can understand the window size?

To some degree this is already shown in the diagnostic plot--- the window is y-axis on the 2D ACF image. I'll update the docstrings to reflect this a bit better.

Overall I think this refactor is super neat There a couple of stuff I want to make sure we address, depending on what you think.

* The `estimate_numax()` function will not work unless the periodogram has been flattened. We could do this if SeismologyButler is called from anything that isn't a SNRPeriodogram, or we could make the user do this themselves beforehand. Flattening out the background is a relatively big deal when it comes to studying the modes so I want to make sure that the fact that this step is happening is clear to the user.

* The `from_lightcurve` staticmethod is super neat! At the moment the kwargs aren't passed in, so we should make sure that if using this staticmethod the user can pass in all the kwargs they would usually pass into `periodogram()`. In particular, `normalization='psd'`, because I haven't tested these tools on `normalization='amplitude'` yet... (in principle that should not be a problem, the code is robust against any frequency unit you throw at it)

* As a follow on to the previous comment, there a lot of important stuff that happens in the Periodogram class that is all done under the hood if a user goes directly to the SeismologyButler. We should make sure its very clear in the docstrings where a user can go to read about the appropriate kwargs.

* This is a tricky one: I don't think the échelle diagram should be limited to the seismology tools per-se. A user may want to represent their periodogram as an echelle diagram with or without deltanu. But this doesn't have to implemented in this PR. A workaround I think would be to have a periodogram.plot_echelle() function, with `period` and `minimum/maximum frequency` kwargs. Then SeismologyButler would have a `plot_echelle` function that called `self.periodogram.plot_echelle(period=dnu, minimum/maximum_frequency = dnu+/-fwhm)`.

* Final comment, I'm not sure on this one: I think it would be useful if the user could tweak the periodogram as they work to get seismic results out, e.g. slicing, smoothing, etc. The smoothing in particular is a big one because I'd like to maintain the `periodogram.smooth().plot_echelle()` and other intuitive functionality like that. I guess with the refactor this becomes `periodogram.smooth().to_seismology().plot_echelle()`, which is a little less intuitive. Thoughts?

lightkurve/seismology.py
```diff
@@ -0,0 +1,908 @@
...
+    @staticmethod
+    def from_lightcurve(lightcurve, **kwargs):
+        warnings.warn("Building a SeismologyButler object directly from a light curve "
+                      "using default periodogram parameters. For further tuneability, "
+                      "create a periodogram object first, using `to_periodogram`.")
+        return SeismologyButler(periodogram=lightcurve.to_periodogram())
+
+    @staticmethod
+    def from_periodogram(periodogram, **kwargs):
+        return SeismologyButler(periodogram=periodogram, **kwargs)
```
barentsen: I don't think there is a need for `from_periodogram` -- users should call the butler constructor directory!

lightkurve/seismology.py
```diff
@@ -0,0 +1,908 @@
...
+                The estimated position of the numax of the power spectrum. This
+                is used to calculated the region autocorrelated with itself.
+
+            window : int or float
+                The width of the autocorrelation window around the central
+                frequency numax.
+
+            frequency_spacing : float
+                The frequency spacing of the periodogram. If none is passed, it
+                is calculated internally. This should never be set by the user.
```
barentsen: If `frequency_spacing` should never be set by the user, is there a use-case for needing this argument?

barentsen:
@christinahedges and I pair-coded on this PR today!  It'll be easier to discuss the changes over Skype with @ojhall94 tomorrow morning, but in brief, we propose two key changes:

1. We moved all the asteroseismology functionality and tests into a dedicated `lightkurve.seismology` sub-package, which provides a `SeismologyButler` class powered by a series of helper functions which enable users to carry out asteroseismic analyses on `Periodogram` objects.  By placing all seismology features into a separate sub-package rather than inside the already-crowded periodogram module, we think it'll be much easier to maintain and expand the seismology features in the future! 
2. We introduced a new `SeismologyQuantity` object, which inherits from `astropy.unit.Quantity`, but has  a few extra attributes that are important for e.g. deltanu or mass estimates.  The extra attributes include `error`, `method`, `diagnostics`, and `diagnostics_plot_method`.

There is definitely still work to do to update the docstrings to reflect these changes, and fix a few failing unit tests, but here is what the usage looks like right now:

Screenshot from 2019-07-16 21-51-32

barentsen:
Note: because we introduced the `lightkurve.seismology` sub-package, you will likely need to execute `python setup.py develop` again to make this sub-package work after pulling from this PR!

lightkurve/seismology/numax_estimators.py
```diff
@@ -0,0 +1,155 @@
...
+    if not hasattr(numaxs, '__iter__'):
+        numaxs = np.asarray([numaxs])
+
+    fs = np.median(np.diff(periodogram.frequency.value))
+    for var, label in zip([np.asarray(window), np.asarray(spacing), numaxs],
+                          ['window', 'spacing', 'numaxs']):
+        if (var < fs).any():
+            raise ValueError("You can't have {} smaller than the "
+                            "frequency separation!".format(label))
+        if (var > (periodogram.frequency[-1].value - periodogram.frequency[0].value)).any():
```
ojhall94: This check doesn't work for `numaxs` when running this code on a periodogram that has been sliced (ie not running from 0 to Nyquist). I'm not sure why this is passing the unit tests on sliced periodograms--- it shouldn't be.

Btw this series of checks is incredibly elegant, I love it!!

lightkurve/seismology/numax_estimators.py
```diff
@@ -0,0 +1,155 @@
...
+    for idx, numax in enumerate(numaxs):
+        acf = utils.autocorrelate(periodogram, numax, window=window, frequency_spacing=fs)      #Return the acf at this numax
+        acf2d[:, idx] = acf                                     #Store the 2D acf
+        metric[idx] = (np.sum(np.abs(acf)) - 1 ) / len(acf)  #Store the max acf power normalised by the length
+
+    # Smooth the data to find the peak
+    # Previous smoothing could be completely wrong, it's based on the length of the array, not the frequency!!!
+    # It needs to be based on the frequency differences in `numaxs`
+    if len(numaxs) > 10:
+        g = Gaussian1DKernel(stddev=100 * np.nanmedian(np.diff(numaxs)))
```
ojhall94: This results in the method failing to recover numax for MS stars. Since the Gaussian1DKernel just takes a stddev in indices and doesn't care about the x-axis units, shouldn't the previous method still work?

ojhall94:
Hey both! I've had a blast playing around with the new seismology butler, and it went over well in a practice talk to the group. I branched off here had to make two minor changes to the code to make it run through, which I've left comments on!

I've got the example in a colab here: https://colab.research.google.com/github/ojhall94/TASC5_lightkurve/blob/master/Notebooks/Colab-Doris-example.ipynb

Speak to you soon!

barentsen:
@ojhall94 Feel free to push those changes as an extra commit to this PR!  (I did a `rebase master` yesterday to make sure we can merge soon, but all should be fine if you pull before comitting.)

lightkurve/seismology/utils.py
```diff
@@ -26,7 +26,9 @@ def __repr__(self):
         return super().__repr__()
 
     def _repr_latex_(self):
-        return "{}: {} (method: {})".format(self.name, super()._repr_latex_(), self.method)
+        return "{}: {} (method: {})".format(self.name,
+                                    '$ {} $'.format(np.round(super().value,2)),
+                                    super().unit._repr_latex_(), self.method)
```
barentsen: This is @christinahedges I'm not sure if this is ideal, the actual value has more decimal places than the repr, so if I start doing math with this value, there will be additional decimal places that get used and might make my final value funny. Should we make the repr e.g. `3.45... ` instead of `3.45`? This problem is kind of small for now and will be taken care of when we can estimate errors.

barentsen:
I just completed a `git rebase master` to resolve the latest merge conflicts.   Everyone please make sure to `git pull`!

barentsen:
I removed three files from this PR which appear to have been committed by accident:
* ACF cluster example.ipynb
* Kepler Test.ipynb
* clusters.csv

(These files will still be recoverable in the commit history of this branch, and I can send someone a copy if need be!)

barentsen:
@ojhall94 Update: today @christinahedges worked her magic to make all the plots look even fancier, while I reviewed the tests and docs.  

It'd be easiest to get this monster PR merged now (Friday) and open smaller PRs for any subsequent changes or improvements. However before we merge it'd be good for you to have a final check to make sure your notebook runs fine, and make any final changes you think are necessary!

I made sure that the new seismology classes/functions will have a page on the online API docs, which you can preview by going into the `docs/` folder and typing `make html`.  There is some work to do to make the pages look even nicer, but at least they'll render and be visible at docs.lightkurve.org!

Let us know what you think by Friday morning California time =)

ojhall94:
Hey both,

Thanks so much for doing this! The new plots are awesome @christinahedges and really are proper diagnostics now! I gave them some minor tweaks, specifically:
- Made sure legends didn't overlap with text
- Added latex symbols for numax and microhertz
- Added extent arrows on the echelle colourbar and placed it closer to the plot (the former makes sure the colourbar limits don't confuse users, as most of the power is > the 99th percentile. The latter makes me happy :) )
- The colourbar label is now updated if scale='log' to reflect that decision.
- I've added a 'smooth_filter_width' kwarg to 'plot_echelle()' which, if given a value, runs 'butler.periodogram.smooth(filter_width=smooth_filter_width)' before plotting the echelle diagram. This is less than ideal, but I'll need this functionality for the talk (its pretty hard to see stuff on an unsmoothed echelle diagram), and its functionality people will look for. We can (and should) iterate on the best way to implement this at a later date.

Also, Martin, Warrick and I discovered a flaw in the ACF approach. When the thing we're autocorrelating has a mean that is not 0, the ACF contains noise that falls off as 1/x! The `autocorrelate()` util now subtracts the mean of the chunk it is about to autocorrelate, removing noise from the ACFs and making them look much nicer (and protecting me from difficult questions). I've explained the justification in all 3 docstrings also.

My talk notebooks run through, everything looks ready to merge!

barentsen:
@ojhall94 I think having a `smooth`-related keyword in `plot_echelle` that triggers `self.periodogram.smooth` is absolutely fine!  (Why is it less than ideal?)  If you say it's hard to see anything on an unsmoothed diagram, you may want to think about whether this smoothing should happen by default within the plot function.

One of the tests seemed to fail for a weird reason.  Re-running now!

ojhall94:
I guess less than ideal because its only passing one kwarg on to periodogram.smooth(), but it has all the functionality it needs so I suppose its ideal in that sense! :)

Found a bug while running implementation tests, fixed it.

barentsen:
@ojhall94 No worries, all looks good!

@christinahedges I plan to merge this and build a release of Python v1.1.0 around lunchtime.  Let me know any concerns before then!

ojhall94:
Implementation tests on Kepler, K2 and TESS data all work a dream! Good to go from a science perspective, imo.

barentsen:
I intend to merge this PR in a few moments!

As a result of a discussion we had in the office today, we decided to rename `SeismologyButler` into `Seismology` to avoid having to explain to users why it is called `Butler`.  Otherwise there are no big changes to report relative to yesterday, other than docstring improvements etc.

barentsen:
One additional change: we though it may be best to use `acf2d` as the name of the method currently offered, to leave the door open for alternative methods that utilize the auto-correlation function to be added in the future.  (The user doesn't really need to specify the method right now, so we can still revise this in the coming weeks.)

barentsen:
@ojhall94 Thanks so much for working on this!  We just released Lightkurve v1.1.0 to pypi which includes this PR.

You may also want to have a look at #548 which was created by @christinahedges today and provides a `LightCurveCollection.stitch()` method as well as an improved `LightCurve.fill_gaps()` implementation, both to aid seismology!

ojhall94:
Awesome, thanks for making me aware of these! Those streamline our presentation loads!

PR #502, opened 2019-04-24
Fix estimate centroid bug

christinahedges:
There's a bug in `tpf.create_threshold_mask()` where the supposed centroid pixel is not actually in the center. We had the dimensions the wrong way around.

I've fixed it

# Pre fix:
image

# Post fix:
image


christinahedges:
@barentsen travis just timed out, if you're happy we can merge this

barentsen:
Travis passed after a restart.  Thank you @christinahedges! :+1:

PR #505, opened 2019-04-27
[Bugfix] Carefully handle TIC / EPIC ID overlap

nksaunders:
This PR addresses a couple of issues. First, @christinahedges pointed out that certain searches returned overlapping targets and claimed a distance of 0 between them.

**Example:**
image

Now, the `_query_mast()` function parses the ID from queries beginning in `EPIC` or `KIC`, and assigns them their `kplr...` or `ktwo...` names. These queries now resolve to the correct objects.

**🚨New bug alert 🚨**
While fixing this issue, I found that there are targets observed by TESS with overlapping TIC and EPIC IDs. Depending on the syntax of the search, either one can be returned.

**Example:**
Screen Shot 2019-04-26 at 5 29 58 PM

**Next step:**
Provide a warning when a target ID is within the overlap range of TIC and EPIC. Also, set the default formatting to `TIC #` when an int is passed into `search_tesscut()`.

### DIFF
--- a/lightkurve/search.py
+++ b/lightkurve/search.py
@@ -616,15 +616,9 @@ def _search_products(target, radius=None, filetype="Lightcurve", cadence='long',
     -------
     SearchResult : :class:`SearchResult` object.
     """
-    if isinstance(target, int):
-        if (0 < target) and (target < 13161030):
-            log.warning("Warning: {} may refer to a different Kepler or TESS target. "
-                        "Please add the prefix 'KIC' or 'TIC' to disambiguate."
-                        "".format(target))
-        elif (0 < 200000000) and (target < 251813739):
-            log.warning("Warning: {} may refer to a different K2 or TESS target. "
-                        "Please add the prefix 'EPIC' or 'TIC' to disambiguate."
-                        "".format(target))
+    if isinstance(target, int) or (isinstance(target, str) and target.isdigit()):
+        log.warning("Warning: {} is ambiguous. Please add a catalog prefix "
+                    "such as 'KIC', 'TIC', or 'EPIC'.".format(target))
 
     observations = _query_mast(target, project=mission, radius=radius)
     log.debug("MAST found {} observations. "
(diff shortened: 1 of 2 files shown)


nksaunders:
@barentsen said he will provide the EPIC ID range, which we can compare to the TIC ID range to constrain the possible overlapping targets. This will help us provide the warning message to targets which only have alternate observations with the same ID number.

barentsen:
* The full range of target identifiers for *observed* Kepler targets is: [757076,	100004300].
* The full range of target identifiers for *observed* K2 targets is: [200000811, 251816811]. 

 I obtained these ranges using the following [MAST CasJobs](http://mastweb.stsci.edu/mcasjobs) queries:
* `SELECT MIN(sci_kepler_id), MAX(sci_kepler_id) FROM kepler_science WHERE sci_archive_class = 'TPL';`
* `SELECT MIN(sci_kepler_id), MAX(sci_kepler_id) FROM k2_science WHERE sci_archive_class = 'KTL';`

lightkurve/search.py
```diff
@@ -711,12 +711,16 @@ def _query_mast(target, radius=None, project=('Kepler', 'K2', 'TESS')):
         # `target_name` under which MAST will know the object to prevent
         # source confusion (see GitHub issue #148).
         target = int(target)
```
barentsen: Problem: if `target` is a string with epic/kic prefix, the line above will raise a `ValueError` and the code below will be skipped until the end of the `except` block we're in.  This is obscure but intentional.

lightkurve/tests/test_search.py
```diff
@@ -295,3 +295,14 @@ def test_indexerror_631():
     # This previously triggered an exception:
     result = search_lightcurvefile("KIC 8462852", sector=15)
     assert len(result) == 1
+    len(search) == 0
```
barentsen: Codacy Issue found: [undefined name 'search'](https://app.codacy.com/app/barentsen/lightkurve/pullRequest?prid=3940605)

barentsen: Codacy Issue found: [Expression "len(search) == 0" is assigned to nothing](https://app.codacy.com/app/barentsen/lightkurve/pullRequest?prid=3940605)

barentsen:
I pushed a commit which refined the solution.

If we merge this PR, calls like `lk.search_lightcurvefile("EPIC 203348744")` will directly search the K2 data archive for that exact target ID rather than resolving to coordinates, which solves the problem of the overlapping target being returned instead ("EPIC 203348270").

This is good, :rotating_light:  HOWEVER I ALSO NOTICED A NEW ISSUE :rotating_light: 

...if we merge this PR, calls like `lk.search_lightcurvefile("KIC 8462852")` will ONLY return Kepler data.  This is a problem because `KIC 8462852` has now been observed by TESS, and we can only show the TESS results to the user if we resolve the target identifier to coordinates.  That's bad! =(

barentsen:
The only good solution I can currently think of to resolve this issue is to add a new keyword, e.g. `resolve`, to let a user determine whether or not the target string will be resolved to coordinates.

For example: 
* `lk.search_lightcurvefile("EPIC 203348744", resolve=False)`
  :arrow_right: this will ONLY return K2 observations made using the mask identifier "203348744"  (i.e. this translates to a MAST portal query on `target_name='ktwo203348744'`).
*  `lk.search_lightcurvefile("EPIC 203348744", resolve=True)`
  :arrow_right: this will resolve the coordinates of the target string to (ra, dec) positions and return ALL Kepler/K2/TESS observations within the search radius around those coordinates.

The default would be `resolve=True`.

barentsen:
This should now have been resolved by #796  :+1:

Issue #507, opened 2019-05-03
Unit tests failing because C11 targets are duplicated in the MAST Portal


barentsen:
Since the recent release of re-processed K2 Campaign 11 data, targets observed during this campaign now appear to have duplicate entries in the MAST portal, which is likely by accident:

![Screenshot from 2019-05-03 11-29-53](https://user-images.githubusercontent.com/817669/57157918-cf2f0d00-6d96-11e9-85b2-14d00c666f77.png)

This is currently causing our `test_search_targetpixelfile` unit test [to fail](https://dev.azure.com/KeplerGO/Lightkurve/_build/results?buildId=43&view=logs&jobId=e3af9475-6656-5c3e-fbc6-269fb81d3f5f&taskId=f61e0797-20a9-5608-7aa4-45ddff81691d&lineStart=40&lineEnd=60&colStart=2&colEnd=1).

I will follow-up offline with @JeffLCoughlin.

barentsen:
This issue appears to have been resolved, as evidenced by the fact that the failing unit test (cf. #508) is passing again. Closing!

JeffLCoughlin:
MAST confirmed this morning they fixed the issue. Hooray.

PR #508, opened 2019-05-03
Restore the unit test that was failing due to #507

barentsen:
A few moments ago I disabled the unit test that was failing due to the minor issue with C11 data at MAST described in #507 (commit: 211f134a969756e6935b078e2930013decef1a67).

This PR restores the unit test that has been temporarily disabled.  We should merge this PR as soon as #507 is resolved, i.e. as soon as MAST removed the duplicate entries for C11 data from its API.


barentsen:
This test passes again; #507 appears to have been resolved. :+1: 

Merging in 3 ... 2 ... 1 ...

PR #510, opened 2019-05-10
Configure `pytest` to use all processors by default

barentsen:
I modified the `pytest` configuration to run the unit tests in parallel by default, using as many processors as available.  This significantly speeds up running the unit tests on a local multi-core machine.

I'm opening this PR to see if CI (Azure) is ok with this change!

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,4 +2,4 @@
 test=pytest
 
 [tool:pytest]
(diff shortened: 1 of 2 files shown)


barentsen:
Interesting!  Some of the CI jobs fail.  I wonder if there is a concurrency problem which needs fixing inside Lightkurve.

PR #513, opened 2019-05-22
Improve the treatment of the `view` parameter in Periodograms

barentsen:
This PR suggests two very minor changes to the `periodogram` module:
1. Ensure that the `view` parameter is validated in a consistent way in both the `Periodogram.__init__()` and `Periodogram.plot()` methods by introducing a single `_validate_view()` method.
2. When `period`, `minimum_period`, or `maximum_period` are passed to `LombScarglePeriodogram.from_lightcurve()`, I propose we set `default_view=period`.  The resulting change in behavior is shown below.

## Old behavior
Screenshot from 2019-05-22 11-25-14

## New behavior
Screenshot from 2019-05-22 11-26-14

@ojhall94 Would you be willing to review?

### DIFF
--- a/lightkurve/periodogram.py
+++ b/lightkurve/periodogram.py
@@ -85,8 +85,8 @@ class Periodogram(object):
             view = self.default_view
         allowed_views = ["frequency", "period"]
         if view not in allowed_views:
-            raise ValueError(("'{}' is an invalid value for view; \n"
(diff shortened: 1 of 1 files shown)


barentsen:
Thank you @ojhall94!!  Please have this medal: 🥇

PR #514, opened 2019-05-22
Bugfix: BLS periodogram was ignoring the `period` parameter

barentsen:
This PR fixes a bug in `BoxLeastSquaresPeriodogram` which causes the `period` parameter to be ignored, caused by the fact that `kwargs.pop("period")` was called multiple times.  A regression test is included as well.

### Example

```python
>>> lc = lk.LightCurve(time=[1, 2, 3], flux=[4, 5, 6])
>>> pg = lc.to_periodogram(method="bls", period=[1, 2, 3, 4, 5])
>>> pg.period
[0.66666667, 1.1428571, 4]d
```

The behavior expected above is for `pg.period` to return `[1, 2, 3, 4, 5]`.

### DIFF
--- a/lightkurve/periodogram.py
+++ b/lightkurve/periodogram.py
@@ -863,27 +863,30 @@ class BoxLeastSquaresPeriodogram(Periodogram):
     @staticmethod
     def from_lightcurve(lc, **kwargs):
         """Creates a Periodogram from a LightCurve using the Box Least Squares (BLS) method."""
-        time_unit = (kwargs.pop("time_unit", "day"))
-        if time_unit not in dir(u):
-            raise ValueError('{} is not a valid unit for time.'.format(time_unit))
-
         try:
             from astropy.stats import BoxLeastSquares
         except ImportError:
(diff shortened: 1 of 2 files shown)


barentsen:
@christinahedges Would you be willing to review and merge if the tests pass?

lightkurve/periodogram.py
```diff
@@ -885,16 +885,16 @@ def from_lightcurve(lc, **kwargs):
...
-                minimum_period = period.min()
-            else:
+            if period is None:
                 minimum_period = np.max([np.median(np.diff(lc.time)) * 4,
                                          np.max(duration) + np.median(np.diff(lc.time))])
-        if maximum_period is None:
-            if 'period' in kwargs:
-                maximum_period = period.max()
             else:
+                minimum_period = np.min(period)
```
christinahedges: You should catch nans somewhere in period

barentsen: Good point!  I re-arranged and edited the user input validation steps now to catch this, and added it to the unit test (c6b0630).

PR #516, opened 2019-05-23
Ensure all important docstrings have an `Examples` section

barentsen:
To further improve Lightkurve's documentation resources, I think it is important that we start to make sure that all popular methods and classes include an `Examples` section in their docstring. Examples included in this way will automatically be rendered as part of our [online API docs](https://docs.lightkurve.org/api) and make the package more accessible.

This PR starts this effort and can serve as an example for other PRs.


barentsen:
There is still more work to do to improve examples (any help is welcome!) -- but I'm going to go and merge this PR so that Lightkurve v1.1 can benefit from improved docs!

PR #522, opened 2019-05-28
Add separate aperture for design matrix construction to PLDCorrector

nksaunders:
When applying the `PLDCorrector`, it is sometimes desired to use one aperture mask to generate the light curve and a separate aperture mask to select pixels to use as basis vectors in the PLD design matrix. `PLDCorrector.correct()` now takes two aperture mask arguments:

 - `aperture_mask`: pixels used in aperture photometry to generate the raw flux light curve
- `pld_aperture`: pixels used when selecting PLD basis vectors

Both of these can take the default lightkurve aperture_mask options: `array-like, 'pipeline', 'all', 'threshold', or None`. 

Thank you @christinahedges for this feature request! Do you think the `aperture_mask` argument should be renamed? I'm hesitant to veer away from the naming convention we use in other functions. Feedback welcome!

### DIFF
--- a/lightkurve/correctors.py
+++ b/lightkurve/correctors.py
@@ -594,14 +594,14 @@ class PLDCorrector(object):
         self.time = tpf.time
 
     def correct(self, aperture_mask=None, cadence_mask=None, gp_timescale=30,
-                use_gp=True, pld_order=2, n_pca_terms=10):
+                use_gp=True, pld_order=2, n_pca_terms=10, pld_aperture=None):
         r"""Returns a PLD systematics-corrected LightCurve.
 
         Parameters
         ----------
         aperture_mask : array-like, 'pipeline', 'all', 'threshold', or None
             A boolean array describing the aperture such that `True` means
-            that the pixel will be used.
+            that the pixel will be used to generate the raw light curve.
             If `None` or 'all' are passed, all pixels will be used.
             If 'pipeline' is passed, the mask suggested by the official pipeline
             will be returned.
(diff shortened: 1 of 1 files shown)


barentsen:
To be more consistent with `aperture_mask`, I wonder if we should call the second parameter `pld_aperture_mask`.

lightkurve/correctors.py
```diff
@@ -672,15 +680,18 @@ def correct(self, aperture_mask=None, cadence_mask=None, gp_timescale=30,
         self.flux_err = self.flux_err[nanmask]
         self.time = self.time[nanmask]
 
+        # parse PLD aperture mask
+        pld_pixels = self.tpf._parse_aperture_mask(pld_aperture)
```
barentsen: :point_up: It may be best to consistently use the suffix `_mask` in all variables that contain a ... mask.

barentsen: (A reader of the code may be tempted to think that `pld_pixels` contain pixel values otherwise.)

barentsen:
I left some thoughts above.  Two more items of feedback are:
* A PR re-factoring `correctors.py` into a sub-package has just been merged (#519).  It will be hard to resolve the merging conflict in this PR, so it may be easiest to pull keplergo/master and start a new branch from master.
* Let's add a unit test, even if all it does is pass some mask to the new parameter.

nksaunders:
@barentsen got it 👍 closing and reopening this PR to avoid merge conflicts.

PR #523, opened 2019-05-28
Add separate aperture for design matrix construction to PLDCorrector

nksaunders:
_(Reopened from #522 to avoid merge conflicts)_

When applying the `PLDCorrector`, it is sometimes desired to use one aperture mask to generate the light curve and a separate aperture mask to select pixels to use as basis vectors in the PLD design matrix. `PLDCorrector.correct()` now takes two aperture mask arguments:

 - `aperture_mask`: pixels used in aperture photometry to generate the raw flux light curve
 - `pld_aperture_mask`: pixels used when selecting PLD basis vectors
Both of these can take the default lightkurve aperture_mask options: `array-like, 'pipeline', 'all', 'threshold', or None`.

Thank you @christinahedges for this feature request!

### DIFF
--- a/lightkurve/tests/test_correctors.py
+++ b/lightkurve/tests/test_correctors.py
@@ -150,3 +150,19 @@ def test_to_corrector():
     tpf = KeplerTargetPixelFile(TABBY_TPF)
     lc = tpf.to_corrector("pld").correct()
     assert len(lc.flux) == len(tpf.time)
+
+
+@pytest.mark.remote_data
+@pytest.mark.skipif(bad_optional_imports, reason="PLD requires celerite and fbpca")
+def test_pld_aperture_mask():
+    """Regression test for #523: does PLDCorrector.correct() accept separate
+    apertures for PLD pixels?"""
+    from .. import KeplerTargetPixelFile
+    from .test_targetpixelfile import TABBY_TPF
+    tpf = KeplerTargetPixelFile(TABBY_TPF)
+    # use only the pixels in the pipeline mask
+    lc_pipeline = tpf.to_corrector("pld").correct(pld_aperture_mask='pipeline')
+    # use all pixels in the tpf
+    lc_all = tpf.to_corrector("pld").correct(pld_aperture_mask='all')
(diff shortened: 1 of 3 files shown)


nksaunders:
@barentsen I believe fixed the broken TESS portion of `test_pld_corrector()` introduced by the TIC renaming. Could you look at that?

lightkurve/tests/test_correctors.py
```diff
@@ -150,3 +151,19 @@ def test_to_corrector():
     tpf = KeplerTargetPixelFile(TABBY_TPF)
     lc = tpf.to_corrector("pld").correct()
     assert len(lc.flux) == len(tpf.time)
+
+
+@pytest.mark.remote_data
+@pytest.mark.skipif(bad_optional_imports, reason="PLD requires celerite and fbpca")
+def test_pld_aperture_mask():
+    """Regression test for #523: does PLDCorrector.correct() accept separate
```
barentsen: Note: most people wouldn't call this a regression test because there hasn't been a regression i.e. bug here, this is simply a new feature.

lightkurve/tests/test_correctors.py
```diff
@@ -150,3 +151,19 @@ def test_to_corrector():
...
+    PLD pixels?"""
+    from .. import KeplerTargetPixelFile
+    from .test_targetpixelfile import TABBY_TPF
+    tpf = KeplerTargetPixelFile(TABBY_TPF)
+    # use only the pixels in the pipeline mask
+    lc_pipeline = tpf.to_corrector("pld").correct(pld_aperture_mask='pipeline')
+    # use all pixels in the tpf
+    lc_all = tpf.to_corrector("pld").correct(pld_aperture_mask='all')
+    # does this improve the correction?
+    assert(lc_all.estimate_cdpp() < lc_pipeline.estimate_cdpp())
```
barentsen: I like this sanity check/test!

barentsen:
This all looks excellent.  Great job!  Merging in 3 ... 2 ... 1 ...

PR #525, opened 2019-06-02
Refactor `PLDCorrector` to be easier to test and modify

barentsen:
This PR proposes to refactor `PLDCorrector` to be more modular, i.e. to provide more atomic methods which can be tested and modified more easily.  For example, by exposing `create_design_matrix` and `create_pymc_model` as separate methods, we will be able to test and modify these operations more easily for different experiments.

The refactor is based on the existing code found in both `lightkurve.correctors.PLDCorrector` and in `happyplanets` (https://github.com/nksaunders/happyplanets/blob/master/happyplanets/correct.py).


nksaunders:
In testing, I found that creating the 3rd order PLD design matrix for target `K2-105` times out, while 1st and 2nd order run fine. This is surprising as the pipeline aperture is not unusually large, and I can't identify a clear source of the issue. 

I'm putting this in writing so we can revisit the issue.

nksaunders:
@barentsen we need to think about where we pass in the aperture mask. The current implementation seems to prohibit the use of `to_corrector("pld")`, because there isn't a clear way to define the aperture in the returned `Corrector` object. Let's try to pass it into `.correct()` as with the previous PLD module.

nksaunders:
This line of code fails for Tabby's star:
```
PyMCPLDCorrector(tpf, pld_aperture_mask='pipeline').correct(gp_timescale=150)
```
@barentsen and I believe this is due to a `gp_timescale` value too large to be flexible enough to fit the large dips. Changing the value to `gp_timescale=30` runs fine.

Potential fixes:
 - Better outlier rejection
 - Different default `gp_timescale` value
 - Catch error when exoplanet is unable to optimize GP hyperparameters

barentsen:
I am going to record here our unit tests encountered the following Theano failure on the Azure pipeline Ubuntu 16.04 images:

```
E   ImportError: Version check of the existing lazylinker compiled file. Looking for version 0.211, but found None. Extra debug information: force_compile=False, _need_reload=True
```
which was in turn caused by:
```
E   Exception: Compilation failed (return status=1): /usr/bin/ld: cannot find -lpython3.6m. collect2: error: ld returned 1 exit status.
```

I believe this may be related due to an issue in the Azure-pipelines `UsePythonVersion@0` task.  My speculation is that it may e.g. not correctly be updating the `LIBRARY_PATH` when using Python versions other than the standard system Python provided by the Azure Ubuntu 16.04 image.

Rather than attempting to resolve this issue, I decided to switch to using Anaconda Python (cf. #528), which resolved the issue.

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+        success, messages = self._check_optional_dependencies()
+        if not success:
+            for message in messages:
+                log.error(message)
+            raise ImportError("\n".join(messages))
+
+        # Input validation: parse the aperture masks to accept strings etc.
+        self.aperture_mask = tpf._parse_aperture_mask(aperture_mask)
+        if pld_aperture_mask is None:
+            self.pld_aperture_mask = self.aperture_mask
```
christinahedges: Do you think the default should be `'all'`?

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+        else:
+            self.pld_aperture_mask = tpf._parse_aperture_mask(pld_aperture_mask)
+        # Generate raw flux light curve from desired pixels
+        raw_lc = tpf.to_lightcurve(aperture_mask=self.aperture_mask)
+        # It is critical to remove all NaNs or the linear algebra below will crash
+        self.raw_lc, self.nan_mask = raw_lc.remove_nans(return_mask=True)
+        self.tpf = tpf[~self.nan_mask]
```
christinahedges: This won't remove nan pixels, only nan cadences. Might add that in a comment.

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
-        self.time = tpf.time
+        try:
+            import theano.tensor as tt
+        except ImportError:
+            success = False
+            messages.append("PLDCorrector requires theano to be installed (`pip install theano`).")
+
+        return success, messages
+
+    def create_first_order_matrix(self):
```
christinahedges: Does this need to be a user facing function? Should this be `_create_first_order_matrix`?

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+        matrix = matrix / np.sum(matrix, axis=-1)[:, None]
+        # If we return matrix at this point, theano will raise a "dimension mismatch".
+        # The origin of this bug is not understood, but copying the matrix
+        # into a new one as shown below circumvents it:
+        result = np.empty((matrix.shape[0], matrix.shape[1]))
+        result[:, :] = matrix[:, :]
+
+        return result
+
+    def create_design_matrix(self, pld_order=1, n_pca_terms=10,
```
christinahedges: Same as above, do users ever need this function?

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+            use_fbpca = False
+            log.warning("PLD systematics correction will run faster if the "
+                        "optional `fbpca` package is installed "
+                        "(`pip install fbpca`).")
+
+        matrix_sections = []  # list to hold the design matrix components
+        first_order_matrix = self.create_first_order_matrix()
+
+        # Input validation: n_pca_terms cannot be larger than the number of cadences
+        n_cadences = len(first_order_matrix)
```
christinahedges: The user never specifies this number, but gets a warning if it's too small. This seems annoying. Can we instead just use

`n_cadences = np.min([len(first_order_matrix), n_cadences])`

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+            log.warning("`n_pca_terms` ({}) cannot be larger than the number of cadences ({});"
+                        "using n_pca_terms={}".format(n_pca_terms, n_cadences, n_cadences))
+            n_pca_terms = n_cadences
+
+        # The original EVEREST paper includes a column vector of ones in the
+        # design matrix to improve the numerical stability (see Luger et al.);
+        # it is unclear whether this is necessary, so this is an optional step
+        # for now.
+        if include_column_of_ones:
+            matrix_sections.append([np.ones((len(first_order_matrix), 1))])
```
christinahedges: Why isn't the array of ones added as the last step? Why is it 1st order, column of ones, 2nd order?

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+            if use_fbpca:  # fast mode
+                components, _, _ = pca(matrix, n_pca_terms)
+            else:  # slow mode
+                components, _, _ = np.linalg.svd(matrix)
+            section = components[:, :n_pca_terms]
+            matrix_sections.append(section)
+
+        return np.concatenate(matrix_sections, axis=1)
+
+    def create_pymc_model(self, design_matrix=None, cadence_mask=None,
```
christinahedges: I can pass in an arbitrary design matrix here. This is super. If I understand correctly, in theory then I can pass in e.g. the CBVs, and get out the best correction.

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+
+        # The cadence mask is applied by inflating the uncertainties in the covariance matrix;
+        # this is because celerite will run much faster if it is able to predict
+        # data for cadences that have been fed into the model.
+        diag[~cadence_mask] += 1e12
+
+        with pm.Model() as model:
+            # Create a Gaussian Process to model the long-term stellar variability
+            # log(sigma) is the amplitude of variability, estimated from the raw flux scatter
+            logsigma = pm.Normal("logsigma", mu=np.log(np.std(lc_flux)), sd=4)
```
christinahedges: The sd is hard coded here...I think I'm happy with that but it would make a difficult error to debug if your priors were very off.

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+            # log(rho) is the timescale of variability with a user-defined prior
+            logrho = pm.Normal("logrho", mu=np.log(gp_timescale_prior), sd=4)
+            # log(s2) is a jitter term to compensate for underestimated flux errors
+            logs2 = pm.Normal("logs2", mu=np.log(np.var(lc_flux)), sd=4)
```
christinahedges: This jitter term seems wrong. This should be the amount of white-ish noise in the light curve. If your don't flatten out long term trends first, you'll get the wrong answer. I think it's better to set this to roughly the CDPP of the light curve.

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+            B = tt.dot(design_matrix.T, model.gp.apply_inverse(lc_flux[:, None]))
+            weights = tt.slinalg.solve(A, B)
+            motion_model = pm.Deterministic("motion_model", tt.dot(design_matrix, weights)[:, 0])
+            pm.Deterministic("weights", weights)
+
+            # Likelihood to optimize
+            pm.Potential("obs", model.gp.log_likelihood(lc_flux - motion_model))
+
+            # Track the flux values of the different diagnostic light curves for ease of use
+            gp_model, gp_model_var = model.gp.predict(return_var=True)
```
christinahedges: The `predict` step here is pretty costly, and I don't think you're actually using it for anything. I think it would be good to try removing it, and just returning the gp object itself, or the best fit hyper parameters, and predicting only once at the end, outside of the model.

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+        with model:
+            solution = xo.optimize(start=start, vars=[model.logs2])
+            # Optimizing parameters separately appears to make finding a solution more likely
+            if robust:
+                solution = xo.optimize(start=solution, vars=[model.logsigma])
+                solution = xo.optimize(start=solution, vars=[model.logrho, model.logsigma])
+                solution = xo.optimize(start=solution, vars=[model.logs2])
+            solution = xo.optimize(start=solution)  # Optimize all parameters
+
+        self.most_recent_solution_or_trace = solution
```
christinahedges: Should there be `most_recent_solution` and `most_recent_trace`? The two are quite different.

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+
+    def correct(self, sample=False, remove_gp_trend=False, **kwargs):
+        """Returns a systematics-corrected light curve.
+
+        Parameters
+        ----------
+        sample : boolean
+            `True` will sample the output of the optimization
+            step and include robust errors on the output light curve.
+        remove_gp_trend : boolean
```
christinahedges: Let's make sure these keywords have easy to understand names, that are consistent between correctors. Since we will also be having a GP corrector, could this be `remove_long_term_trends` or similar? Don't want people to get confused.

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+            Light curve object corresponding to the model variable.
+        """
+        lc = self.raw_lc.copy()
+        # If the model was sampled, use the mean and std of the flux values.
+        if isinstance(solution_or_trace, pm.backends.base.MultiTrace):
+            lc.flux = np.nanmean(solution_or_trace[variable], axis=0)
+            lc.flux_err = np.nanstd(solution_or_trace[variable], axis=0)
+        else:  # Otherwise, use the maximum likelihood flux values.
+            lc.flux = solution_or_trace[variable]
+        return lc
```
christinahedges: Can you also change the `label` attribute to "PLD Corrected targetid", so that we can tell the difference? Also, you could add the motion model to the `meta` attribute, so that we can keep hold of it and use it down the line.

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+        if model is None:
+            model = self.create_pymc_model(**kwargs)
+        if start is None:
+            start = model.test_point
+
+        with model:
+            solution = xo.optimize(start=start, vars=[model.logs2])
+            # Optimizing parameters separately appears to make finding a solution more likely
+            if robust:
+                solution = xo.optimize(start=solution, vars=[model.logsigma])
```
christinahedges: These seem to be in the wrong order it should probably be

`logrho`
`logrho, logsigma`
`logrho, logsigma, logs2`

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+                         step_kwargs=dict(target_accept=0.9),
+                         chains=chains)
+            # Sample the parameters
+            trace = sampler.sample(draws=draws, chains=chains)
+
+        self.most_recent_solution_or_trace = trace
+        return trace
+
+    def correct(self, sample=False, remove_gp_trend=False, **kwargs):
+        """Returns a systematics-corrected light curve.
```
christinahedges: Let's put all the **kwargs in the docstring of the `correct` function. e.g. it's hard to find where to put in `pld_order`.

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+    .. [6] Exoplanet documentation, https://exoplanet.readthedocs.io/en/stable/
+    .. [7] PyMC3 documentation, https://docs.pymc.io
+
+    Parameters
+    ----------
+    tpf : `TargetPixelFile` object
+        The pixel data from which a light curve will be extracted.
+    aperture_mask : 2D boolean array or str
+        The pixel aperture mask that will be used to extract the raw light curve.
+    pld_aperture_mask : 2D boolean array or str
```
christinahedges: This keyword is not very descriptive. How about `design_matrix_mask`.

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
-            If 'threshold' is passed, all pixels brighter than 3-sigma above
-            the median flux will be used.
+        design_matrix : np.ndarray
+            Matrix of shape (n_cadences, n_regressors) used to create the
+            motion model.  If `None`, then the output of this object's
+            `create_design_matrix` method will be used.
+        cadence_mask : np.ndarray
+            Boolean array to mask cadences. Cadences that are False will be excluded
+            from the model fit.  If `None`, then all cadences will be used.
+        gp_timescale_prior : int
```
christinahedges: Why does this have to be an int? Prior in what? Prior in days or cadences?

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+        with pm.Model() as model:
+            # Create a Gaussian Process to model the long-term stellar variability
+            # log(sigma) is the amplitude of variability, estimated from the raw flux scatter
+            logsigma = pm.Normal("logsigma", mu=np.log(np.std(lc_flux)), sd=4)
+            # log(rho) is the timescale of variability with a user-defined prior
+            logrho = pm.Normal("logrho", mu=np.log(gp_timescale_prior), sd=4)
+            # log(s2) is a jitter term to compensate for underestimated flux errors
+            logs2 = pm.Normal("logs2", mu=np.log(np.var(lc_flux)), sd=4)
+            kernel = xo.gp.terms.Matern32Term(log_sigma=logsigma, log_rho=logrho)
+
```
christinahedges: Add the mean as a parameter, otherwise you will end up with lots of funny negatives as discussed in office.

It will go super negative.

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,199 +116,509 @@ class PLDCorrector(object):
...
+            # multiply all pairs of pixels, order=3 will multiple triples, etc.
+            matrix = np.product(list(multichoose(first_order_matrix.T, order)), axis=1).T
+            # This product matrix becomes very big very quickly, so we reduce
+            # its dimensionality using PCA.
+            if use_fbpca:  # fast mode
+                components, _, _ = pca(matrix, n_pca_terms)
+            else:  # slow mode
+                components, _, _ = np.linalg.svd(matrix)
+            section = components[:, :n_pca_terms]
+            matrix_sections.append(section)
```
christinahedges: Make sure your nth order design matrix columns are normalized.

lightkurve/correctors/pldcorrector.py
nksaunders: I'm making this a hidden function, but leaving `create_design_matrix` so users can more easily edit design matrices and append new columns. Calling `create_design_matrix(pld_order=1)` has the same effect as calling this function.

lightkurve/correctors/pldcorrector.py
nksaunders: This check should actually ensure that `n_pca_terms` is not greater than the number of columns in the design matrix, so I've changed the name to `n_pixels`. This warning will only appear if the user passes in a very small PLD aperture mask (fewer than 4 pixels), which I don't think will be very common, and it prevents the code from breaking during PLD.

Do you think we should only give the warning if the user specifies `n_pca_terms`?

lightkurve/correctors/pldcorrector.py
nksaunders: That makes it sound like we're masking the design matrix itself instead of using an aperture mask on pixels. Maybe `design_matrix_aperture_mask`? It's longer but more descriptive. I'll change it to that for now, let me know if you have other suggestions.

lightkurve/correctors/pldcorrector.py
nksaunders: 👍👍👍

lightkurve/correctors/pldcorrector.py
nksaunders: After some experimentation this afternoon, it seems like this step is unnecessary, so I've removed the column of ones. I think this is because we're simultaneously fitting the GP and the mean along with the PLD coefficients, which removes the need to include the assumption that the light curve is otherwise "flat" without the GP trend.

nksaunders: Ok, I take this back. I found a case where the column of ones prevents a crash:

When the design matrix is small in both dimensions, meaning few cadences (<~500) and few column vectors (<~25). One of the tests fit this description and started failing when I removed the column of ones. 

The error that Theano raises is very opaque, so I've created a `LightkurveError` class and raise that with a more informative warning in this case. @barentsen please weigh in on this choice (see the `optimize` call in the `correct` function).

I'm putting the option to include the column of ones back in, but as the last column (though this doesn't seem to make a difference).

lightkurve/correctors/pldcorrector.py
```diff
@@ -82,43 +117,140 @@ class PLDCorrector(object):
...
+            for message in messages:
+                log.error(message)
+            raise ImportError("\n".join(messages))
+
+        # Input validation: parse the aperture masks to accept strings etc.
+        self.aperture_mask = tpf._parse_aperture_mask(aperture_mask)
+        if design_matrix_aperture_mask is None:
+            self.design_matrix_aperture_mask = tpf._parse_aperture_mask('all')
+        else:
+            self.design_matrix_aperture_mask = tpf._parse_aperture_mask(design_matrix_aperture_mask)
```
barentsen: I recommend simplifying lines 150-153 as follows:
```python
if design_matrix_aperture_mask is None:
    design_matrix_aperture_mask = 'all'
self.design_matrix_aperture_mask = tpf._parse_aperture_mask(design_matrix_aperture_mask)
```

nksaunders:
Now that we're correctly normalizing the higher order PLD vectors, applying PLD to TESS observations no longer meaningfully improves the precision of the corrected light curve. This tells me that we may have been overfitting up to this point by finding false correlations between the higher order basis vectors. I removed the tests for TESS targets that checked for an improvement in CDPP, but kept a simple test to make sure it still _can_ be applied to TESS observations.

barentsen:
@nksaunders Would you be willing to post the output of `plot_diagnostics` for the TESS target you mention?

nksaunders:
> @nksaunders Would you be willing to post the output of `plot_diagnostics` for the TESS target you mention?

I take back what I said... I was able to get it working well for TESS targets again. The problem seems to be that the matrix `A` was _always_ ill-conditioned for TESS observations (maybe due to the short cadence?), and it was necessary to implement some sort of regularization.

After struggling with computationally heavy versions of L2 normalization, I was able to figure out a quick method for ridge regression. I create a tensor array with the indices of the diagonal of `A`, and set a new subtensor with small numbers added along the diagonal. This is quick, and has completely removed the *very* verbose warning about ill-conditioned matrices during the `tt.slinalg.solve` step.

barentsen:
@nksaunders wow!

lightkurve/correctors/pldcorrector.py
```diff
@@ -129,152 +262,543 @@ def correct(self, aperture_mask=None, cadence_mask=None, gp_timescale=30,
...
+            # log(sigma) is the amplitude of variability, estimated from the raw flux scatter
+            logsigma = pm.Normal("logsigma", mu=np.log(np.std(lc_flux)), sd=5)
+            # log(rho) is the timescale of variability with a user-defined prior
+            logrho = pm.Normal("logrho", mu=np.log(gp_timescale_prior),
+                               sd=np.log(fractional_prior_width*gp_timescale_prior))
+            # Enforce that the scale of variability should be no shorter than 0.5 days
+            pm.Potential("logrho_prior", tt.switch(logrho > np.log(0.5), 0, np.inf))
+            # log(s2) is a jitter term to compensate for underestimated flux errors
+            # We estimate the magnitude of jitter from the CDPP (normalized to the flux)
+            s2_mu = self.raw_lc.estimate_cdpp() * 1e-6 * mean
```
christinahedges:
```suggestion
            s2_mu = self.raw_lc.estimate_cdpp() * 1e-6 * np.median(self.raw_lc.flux)
```

I'm finding that having the cyclic "`mean` is a distribution that also sets the distribution for `logs2`" hard to deal with, and my optimization seems better when the mean of `logs2` is just set to a value.

nksaunders:
After talking to Dan FM and Dan Hey at KASC, I've reverted the priors to broad normals instead of uniform distributions. The hard boundaries in parameter space are costly to optimize, and performance is vastly increased by the new parameterization.

barentsen:
I'm going to close this PR because we'll want to revise `PLDCorrector` to use the new `RegressionCorrector` framework (cf. #613).  The code attached to this PR will continue to remain immensely useful for that purpose! :+1::+1:

PR #528, opened 2019-06-06
[WIP] Modify azure-pipelines continuous integration to use Anaconda Python

barentsen:
Over in #525 we are encountering a tricky theano compilation error which can likely be resolved by using Anaconda Python instead of relying on the Ubuntu 16.04 system Python.

This PR attempts to make that change.

### DIFF
--- a/.azure-template.yml
+++ b/.azure-template.yml
@@ -1,9 +1,23 @@
 steps:
 
-- task: UsePythonVersion@0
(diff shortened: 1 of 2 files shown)


barentsen:
This solution appears to resolve the issue encountered in #525.  I'm closing this PR as I ended up merging this branch into #525.

PR #532, opened 2019-06-14
Allow LightCurve objects to be added and multiplied

barentsen:
## Problem

At present, it is possible to perform math operations involving a `LightCurve` object and a number or array of numbers, e.g.:

```python
>>> from lightkurve import LightCurve
>>> lc = LightCurve(time=[1, 2, 3], flux=[1, 2, 3])
>>> (lc + 5).flux
array([6, 7, 8])
```

However, adding adding or multiplying two LightCurve objects together yields nonsensical results. Example:

```python
>>> lcsum = lc + lc
>>> lcsum.flux
array([<lightkurve.lightcurve.LightCurve object at 0x7f29b06589e8>,
       <lightkurve.lightcurve.LightCurve object at 0x7f29b0658a20>,
       <lightkurve.lightcurve.LightCurve object at 0x7f29b0658cf8>],
      dtype=object)
>>> lcsum.flux[0].flux
array([2, 3, 4])
``` 

## Solution

This PR fixes support for math operations involving two LightCurve objects.  Example:

```python
>>> lcsum = lc + lc
>>> lcsum.flux
array([2, 4, 6])
```

## Notes

* Standard [uncertainty propagation rules](https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Example_formulae) are being applied.
* If the two `LightCurve` objects being added or multiplied together do not have the same length, a `ValueError` will be raised.
* If the two `LightCurve` objects do not have the same `time` values, a `LightkurveWarning` will be raised.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -131,44 +131,81 @@ class LightCurve(object):
         return arr
 
     def __getitem__(self, key):
-        copy_self = copy.copy(self)
+        copy_self = self.copy()
         copy_self.time = self.time[key]
         copy_self.flux = self.flux[key]
         copy_self.flux_err = self.flux_err[key]
         return copy_self
 
+    def __len__(self):
+        return len(self.time)
+
     def __add__(self, other):
-        copy_self = copy.copy(self)
-        copy_self.flux = copy_self.flux + other
-        return copy_self
+        newlc = self.copy()
+        if isinstance(other, LightCurve):
+            if len(self) != len(other):
+                raise ValueError("Cannot add LightCurve objects because "
+                                 "they do not have equal length ({} vs {})."
+                                 "".format(len(self), len(other)))
+            if np.any(self.time != other.time):
+                warnings.warn("Two LightCurve objects with inconsistent time "
+                              "values are being added.")
+            newlc.flux = self.flux + other.flux
+            newlc.flux_err = np.hypot(self.flux_err, other.flux_err)
+        else:
+            newlc.flux = self.flux + other
+        return newlc
 
     def __radd__(self, other):
         return self.__add__(other)
 
     def __sub__(self, other):
-        return self.__add__(-other)
+        return self.__add__(-1 * other)
 
     def __rsub__(self, other):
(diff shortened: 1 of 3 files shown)


lightkurve/lightcurve.py
```diff
@@ -103,44 +103,81 @@ def _validate_array(self, arr, name='array'):
...
+        newlc = self.copy()
+        if isinstance(other, LightCurve):
+            if len(self) != len(other):
+                raise ValueError("Cannot add LightCurve objects because "
+                                 "they do not have equal length ({} vs {})."
+                                 "".format(len(self), len(other)))
+            if np.any(self.time != other.time):
+                warnings.warn("Two LightCurve objects with inconsistent time "
+                              "values are being added.")
+            newlc.flux = self.flux + other.flux
```
christinahedges: Can you add some more catches here? e.g. what if I try to add a `TessLightCurve` and a `KeplerLightCurve`? What if I want to add two light curves with the `metadata` dictionary filled with values? What happens to the attributes of the light curve? (e.g. column, row, centroid)

gully: I'd say make it `np.allclose()` for the time values, since times can be slightly spatially dependent due to the differences in BKJD conversion, and instead of a warning, raise an error.  I think its fair to only support adding of time-aligned lightcurves.  That change will address the Kepler/Tess lightcurves.

The issue with how to handle metadata is tricky!  We could elect to pass on the metadata from only one of the lightcurves.  But which one?  How to deal with the reverse?

gully:
Oops, I didn't actually rebase, make sure it's a squash and merge...

barentsen:
I intend to go ahead and merge this PR, because there are many useful applications for adding/subtracting/multiplying light curves together, and the existing behavior was for such operations to crash with an unclear exception.

An open question was "how to deal with metadata".  There is no generic way to merge inconsistent meta data (e.g. a light curve can only have one unique target label). For now the behavior is that the meta data of the leading operating (e.g. the left hand side of an addition) will be preserved, but we can revisit this in the future if new insights emerge.

Issue #531, opened 2019-06-13
Add a `test_synthetic_data.py` module using scope-generated TPFs


barentsen:
This issue describes a project which I would like @gully to work on when he returns from leave (and I may be gone).

In brief, I would like us to use @nksaunders' excellent [scope package](https://github.com/nksaunders/scope) to generate three synthetic TPF files, and then use those synthetic files to add a new unit test module. Specifically, I propose the following two-phase plan:

## Phase 1: Generate three synthetic TPF files

The first step is to use `scope` to generate three synthetic TPFs with the following requirements:

- The three TPFs shall each contain a single star with K2-like noise and motion.
- The three TPFs shall use different input light curves:
  * `synthetic-k2-flat.targ.fits` shall use a flat light curve;
  * `synthetic-k2-sinusoid.targ.fits` shall use a low-frequency sinusoid (e.g. 5-day period);
  * `synthetic-k2-planet.targ.fits` shall use a short-period Neptune-sized planet (e.g. Kp = 14; planet period = 5 days; transit depth = 2000ppm).
- The TPFs shall include the input star position and light curves in the FITS file, e.g. as part of an extra "ground truth" extension, so that we can easily verify signal recovery.  Scope may already do this out of the box. If not, I recommend preparing a scope PR to add this feature.
- The TPFs shall be sufficiently small to be added to the git repository, i.e. a few hundred Kb max.
- The TPFs shall be added to the Lightkurve repository under `lightkurve/tests/data/synthetic` alongside a README file which documents how the files were generated.

## Phase 2: Add unit test module

Once the three synthetic files are generated, the second step will be to create a new test module called `lightkurve/tests/test_synthetic_data.py` which performs the following sanity checks:

1. Can each of the three input light curves be recovered to a good level of accuracy after applying `SFFCorrector`?
2. Same for `PLDCorrector`?
3. Can the sinusoid period successfully be recovered from `synthetic-k2-sinusoid.targ.fits` using our `LombScarglePeriodogram` class?
4. Can the Neptune-sized planet period and depth be recovered from `synthetic-k2-planet.targ.fits` using our `BoxLeastSquaresPeriodogram` class?
5. Can the star positions be recovered to a good level of accuracy using `tpf.estimate_centroids()`?

There are a lot more tests and data challenges we can imagine executing using `scope`, but let's start with the simple TPFs and tests described above!

gully:
Progress so far: 


[...]

PR #534, opened 2019-06-20
Test synthetic data with known injected signals

gully:
Work in progress, aimed at Issue #531


gully:
I am creating the test TPFs in a notebook on a branch of my fork of scope:
https://github.com/gully/scope/blob/lightkurve_test_tpfs/notebooks/scope_experiments_01.ipynb

gully:
OK, I currently have three Kepler-Style TPFs for synthetic data, but their MB sizes exceed the desired few hundred KB by over 10x:

```bash
$ du -hs *
1.2M	synthetic-k2-flat.targ.fits
1.2M	synthetic-k2-planet.targ.fits
1.2M	synthetic-k2-sinusoid.targ.fits
```

I will investigate strategies for shrinking the file size, perhaps tar.gz-ing each fits file, or reducing the numerical precision of the vector data.

gully:
Update: I got to 392K per fits file with gzip.

barentsen:
@gully How many cadences/pixels do the TPFs include?  If we can get these files down to e.g. 100kb that would be even better.  Would the tests still be equally realistic if e.g. we cut the number of cadences in half, or would you recommend against doing so?

gully:
Currently using 1000 cadences.  500 cadences would be about 1/7th of a typical campaign.  I suspect the machinery could still work.

barentsen:
@gully I'd go for TPFs with fewer cadences if it doesn't invalidate the tests, given that we'll ship these files with the source distribution!

gully:
Hmm, seems like the python2 test fails because of an astropy version problem:
`ImportError: BLS requires AstroPy v3.1 or later`

barentsen:
AstroPy dropped support for Python 2 *before* it added the BLS feature, and therefore Lightkurve also does not support BLS features when used in Python 2.

gully:
Tasks remaining:
- [ ] Can the star positions be recovered to a good level of accuracy using `tpf.estimate_centroids()`?

lightkurve/tests/test_synthetic_data.py
```diff
@@ -0,0 +1,188 @@
+from __future__ import division, print_function
+
+import numpy as np
+
+from astropy.utils.data import get_pkg_data_filename
+import pytest
+import warnings
```
barentsen: `warnings` doesn't seem to be used, so this can go.

lightkurve/tests/test_synthetic_data.py
```diff
@@ -0,0 +1,188 @@
...
+
+from ..targetpixelfile import KeplerTargetPixelFile, TessTargetPixelFile
```
barentsen: no need to import `TessTargetPixelFile` I think?

lightkurve/tests/test_synthetic_data.py
```diff
@@ -0,0 +1,188 @@
...
+from ..correctors import SFFCorrector, PLDCorrector
+
+filename_synthetic_sine = get_pkg_data_filename("data/synthetic/synthetic-k2-sinusoid.targ.fits.gz")
+filename_synthetic_transit = get_pkg_data_filename("data/synthetic/synthetic-k2-planet.targ.fits.gz")
+filename_synthetic_flat = get_pkg_data_filename("data/synthetic/synthetic-k2-flat.targ.fits.gz")
+
+
+lacks_BLS = False
```
barentsen: This isn't a big deal, but PEP8 convention is to use `lower_case_with_underscores` for all variable names.

lightkurve/tests/test_synthetic_data.py
```diff
@@ -0,0 +1,188 @@
...
+    pg = cor_lc.to_periodogram(method='bls', minimum_period=1, maximum_period=9,
+                               frequency_factor=0.05, duration=np.arange(0.1, 0.6, 0.1))
+    ret_period = pg.period_at_max_power.value
+    threshold = 0.05
+    assert ((ret_period > true_period*(1-threshold)) &
+            (ret_period < true_period*(1+threshold)) )
+
+    # Verify that we get the transit depth to within a factor of 50%
+    assert ((pg.depth_at_max_power > true_depth/1.5) &
+            (pg.depth_at_max_power < true_depth*1.5))
```
barentsen: 50% seems a bit generous here; we can't claim that our SFF works well if it overfits by 50%.  At minimum let's reduce this to 20%  (which still passes), and ideally, let's plot detrended vs ideal light curve to check that nothing crazy is going on here?

lightkurve/tests/test_synthetic_data.py
```diff
@@ -0,0 +1,188 @@
...
+    ret_period = pg.period_at_max_power.value
+    threshold = 0.05
+    assert ((ret_period > true_period*(1-threshold)) &
+            (ret_period < true_period*(1+threshold)) )
+
+    # Verify that we get the transit depth to within a factor of 50%
+
+    print(pg.depth_at_max_power, true_depth)
+    assert ((pg.depth_at_max_power > true_depth/1.5) &
+            (pg.depth_at_max_power < true_depth*1.5))
```
barentsen: I see PLD only recovers the depth to ~15% or so on my machine, so let's make this test more stringent?  Ideally, can you post a quick plot of ideal vs detrended light curve to make sure nothing crazy is going on?

lightkurve/tests/test_synthetic_data.py
```diff
@@ -0,0 +1,188 @@
...
+filename_synthetic_flat = get_pkg_data_filename("data/synthetic/synthetic-k2-flat.targ.fits.gz")
+
+
+lacks_BLS = False
+try:
+    from astropy.stats.bls import BoxLeastSquares
+except ImportError:
+    lacks_BLS = True
+
+def test_sine_SFF():
```
barentsen: PEP8 also stipulates `lower_case_with_underscores` for all function names.  It's good to be consistent because then we don't have to think about case.

lightkurve/tests/test_synthetic_data.py
```diff
@@ -0,0 +1,188 @@
...
+
+    # Verify that we get the period within ~20%
+    pg = cor_lc.to_periodogram(method='lombscargle', minimum_period=1,
+                                 maximum_period=10, oversample_factor=10)
+    ret_period = pg.period_at_max_power.value
+    threshold = 0.2
+    assert ((ret_period > true_period*(1-threshold)) &
+            (ret_period < true_period*(1+threshold)) )
+
+    # Verify that we get the amplitude to within a factor of 2
```
barentsen: We should be able to do better than a factor 2?  Can't we?

lightkurve/tests/test_synthetic_data.py
```diff
@@ -0,0 +1,188 @@
...
+
+    cdpp_improvement = lc.estimate_cdpp()/cor_lc.estimate_cdpp()
+    assert cdpp_improvement > 10.0
+
+
+def test_centroids():
+    """Test the estimate centroid command"""
+    filename_synthetic_sine = get_pkg_data_filename("data/synthetic/synthetic-k2-sinusoid.targ.fits.gz")
+    filename_synthetic_transit = get_pkg_data_filename("data/synthetic/synthetic-k2-planet.targ.fits.gz")
+    filename_synthetic_flat = get_pkg_data_filename("data/synthetic/synthetic-k2-flat.targ.fits.gz")
```
barentsen: The filenames are defined globally, so can get rid of lines 174-176.

lightkurve/tests/test_synthetic_data.py
gully: Since BLS approximates a true transit shape, I will modify the benchmark from the coarse Rp/Rs (existing), to a mean of the in-transit points, which I will define as anything within, say, two hundred PPM of the deepest part of the transit.

gully:
Tests are failing for unrelated reasons: likely because Vizier is down---
`Either no sources were found in the query region or Vizier is unavailable`

barentsen:
This is really awesome, thanks @gully! :+1: :+1: :+1:

I intend to merge once the tests pass.  (Just restarted them.)

PR #537, opened 2019-06-24
Added the cutout method to the TargetPixelFile class

christinahedges:
I added a method to trim a TPF in the spatial dimensions. The idea is that this will eventually allow us to index tpfs like we can index light curves. Also this is handy for "zooming in" on large tpfs e.g. superstamps.


barentsen:
@christinahedges Did you intend to edit `2.06-identify-rolling-band.ipynb` as part of this PR, or is that an accident?

barentsen:
I believe column and row are mixed up in this version of the code.  This PR is good to merge otherwise.

barentsen:
The docstring claims center and size are specified in (column, row).  However using e.g. `size=(1, 3)` returns a TPF with 3 columns and 1 row.  I was expecting 1 column and 3 rows:

Screenshot from 2019-06-29 15-33-50

barentsen:
I pushed an extra commit which changes the behavior of `tpf.cutout()` to interpret the `center` and `size` arguments as a tuple of (col, row) instead of (row, col).

Example:
Screenshot from 2019-07-15 15-06-50

Issue #542, opened 2019-07-01
Centroiding doesn't work very well on faint, highly variable targets.


christinahedges:
The current implementation of centroids can fail really weirdly with some faint targets that are highly variable. In this case, it may be because the tpf background is actually negative.

![image](https://user-images.githubusercontent.com/14965634/60458929-af7c5f80-9bf4-11e9-9e42-7bc709de033c.png)

We should be using @mirca's `muchbettermoments`: https://github.com/mirca/muchbettermoments

![image](https://user-images.githubusercontent.com/14965634/60458937-b2775000-9bf4-11e9-916d-1d2da9ba5e79.png)

barentsen:
Excellent find, thank you @christinahedges!

And thank you @mirca for solving problems before they appear!

mirca:
credits to @dfm 👍

barentsen:
@mirca Do you think it is best to add `muchbettercentroids` as a dependency, or to borrow the function in Lightkurve?  (I'm trying to trade-off having a large number of dependencies vs avoiding code duplication.  I guess either solution is fine though!)

mirca:
@barentsen I think moving the function over to ``lightkurve`` would be much better :)

barentsen:
The quadratic centroiding method from `muchbettermoments` is now available in Lightkurve.

To check whether we can resolve this issue, I tried it on the target discussed at the top (EPIC 203850058).

I find that `method='quadratic'` yields more robust results when using `aperture_mask='all'`, though it is interesting to note that quadratic centroiding correlates in a non-linear way with `pos_corr`:

![quad1](https://user-images.githubusercontent.com/817669/67048785-b074c680-f0e9-11e9-9c17-5bac0a7a0e12.png)

It is also worth nothing that the performance of `methods='moments'` is much better when using the much smaller `aperture_mask='pipeline'`:

![quad2](https://user-images.githubusercontent.com/817669/67048937-fa5dac80-f0e9-11e9-8edb-96e425c8e396.png)

We'll have to gain more experience with quadratic centroiding to see when and where it should be used, but for the time being I think this issue can be closed!

mirca:
wow, thanks for showing this comparison!

PR #544, opened 2019-07-07
Add centroid estimator based on Vakili & Hogg 2016

mirca:
Hi y'all,

this PR adds the centroid estimator designed by [Vakili & Hogg 2016](https://arxiv.org/abs/1610.05873).

I've compared it with the classical method of moments using data from Kepler-10 and the results look consistent.  Tests on wild/faint targets have been done by @christinahedges in #542. 

Let me know your thoughts!

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -538,7 +538,7 @@ class TargetPixelFile(object):
 
     def _estimate_centroids_via_quadratic(self, aperture_mask):
         """Estimate centroids by fitting a 2D quadratic to the brightest pixels;
-        this is a hqelper method for `estimate_centroids()`.
+        this is a helper method for `estimate_centroids()`.
 
         This method will fit a simple 2D second-order polynomial
         $P(x, y) = a + bx + cy + dx^2 + exy + fy^2$
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -489,6 +489,7 @@ class TargetPixelFile(object):
         * `method='quadratic'` will fit a two-dimensional, second-order
           polynomial to the 3x3 patch of pixels centered on the brightest pixel
           inside the ``aperture_mask``, and return the peak of that polynomial.
+          Following Vakili & Hogg 2016 (ArXiv:1610.05873, Section 3.2).
 
         Parameters
         ----------
@@ -581,7 +582,7 @@ class TargetPixelFile(object):
                       [1,  1,  1, 1,  1, 1]])
         # We also pre-compute $(A^t A)^-1 A^t$, cf. Eqn 21 in Vakili & Hogg.
         At = A.transpose()
-        Aprime = np.linalg.inv(At @ A) @ At
+        Aprime = np.linalg.inv(At.dot(A)).dot(At)  # np.linalg.inv(At @ A) @ At
 
         # Loop through each cadence
         for i in range(len(self.time)):
@@ -594,7 +595,7 @@ class TargetPixelFile(object):
 
             # Step 2: fit the polynomial $P = a + bx + cy + dx^2 + exy + fy^2$
             # following Equation 21 in Vakili & Hogg.
-            a, b, c, d, e, f = Aprime @ z_.flatten()
+            a, b, c, d, e, f = Aprime.dot(z_.flatten())  # Aprime @ z_.flatten()
 
             # Step 3: analytically find the function maximum,
             # following https://en.wikipedia.org/wiki/Quadratic_function
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -27,7 +27,8 @@ from .lightcurve import KeplerLightCurve, TessLightCurve
 from .prf import KeplerPRF
 from .utils import KeplerQualityFlags, TessQualityFlags, \
                    plot_image, bkjd_to_astropy_time, btjd_to_astropy_time, \
-                   LightkurveWarning, detect_filetype, validate_method
+                   LightkurveWarning, detect_filetype, validate_method, \
+                   centroid_quadratic
 
 __all__ = ['KeplerTargetPixelFile', 'TessTargetPixelFile']
 log = logging.getLogger(__name__)
@@ -521,7 +522,8 @@ class TargetPixelFile(object):
             return self._estimate_centroids_via_quadratic(aperture_mask=aperture_mask)
 
     def _estimate_centroids_via_moments(self, aperture_mask):
-        """Helper method for `estimate_centroids()`."""
+        """Compute the "center of mass" of the light based on the 2D moments;
+        this is a helper method for `estimate_centroids()`."""
         aperture_mask = self._parse_aperture_mask(aperture_mask)
         yy, xx = np.indices(self.shape[1:]) + 0.5
         yy = self.row + yy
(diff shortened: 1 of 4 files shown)


barentsen:
Wowww this looks awesome, and all tests pass!  Hope to do a careful review & merge later today!

Thank you!! :heart: :heart: :heart: :heart: :heart:

lightkurve/targetpixelfile.py
```diff
@@ -489,9 +489,55 @@ def centroids(self, **kwargs):
                       LightkurveWarning)
         return self.estimate_centroids(**kwargs)
 
-    def estimate_centroids(self, aperture_mask='pipeline'):
+    def estimate_centroids(self, aperture_mask='pipeline', method='moments'):
```
barentsen: Let's make `quadratic` the default `method`, or do you think `moments` is significantly more robust?

mirca:
Is `pos_corr` in pixel units? I got very small (or negative values) of it.

barentsen:
Yes, `pos_corr` is in pixels.  The difference between `pos_corr` and the measured centroids should be a simple constant offset, because `pos_corr` encodes the residual of the WCS solution in the header (which is fixed for all frames) versus the *actual* position of a given frame.

barentsen:
@mirca Here is an example cadence where the quadratic centroid is definitely incorrect, but I haven't yet been able to figure out why!

Screenshot from 2019-07-09 16-40-43

mirca:
@barentsen yep, thanks for investigating! I'll take a look!

barentsen:
@mirca Any idea why the example posted above fails?  I can't seem to figure it out!

mirca:
I'll take a look on this ASAP

barentsen:
I think the way the parameters (a, b, c, d, e, f) are computed may be wrong.

Following Vakili & Hogg (https://arxiv.org/pdf/1610.05873.pdf), I derive the coefficients from Equation 21 as follows:

```python
from sympy import Matrix

# Matrix A as defined by Equation 20 in Vakili & Hogg (arxiv:1610.05873);
# the columns are: 1, x, y, x**2, xy, y**2.
A = Matrix([[1, 1, 1, 1, 1, 1],
            [1, 2, 1, 4, 2, 1],
            [1, 3, 1, 9, 3, 1],
            [1, 1, 2, 1, 2, 4],
            [1, 2, 2, 4, 4, 4],
            [1, 3, 2, 9, 6, 4],
            [1, 1, 3, 1, 3, 9],
            [1, 2, 3, 4, 6, 9],
            [1, 3, 3, 9, 9, 9]])
At = A.transpose()

# Coefficients of the free parameters {a, b, c, d, e, f}
# following Equation 21 in Vakili1 & Hogg.
(At*A).inv() * At
```
Result:
Screenshot from 2019-10-03 21-34-51

The coefficients in that matrix do not appear to match those in the code, i.e. here:

https://github.com/KeplerGO/lightkurve/pull/544/files#diff-cbdf227627021b7f9d0bca4efb5c2113R567-R575


(But it is very likely I am doing something wrong?)

mirca:
@barentsen we considered ``(0, 0)`` to be the coordinate (on the pixel-space) of the center of the brightest pixel,  and I guess you assumed it to be (2, 2)? Although I haven't verified it, I guess it shouldn't matter one way or the other up to a translation of the pixel-space coordinates.

Screenshot 2019-10-04 at 7 54 19 PM

mirca:
See this example:
```
import numpy as np

Z = np.zeros((3, 3))
Z[1, 1] = 1

A = np.array([[1, -1, -1, 1,  1, 1],
              [1,  0, -1, 0,  0, 1],
              [1,  1, -1, 1, -1, 1],
              [1, -1,  0, 1,  0, 0],
              [1,  0,  0, 0,  0, 0],
              [1,  1,  0, 1,  0, 0],
              [1, -1,  1, 1, -1, 1],
              [1,  0,  1, 0,  0, 1],
              [1,  1,  1, 1,  1, 1]])
At = A.transpose()
X = np.linalg.inv(At @ A) @ At @ Z.flatten()
a, b, c, d, e, f = X
det = 4 * d * f - e ** 2
print(det)

xm = - (2 * f * b - c * e) / det
ym = - (2 * d * c - b * e) / det
print(xm, ym)

B = np.array([[1, 1, 1, 1, 1, 1],
              [1, 2, 1, 4, 2, 1],
              [1, 3, 1, 9, 3, 1],
              [1, 1, 2, 1, 2, 4],
              [1, 2, 2, 4, 4, 4],
              [1, 3, 2, 9, 6, 4],
              [1, 1, 3, 1, 3, 9],
              [1, 2, 3, 4, 6, 9],
              [1, 3, 3, 9, 9, 9]])
Bt = B.transpose()

X_tilde = np.linalg.inv(Bt @ B) @ Bt @ Z.flatten()
a, b, c, d, e, f = X_tilde
det = 4 * d * f - e ** 2
print(det)

xm = - (2 * f * b - c * e) / det
ym = - (2 * d * c - b * e) / det
print(xm, ym)
```

barentsen:
I confirm that matrices `A` and `B` as defined above give consistent results, although there is of course a fixed column/result offset (+2/+2) in the result obtained with `B`.

**Good news**: if I remove the hard-coded analytical solutions for a/b/c/d/e/f that are currently in the code (lines 567-575), and replace it with `X = np.linalg.inv(At @ A) @ At @ Z.flatten()`, then the results magically become consistent with the old-school moment-based centroids.

So, I do think the analytical solution hard-coded here and in `muchbettermoments` is incorrect, but it'll be easy to fix.

barentsen:
@mirca I notice that Vakili & Hogg 2016 smooth the image by correlating it with the PSF (i.e. they apply a matched filter) before fitting the 2D polynomial.

Our implementation doesn't do this.  Do you think it matters?  It's certainly much easier not to have to do this.

lightkurve/targetpixelfile.py
```diff
@@ -515,11 +515,17 @@ def estimate_centroids(self, aperture_mask='pipeline', method='moments'):
...
-        """
-        This function approximates a star profile by a second order polynomial
-        and returns the coordinate of the peak flux.  For the mathematical
-        motivation and the details around this technique, please refer to
-        Vakili, M., & Hogg, D. W. 2016, ArXiv, 1610.05873.
+        """Estimate centroids by fitting a 2D quadratic to the brightest pixels.
+
+        This method will fit a simple 2D second-order polynomial
+        $P(x, y) = a + bx + cy + dx^2 + exy + fy^2$
+        the the 3x3 patch of pixels centered on the brightest pixel within
```
mirca: "the the"  --> "to the"

barentsen: Thanks, fixed it!

mirca:
@barentsen I guess the matrix multiplication operator (``@``) is not available for Python 2.7

mirca:
> @mirca I notice that Vakili & Hogg 2016 smooth the image by correlating it with the PSF (i.e. they apply a matched filter) before fitting the 2D polynomial.
> 
> Our implementation doesn't do this. Do you think it matters? It's certainly much easier not to have to do this.

@barentsen I think it matters. Specially in low SNR cases. But I guess an estimate of the PSF is not available is our case?

lightkurve/targetpixelfile.py
```diff
@@ -503,6 +534,87 @@ def estimate_centroids(self, aperture_mask='pipeline'):
...
+                      [1,  1, -1, 1, -1, 1],
+                      [1, -1,  0, 1,  0, 0],
+                      [1,  0,  0, 0,  0, 0],
+                      [1,  1,  0, 1,  0, 0],
+                      [1, -1,  1, 1, -1, 1],
+                      [1,  0,  1, 0,  0, 1],
+                      [1,  1,  1, 1,  1, 1]])
+        # We also pre-compute $(A^t A)^-1 A^t$, cf. Eqn 21 in Vakili & Hogg.
+        At = A.transpose()
+        Aprime = np.linalg.inv(At.dot(A)).dot(At)  # np.linalg.inv(At @ A) @ At
```
mirca: @barentsen I have a tiny objection here.  I would use ``matmul`` in place of ``dot``, see the second bullet in the
numpy docs https://docs.scipy.org/doc/numpy/reference/generated/numpy.dot.html

lightkurve/targetpixelfile.py
```diff
@@ -503,6 +534,87 @@ def estimate_centroids(self, aperture_mask='pipeline'):
...
+            # Step 1: identify the patch of 3x3 pixels (z_)
+            # that is centered on the brightest pixel (xx, yy)
+            data_mask = self.flux[i] * aperture_mask
+            arg_data_max = np.argmax(data_mask)
+            yy, xx = np.unravel_index(arg_data_max, data_mask.shape)
+            z_ = data_mask[yy-1:yy+2, xx-1:xx+2]
+
+            # Step 2: fit the polynomial $P = a + bx + cy + dx^2 + exy + fy^2$
+            # following Equation 21 in Vakili & Hogg.
+            a, b, c, d, e, f = Aprime.dot(z_.flatten())  # Aprime @ z_.flatten()
```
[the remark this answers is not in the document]
mirca: same comment here about `matmul` and `dot`.

lightkurve/targetpixelfile.py
```diff
@@ -503,6 +534,89 @@ def estimate_centroids(self, aperture_mask='pipeline'):
...
+            # that is centered on the brightest pixel (xx, yy)
+            data_mask = self.flux[i] * aperture_mask
+            arg_data_max = np.argmax(data_mask)
+            yy, xx = np.unravel_index(arg_data_max, data_mask.shape)
+            z_ = data_mask[yy-1:yy+2, xx-1:xx+2]
+
+            # Step 2: fit the polynomial $P = a + bx + cy + dx^2 + exy + fy^2$
+            # following Equation 21 in Vakili & Hogg.
+            # In Python 3 this can become `Aprime @ z_.flatten()`
+            a, b, c, d, e, f = np.matmul(Aprime, z_.flatten())
```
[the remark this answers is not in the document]
barentsen: Ignoring this codacy warning because the unused variable adds clarity to the implementation.

barentsen:
> @barentsen what do you think? Maybe we could expose it as a standalone function in `lightkurve`?

I'd be happy with that!  The only drawback I see is that the standalone function would have to receive the pixel values as input, which probably means we'll end up having to copy pixels in memory as opposed to directly accessing `self.flux`.  However the speed impact is probably very small -- I may play with / test this.

barentsen:
Using a Kepler-10 TPF with 473 cadences, I am finding that separating the quadratic centroiding function into a standalone function slows down `tpf.estimate_centroid(method='quadratic')` from 33ms to 48ms.

That's a significant slowdown, but given that we are talking about milliseconds, the speed decrease really shouldn't matter!

barentsen:
I'm going to go ahead and merge this!

I note that `method='moments'` remains the default for now, until we gather some more hands-on experience with the characteristics of the new  `method='quadratic'` option.

mirca:
@barentsen awesome, thanks!!!

PR #545, opened 2019-07-08
DOC: update kepler-10 doc index page example

mirca:
Hi all,

the current example on the index page gives me the following plot:
Screen Shot 2019-07-08 at 1 36 26 PM

I propose we tweak the example by using the light curve from quarter 5, which has more data points:
Screen Shot 2019-07-08 at 1 35 41 PM

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -37,7 +37,8 @@
 
     import lightkurve as lk
 
-    pixels = lk.search_targetpixelfile("Kepler-10", quarter=5).download()
+    pixels = lk.search_targetpixelfile("Kepler-10",
+                                       quarter=5).download()
     pixels.plot()
 
     lightcurve = pixels.to_lightcurve()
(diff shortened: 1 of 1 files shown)


barentsen:
Thank you @mirca!!!  Yess, this is much better, however it will overflow the box in my browser:

### Before:
Screenshot from 2019-07-08 09-48-54

### After:
Screenshot from 2019-07-08 09-48-38

Hmmmm how to solve this?  Or should I not worry about the box overflowing?

mirca:
@barentsen I've placed the quarter argument in the line below. What do you think?

barentsen:
I'm wondering if we want to use a TESS target instead, which will make Lightkurve look very modern and means we don't have to specify a quarter!

mirca:
@barentsen sounds good

christinahedges:
I'm closing this because it is out of date, our tutorials have been updated since this PR. Thanks for the playtest of the tutorials though @mirca!

mirca:
awesome @christinahedges!

Issue #539, opened 2019-06-27
`search` needs more `log.debug()` statements


barentsen:
[Over on Twitter](https://twitter.com/kellecruz/status/1144310082613723137) it is rightly pointed out that it can be hard to diagnose why `search_targetpixelfile` hangs.  I propose we add a number of useful `log.debug()` statements to the `search` module such that one can diagnose issues more easily, e.g. using:
```python
import lightkurve as lk
lk.log.setLevel("DEBUG")
lk.search_targetpixelfile(...)
``` 

Everything is already in place to support the example above, we just need to add `log.debug()` statements in useful places. Optionally, we can consider supporting debug mode using:

```python
lk.search_targetpixelfile(..., verbose=True)
```

barentsen:
(we need to look into an optional download progress bar as well)

barentsen:
We merged #547 today, which enables users to see status messages during search and download operations by calling `lightkurve.log.setLevel("DEBUG")` prior to executing search or download functions.

This will be part of the Lightkurve v1.1 release.

barentsen:
Example of the new functionality:
![Screenshot from 2019-07-18 12-34-43](https://user-images.githubusercontent.com/817669/61486503-76641f00-a958-11e9-9d6c-dabf2b77598c.png)

PR #547, opened 2019-07-18
Add `log.debug` messages during data search and download

barentsen:
This PR provides the user with the option to follow the progress of data search and download operations by turning on `lightkurve.log.setLevel("DEBUG")`.  This addresses #539 and should aid the debugging of search and download problems.

Example:
Screenshot from 2019-07-17 19-32-02

Any messages I missed?

### DIFF
--- a/lightkurve/search.py
+++ b/lightkurve/search.py
@@ -110,10 +110,13 @@ class SearchResult(object):
         # if the SearchResult row is a TESScut entry, then download cutout
         if 'FFI Cutout' in table[0]['description']:
             try:
+                log.debug("Started downloading TESSCut for '{}' sector {}."
+                          "".format(table[0]['target_name'], table[0]['sequence_number']))
                 path = self._fetch_tesscut_path(table[0]['target_name'],
                                                 table[0]['sequence_number'],
                                                 download_dir,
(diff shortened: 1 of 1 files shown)


lightkurve/search.py
```diff
@@ -667,6 +676,7 @@ def _query_mast(target, radius=None, project=['Kepler', 'K2', 'TESS']):
                 # suppress misleading AstropyWarning
                 warnings.simplefilter('ignore', AstropyWarning)
                 from astroquery.mast import Observations
+                log.debug("Started querying MAST for coordinates='{} {}'.".format(ra, dec))
```
nksaunders: This may be confusing if a user enters a `target` name and gets a message about querying coordinates... Since this only happens during a cone search it might make sense to specify 
```python
log.debug("Started querying MAST for observations within {} arcsec "
          "of target coordinates='{} {}'.".format(radius.to(u.arcsec), ra, dec))
```

If this is getting too wordy, don't worry about it. Just a thought!

barentsen: I like it!  I'm going to add the `radius` to all debug messages in search, because I think it's always being used in the query.

PR #548, opened 2019-07-19
Better collections and fill gaps for asteroseismology

christinahedges:
Updates:

* Collections have nicer plot functions
* Collections have a more efficient repr
* Collections have a new `stitch` method
* lc.append now has an `inplace` keyword argument, which helps stitching
* SAP_FLUX has the CROWDSAP and FLFRCSAP correction applied inside light curve
* `fill_gaps` now takes varying time deltas into account
* `fill gaps` now adds in reasonable white noise instead of point values


lightkurve/collections.py
```diff
@@ -95,6 +188,36 @@ class LightCurveFileCollection(Collection):
...
 
+    @property
+    def PDCSAP_FLUX(self):
+        return LightCurveCollection([lcf.PDCSAP_FLUX for lcf in self])
+
+    @property
+    def SAP_FLUX(self):
+        return LightCurveCollection([lcf.SAP_FLUX for lcf in self])
+
+    def stitch(self, offset=0.1, **kwargs):
```
barentsen: We'll need a short but sweet docstring here.

lightkurve/collections.py
```diff
@@ -95,6 +188,36 @@ class LightCurveFileCollection(Collection):
...
+        try:
+            warnings.warn('Stitching a `LightCurveFileCollection` which contains both SAP and '
+                         'PDCSAP_FLUX. Plotting PDCSAP_FLUX. You can remove this warning by '
+                         'using `LightCurveFileCollection.PDCSAP_FLUX.stitch()`.',
+                         LightkurveWarning)
```
barentsen: I'm worried that warnings like this will make people insensitive to warnings.  Do we really need to alert the user here?

lightkurve/lightcurve.py
```diff
@@ -271,7 +271,7 @@ def show_properties(self):
                     idx += 1
         output.pprint(max_lines=-1, max_width=-1)
 
-    def append(self, others):
+    def append(self, others, inplace=False):
```
barentsen: nice, love the inplace!

lightkurve/lightcurve.py
```diff
@@ -300,7 +304,8 @@ def append(self, others):
                 new_lc.centroid_col = np.append(new_lc.centroid_col, others[i].centroid_col)
             if hasattr(new_lc, 'centroid_row'):
                 new_lc.centroid_row = np.append(new_lc.centroid_row, others[i].centroid_row)
-        return new_lc
+        if not inplace:
+            return new_lc
```
barentsen: Should we `return new_lc` regardless of whether or not `inplace=True`?  It makes the API more consistent/predictable if we always return something, and it enables chaining.

lightkurve/collections.py
christinahedges: Yes because we're making an assumption that they want PDCSAP_FLUX when we'd rather they were explicit.

barentsen:
This looks awesome.  Thank you @christinahedges!

PR #555, opened 2019-08-01
[WIP] Adding a "waterfall" plot to the FoldedLightCurve Class

christinahedges:
We can add a nifty "waterfall" plot to the `FoldedLightCurve` class which will let us see the folded light curve as a "stack". This is useful for identifying TTVs and looking for systematics in the light curve, as well as checking out star spot evolution etc. 

Current behavior is that longer light curves maker taller waterfall plots. 

This is WIP

- [ ] Fix the plot rescaling
- [ ] Change the ylabels
- [ ] Add unit tests
- [ ] Check edge cases


christinahedges:
Closing this PR, as it has been replaced with a clean rebased version #682

Issue #554, opened 2019-08-01
No data found for some TOIs in Cycle1


jlillo:
#### Problem description
I am trying to retrieve TOI lightcurves extracted by the TESS pipeline with lightkurve, so the PDCSAP_flux. In more cases than expected I get the message "No data found for target [TIC_ID]". 

#### Example

```python
from lightkurve import search_lightcurvefile
lcfs = search_lightcurvefile(231663901).download_all(quality_bitmask='none')
```

#### Expected behavior
I would have expected this files to exist at least for all TOIs found by the TESS team.

#### Environment:
- platform: Mac OS sierra
- lightkurve version v1.1.0 running on Python 2.7:
- installation method (e.g. pip, conda, source): pip

barentsen:
Thank you for reporting this issue @jlillo!
 
The problem is that 231663901 is also a K2 EPIC ID at a completely different position in the sky.  The solution is to search for `"TIC 231663901"` instead of `231663901`.

We definitely need to update lightkurve to provide the user with a warning in this situation!

jlillo:
Oh! That explains that. Thanks a lot @barentsen!

Issue #543, opened 2019-07-02
Overlapping EPIC and TIC identifiers cause confusion in data searches


barentsen:
@MNGuenther reported via e-mail that this query incorrectly returns zero results:

```lk.search_lightcurvefile(220504623, mission='TESS')```

It looks like this behavior is explained by the fact that the MAST name resolver returns the coordinates for "EPIC 220504623" when 220504623 is passed.  Indeed, using `lk.search_lightcurvefile("TIC 220504623", mission='TESS')` resolves the issue.

Conclusion: when integer IDs without prefix are passed, we should encourage users to add the prefix by emitting a `LightkurveWarning`.

gully:
If the mission is user-supplied and the prefix is not supplied---as in the example above---can we pre-pend the default prefix as a string?  So the warning could say instead "we are assuming you want to search for `TIC 220504623`", and the proceed to search for `TIC 220504623`.

The vast majority of the time, a user providing a mission will be assuming that mission's ID system.

gully:
Just realized the proposal above is identical to a discussion elsewhere:
https://github.com/KeplerGO/lightkurve/issues/533#issuecomment-505155418

barentsen:
I believe this issue has been solved in an effective way by #558, which ensures a helpful warning is issued if the ID is ambiguous, so I think this issue can be closed.

(But please do comment or re-open if you think #558 does not solve this adequately!)

PR #558, opened 2019-08-01
Search: add a warning when a target ID is ambiguous.

barentsen:
This PR addresses #505, #543, #554 by adding a warning when someone searches for an integer TIC ID that could also be an EPIC ID.

Example:
Screenshot from 2019-08-01 13-55-11

### DIFF
--- a/lightkurve/search.py
+++ b/lightkurve/search.py
@@ -556,6 +556,10 @@ def _search_products(target, radius=None, filetype="Lightcurve", cadence='long',
     -------
     SearchResult : :class:`SearchResult` object.
     """
+    if isinstance(target, int) and (200000000 < target) and (target < 300000000):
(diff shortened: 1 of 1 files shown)


barentsen:
It looks like the exact ID ranges are as follows:

TIC ID range = [1, 10005000540].
KIC ID range = [1, 13161029].
EPIC ID range = [200000811, 251813738].

I obtained this using the following MAST CasJobs queries:
`SELECT MIN(ID), MAX(ID) FROM CatalogRecord;`
`SELECT MIN(kic_kepler_id), MAX(kic_kepler_id) FROM kepler_input_catalog;`
`SELECT MIN(ktc_k2_id), MAX(ktc_k2_id) FROM k2_target_catalog;`

Issue #556, opened 2019-08-01
LightCurveFiles from tpf.interact lack SAP_FLUX label


christinahedges:
We currently can't use the `LightCurveFiles` created by `tpf.interact` because the flux extension isn't readable. 

We can fix this by either changing the `LightCurveFile` class to be able to read a "FLUX" extension from the fits file, or changing `tpf.interact` to output fits files with "SAP_FLUX" as an extension. I think I prefer the second option. 

![image](https://user-images.githubusercontent.com/14965634/62320323-c19c2880-b454-11e9-9b5d-01d3ec4b23f6.png)

barentsen:
I think it makes sense for `interact` to write a `SAP_FLUX` extension, because it is using simple aperture photometry after all.  @gully are you happy with that?  Do you want to go ahead and make that change?

gully:
I had some reservation about overloading `SAP_FLUX`, which has historically mapped onto the pipeline aperture.  I also had some reservation, since some degree of pre-processing could have been performed.

Since some methods expect `SAP_FLUX`, though, it indeed may be useful to just define it for convenience.  What methods presently expect SAP_FLUX?

christinahedges:
It's mostly just so we can interact with it like a pipeline file, i.e. `lcf.SAP_FLUX` is something people might expect to do, because it's in our docs. Alternately we could allow for `lcf.FLUX`?

barentsen:
My two cents is that the *Lightkurve philosophy* is to make working with the data as easy as possible, while trusting the user to be smart enough to understand where data comes from and what the algorithms are doing.

I think the implication is to adopt `SAP_FLUX` as the standard extension name for simple aperture photometry, and leave it to the user to understand where the file they are using came from.

gully:
OK, `FLUX` is currently hardcoded into `to_fits()`:

```python
        def _make_lightcurve_extension(extra_data={}):
            """Create the 'LIGHTCURVE' extension (i.e. extension #1)."""
            # Turn the data arrays into fits columns and initialize the HDU
            cols = []
            if ~np.asarray(['TIME' in k.upper() for k in extra_data.keys()]).any():
                cols.append(fits.Column(name='TIME', format='D', unit=self.time_format,
                                        array=self.time))
            if ~np.asarray(['FLUX' in k.upper() for k in extra_data.keys()]).any():
                cols.append(fits.Column(name='FLUX', format='E',
                                        unit='counts', array=self.flux))
```

[...]

PR #559, opened 2019-08-02
Interact will provide SAP_FLUX label instead of FLUX label

gully:
Addresses #556

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1287,7 +1287,7 @@ class LightCurve(object):
             out_mask = None
         return out_mask
 
-    def to_fits(self, path=None, overwrite=False, **extra_data):
+    def to_fits(self, path=None, overwrite=False, flux_label='FLUX', **extra_data):
         """Writes the light curve to a FITS file.
 
         Parameters
@@ -1296,6 +1296,8 @@ class LightCurve(object):
             If set, location where the FITS file will be written.
         overwrite : bool
             Whether or not to overwrite the file
+        flux_label : str
+            The name of the label for the FITS extension, e.g. SAP_FLUX or FLUX
         extra_data : dict
             Extra keywords or columns to include in the FITS file.
             Arguments of type str, int, float, or bool will be stored as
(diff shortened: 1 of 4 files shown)


gully:
Pretty sure the tests fail due to remote data / connection issue.

barentsen:
The unit test failures were due to the `--cov=com` pytest option, which I copied from an Azure example but shouldn't be there.  I think it's fixed now.

barentsen:
Codacy is unhappy because `Kepler/TessLightCurve.to_fits()` have an `aperture_mask` argument whereas `LightCurve.to_fits()` does not.  This is syntactically allowed but it is a yellow flag.  Do we have a good answer to the question: "why do we support saving an aperture mask for Kepler/TESS Light Curves but not generic ones?"

lightkurve/lightcurve.py
```diff
@@ -1287,7 +1287,7 @@ def _boolean_mask_to_bitmask(self, aperture_mask):
             out_mask = None
         return out_mask
 
-    def to_fits(self, path=None, overwrite=False, **extra_data):
+    def to_fits(self, path=None, overwrite=False, flux_label='FLUX', **extra_data):
```
barentsen: I think we may want to call this `flux_column_name` or `column_name` (instead of `flux_label`), to make it more clear what this refers to.

gully:
> "why do we support saving an aperture mask for Kepler/TESS Light Curves but not generic ones?"

Imagine you make a `LightCurve` object from, say, ASASSN data.  `aperture_mask` would be ill-defined.  In general, 3D datacube objects are only defined for TESS and Kepler/K2, so aperture masks only have meaning in the context of 3D data cubes.  With that said, there's nothing illegal about placing a kwarg there, and users can ignore or use if they have a purpose for it.  No strong feelings here.  Slightly surprised that overriding a method of a subclass with more kwargs is even a yellow flag: different subclasses have different needs.

barentsen:
I think @codacy-bot's logic here is that "if a method requires different arguments, it's probably not the same method and should have a different name".  However there's definitely no strict rules about this, and it's not the focus of this PR, so we can leave as is!

barentsen:
This looks good.  I intend to merge in a few minutes!

barentsen:
Thanks @gully!

PR #561, opened 2019-08-06
[WIP] Refactor `PLDCorrector` and add `GPCorrector`

barentsen:
This PR proposes to add a `GPCorrector` object which makes it easy to fit and remove a GP from a light curve.

This was trio-coded with @christinahedges and @nksaunders!


barentsen:
@nksaunders Looks like the original PLDCorrector had a `use_gp=True` option.  Do you think we should support disabling the GP in some way, or get rid of this option altogether?  (I don't feel strong either way; I'm reminded of this option because its is causing some unit tests to fail.)

(Please make sure to pull first thing this morning; I fixed the codacy issues and rebased on master.)

barentsen:
Note from in-person discussion: predicting the variance with celerite appears to be slow, so we'll want to introduce a `propagate_errors` option which we may want to set to `False` by default.

nksaunders:
Improvements to the implementation from this morning's discussion with @christinahedges, @barentsen, @gully, and @nksaunders (to be added either before this is merged or as a future PR):

 - Split light curve into two sections to individually capture distinct noise properties of each half
 - Improve design matrix by using neighboring targets
 - Fix normalization to remove any stellar signal from design matrix column vectors
 - `GPCorrector` takes `cadence_mask` in the constructor, while `PLDCorrector` and `SFFCorrector` take it as an argument in the `correct` method

barentsen:
I'm going to close this PR because we'll want to revise `PLDCorrector` to use the new `RegressionCorrector` framework (cf. #613).  The code attached to this PR will continue to remain immensely useful for that purpose! :+1::+1:

PR #562, opened 2019-08-06
Create option to show number of cadence on TPF plot

zabop:
I created an option in the plot function in the TargetPixelFile class which enables users to print the number of the frame which they are plotting if they want to, by setting showcadencenum  = True:

pixelfile.plot(frame=500, showcadencenum=True)

A short video about how it looks like: https://www.youtube.com/watch?v=IV9U_rbJNn4

Screenshots showing functionality with and without showcadencenum = True are below.

show_frame_number_on_tpf_plot
zoomed_in_show_frame_number_on_tpf_plot


barentsen:
Thank you @zabop!  :+1: :+1: :+1: This is an excellent idea.  We will discuss this proposal in the team later today!  

My first reaction:
* I totally agree `tpf.plot` should show the frame number.
* I wonder if it should do this by default instead of via an optional argument?
* It may be best to display the cadence number (i.e. `self.cadenceno[frame]`) rather than the frame number, because the frame number will be different when e.g. a user applies a different quality mask.

Again, thank you for this contribution! :+1::+1::+1:

PS: If you are specifically trying to make movies of TPFs with frame number or time overlaid, you may be interested in `k2flix`: https://barentsen.github.io/k2flix/

zabop:
Hi @barentsen !
Thanks for the positive reaction!

About the second bullet point: in this implementation it is not showing by default, but this can be changed easily.

About third bullet point: Agree, `cadenceno` is better.

barentsen:
@zabop Would you like to go ahead and make the following two changes:
* Show `self.cadenceno[frame]` instead of frame.
* Remove the `showcadencenum` option, let's just show it by default!

...then we'd be delighted to merge!

zabop:
Hi @barentsen! 
Alright, I think the changes are done now! They look like this:
image
Let me know if there are more issues.
Cheers

zabop:
👍👍welcome ;)✌️

PR #567, opened 2019-08-09
Add support for stingray?

barentsen:
This PR explores adding `to_stingray()` and `from_stingray()` methods to our `LightCurve` object to enable easy access to features of that package!

To do: find a good example use case for these methods to add as part of an interoperability tutorial?

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1321,8 +1321,9 @@ class LightCurve(object):
     def to_stingray(self):
         """Returns a `stingray.Lightcurve` object.
 
-        This feature requires Stingray to be installed.
-        An `ImportError` will be raised if this version is not available.
+        This feature requires `Stingray <https://stingraysoftware.github.io/>`_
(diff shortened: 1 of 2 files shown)


lightkurve/lightcurve.py
```diff
@@ -1131,6 +1131,36 @@ def from_timeseries(ts):
...
+        try:
+            from stingray import Lightcurve as StingrayLightcurve
+        except ImportError:
+            raise ImportError("You need to install Stingray to use "
+                              "the LightCurve.to_stringray() method.")
+        return StingrayLightcurve(time=self.time, counts=self.flux,
+                                  err=self.flux_err, input_counts=False)
+
+    @staticmethod
+    def from_stingray(lc):
```
mirca: @barentsen I think it should be `ts` instead of `lc`?

barentsen:
I incorporated @mirca's feedback and checked that the unit tests pass. I'm going to merge this for now as an experimental feature!

Issue #569, opened 2019-08-13
Support light curves centered around zero


keatonb:
#### Problem description

For light curves input by the user, it is not uncommon for the time series flux to be normalized and mean-subtracted so that the relative flux is centered around zero.

This causes problems when lightkurve attempts to re-normalize (often unnecessarily) the flux by dividing the median which is very close or equal to zero.

The minimum thing to add is a warning of possible bad behavior when ```normalize``` is called on a light curve that contains negative values.

We should also remove unnecessary calls to lc.normalize(), such as the one in ```periodogram.from_lightcurve``` that has no practical effect.  I will submit a pull request for this soon.  Perhaps auto-renormalization could also not be the default where it's an option, like in ```lightcurve.plot``` (change default to ```normalize=False```).

Has zero-centered flux been previously considered for the lightkurve internal default?

#### Example

```python
import numpy as np
import lightkurve as lk

time = np.arange(100)
amp = 0.1 #What we want to measure
freq = 0.08412
flux = amp*np.cos(2*np.pi*np.arange(100)*freq)

lc = lk.LightCurve(time,flux)

ls = lc.to_periodogram(normalization='amplitude')
print(ls.max_power)
```
Gives... 23627464 ppm

#### Expected behavior

Expect to measure close to the input amplitude of 1e5 ppm.  Renormalization under the hood (division by small number) causes the variability to blow up.

#### Environment

-  platform (e.g. Linux, OSX, Windows): OSX
-  lightkurve version (e.g. 1.0b6): 1.0.1
-  installation method (e.g. pip, conda, source): conda

barentsen:
Thanks @keatonb!  It is certainly true that Lightkurve hasn't been carefully tested to work on zero-centered light curves so far.

PS: Do you have an understanding of the pros-and-cons of using zero-centered vs one-centered normalized light curves?  Do you reckon folks use one or the other as a matter of taste, or is there more to the choice?

keatonb:
Residuals are basically zero-centered, and the analysis of residuals is of standard interest for time series astronomy (e.g., prewhitening of pulsation signals).  Starting zero-centered makes the analysis of residuals consistent with the initial light curve analysis.  

Zero-centered data can also have the same units as the periodogram (parts-per-million or percent or whatever), which is nice and may help us keep track of units consistently through lightkurve.  Perhaps this issue could be stated more broadly, even, as "support relative units for flux."

barentsen:
[...]

PR #570, opened 2019-08-13
do not normalize light curve to compute peridogram

keatonb:
Removing an unnecessary (and problematic; #569) call to normalize a lightcurve before computing the periodogram.  The Lomb Scargle algorithm used can compute the same spectrum from an unnormalized light curve.

### DIFF
--- a/lightkurve/periodogram.py
+++ b/lightkurve/periodogram.py
@@ -737,12 +737,6 @@ class LombScarglePeriodogram(Periodogram):
                           LightkurveWarning)
             maximum_frequency = kwargs.pop("max_frequency", None)
 
-        # Make sure the lightcurve object is normalized
-        if not np.in1d(lc.flux, lc.normalize().flux).all():
-            warnings.warn("Input light curve will be normalized.",
-                          LightkurveWarning)
-            lc = lc.normalize()
-
         # Check if any values of period have been passed and set format accordingly
         if not all(b is None for b in [period, minimum_period, maximum_period]):
             default_view = 'period'


barentsen:
Thanks @keatonb!

@ojhall94 @christinahedges Do either of you remember why we added the call to `lc.normalize()` inside `Periodogram.from_lightcurve()`?

(There appears to be a unit test failure that is unrelated to this PR, I'll investigate.)

gully:
It's for the purposes of making the units work out--- `ppm^2 / Hz` is assumed for all units, since after normalization there's a factor of `1 x 10^6`.  I think we could/should support different normalizations, since folks may wish to have units of `variance / Hz` for example.

gully:
See the `flux_scaling` parameter for where the factor of 1 million comes in to convert signals to parts-per-million:
 https://github.com/KeplerGO/lightkurve/blob/b63b7d338ab57149179ebea6071d605e398dc0e5/lightkurve/periodogram.py#L838

keatonb:
> @ojhall94 @christinahedges Do either of you remember why we added the call to `lc.normalize()` inside `Periodogram.from_lightcurve()`?
> 
Could be related to the way the periodogram used to be normalized to enforce agreement with Parseval's theorem.  I changed this in #491 but didn't identify that these lines were no longer needed (or desired).

gully:
There's nothing holy about ppm, we should consider supporting the proper dimensional analysis if we know the input units (e/s, etc).

keatonb:
I agree with @gully that ideally units are inferred from the light curve flux units, but that sounds like hard work.

Could at least a ```normalize=True``` keyword be added to ```from_lightcurve``` like it is in ```_create_plot```?  The language may get confusing since ```from_lightcurve``` also has a ```normalization``` keyword, where one would refer to the light curve and the other to the periodogram.  

Another alternative is that the ```normalize``` function is rewritten to handle zero-centered data.  Perhaps if there are any negative values in the light curve we add one before renormalizing and then subtract it again?  We should probably also not trust this to work in all cases, as very low signal-to-noise light curves in regular flux units may contain negative values.

ojhall94:
Hi Keaton! I've had a look at your issue #569 as well, super happy you raised this. There's so much stuff I missed so I'm glad people are paying attention :) To answer some of your questions:

> Do either of you remember why we added the call to lc.normalize() inside Periodogram.from_lightcurve()?

As Gully pointed out, the `from_lightcurve()` function was originally built to just return power in units of `ppm^2/muHz`, so required fractional flux values. I considered including it under the hood harmless; if somebody had called `lc.normalize()` themselves before calling `lc.to_periodogram()`, it would just divide the data by 1 again.

I hadn't considered the use case for zero-centered data, but Keaton is right, it's completely incompatible with `from_lightcurve()`. 

I think rewriting the `normalize()` function to handle zero-centered data is the path of least resistance here. Instead of calling `lc.normalize()` internally, it could be a `periodogram` utility that checks where the data is centered.

Units would still have to be in `ppm` of course. As Gully says, support for the physical parameters in the future (even just magnitudes) would be ideal.

barentsen:
I prepared a proposal to modify the behavior of `lc.normalize()` over in #589.  Please have a look and leave any comments.  (If that PR is merged, we'll definitely want to make the normalization step optional in periodogram.from_lightcurve!)

barentsen:
To resolve this issue, I propose we do remove the call to `lc.normalize()` inside the periodogram, i.e. I am happy to merge this PR.

However, I do wonder if we need some sort of warning as follows:
```python
if np.abs(1. - np.nanmedian(lc.flux)) < 0.5*np.nanstd(lc.flux):
    warnings.warn("The input light curve does not appear to be normalized, "
                  "this has consequence XXX.",	
                  LightkurveWarning)
```

Is such a warning desirable?  What should the exact words be to explain the consequence of not normalizing? (@keatonb @ojhall94?)

gully:
@barentsen don't forget about the `flux_scaling`, which should be moved if the normalization is turned off.  The PPM units only work out if `normalize` is called, *and* the lightcurve is subsequently multiplied by a factor of 1 million.  Alternatively, you can state: 
                 "The input light curve does not appear to be normalized, "
                  "The Periodogram units are unknown.  Plots and other methods will continue to erroneously report all units as parts per million (PPM)."  
...which is unsatisfactory.  To support the correct input units would take some more sophisticated unit handling, probably an optional unit field for lightcurve flux to support (e/s) or (counts / time stamp)?  For now with Kepler/TESS, we know the units so may be easier, but a general LightCurve object typically has undefined units.

barentsen:
Should this instead be dealt with as as part of the `normalization` argument then?  Example:
* `normalization='amplitude'` will return power in flux units;
* `normalization='amplitude-normalized'` will normalize and return power in units ppm;
* `normalization='psd'` will return power in units variance / frequency;
* `normalization='psd-normalized'` will return power in units ppm^2 / frequency.

Does that make sense?

barentsen:
Offline, @gully suggested we may want to name the options `amplitude`, `amplitude-ppm`, `psd`, and `psd-ppm`.

keatonb:
I think showing a warning about units in to_periodogram when a light curve is passed that does not appear to be pre-normalized is a good idea.  Rather than trying to force a normalization in the function, encourage the user to pass a normalized light curve if they want their periodogram in relative units.

We might generally not want to make the decision that ppm is the user's preferred units (I usually want ppt or percent).  If the user passes an unnormalized light curve they may want to know the amplitude of signals in flux units.  Why not let the input units of the light curve be the output unit of the periodogram?  A periodogram of a normalized light curve can be easily put into units of ppm, ppt, or percent by multiplying the result of .to_periodogram by 1e6, 1e3, or 1e2.

barentsen:
I opened #591 to enable `LightCurve` objects to keep track of the flux units.  This should enable periodogram objects to automatically determine the correct units of power.

barentsen:
I finished work on #591, which keeps track of light curve flux units more properly, and changes `LombScarglePeriodogram` to adopt whatever units the light curve is in.

If we merge that PR, the behavior will become as shown below:

Screenshot from 2019-09-10 07-53-25

@keatonb @ojhall94 @christinahedges @gully Does the behavior of the units in the example above make sense to you all?

ojhall94:
Hey @barentsen!

Thanks for doing this work! Super exciting to see Astropy units carried through the whole process. My only comments:
- PR #591 will break any pipelines/tutorials/etc. that depend on the current normalization options. We might want to consider adding a warning, and/or updating tutorials in the same PR to minimise confusion for users.
- I don't know the answer to this (@keatonb, @danhey?), but does anybody ever use amplitudes of `e-/s` when studying oscillations? If not, we may want to consider setting default kwargs to make the default power units something that users would recognise.

barentsen:
Thanks for the feedback @ojhall94!!

My thoughts:
* We should definitely update the tutorials -- I intend to work on that today.  I don't feel too bad about this change because my understanding is that it should only affect units, not the intrinsic behavior. We should of course highlight it prominently in the change log and similar communications.
* Periodogram will now use whatever units the light curve is in, so the only reason it shows "electron/second" is because that is the unit used by Kepler and TESS data products.  (Admittedly many publications show "counts" or "counts/second" in their plots, which is technically incorrect.)  The only alternative to this is to normalize by default, which is of course what I'm trying to avoid?

keatonb:
This is really great @barentsen !!

Purely for convenience, I have one small suggestion. I think changing units when computing the periodogram would be more natural if the ``to_periodogram`` call had a ``unit`` keyword, which may just access ``normalize`` internally.  Then if the user wants to view the light curve in different units (say e-/s) than the periodogram (say, ppm), they can call ``lc.to_periodogram(unit='ppm')`` which would give the same output as  ``lc.normalize(unit='ppm').to_periodogram()``.

 I think units in = units out is the most predictable behavior.  I agree that most people won't want a periodogram in e-/s, but they probably also don't want the light curve in e-/s.  Periodogram of a normalized lightcurve gives a normalized periodogram.  I prefer avoiding making choices for the user that may abstract away exactly how the data are being manipulated.

Thinking about this has also helped me think of a way to handle zero-centered light curves, which I'll comment on over at #569

keatonb:
@ojhall94 
> As a user of ppm squared per microhertz, `lc.normalize(unit='ppm').to_periodogram('psd')` is more intuitive and much easier to write than specifying that in a unit kwarg. I think a unit kwarg would be neat, but in practice I think the way @barentsen set this up is best.

``psd`` isn't a unit.  It just says you want the power spectral density rather than absolute power.  That could be ppm^2/muHz or ppt^2/muHz or, as you have pointed out is a little crazy but still valid, e-^2/(s^2 muHz).  I'm suggesting that only for convenience a periodogram keyword be provided as an option, so `lc.to_periodogram('psd', unit='ppm')` do the same as above, calling `normalize` under the hood only if specifically requested.  The default in `to_periodogram` should be `unit=None`.

ojhall94:
I agree with @barentsen, that does work nicely, although the syntax you showed in your previous comment I'm also happy with, personally!

keatonb:
I think the confusing thing is actually something else, which makes me regret how we settled the issue of normalization options.  To me,`normalization='psd', unit='ppm'` makes sense because `ppm` is the amplitude unit, and the `psd` part says to square it and give the density.

What's more confusing is that if you calculate the periodogram with `normalization='amplitude'`, the amplitude spectrum values are stored in an attribute called `power` even though these are not powers.  I think we would be most complete to let the periodogram object always have two attributes, `power` and `amplitude`, the latter just being the square root of the former.  Then the `normalization` keyword only says whether we want a power spectral _density_ (`psd`) or absolute power units (`absolute`, or `ps`?), and the unit keyword can be something like `amp_unit` (to match `freq_unit`).  I prefer this to specifying all of the allowed combinations of normalization and amplitude units.

barentsen:
Thanks for sharing that thought @keatonb!  It may take me another day to respond because I decided I need to read Jake Vanderplas' review of the Lomb Scargle periodogram (https://arxiv.org/pdf/1703.09824.pdf).

keatonb:
Question about the above: If you do `np.median(lc.normalize(unit='ppm').flux)` do you get something close to 1e6?

barentsen:
Yep, I confirm:
*`np.median(lc.normalize(unit='ppm').flux)` equals 1e6;
*`np.median(lc.normalize(unit='ppt').flux)` equals 1e3;
*`np.median(lc.normalize(unit='percent').flux)` equals 1e2.

Is that the expected behavior?

PR #571, opened 2019-08-13
Correct target pixel file factor to cut out same pixels 

amcody:
I have adjusted the wcs sent to Cutout2D so that the same detector pixels are sent to each cadence of a newly created target pixel file. This solves the problem of having a star jump around unnecessarily in its mask.


amcody:
I believe the TPF factor is back in working order; I have tested it on the C9 planet as well as another random star, and not only do the individual frames no jiggle unnecessarily, but the detector coordinates also come out correct. Please check once the tests complete, thanks!

barentsen:
@amcody The tests pass (:+1:) but codacy identified a few issues which would be good to fix, in particular, it looks like `test_tpf_wcs_from_images` is defined twice and `Using variable 'images' before assignment.`.  Can you take a look at fixing those?  (Should be quick/easy to fix!)

amcody:
@barentsen: Quality review now passes! I expect everything else to be ok momentarily. Could you have a look and merge if satisfied?

PR #580, opened 2019-08-20
Fixed fits file writing so that it also writes the centroid_col and centroid_row attributes

christinahedges:
When we save light curves with `to_fits` we are not passing along `MOM_CENTRX`, which we use for the centroids. This fixes that bug.


barentsen:
Thank you Dr @christinahedges!!  You get a medal!  :1st_place_medal: 

Merging!

PR #589, opened 2019-09-05
Issue a warning when normalization makes no sense

barentsen:
The `LightCurve.normalize()` method normalizes a light curve by dividing the flux by the median value.  However, this does operation does not make any sense for zero-centered light curves, or for light curves with a negative median.

This PR modifies `normalize()` such that an exception is raised in the following situations:
* the median flux is zero, or within one standard deviation from zero;
* the median flux is negative.

Question for feedback: is raising a `LightkurveError` exception the right behavior?  Alternative options are:
* issue a warning and divide by the median anyway;
* issue a warning and return the original light curve.

Thoughts?

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -23,7 +23,7 @@ from astropy import units as u
 from . import PACKAGEDIR, MPLSTYLE
 from .utils import (
     running_mean, bkjd_to_astropy_time, btjd_to_astropy_time,
-    LightkurveWarning, LightkurveError, validate_method
+    LightkurveWarning, validate_method
 )
 
 __all__ = ['LightCurve', 'KeplerLightCurve', 'TessLightCurve']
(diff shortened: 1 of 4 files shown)


christinahedges:
I think I have a slight preference for a warning and normalizing anyway. There are some normalized light curves that do go through zero.

We also have normalize in the plot function so if I do 
`(lc.normalize() - 1).plot()`

This is going to raise an exception?

gully:
I'm a bit squeamish about some of these conditionals... For example, what happens if I have a mostly flat lightcurve (flux = 1.000 ) with gigantic (flux > 10^6 ) superflare lasting a few hundred cadences.  A human and the existing system would both normalize it to the median (flux = 1.000).  The new system would raise an exception, since the standard deviation would render a median consistent with zero to "one sigma".  

More food for thought: What if `normalize` supported a kwarg `standard_deviation=True` (or equivalent naming scheme) that optionally normalizes by the standard deviation rather than the median.  Normalizing by the standard-deviation should be well-defined for real-world noise-possessing zero-centered lightcurves.  

Or maybe `zero_centered=True` would do the behavior in this PR, and `zero_centered=False` would do the old behavior?

No strong feelings here either way, just some mild reluctance to make normalize too "smart".

barentsen:
I intend to go ahead and turn the exception into a warning.  That will also address the fact that freak light curves may exist (cf. Gully's example).

barentsen:
I edited the PR to emit a warning rather than an exception.  I intend to merge this PR tonight unless anyone objects.

PR #591, opened 2019-09-06
Add a `LightCurve.flux_unit` property to keep track of flux units

barentsen:
@gully and I think it may be wise to add a `flux_unit` property to the `LightCurve` class for four reasons:
1. it enables us to get rid of hard-coding "electron/second" in `LightCurve._create_plot()`;
2. it enables normalized light curves to be identified because they'll have `flux_unit=Unit('dimensionless')`;
3. it enables periodogram classes to determine the units of power;
4. it will enable interoperability with astropy timeseries and quantities.

Items 1, 2, 3 are already addressed as part of this PR.

By default, Kepler and TESS light curves will be instantiated with `flux_unit=u.Unit('electron/second')`.

Thoughts?


barentsen:
I slightly expanded the scope of this PR to add support for letting a user instantiate a LightCurve object using AstroPy `Time` and `Quantity` objects.

gully:
This PR allows a user to override the `flux_unit` attribute:

```python
lc.flux_quantity[0]
207585.94 e/s
 
lc.flux_unit = u.electron/u.hour

lc.flux_quantity[0]
207585.94 e/h

```

resulting in `lc.flux_quantity` possessing the new units, with no change to the flux_quantity values.  If someone wants to change the default units of the lightcurve *and* update the values contained in `lc.flux`, they have to also manually scale the flux values.  

That's the expected and desired behavior from my perspective, but I can imagine some folks expecting an override of the flux unit to percolate to the `lc.flux_quantity` and/or `lc.flux` array in a unit conversion sense.

gully:
I get an error when I plot a periodogram with normalization set to 'psd':

```
pg =lc.to_periodogram(normalization='psd')
pg.plot()

<matplotlib.axes._subplots.AxesSubplot at 0x1c39c7f358>
Error in callback <function install_repl_displayhook.<locals>.post_execute at 0x10f612950> (for post_execute):
ParseFatalException                       Traceback (most recent call last)
ParseFatalException: Double superscript (at char 0), (line:1, col:1)

ParseSyntaxException                      Traceback (most recent call last)
~/anaconda3/envs/lightkurve_base/lib/python3.6/site-packages/matplotlib/mathtext.py in parse(self, s, fonts_object, fontsize, dpi)
[...]
```

barentsen:
The periodogram plotting issue identified by @gully above is an AstroPy unit representation bug.  I opened an AstroPy PR to try and resolve this (https://github.com/astropy/astropy/pull/9218), and added a workaround in this PR for the time being:

https://github.com/barentsen/lightkurve/blob/108733ec27521a428d637303f26638a97748aea8/lightkurve/periodogram.py#L303-L307

Thanks @gully!

barentsen:
Tests are passing and @christinahedges completed an in-person code review.

I intend to merge this now so we can try out this new behavior on master before the next release.

Issue #586, opened 2019-08-28
test_interact.py::test_interact_sky_functions() trips interact.py::make_tpf_figure_elements into logs10 errors


texadactyl:
#### Problem description
A zero pedestal results when interact.py::make_tpf_figure_elements() defaults pedestal=0.
test_interact.py::test_interact_sky_functions() calls make_tpf_figure_elements() without passing a pedestal value.
Then, ```vlo, lo, hi, vhi = np.nanpercentile(tpf.flux - pedestal, [0.2, 1, 95, 99.8])``` returns
vlo=-9.479096405029297, lo=-7.892417116165161, hi=445.12173614501944, vhi=2282.0120917968748
Subsequently, ```vstep = (np.log10(vhi) - np.log10(vlo)) / 300.0``` throws this:
```RuntimeWarning: invalid value encountered in log10```

Changing ```fig1, slider1 = make_tpf_figure_elements(tpf, tpf_source)```
to ```fig1, slider1 = make_tpf_figure_elements(tpf, tpf_source, -10)``` in test_interact.py::test_interact_sky_functions()
stopped the log10 errors since vlo=0.5209035949707032, lo=2.107582883834839, hi=455.12173614501944, vhi=2292.012091796875.  Also, it allowed the test to proceed but I do not know what the author of this test intended.

#### Environment
-  platform: Xubuntu 18.04.3
-  lightkurve version: downloaded from github today
-  installation method: downloaded from github today

#### Comments
I would suggest that make_tpf_figure_elements() be amended to protect itself from negative values of vlo, lo, hi, and vhi.  Maybe just assert and throw an exception if negative?

Issue #585, opened 2019-08-28
test_periodogram::test_assign_periods() causes sqrt errors


texadactyl:
#### Problem description
test_assign_periods(), in its call to lc.to_periodogram(frequency=frequency), fails to pass the freq_units.  I was able to stop the sqrt() errors and allow the test to proceed as intended by supplying matching units: 
```
frequency = np.arange(1, 100) * u.Hz
p = lc.to_periodogram(frequency=frequency, freq_unit=u.Hz)
```

#### Environment
-  platform: Xubuntu 18.04.3
-  lightkurve version: downloaded from github today
-  installation method: downloaded from github today

#### Comments
I would suggest that from_lightcurve() in both BoxLeastSquaresPeriodogram and LombScarglePeriodogram (periodogram.py) protect themselves from erroneous frequency input by requiring freq_unit be specified when frequency is specified.

gully:
The docstring for `from_lightcurve` states the defaults for frequency units, which depends on the *normalization*:

>        frequency :  array-like
            The regular grid of frequencies to use. If given a unit, it is
            converted to units of freq_unit. If not, it is assumed to be in
            units of freq_unit. This over rides any set frequency limits.

>          freq_unit : `astropy.units.core.CompositeUnit`
            Default: None. The desired frequency units for the Lomb Scargle
            periodogram. This implies that 1/freq_unit is the units for period.
            With default normalization ('amplitude'), the freq_unit is set to
            1/day, which can be overridden. 'psd' normalization will set
            freq_unit to microhertz.

The context-dependent units is a bit funky, but was designed to satisfy the disparate conventions in different astronomical subfields.  We may be revisiting the units as part of discussion of #570.

For the purposes of suppressing a warning, it's fine to explicitly add freq_units, as long as the test coverage remains undiminished.

gully:
By the way, a related problem is that in Jupyter notebooks (and maybe most editors?), one cannot easily view the informative docstring from the convenience functions, so I worry that a lot of these valuable `from_lightcurve` docstrings are missed by typical users that use `to_periodogram`.

texadactyl:
I understand the defaults.  The goal is not just to suppress a warning; if the frequency units are not consistent with the frequency value, then the test result is invalid.

I believe that the above code change allows the test to complete as intended.

PR #592, opened 2019-09-12
Suppress warnings during tests

gully:
This PR is aimed at suppressing warnings when running the unit test.  See Issues #586, #585, and others.

### DIFF
--- a/lightkurve/tests/test_interact.py
+++ b/lightkurve/tests/test_interact.py
@@ -90,7 +90,7 @@ def test_custom_exported_filename():
 
 @pytest.mark.skipif(bad_optional_imports, reason="requires bokeh")
 def test_max_cadences():
-    """Can we provide a custom lightcurve to show?"""
+    """Test what is the max number of cadences allowed"""
     import bokeh
     with warnings.catch_warnings():
         # Ignore the "TELESCOP is not equal to TESS" warning
--- a/lightkurve/tests/test_lightcurve.py
+++ b/lightkurve/tests/test_lightcurve.py
@@ -290,7 +290,8 @@ def test_custom_lightcurve_file(path, mission):
     if mission == "K2":
         lcf_custom = KeplerLightCurveFile(path)
     elif mission == "TESS":
-        lcf_custom = TessLightCurveFile(path)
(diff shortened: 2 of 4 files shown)


gully:
This warning suppression is going great!  Thanks @texadactyl for guiding us to these issues.

I'm having trouble running the "raise error on warning trick" that @barentsen showed me for the special case of `RuntimeWarning`.  What API tree hierarchy should I replace in the `????` placeholder?

`py.test -v test_targetpixelfile.py -W error::????.RuntimeWarning`

gully:
By the way, the coverage is almost guaranteed to drop because we are now avoiding warnings in the first place.  The strategy to raise coverage should be to [catch the warnings](https://docs.pytest.org/en/latest/warnings.html#warns) with pytest, which I've done only sparingly in this work so far.

gully:
Hey @barentsen this PR is a step in the right direction and future-proofs some now-failing tests due to TESS data making Kepler-10 have more search results than before.  It's not 100% silencing all warnings, and it drops the coverage significantly.

gully:
Quick check of `py.test` with no remote data

**master** 117 passed, 47 skipped, 25 warnings in 17.38 seconds
**this PR** 118 passed, 47 skipped, 13 warnings in 17.78 seconds

barentsen:
Thanks @gully!  This is all look spiffy! :+1:

texadactyl:
@gully @barentsen 

My only concern is that by suppressing warnings (normally highly inadvisable) about log() or sqrt() failures, the test in question is not completing the intended exercize.  If you are confident that this is not the case, then okay.  

As I have said before, I understand that you folks are primarily scientists and need to prioritize your time.

barentsen:
@texadactyl Agreed, we have to be very careful with silencing `RuntimeWarning` messages related to math operations.  (And if we silence them, we need to have sufficient inline comments explaining why it is appropriate if at all.)

Fortunately the ~12 warnings that were fixed by this PR are unrelated to e.g. log or sqrt failures.  Those `RuntimeWarnings` are amongst the ~13 warnings that still remained after this PR was merged.

PR #596, opened 2019-09-18
Update stale tutorials

gully:
The tutorials get stale after a while.  This PR updates some of them.


gully:
Directive from Geert: 
- Do not worry about bloating the repo size by re-committing large images in the tutorials, we can deal with that later if it's a problem.
- Do not update the periodogram tutorials, those are handled in a new PR.

gully:
I think I've updated all tutorials now, mostly with superficial changes, but some new stuff like `stitch` and `tesscut`.  Content-level changes will probably come as part of looking at what tutorials we want to add.

I skipped the PRF photometry tutorial-- it takes a while to run and everything seemed to work as-is, not to mention the fairly expert-level of knowledge needed to understand that tutorial to completion.  I also skipped the Periodogram tutorials as instructed.

gully:
Codacy is complaining about my whitespace formatting in a README.  Other than that, this is ready for review.

The final outstanding task would be to: 
- [ ] run the `make` documentation to get these updates onto the docs.

barentsen:
@gully I suspect you have a setting somewhere which is causing your notebook figures to have a larger resolution than when e.g. Christina or I produce them, e.g. the first figure in `01-what-are-lightcurves` is 1511px wide.

I think that is fine and maybe even desirable, but it incidentally revealed a bug in the doc theme's CSS which is messing up the aspect ratio of large figures.  I am going to need to replace this:
```css
.nboutput img {
    max-width: 560px;
    height: auto;
}
```

with this:
```css
.nboutput img {
    max-width: 560px;
    height: auto !important;
}
```

barentsen:
CSS fixed!  For future reference, the change has to be made in the repository of the sphinx theme we use (`amunra`), i.e. this file was changed:

https://github.com/barentsen/amunra-sphinx-theme/blob/master/amunra_sphinx_theme/static/css/amunra.css

gully:
Great catch, thanks.  Yes my resolution defaults are configures to retina displays.

barentsen:
Please go ahead and merge if you are happy!

Issue #595, opened 2019-09-17
The search_tesscut() method appears broken


AngryKiwiNoises:
#### Problem description
Search_tesscut() rarely works properly for me. It used to work with at least some regularity. As of a couple weeks ago, the only hitch was that it would not download a cutout if that same cutout had been downloaded immediately prior. Now I can't get it to work at all, and I've kind of hit a wall on my research project because of it. I honestly can't think of anything I've changed that would cause this to happen. Most of the targets I need to analyze don't turn up any results using search_targetpixelfile(), otherwise I'd just use that. 

#### Example

```python
import lightkurve
import numpy
import matplotlib as plt
from lightkurve import search_tesscut

#Search for and download target pixel file
TICarray = ['120896927', '178155732', '410214986', '441462736', '55652896', '290131778', '29857954', '394137592']

i = 0

searchparam = 'TIC ' + TICarray[i]
print('Searching...')
targetpixelfile = search_tesscut(searchparam).download(quality_bitmask='none')

#If the target pixel file was found and downloaded
#change it to a light curve file, plot it, and put it in a folder
if type(targetpixelfile) == lightkurve.targetpixelfile.TessTargetPixelFile:
    
    print('Found. Plotting...')
```

#### Expected behavior
It should download a cutout if there are search results for the TIC ID.

#### Actual behavior
```python
SearchError: Unable to download FFI cutout. Desired target coordinates may be too near the edge of the FFI.
```
This happens even when I've been able to successfully download a cutout in the past.
#### Environment

-  platform (e.g. Linux, OSX, Windows): Windows 10
-  lightkurve version (e.g. 1.0b6): v1.1
-  installation method (e.g. pip, conda, source): conda

gully:
I confirm that there are one or two bugs here.  First, the the MAST TESScut service appears to timeout due to high demand, and the default SearchError misleadingly states that the main interpretation could be a field too close to the edge.  A fix should be to parse the HTTP code to see if a timeout occurred.  Another, unrelated problem may be occurring if you are using the development version of astroquery, but this latter problem has not been confirmed.

gully:
A temporary workaround-- you can visually inspect the `urllib` HTTP codes:

```python
lightkurve.log.setLevel('INFO')
```

If you see `504 Gateway Timeout error` HTTP status code, you can surmise that the MAST servers were getting overwhelmed with traffic.

gully:
[...]

PR #598, opened 2019-09-19
Support astroquery v0.3.10

barentsen:
Over in #595 it transpired that the current development version of `astroquery` renamed `MastClass()._resolve_object()` into `MastClass().resolve_object()`.

This PR makes Lightkurve robust against the change, i.e. it will try both options.

/cc @mustaric

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,8 @@
 - Changed `LightCurve.normalize()` to emit a warning if the light curve appears
   to be zero-centered. [#589]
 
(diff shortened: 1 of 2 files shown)


barentsen:
@gully can you review and merge if tests pass? Thx++

gully:
@mustaric this is now live on `master` so you can follow our [developer version installation guide](https://docs.lightkurve.org/about/install.html#installing-the-development-version) to get the bleeding edge lightkurve.  Depending how you previously installed lightkurve, you may wish to uninstall the old version.

PR #579, opened 2019-08-20
[WIP] Analytic Corrector for spacecraft motion systematics

christinahedges:
### What is it?
A corrector for removing systematics using linear algebra. 

### Working on:

* Error propagation
* BIC to understand the correct number of components
* Full integration testing for two new correctors
* Full TESS compatibility
* Tutorials

PR #599, opened 2019-09-20
[WIP] Add `DesignMatrix` and `RegressionCorrector` classes

barentsen:
Based on several discussions and impressive prototypes developed by @christinahedges (#579) and @nksaunders (#525, #561), it has emerged that `lightkurve.correctors` would benefit from having generic  `DesignMatrix` and `RegressionCorrector` classes to provide the generic building blocks for various systematic removal methods.

For example, it is envisaged that the SFF, PLD, and CBV corrector classes could eventually all subclass from `RegressionCorrector`, because they all use linear regression to correlate a light curve against a "design" matrix of column vectors which are known to correlate with a set of systematics/trends one seeks to remove.

## Envisaged usage example

```python
dm1 = DesignMatrix({'centroid_col': np.ones(size),
                    'centroid_row': np.ones(size)},
                    name='motion_systematics')
dm2 = DesignMatrix({'temperature': np.ones(size)},
                    name='spacecraft_systematics')
dmcol = DesignMatrixCollection([dm1, dm2])
corr = RegressionCorrector(lc, dmcol)
lc_corrected = corr.correct()
```

We envisage that a dedicated `DesignMatrix` and `DesignMatrixCollection` class will provide a number of useful features, e.g.:
```python
DesignMatrix.columns
DesignMatrix.shape
DesignMatrix.name
DesignMatrix.values
DesignMatrix.plot()
DesignMatrix.model(weights)  # given the fitted weights, what is the noise model?
DesignMatrix.split([x_values])  # accommodate "break points" in a light curve
DesignMatrix.whiten()  # e.g. transform columns to all have mean zero and std dev one?
DesignMatrix['column_name'].values

DesignMatrixCollection.shape
DesignMatrixCollection.info()
DesignMatrixCollection.plot()
DesignMatrixCollection['submatrix_name'].plot()
DesignMatrixCollection['submatrix_name'].model(weights)
DesignMatrixCollection['submatrix_name'][mask].model(weights)
DesignMatrixCollection['submatrix_name'].plot()
```


barentsen:
I think we may want to implement `RegressionCorrector` using scikit-learn's `sklearn.linear_model` package, so we can take advantage of advanced regularization methods.

Right now we solve the model coefficients as follows:
```python
    A = np.dot(X.T, X)
    B = np.dot(X.T, y)
    coefficients = np.linalg.solve(A, B)
```

The equivalent way to do this with scikit-learn is:
```python
    reg = linear_model.LinearRegression(fit_intercept=False).fit(X, y)
    coefficients = reg.coef_
```

The advantage of the latter is that we can replace `LinearRegression` with e.g. `Ridge` or `Lasso` in the example above to take advantage of L2 or L1 regularization (respectively) without having to implement it with numpy.

barentsen:
I am closing this PR because the commits have merge conflicts that made rebasing on master really tricky.  I squashed this PR into a single commit and opened PR #613 -- we will continue work there!

PR #602, opened 2019-10-03
Fix for Issue #594-- rectify before the reshape in echelle_plot

### LINKED ISSUE #594: Echelle diagrams are mirrored
#### Problem description
G'day all. I'm really enjoying the new seismology (and echelle!) code in lightkurve. However, it looks like the echelle diagram produced by plot_echelle is mirrored about the x-axis. You can see this with a simple test as below;

#### Example

```python
import lightkurve as lk
from scipy.signal import find_peaks

lc = lk.search_lightcurvefile('KIC 11615890', cadence='long').download_all().stitch().remove_nans()
pg = lc.to_periodogram()
ax = pg.plot()
plt.show()

cut = pg.frequency.value > 10
p = find_peaks(pg.power.value[cut], height=5, distance=100)[0]
ax.scatter(pg.frequency.value[cut][p], pg.power.value[cut][p], c='r', marker='v')
plt.show()

seis = pg.flatten().to_seismology()
seis.estimate_numax()
seis.estimate_deltanu()

ax = seis.plot_echelle()
ax.scatter(pg.frequency.value[cut][p] % seis.deltanu.value, pg.frequency.value[cut][p], c='r', marker='x')
```
#### Expected behavior
The peaks pulled directly from the light curve modulo the returned delta nu should align with the ridges. This is not the case.

Here are the relevant plots from above:
Screen Shot 2019-09-14 at 12 36 39 pm
Screen Shot 2019-09-14 at 12 38 17 pm

Shameless plug, but the ridges are properly aligned in my `echelle` code (though no where near as pretty!):
Screen Shot 2019-09-14 at 12 39 52 pm

#### Environment

-  platform (e.g. Linux, OSX, Windows): OSX
-  lightkurve version (e.g. 1.0b6): 1.1.1

gully:
Hi all,

This PR fixes #594.  Basically what was happening is that if you reshape with partially-filled rows, the faction of filled row phase shifts all subsequent rows by the size of the fraction.  You need to start and end your frequency array near where the modulus is zero.  

An alternative solution would have been to pad the partially-filled rows with NaNs up to the previous modulus=0 frequency.  That approach has some advantages, and could be implemented instead of this one if there is user interest.

Here is the revised figure.

image

Thanks to @danhey for raising this issue, and thanks for making `echelle` source code available.  I used it in the debugging process for this PR.

Cheers,
-gully

### DIFF
--- a/lightkurve/seismology/core.py
+++ b/lightkurve/seismology/core.py
@@ -13,6 +13,8 @@ from . import utils, stellar_estimators
 from ..periodogram import SNRPeriodogram
 from ..utils import LightkurveWarning, validate_method
 
+from scipy.signal import find_peaks
+
 log = logging.getLogger(__name__)
 
 __all__ = ['Seismology']
@@ -94,6 +96,7 @@ class Seismology(object):
             except AttributeError:
                 raise AttributeError("You need to call `Seismology.estimate_deltanu()` first.")
         return deltanu
+
 
     def plot_echelle(self, deltanu=None, numax=None,
                      minimum_frequency=None, maximum_frequency=None,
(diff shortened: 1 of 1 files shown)


gully:
Double thanks to Dan Hey for pointing out the find_peaks function, which I used in the solution to find the near-modulus-zero points.

lightkurve/seismology/core.py
```diff
@@ -210,17 +213,25 @@ def plot_echelle(self, deltanu=None, numax=None,
...
+        # This implementation will truncate any partially-filled rows.
+
+        mod_zeros = find_peaks( -1.0*(ff % deltanu) )[0]
+        start, approx_end = mod_zeros[0], mod_zeros[-1]
+        approx_samples = (approx_end - start)
+        n_rows = len(mod_zeros) - 1
+        n_columns =  approx_samples // n_rows
+        end = start + n_columns*n_rows
+
+        ep = np.reshape(pp[start : end], (n_rows, n_columns))
```
barentsen: Can you add a few more lines of comments to explain what is happening here?  I don't understand!

gully:
I can conceive of corner cases where my solution would go wrong.  An off-by-one error could result if your frequency grid starts and ends exactly with modulus=0 points?  Not sure.  My hunch is that scenario is rare based on typical deltanu precisions anyways, so I'm going to merge.

PR #609, opened 2019-10-11
Allow `bkjd_to_astropy_time` and `btjd_to_astropy_time` to accept any iterable (fixes #607)

### LINKED ISSUE #607: btjd_to_astropy_time() does not accept some float arrays
#### Problem description

Per [documentation](https://docs.lightkurve.org/api/lightkurve.utils.btjd_to_astropy_time.html#lightkurve.utils.btjd_to_astropy_time), `utils.btjd_to_astropy_time()` is supposed to accept an array of floats.

However, the function only accepts some types of arrays. See example below.

So either the documentation is inaccurate, or the implementation has bugs.

I expect the resolution would also apply to the peer `bkjd_to_astropy_time()`

#### Example

```python
import numpy as np
from lightkurve.utils import btjd_to_astropy_time

t1 = [1.1, 2.2, 3.3]
btjd_to_astropy_time(t1) # failed 

t2 = np.array(t1, dtype='float')
btjd_to_astropy_time(t2) # worked
```
The error is:
```
--> 370     jd = btjd + bjdref
...
TypeError: can only concatenate list (not "float") to list
```

#### Environment

-  platform (e.g. Linux, OSX, Windows): Windows, Jupyter, Python 3.7
-  lightkurve version (e.g. 1.0b6): 1.2.0

barentsen:
This PR addresses #607.

### DIFF
--- a/lightkurve/tests/test_utils.py
+++ b/lightkurve/tests/test_utils.py
@@ -111,7 +111,11 @@ def test_import():
 
 def test_btjd_bkjd_input():
     """Regression test for #607: are the bkjd/btjd functions tolerant?"""
-    for user_input in [0, [0], np.array([0])]:
+    # Kepler
+    assert bkjd_to_astropy_time(0).value == 2454833.
+    for user_input in [[0], np.array([0])]:
         assert_array_equal(bkjd_to_astropy_time(user_input).value, np.array([2454833.]))
-    for user_input in [0, [0], np.array([0])]:
+    # TESS
+    assert btjd_to_astropy_time(0).value == 2457000.
+    for user_input in [[0], np.array([0])]:
         assert_array_equal(btjd_to_astropy_time(user_input).value, np.array([2457000.]))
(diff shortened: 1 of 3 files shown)


lightkurve/tests/test_utils.py
```diff
@@ -106,3 +107,11 @@ def test_import():
...
     from .. import utils
     assert hasattr(utils, "btjd_to_astropy_time")
+
+
+def test_btjd_bkjd_input():
+    """Regression test for #607: are the bkjd/btjd functions tolerant?"""
+    for user_input in [0, [0], np.array([0])]:
+        assert_array_equal(bkjd_to_astropy_time(user_input).value, np.array([2454833.]))
+    for user_input in [0, [0], np.array([0])]:
+        assert_array_equal(btjd_to_astropy_time(user_input).value, np.array([2457000.]))
```
orionlee: It might be a nitpick... But if the input is a scalar (case `btjd == 0` in the test), I think typical developers would expect the output to be a scalar as well.

barentsen: That's a fair comment.  Let's make it behave that way!

PR #611, opened 2019-10-18
Add ability to download TESS FFI light curves from Eleanor

nksaunders:
I wrote this function to save myself time when I needed a corrected light curve from a TESS FFI, and I figured it would be useful to others! This PR adds the ability to download the Eleanor pipeline FFI light curves and return them as a `LightCurveCollection`.

There are a few features that definitely need be addressed:
- [ ] Add a `search_eleanor` function, and treat these search results the same as the current `SearchResult` objects
- [ ] All light curves are currently normalized to allow them to be stitched together without discontinuities, but this should probably be optional
- [ ] Ability to specify desired sectors instead of downloading all available sectors
- [ ] Ability to specify desired corrections applied to light curves by Eleanor and which light curves are returned

This raises the larger question — should `lightkurve` support high level science products and pipeline products? Feel free to weigh in or suggest changes!

### DIFF
--- a/lightkurve/search.py
+++ b/lightkurve/search.py
@@ -3,6 +3,7 @@ from __future__ import division
 import os
 import logging
 import warnings
+import re
 
 import numpy as np
 from astropy.table import join, Table, Row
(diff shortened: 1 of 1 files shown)


barentsen:
Thanks for pasting the example @nksaunders!

I note this feature wraps `eleanor.TargetData`.  Does that class do local systematics removal, or does it get a pre-computed light curve from somewhere?

barentsen:
Update: Lightkurve v2.x adds the `lightkurve.io` sub-package, which is intended to be the location where we implement readers for both official and community data products such as Eleanor.

I'd like to propose closing this PR for now in favor of opening a (future) new PR which adds Eleanor support to `lightkurve.io`.

Thanks for working on this!! :+1:

PR #612, opened 2019-10-18
include "ls_method" keyword in docstrings

keatonb:
Information about the `ls_method` keyword was missing from the documentation.


barentsen:
This is a perfect change.  Thanks @keatonb!!

PR #624, opened 2019-11-08
Add Oliver Hall's asteroseismology tutorial & related fixes

barentsen:
This PR:
* Adds an asteroseismology tutorial based on [this one](https://github.com/ojhall94/TASC5_lightkurve/blob/master/Notebooks/Rudy-example.ipynb) originally presented by @ojhall94 at KASC/TASC earlier this year.
* Renames the existing asteroseismology tutorial to "How to improve the signal-to-noise of Periodograms?" (it didn't actually discuss seismology!)
* Modifies `DesignMatrix` to only emit a warning if the rank of the matrix is less than half the number of columns, to avoid warnings being raised in *every* call to SFF.


gully:
Cool tutorials!  I'd say they're fine to launch as-is.  Here are some small comments/changes. 

Changes made to "How to optimize periodogram SNR" tutorial:
- Plots in cell 11 and 12 are very similar just slightly different x limits, so skip cell 11
- Silence `UserWarning: Creating legend with loc="best" can be slow with large amounts of data.` when calling `.plot()` by setting `loc='best'` in the legend.
- The final C16 aperture now selected with `threshold` to get a better SNR than `all`.

Other ideas:
- Could add references or links for SFF algorithm and asteroseismology.
- Plot figures could be at higher resolution to accommodate retina macs (dpi=300)

barentsen:
Thanks @gully!!

PR #613, opened 2019-10-18
Add new `DesignMatrix`, `RegressionCorrector`, and `SFFCorrector` classes

barentsen:
**This PR replaces #599 which suffered from complicated merge conflicts.**

Based on several discussions and impressive prototypes developed by @christinahedges (#579) and @nksaunders (#525, #561), it has emerged that `lightkurve.correctors` would benefit from having generic  `DesignMatrix` and `RegressionCorrector` classes to provide the generic building blocks for various systematic removal methods which can be formulated as a linear regression problem.

For example, it is envisaged that the SFF, PLD, and CBV corrector classes could eventually all subclass from `RegressionCorrector`, because they all use linear regression to correlate a light curve against a "design" matrix of column vectors which are known to correlate with a set of additivate systematics/trends one seeks to remove.

The `RegressionCorrector` will also provide an easy way to remove TESS background by regressing against ... background pixels!

## Envisaged usage example

```python
dm1 = DesignMatrix({'centroid_col': np.ones(size),
                    'centroid_row': np.ones(size)},
                    name='motion_systematics')
dm2 = DesignMatrix({'temperature': np.ones(size)},
                    name='spacecraft_systematics')
dmcol = DesignMatrixCollection([dm1, dm2])
corr = RegressionCorrector(lc, dmcol)
lc_corrected = corr.correct()
corr.diagnose()  # Produces helpful plots
```

We envisage that a dedicated `DesignMatrix` and `DesignMatrixCollection` class will provide a number of useful features, e.g.:
```python
DesignMatrix.columns
DesignMatrix.shape
DesignMatrix.name
DesignMatrix.values
DesignMatrix.plot()
DesignMatrix.model(weights)  # given the fitted weights, what is the noise model?
DesignMatrix.split([x_values])  # accommodate "break points" in a light curve
DesignMatrix.whiten()  # e.g. transform columns to all have mean zero and std dev one?
DesignMatrix['column_name'].values

DesignMatrixCollection.shape
DesignMatrixCollection.info()
DesignMatrixCollection.plot()
DesignMatrixCollection['submatrix_name'].plot()
DesignMatrixCollection['submatrix_name'].model(weights)
DesignMatrixCollection['submatrix_name'][mask].model(weights)
DesignMatrixCollection['submatrix_name'].plot()
```

PR #626, opened 2019-11-10
Update the SFF tutorials to reflect the refactored `SFFCorrector` class

barentsen:
We recently merged a refactored version of `SFFCorrector` into `master` (#613), but we forgot to update the three SFF-specific tutorials to reflect this change.

This PR updates these tutorials in preparation for the v1.4 release.


barentsen:
Note: it looks like we have to remove the tutorial "Replicating Vanderburg & Johnson 2014 using lightkurve" (`04-replicate-vanderburg-2014-lightkurve`) for now, for two reasons:
1. the new `SFFCorrector` works in a different (faster!) way using the `RegressionCorrector`;
2. the new corrector class does not provide the hidden methods the existing tutorial depends on.

PR #627, opened 2019-11-12
Make unit tests robust against TESSCut HTTP 504 timeouts

barentsen:
Occasionally, the TESSCut service will return a `504 Gateway Timeout error` response when it is overloaded.  This PR aims to prevent our unit tests from failing in this situation.

Curious to see whether the tests will pass!


barentsen:
(Ignoring the `Lightkurve-Linux-Tests` tests because I am still in the process of configuring those!)

PR #633, opened 2019-11-19
Ensure TessLightCurve(File) objects carry common properties

barentsen:
`TessLightCurveFile.get_lightcurve()` did not pass several common properties (mission, sector, camera, ccd, ra, dec) to the `TessLightCurve` objects it returned.  As a result, the collection `__repr__` did not work correctly, as shown below.  This PR fixes the issue.

## Before this PR
Screenshot from 2019-11-18 16-54-44

## After this PR
Screenshot from 2019-11-18 16-54-17

### DIFF
--- a/lightkurve/lightcurvefile.py
+++ b/lightkurve/lightcurvefile.py
@@ -299,10 +299,7 @@ class KeplerLightCurveFile(LightCurveFile):
         self.quality_mask = KeplerQualityFlags.create_quality_mask(
                                 quality_array=self.hdu[1].data['SAP_QUALITY'],
                                 bitmask=quality_bitmask)
-        try:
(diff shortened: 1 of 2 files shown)


lightkurve/lightcurvefile.py
```diff
@@ -491,7 +525,12 @@ def get_lightcurve(self, flux_type, centroid_type='MOM_CENTR'):
                 quality_bitmask=self.quality_bitmask,
                 cadenceno=self.cadenceno,
                 targetid=self.targetid,
-                label=self.hdu[0].header['OBJECT'])
+                label=self.hdu[0].header['OBJECT'],
```
christinahedges: why is this not `self.get_keyword('object')`

PR #604, opened 2019-10-08
Streamline the TPF and lightcurve base classes for broad use cases

gully:
Will make it possible to use lightkurve on Spitzer data, for example.  E.g. #603

PR #641, opened 2019-11-26
TPF factory overhaul with Spitzer demo

gully:
Altering the workflow to base off of `origin/spitzer_tpfs` rather than `gully/spitzer_tpfs` to get the GitHub actions to work.

Builds off #604


gully:
I migrated `from_fits_images` to `factory.py`.  Here is a path towards an overhaul:

1. (optionally) Migrate factory related tests to a new path `test_factory.py`
2. Update the tests with backwards-breaking behavior 
    - Check for and flag calls to heritage `from_fits_images`, raise deprecation warning/error
    - Test for this warning/error
3. Remove static method decorator from `from_fits_images`:
    - replace with regular method
    - Update `from_fits_images` internal references to factory with `self`
    - De-deplicate common arguments and key-word arguments between factory and `from_fits_images`.
    - Make new tests that look for the new behavior
4. Break up the large function into more manageable functions
5. Identify mission-specific functionality and replace with custom mission-specific dataloaders:
    - Make a Spitzer data loader
    - Make a Kepler/TESS data loader?
6. Verify that the delivered TPFs have standardized metadata mimicking Kepler/TESS, and potentially introducing new metadata header keywords if necessary
    - `RA_obj`,`Dec_obj`, `time_format`, `time_scale`
7. Specify how Spitzer WCS will map to poscorr
    - Should we use WCS differential or the `diff_x`, `diff_y` header columns?

lightkurve/factory.py
```diff
@@ -275,6 +281,154 @@ def _make_aperture_extension(self):
         hdu.header['EXTNAME'] = 'APERTURE'
         return hdu
 
+
+    @staticmethod
+    def from_fits_images(images, position, size=(11, 11), extension=1,
```
barentsen: I think this should now be a standalone function rather than a `@staticmethod`.

`TargetPixelFileFactory.from_fits_images()` does not make much sense because the method returns a `TargetPixelFile` rather than a `TargetPixelFileFactory`.  (And, in general, people find static methods confusing.)

barentsen:
I have consolidated this and the related PRs into a single, new PR, see here:
https://github.com/KeplerGO/lightkurve/pull/678

PR #645, opened 2019-12-04
Bug fix for early versions of astropy lacking support for new bin features

### LINKED ISSUE #644: Importing Lightkurve v1.5 fails to import due to missing `astropy.stats.calculate_bin_edges`
In Lightkurve v1.5 (via #629) we added `from astropy.stats import calculate_bin_edges` to `lightcurve.py` to enable custom binning in the `lc.bin()` method.

This is problematic because the `astropy.stats.calculate_bin_edges` function is only available in Astropy >3.1 and >2.10 (both released just ~one year ago), whereas Lightkurve currently only requires `astropy>=1.3`.

As a result, users with slightly older versions of Astropy installed will see the following error:

```python
>>> import lightkurve
ImportError                               Traceback (most recent call last)
<ipython-input-1-23246d7cf2d6> in <module>()
----> 1 import lightkurve as lk

~/dev/lightkurve/lightkurve/__init__.py in <module>()
     21 from .version import __version__
     22 from .prf import *
---> 23 from .lightcurve import *
     24 from .lightcurvefile import *
     25 from .correctors import *

~/dev/lightkurve/lightkurve/lightcurve.py in <module>()
     16 from copy import deepcopy
     17 
---> 18 from astropy.stats import sigma_clip, calculate_bin_edges
     19 from astropy.table import Table
     20 from astropy.io import fits

ImportError: cannot import name 'calculate_bin_edges' from 'astropy.stats' (/Users/gbarents/bin/anaconda/lib/python3.7/site-packages/astropy/stats/__init__.py)
```

gully:
Closes #644

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,15 @@
 1.6.0 (unreleased)
 ==================
 
+
+
+1.5.2 (2019-12-04)
+==================
+
+- Fixed a bug introduced in Lightkurve v1.5 which raised an error upon import
+  for installs with astropy version <=3.1 or <=2.10, due to missing
+  calculate_bin_edges. [#644]
+
 - Fixed a bug which caused the positions of stars in ``tpf.interact_sky()`` to
   be off by one pixel. [#638]
 
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -15,7 +15,7 @@ from scipy.stats import binned_statistic
 from matplotlib import pyplot as plt
 from copy import deepcopy
 
-from astropy.stats import sigma_clip, calculate_bin_edges
+from astropy.stats import sigma_clip
 from astropy.table import Table
 from astropy.io import fits
 from astropy.time import Time
@@ -898,8 +898,10 @@ class LightCurve(object):
             samples rather than index; overrides the `binsize=` if given.
             If ``bins`` is an int, it is the number of bins. If it is a list
             it is taken to be the bin edges. If it is a string, it must be one
-            of  'blocks', 'knuth', 'scott' or 'freedman'. See
-            `~astropy.stats.histogram` for a description of each method.
+            of  'blocks', 'knuth', 'scott' or 'freedman' if your AstroPy version
+            supports string input to `~astropy.stats.histogram`. Older versions
+            of AstroPy will instead default to the `~numpy.histogram_bin_edges`
+            method.
         method: str, one of 'mean' or 'median'
             The summary statistic to return for each bin. Default: 'mean'.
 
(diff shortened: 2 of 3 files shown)


gully:
Tests should fail for reasons known from discussions in the issue.

lightkurve/lightcurve.py
```diff
@@ -920,6 +922,17 @@ def bin(self, binsize=None, bins=None, method='mean'):
...
+        # Early versions of astropy do not have calculate_bin_edges
+        try:
+            from astropy.stats import calculate_bin_edges
+            raise ImportError #for debugging purposes
+        except ImportError:
+            from astropy import __version__ as astropy_version
+            log.info("Version {} of astropy does not support calculate_bin_edges,"
+                  " falling back to numpy analog.  Upgrade astropy to version to"
+                  " >3.1 or >2.10 for consistent string input".format(astropy_version))
+            from numpy import histogram_bin_edges as calculate_bin_edges
```
barentsen: Mmmm I really think we should make a firm choice to use either the astropy or the numpy version in Lightkurve, and not fall back from one to the other, because that might create surprising behavior!

gully: Agreed!

gully:
Okay, check out the new fix.  It'll be good enough for a bug fix, I think.  It won't crash upon import, only if you try to pass `bins=` to the `LightCurve.bin()` method or one of its subclasses.

gully:
I have no way to test for the ImportError, since we don't have old astropy versions as part of our test suite build matrix.  The existing tests *will* execute the workaround for `calculate_bin_edges`.

gully:
Yay! Tests pass.  I will update the changelog.

gully:
Should I base off of master or `1.5.1`?  I'll go with master for now, and move the interact_sky bug fix into `1.5.2`.

lightkurve/lightcurve.py
```diff
@@ -953,7 +964,9 @@ def bin(self, binsize=None, bins=None, method='mean'):
         if bins is None:  # use ``binsize```
             n_bins = self.flux.size // binsize
             bin_by_array = np.arange(len(self.time))
-            bin_edges = calculate_bin_edges(bin_by_array, bins=n_bins)
+            #bin_edges = calculate_bin_edges(bin_by_array, bins=n_bins)
+            bin_edges = np.linspace(bin_by_array.min(), bin_by_array.max(),
+                        n_bins + 1, endpoint=True)
```
barentsen: @gully Did you verify that `np.linspace` and `calculate_bin_edges` do the same thing here?  It would be good to know if they don't!

CHANGES.rst
```diff
@@ -1,6 +1,15 @@
...
+
+1.5.2 (2019-12-04)
+==================
+
+- Fixed a bug introduced in Lightkurve v1.5 which raised an error upon import
+  for installs with astropy version <=3.1 or <=2.10, due to missing
+  calculate_bin_edges. [#644]
+
 - Fixed a bug which caused the positions of stars in ``tpf.interact_sky()`` to
   be off by one pixel. [#638]
```
barentsen: Yep, this looks good, the interact_sky bug can be part of v1.5.2 as well.

lightkurve/lightcurve.py
gully: Yes, it's ultimately what the `bins=` parameter does in numpy when handed a number of bins.

gully:
Ok, ready to merge.

PR #661, opened 2020-01-08
Fix unit test failures with NumPy v1.8 and Scipy v1.4

barentsen:
The GitHub Actions unit tests are currently failing because there has been a small change in the error message yielded by the library's `assert_array_equal` function.

This PR should fix it!

### DIFF
--- a/lightkurve/tests/test_lightcurve.py
+++ b/lightkurve/tests/test_lightcurve.py
@@ -409,6 +409,7 @@ def test_cdpp_tabby():
     assert(np.abs(lc.estimate_cdpp() - lcf.header(ext=1)['CDPP6_0']) < 30)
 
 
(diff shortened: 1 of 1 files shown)


barentsen:
Mmmm there are still unit tests failures remaining.

It looks like `test_bin` in `test_lightcurve.py` is failing with the new SciPy v1.4, because a new exception is raised if the values being binned contain NaN:

https://github.com/scipy/scipy/blob/729efb880058cf841a561c733910f2aefb0a9cf6/scipy/stats/_binned_statistic.py#L518

We may have to revise `lc.bin()` to accommodate. /cc @gully

barentsen:
I've temporarily skipped two tests, `test_bin` and `test_bin_quality`, while we await to see how SciPy issue #11365 pans out.

PR #662, opened 2020-01-16
Some small improvements to the tutorials.

jcsmithhere:
Hello There!

Here's my first small attempt to work on the tutorials. In addition to my changes in the pull request I have several observation/recommendations in several of the tutorials listed below. I could work on some of these later but want to share them with you:

*********************
In "How to interactively inspect Target Pixel Files and Light Curves?" tutorial:

The first demo animation does not appear to be for HL Tau.

*********************
In Identifying rolling bands tutorial:

In the first video we are to see features at the “2 second” mark but the video does not have a second counter to reference where is 2 seconds. Might also be good to mention the rolling band is a roughly horizontal band in the video. It might be possible someone might confuse the asteroids with what you are referring to as rolling bands.

Also, emphasize this is ONLY a Kepler/K2 artifact.

*********************
In CorrectorClass tutorial:

Appending a constant to a design matrix is a standard thing to do. Consider doing that by default and allowing an option to NOT append to it.

When removing negative flux from the model why subtract off 5th percentile? Why not just subtract off min to get all values positive? Or, switch to a normalized basis, perform the systematic correction, then denormalize back, then adjust the secular term.

The example of Boyajian's Star might not be ideal for a TESS 2 min example because the PDC light curve has a lower CDPP than the RegressionCorrector residual (811 vs 836). In any event, we should probably provide a tutorial where we discuss how to diagnose when to use the PDC light curves versus generating your own. 

A lot more can be done with the pca method. Two ideas: 1) consider adding method to use BIC or AIC or something to select the number of principle components to use. 2) Consider taking cuts on the design matrix before applying PCA, for example, to only keep highly correlated design matrix columns before finding the principle components.

*********************
In SFFCorrector tutorial:

Link to tutorial to tune parameters is not working.

*********************
in KeplerCBVCorrector tutorial:

So, search_ligthcurvefile also reads in the CBV FITS files?

Removing NaNs changes the length of the time series. Doing so means the CBVs no longer line up with the light curve. We should fill gaps instead. Looks like the fill_gaps method can use some improvements.

In plot_cbvs, consider reversing the order of the offset so that the curves are in the same vertical order as the legend.


barentsen:
Thank you so much for your contribution @jcsmithhere !! :+1:

We will try to review and merge these suggestions asap.

(Note: we are having an issue with the unit tests right now, which explains why they are failing, i.e. all the "checks are failing" messages below are not caused by your changes in any way.)

christinahedges:
Hey Jeff,

Thanks for the input, I'm closing this PR and opening another one with these changes. In response to your comments:

> The first demo animation does not appear to be for HL Tau.

It is, the EPIC number is the K2 designation for HL Tau

> Appending a constant to a design matrix is a standard thing to do. Consider doing that by default and allowing an option to NOT append to it.

This is a comment on the API, not the tutorial. If you feel strongly about this I would encourage you to open an issue. Same with your other comment on additional metrics for RegressionCorrector, I'd be very happy to discuss them on a separate issue.

> When removing negative flux from the model why subtract off 5th percentile? Why not just subtract off min to get all values positive? Or, switch to a normalized basis, perform the systematic correction, then denormalize back, then adjust the secular term.

The minimum flux in some light curves could be zero, and so this is to be more robust to your average user. The tutorial should explain and illustrate why we are doing this.

> The example of Boyajian's Star might not be ideal for a TESS 2 min example because the PDC light curve has a lower CDPP than the RegressionCorrector residual (811 vs 836). In any event, we should probably provide a tutorial where we discuss how to diagnose when to use the PDC light curves versus generating your own.

Boyajian's star was chosen because it is popular and an easy case to demonstrate why scattered light needs to be removed. A tutorial on PDC flux would be excellent.

I think your CBV corrector comments are also on the API, not the tutorial, and I think we're solving them in your PR #672. 

I'm closing this and opening a new PR, thanks for playtesting our tutorials!! 👍

PR #667, opened 2020-01-24
[Bugfix] Fixing interact bug where empty pipeline aperture masks cause it to crash.

christinahedges:
In tesscut TPFs the aperture mask is set to be all False. In interact, if there is an all False aperture mask it doesn't actually work. I've added a fix that will create a threshold mask if there isn't one, or failing that use all the pixels.


barentsen:
This is a good change that can be merged, though I don't understand why the windows test is failing.  Re-running...

Issue #669, opened 2020-01-26
Add option to mean normalize light curves


jcsmithhere:
Some methods prefer zero-centered mean normalization over 1-centered median normalization.

Add option to LightCurve.normalize() to perform zero-centered median normalization.

1 - y / y_{bar}

During mean normalization it is also good to have an option to mask poor quality cadences when computing the mean. But also have an option to NOT mask any cadences. Doing so, can introduce a bias in the zero-centering, which can sometimes be bad.

PR #675, opened 2020-01-29
Add a `method='mean'` option to `LightCurve.normalize()`

barentsen:
This PR adds a `method='mean'` option to `LightCurve.normalize()` in response to #669.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 - Added a ``scale='linear'`` option to ``TargetPixelFile.interact()`` to show
   pixels using a linear stretch. The default is ``scale='log'``. [#664]
 
+- Added a ``method='mean'`` option to ``LightCurve.normalize()``. [#669]
+
 - Changed the behavior of sigma-clipping inside ``RegressionCorrector``. [#654]
 
 - Fixed a bug which caused ``LightCurve.show_properties()`` to raise a
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -605,11 +605,11 @@ class LightCurve(object):
                                 label=self.label,
                                 meta=self.meta)
 
-    def normalize(self, unit='unscaled'):
+    def normalize(self, unit='unscaled', method='median'):
         """Returns a normalized version of the light curve.
 
         The normalized light curve is obtained by dividing the ``flux`` and
-        ``flux_err`` object attributes by the by the median flux.
+        ``flux_err`` object attributes by the median or mean flux.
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -4,6 +4,8 @@
 - Added a ``scale='linear'`` option to ``TargetPixelFile.interact()`` to show
   pixels using a linear stretch. The default is ``scale='log'``. [#664]
 
+- Added a ``method='mean'`` option to ``LightCurve.normalize()``. [#669]
```
christinahedges: Good job!

lightkurve/lightcurve.py
```diff
@@ -605,11 +605,11 @@ def fold(self, period, t0=None, transit_midpoint=None):
                                 label=self.label,
                                 meta=self.meta)
 
-    def normalize(self, unit='unscaled'):
+    def normalize(self, unit='unscaled', method='median'):
```
christinahedges: Should we be able to pass a method here as well? e.g.
`lc.normalize(method=np.mean)`
`lc.normalize(method=astropy.stats.sigma_clipped_stats)`

lightkurve/lightcurve.py
```diff
@@ -642,6 +645,7 @@ def normalize(self, unit='unscaled'):
             from zero.
         """
         validate_method(unit, ['unscaled', 'percent', 'ppt', 'ppm'])
+        validate_method(method, ['median', 'mean'])
```
christinahedges: If we do allow users to pass a function to the `method` keyword argument this will break.

lightkurve/lightcurve.py
christinahedges: I understand this a bit of a fancy request, so I'm cool if we don't actually do it.

I think I would rather rely on `astropy`'s `sigma_clipped_stats` method for normalization, because then we could pass kwargs like `sigma` and `niters` in order to sigma clip and `mask` in order to mask...@barentsen what do you think?

lightkurve/lightcurve.py
```diff
@@ -673,8 +677,12 @@ def normalize(self, unit='unscaled'):
 
         # Create a new light curve instance and normalize its values
         lc = self.copy()
-        lc.flux = lc.flux / median_flux
-        lc.flux_err = lc.flux_err / median_flux
+        if method == 'median':
+            denominator = median_flux
+        elif method == 'mean':
+            denominator = np.nanmean(self.flux)
+        lc.flux = lc.flux / denominator
```
christinahedges: Please replace with 
`lc /= denominator` 
We have error propagation in the divide, and I'd rather we didn't have two places in the code where we could introduce bugs

PR #683, opened 2020-02-04
Update tutorials

christinahedges:
Reran all tutorials and updated based on comments from @jcsmithhere.


barentsen:
Thank you @christinahedges !!

PR #693, opened 2020-02-20
Add unit test: can files be opened from cache on Windows?

barentsen:
Via e-mail, a user reported encountering a `FileNotFoundError` when downloading a file from cache in Windows.  This test test attempts to reproduce the problem.

### DIFF
--- a/lightkurve/tests/test_search.py
+++ b/lightkurve/tests/test_search.py
@@ -309,3 +309,6 @@ def test_windows_cache_bug():
     lcf1 = search.download()
     lcf2 = search.download()
     assert "cache" in lcf2.path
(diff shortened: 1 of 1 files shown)


barentsen:
Interestingly, this unit test suggests there is no issue with opening files on Windows.  It even confirms the following path works:

```C:\Users\runneradmin\.lightkurve-cache/mastDownload\Kepler\kplr010848459_lc_Q011111110111011101\kplr010848459-2010009091648_llc.fits```

(note the backslash in the path, which is a small bug in astroquery)

I've asked the user to reconfirm that the issue is still taking place, and inform me of the Windows version if so!

PR #481, opened 2019-03-29
Cache TESSCut downloads to speed up file access

nksaunders:
Fixes #474: Build path name for TESSCut files before downloading and check if they exist. If the desired TESSCut has previously been downloaded (with the same coordinates and cutout size), the path to the target will be returned instead of re-downloading.

PR #696, opened 2020-02-25
Bugfix: Tesscut caching mechanism can return the wrong sector

barentsen:
The TessCut caching mechanism introduced via #481 contained a bug.  If a sector 1 cut-out was available in the cache, it would always return the sector 1 cut-out rather than the desired sector. 

This is because the following line:

https://github.com/KeplerGO/lightkurve/blob/cdf6acea8be68413d91df5a1064fc40771b1aeed/lightkurve/search.py#L367

should have said `sec['sector'] == sector` instead of `sec['sector'] == 1`.

This bug only affected master.  This feature was not part of the most recent release yet (v1.8.0), so no harm done!  This PR fixes the bug and adds a regression test.

### DIFF
--- a/lightkurve/search.py
+++ b/lightkurve/search.py
@@ -364,7 +364,7 @@ class SearchResult(object):
         # build path string name and check if it exists
         # this is necessary to ensure cutouts are not downloaded multiple times
         sec = TesscutClass().get_sectors(coords)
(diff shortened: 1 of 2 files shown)


nksaunders:
Thanks for catching this @barentsen! I realized my mistake last week and went to open a PR, but you beat me to it!

PR #714, opened 2020-03-24
[WIP] Added ability to query TPFs and light curves for known solar system objects.

christinahedges:
Some of the folks looking at TESS FFI light curves seem to have some trouble distinguishing between flares and solar system objects. This is a pretty easy query, so I made a helper function for lightkurve so that users can easily check whether there is a known SSO in their data.

@barentsen, what do you think?

`lc.query_solarsystem_objects()` will return a pandas dataframe with all the solar system objects in at the correct times. Default behavior is to run only on outliers in the light curve/target pixel file, but users can ask for `cadence_mask=all` or specify a boolean array to find particular times.


barentsen:
Thanks @christinahedges!  I think this would be very useful for sure!

To reduce complexity, I suggest:
1. Creating a more light-weight SkyBot wrapper function, e.g. ` utils._query_solar_system_objects(ra, dec, radius, location, time)`, which returns solar system objects given a cone/time/location.  This wrapper should know nothing about target pixel files or light curves.
2. Expand the `TargetPixelFile.query_solar_system_objects()` and `LightCurve.query_solar_system_objects()` methods to include the logic related to radius and cadence selection. These methods will call the lightweight wrapper.

To help convince us that it works, would it be useful to add a `mark_solar_system_objects` option to `TargetPixelFile.plot()` which would put a circle around known SSOs?

christinahedges:
Thanks for the great comments @barentsen, I've refactored based on your suggestions. I like the idea of having a way to plot these, but I don't want to bloat the plotting API with keywords that don't get used often. The other reason is I'm learning the service is a little fickle.

It seems like executing the query too often returns an empty dataframe. I've started downloading using astropy's caching feature. This is helpful, but if the first time you download the data you get a spurious NULL, it will be permanently in the cache which is hard to debug.

What do you think?

christinahedges:
Yeah so the tests are failing because the skybot service returned a null result today......

barentsen:
It would be good to understand the spurious NULL a bit more.  I wonder if the service actually returns an http error code or message.  If that's the case, it would be good to somehow raise an exception rather than silently return an empty frame.

Do you have an example URL we can look at that shows the issue?

lightkurve/lightcurve.py
```diff
@@ -1070,6 +1072,65 @@ def estimate_cdpp(self, transit_duration=13, savgol_window=101,
...
+
+        elif not isinstance(cadence_mask, np.ndarray):
+            raise ValueError('Pass a cadence_mask method or a cadence_mask')
+
+        if (location == 'kepler') | (location == 'k2'):
+            pixel_scale = 4
+        if location == 'tess':
+            pixel_scale = 27
+
+        if radius == None:
```
barentsen:
```suggestion
        if radius is None:
```

lightkurve/lightcurve.py
```diff
@@ -1070,6 +1072,65 @@ def estimate_cdpp(self, transit_duration=13, savgol_window=101,
...
+        if cadence_mask == 'all':
+            cadence_mask = np.ones(len(self.time)).astype(bool)
+
+        elif not isinstance(cadence_mask, np.ndarray):
+            raise ValueError('Pass a cadence_mask method or a cadence_mask')
+
+        if (location == 'kepler') | (location == 'k2'):
+            pixel_scale = 4
+        if location == 'tess':
+            pixel_scale = 27
```
barentsen: Hard-coding the pixel scale isn't very generic (e.g. what about TPFs created using the factory).

I suggest adding a `tpf.pixel_scale` property, which could `return tpf.hdu[1].header['11PC5'] * 3600.`.

christinahedges:
Yeah of course, here is a demo URL:

http://vo.imcce.fr/webservices/skybot/skybotconesearch_query.php?-mime=text&-ra=38.3919571663379&-dec=50.1529169838597&-bd=0.1&-loc=TESS&EPOCH=2458793.452819824

This is what I get on my first attempt:

image

Yesterday when I kept accessing it I got an error, but I didn't save it. I'll post here if I can reproduce it!

lightkurve/lightcurve.py
christinahedges: This seems like a larger fix, because we would need to then think about how this propagates to lightcurves. Pixel scale is also hard coded in the interact functions, can I add this as an issue to fix for our next version?

christinahedges:
I think the tests are now failing because of interact, not because of the PR. Also thank you for your comment about pixel scale, I think this might be an issue that needs fixing but maybe in a different PR? Interact also uses hardcoded pixel scales.

lightkurve/lightcurve.py
```diff
@@ -1070,6 +1072,64 @@ def estimate_cdpp(self, transit_duration=13, savgol_window=101,
         cdpp_ppm = np.std(mean) * 1e6
         return cdpp_ppm
 
+    def query_solar_system_objects(self, cadence_mask='outliers', radius=None, sigma=3, cache=True):
+        '''Query to find what solar system are in a lightcurve or targetpixel file.
```
barentsen: Please use triple double-quotes for docstring, and update/clarify the description (e.g. the reference to a target pixel file is old).

lightkurve/lightcurve.py
```diff
@@ -1070,6 +1072,64 @@ def estimate_cdpp(self, transit_duration=13, savgol_window=101,
...
+
+        Parameters
+        ----------
+        cadence_mask : str or bool
+            mask in time to select which frames or points should be searched for SSOs.
+            Default "outliers" will search for SSOs at points that are `sigma` from the mean.
+            "all" will search all cadences. Pass a boolean array with values of "True"
+            for times to search for SSOs.
+        radius : optional, float
+            Radius to search for bodies. If None, will set a sensible value.
```
barentsen: Can you clarify `sensible value` a bit?  I think it would be good to summarize what this function does in a few more sentences at the top of this docstring as well, e.g. the fact that it searches within a certain radius at certain times etc.

lightkurve/lightcurve.py
```diff
@@ -1070,6 +1072,64 @@ def estimate_cdpp(self, transit_duration=13, savgol_window=101,
...
+        This function relies on The Virtual Observatory Sky Body Tracker (skybot) service which can be found here:
+            http://vo.imcce.fr/webservices/skybot/
+
+        Parameters
+        ----------
+        cadence_mask : str or bool
+            mask in time to select which frames or points should be searched for SSOs.
+            Default "outliers" will search for SSOs at points that are `sigma` from the mean.
+            "all" will search all cadences. Pass a boolean array with values of "True"
+            for times to search for SSOs.
```
barentsen: I think we should emphasize the fact that only *outlier* values are searched by default at the top of the docstring, i.e. expand the docstring with a few more sentences.

lightkurve/lightcurve.py
```diff
@@ -1070,6 +1072,64 @@ def estimate_cdpp(self, transit_duration=13, savgol_window=101,
...
+        '''
+
+        location = self.mission.lower()
+
+        if isinstance(cadence_mask, str):
+            if cadence_mask == 'outliers':
+                cadence_mask = self.remove_outliers(sigma=sigma, return_mask=True)[1]
+
+        if cadence_mask == 'all':
+            cadence_mask = np.ones(len(self.time)).astype(bool)
```
barentsen: I think you wanted to indent lines 1107-1108?

lightkurve/lightcurve.py
```diff
@@ -1070,6 +1072,64 @@ def estimate_cdpp(self, transit_duration=13, savgol_window=101,
...
+            outliers.
+
+        Returns
+        -------
+        result : pandas.DataFrame
+            DataFrame containing the list objects in frames that were identified to contain
+            SSOs.
+        '''
+
+        location = self.mission.lower()
```
barentsen: Generic `LightCurve` objects do not have a mission keyword.  One solution is to make location a keyword argument which can default to the mission location if the mission keyword is set.

lightkurve/lightcurve.py
```diff
@@ -1070,6 +1072,64 @@ def estimate_cdpp(self, transit_duration=13, savgol_window=101,
...
+        elif not isinstance(cadence_mask, np.ndarray):
+            raise ValueError('Pass a cadence_mask method or a cadence_mask')
+
+        if (location == 'kepler') | (location == 'k2'):
+            pixel_scale = 4
+        if location == 'tess':
+            pixel_scale = 27
+
+        if radius is None:
+            radius = (15 * (pixel_scale))*u.arcsecond.to(u.deg)
```
barentsen: Why 15?  Even if it's an arbitrary choice, let's add a comment to say it is.

lightkurve/lightcurve.py
```diff
@@ -1070,6 +1072,64 @@ def estimate_cdpp(self, transit_duration=13, savgol_window=101,
...
+        for idx, time in enumerate(tqdm(self.astropy_time.jd[cadence_mask], desc='Querying for SSOs')):
+            res = _query_solar_system_objects(ra=self.ra, dec=self.dec, time=time,
+                                              location=location, radius=radius, cache=cache)
+            if isinstance(res, pd.DataFrame):
+                res['cadenceno'] = self.cadenceno[cadence_mask][idx]
+                if df is None:
+                    df = res
+                else:
+                    df = df.append(res)
+        return df
```
barentsen: Could we extend `utils._query_solar_system_objects` to accept an array of times, so we don't have to repeat the `df.append(res)`-related logic in both lightcurve.py and targetpixelfile.py?

lightkurve/lightcurve.py
christinahedges: Generic light curves don't have the `RA` or `Dec` attributes so it won't work anyway at the moment. I can make this specifically a method of `TESSLightCurve` and `KeplerLightCurve`?

christinahedges:
Thanks for the great comments @barentsen they really helped!

barentsen:
This PR needs a bit more work:
* I tried to make the docstring more explanatory in `LightCurve`, but `TargetPixelFile` has yet to be adapted in a similar way.
* Error handling isn't robust yet.  For example, when the service fails because the kernel doesn't cover the epoch, the user gets a very unclear error message.  Example:
```python
import lightkurve as lk
lc = lk.search_lightcurvefile(249780361).download().PDCSAP_FLUX
lc.query_solar_system_objects(sigma=10)
```
yields:
```
KeyError: "['Num', 'Mv', 'Name', 'Class'] not in index"
```

barentsen:
Recording here that it looks like any SkyBot searches with Kepler (C55) as the location are failing for dates prior to 2018.  We suspect SkyBot is using a limited Kepler spice kernel.

For example, this URL:
```http://vo.imcce.fr/webservices/skybot/skybotconesearch_query.php?-mime=text&-ra=285.679391&-dec=50.2413&-bd=0.016666666666666666&-loc=C55&EPOCH=2454953.6004491695```

Returns:
```
# Flag: -1
# Ticket: 166271270728715834
SkyBoT asteroid conesearch -> numIntAstromJ2000_para[1]: computeEphemeris: *** Error: epoch of computation not covered by the orbit:
*** ... requested epoch ...: 2454953.6019812063              
*** ... available time span: 2018-05-30T13:15:51.000 (UTC) - 2043-05-30T19:15:51.000 (UTC)
```

**Comment by barentsen** [comment]
I verified that the Kepler spacecraft position kernel posted at https://archive.stsci.edu/missions/k2/spice/spk_2018290000000_2018306220633_kplr.bsp works fine from launch (2009) through Dec 2019.

Code snippet I used to test the kernel:

```python
import spiceypy as spice

spice.furnsh("spk_2018290000000_2018306220633_kplr.bsp")  # Kepler
spice.furnsh("naif0012.tls")  # Leap seconds
spice.furnsh("de432s.bsp")  # Planets

utc = ['Apr 1, 2009', 'Dec 1, 2019']
et1, et2 = spice.str2et(utc[0]), spice.str2et(utc[1])
stepsize = 4000
times = [x*(et2-et1)/step + et1 for x in range(step)]

positions, _ = spice.spkpos('-227', times, 'J2000', 'NONE', 'SOLAR SYSTEM BARYCENTER')

x = positions[:, 0]
y = positions[:, 1]
z = positions[:, 2]

plt.plot(x, y)
```

christinahedges:
The SPICE kernel has been updated at skybot, these queries all now work. Hooray!

christinahedges:
Here's a demo use case in action, seems like it might be worth having a `return_mask` option:

image

barentsen:
This looks good to merge.  Thank you @christinahedges!

PR #715, opened 2020-03-24
Speed up TESS data product searches by a factor 10x

barentsen:
**Issue**: one of our beloved users (@dfm) pointed out to me that searches for TESS light curve and target pixel files have become incredibly slow.

**Cause**: a huge number of TESS FFIs records now exist in the MAST Portal database. Whenever someone calls `search_lightcurvefile` or `search_targetpixelfile`, Lightkurve unnecessarily downloads all the FFI records for the requested target before filtering them out locally.

**Solution**: we need to make sure we specify `dataproduct_type=['cube', 'timeseries']` as a parameter in our calls to `astroquery.mast.Observations.query_criteria()`. This prevents FFI records from being returned (they have `dataproduct_type='image'`).

This PR makes this change.  The speed up is pretty incredible.  For example, `search_lightcurvefile("Pi Men")` now takes 2 seconds instead of 2 minutes.

### Before:

Screenshot from 2020-03-24 14-32-34

### After:

Screenshot from 2020-03-24 14-36-20

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,9 @@
 1.9.1 (unreleased)
 ==================
 
+- Increased the speed of ``search_lightcurvefile()`` and
+  ``search_targetpixelfile()`` by a factor ~10. [#715]
+
 - Fixed a bug in `LightCurve.bin()` which caused the method to fail if the
   ``quality`` array has a floating point data type. [#705]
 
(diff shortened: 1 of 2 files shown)


barentsen:
@dfm The speed-up has been released as part of Lightkurve v1.9.1 last night.  Upgrading will help your code!

PR #722, opened 2020-04-04
[WIP] Added support for scipy.sparse in DesignMatrix and RegressionCorrector to speed up regression on large, sparse matrices

christinahedges:
`scipy.sparse` provides a great tool for doing matrix maths on very sparse matrices, which for very large/sparse matrices can speed up the following lines from `RegressionCorrector`:

```python
sigma_w_inv = X.T.dot(X/flux_err[:, None]**2)
B = X.T.dot(flux/flux_err**2)
```

(For a full Kepler light curve `X` can easily be 80000 x 2000.)

The implementation doesn't slow down `RegressionCorrector` for small values of X:
image

But for large values of X (i.e. with an entire Kepler dataset) we can increase the speed by a factor 2x:

image

This is a fairly minor tweak to `RegressionCorrector` and `DesignMatrix` to allow for sparse matrix maths. `DesignMatrix` still relies on users supplying a (non-sparse) matrix, pandas dataframe or dictionary. What this means in practice is that we're not getting any of the memory savings from using sparse, because we still store the large matrix in memory. We are only saving time during the `RegressionCorrector` call to `_fit_coefficients`. This is still worth it, but in future it would be great to improve the `DesignMatrix` API specifically for sparse matrices.

To do this in practice, I would need to remove the `patsy` dependence in `create_spline_dm`, because that function is relied on heavily in most detrending and does not currently support scipy sparse.

### DIFF
--- a/lightkurve/correctors/designmatrix.py
+++ b/lightkurve/correctors/designmatrix.py
@@ -69,7 +69,7 @@ class DesignMatrix():
                 self.columns = columns
             else:
                 if hasattr(X, 'columns'):
-                    self.columns = X.columns
+                    self.columns = list(X.columns)
 
             if self._sparse:
                 self.X = lil_matrix(X.values)
@@ -169,12 +169,12 @@ class DesignMatrix():
             A new design matrix with shape (n_rows, len(row_indices)*n_columns).
         """
         if not hasattr(row_indices, '__iter__'):
-            row_indices = list(row_indices)
+            row_indices = [row_indices]
+
+        # You can't split on the first or last index
+        row_indices = list(np.asarray(row_indices)[~np.in1d(row_indices, [0, self.shape[0]])])
         if len(row_indices) == 0:
             return self
-        if (row_indices == [0]) or (row_indices == [self.shape[0]]):
-            return self
-
         if inplace:
             dm = self
         else:
@@ -187,15 +187,17 @@ class DesignMatrix():
             dm.df = pd.concat([((dm.df * np.atleast_2d(np.in1d(x, idx).astype(int)).T)).add_suffix('_{}'.format(jdx)) for jdx, idx in enumerate(np.array_split(x, row_indices))], axis=1)
             non_zero = dm.df.sum(axis=0) != 0
             dm.df = dm.df[dm.df.columns[non_zero]]
-            dm.df = dm.df.values
+            dm.df = dm.df
             if not self._sparse:
                 dm.X = dm.df.values
+            dm.columns = list(dm.df.columns)
         if issparse(dm.X):
             dm.X = hstack([dm.X.multiply(lil_matrix(np.in1d(x, idx).astype(int)).T) for idx in np.array_split(x, row_indices)], format='lil')
             non_zero = dm.X.sum(axis=0) != 0
             non_zero = np.asarray(non_zero).ravel()
             dm.X = dm.X[:, non_zero]
-
+            if dm.columns is not None:
+                dm.columns = list(np.asarray([['{}_{}'.format(c, idx) for c in dm.columns] for idx in range(len(row_indices))]).ravel())
         dm.prior_mu = dm.prior_mu[non_zero]
         dm.prior_sigma = dm.prior_sigma[non_zero]
         return dm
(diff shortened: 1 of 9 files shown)


lightkurve/correctors/designmatrix.py
```diff
@@ -42,13 +43,16 @@ class DesignMatrix():
...
     def __init__(self, df, columns=None, name='unnamed_matrix', prior_mu=None,
-                 prior_sigma=None):
+                 prior_sigma=None, sparse=False):
         if not isinstance(df, pd.DataFrame):
             df = pd.DataFrame(df)
         if columns is not None:
             df.columns = columns
         self.df = df
         self.name = name
+        self._sparse = sparse
```
barentsen, line 53: Is there a reason for the underscore (i.e. `_sparse` rather than `sparse`)?  Given that it is a constructor argument, I think `sparse` should be a normal attribute.

lightkurve/correctors/designmatrix.py
```diff
@@ -224,12 +239,19 @@ def append_constant(self, prior_mu=0, prior_sigma=np.inf):
...
+            self.prior_sigma = np.append(self.prior_sigma, prior_sigma)
+            if self._sparse:
+                self._sparse_values = hstack([self.values, csr_matrix(np.ones(self.values.shape[0])).T], format='csr')
+            return self
+        else:
+            df = self.df.copy()
+            df.insert(self.df.shape[1], 'offset', 1, allow_duplicates=False)
+            return DesignMatrix(df, prior_mu=np.append(self.prior_mu, prior_mu),
+                                prior_sigma=np.append(self.prior_sigma, prior_sigma), name=self.name,
+                                sparse=self._sparse)
```
barentsen, line 254: There's quite a lot of code duplication in the function above.  Could we refactor it like this:
```python
if inplace:
    dm = self
else:
    dm = self.copy()
# insert pca logic here
return dm
```

lightkurve/correctors/designmatrix.py
```diff
@@ -212,10 +224,13 @@ def pca(self, nterms=6):
         # we find this to be too few, and that n_iter=10 is still fast but
         # produces more stable results.
         from fbpca import pca  # local import because not used elsewhere
-        new_values, _, _ = pca(self.values, nterms, n_iter=10)
-        return DesignMatrix(new_values, name=self.name)
+        if self._sparse:
+            new_values, _, _ = pca(self.values.toarray(), nterms, n_iter=10)
+        else:
+            new_values, _, _ = pca(self.values, nterms, n_iter=10)
+        return DesignMatrix(new_values, name=self.name, sparse=self._sparse)
```
barentsen, line 231: We could reduce code duplication above by adding a `self.get_values()` method to the class.  That method would take take care of returning either `self.values.toarray()` or `self.values` so other methods don't need to worry.

lightkurve/correctors/designmatrix.py
```diff
@@ -246,6 +268,8 @@ def _validate(self):
     @property
     def rank(self):
         """Matrix rank computed using `numpy.linalg.matrix_rank`."""
+        if self._sparse:
+                return np.linalg.matrix_rank(self.values.toarray())
         return np.linalg.matrix_rank(self.values)
```
barentsen, line 273: This method would also benefit from `self.get_values()`.

lightkurve/correctors/designmatrix.py
christinahedges: I don't want people to be able to (usually) set `sparse` to `True` after the fact, I'd like them to construct a new matrix. This is because some small amount of the logic for `DesignMatrixCollection` breaks if you parse it a non-sparse matrix first and then change it. So I made it "hidden" with a `_`. Does this sound like a good idea?

lightkurve/correctors/designmatrix.py
christinahedges: Yeah this is great, I like it.

lightkurve/correctors/designmatrix.py
christinahedges: Great idea.

christinahedges: Ok so I'm thinking about this. Sometimes we want the sparse values, because we're going to do the linear algebra. But sometimes we want the numpy array of values, e.g. in the case of PCA. How do you want me to distinguish between these cases?

christinahedges: e.g. there could be a 

`dm.X` for values OR sparse array, whatever is available 
and
`dm.values` for ALWAYS numpy arrays

barentsen: We should try to find a simple and consistent solution for this.  Should `DesignMatrix` perhaps always be sparse, with a method to populate a dense matrix as needed?

christinahedges:
I haven't fixed the issues yet, but I have managed to write a faster spline implementation with way better memory usage. This is the result of using the two different implementations (new "sparse" and old "dense" version) I'm fitting a spline with 1000 knots to the full light curve of Kepler-10. 

The new implementation is a 2x speed up and uses 25% of the memory.

image

christinahedges:
Below is an example for how the sparse splines affect the time to compute SFF for a full Kepler light curve

image

christinahedges:
This PR is being refactored in #732, and so I'm closing this for now.

Issue #713, opened 2020-03-19
Feature request/proposal: Easy access to exposure time property


dfm:
I'm often interested in accessing the exposure time value for light curve or target pixel objects so that I can integrate over the exposure time when computing a model to fit. I would normally do something like:

```python
tpf_file = lk.search_targetpixelfile("TIC 261136679", sector=1).download()
with tpf_file.hdu as hdu:
    tpf_hdr = hdu[1].header

texp = tpf_hdr["FRAMETIM"] * tpf_hdr["NUM_FRM"]
texp /= 60.0 * 60.0 * 24.0
```

but that doesn't seem very "lightkurvian". I couldn't find any easier way to access this and I thought I might implement it, but I figured I should get in touch first. I have a few questions:

1. Does something like this already exist?
2. If not, in terms of API, I'd suggest having a property on LC/TP files that gives a single scalar exposure time and then on the light curves and target pixel objects themselves have a column that is the same length as the number of cadences that gives the exposure time for each observation, because in general this doesn't need to be the same for all times. Thoughts?

Thanks for all your work on this!!

mrtommyb:
What property is most useful, the time spent collecting photons or the total duration of the cadence (including time spent reading out).

dfm:
That's actually a slightly subtle question. I believe that given the way Kepler and TESS collect data in smaller sub exposures we want the total duration including read out. In principle I would want to know exactly what times were spent collecting photons (not just how long!), but I don't think that anyone uses that in practice.

dfm:
Another way to put it - the thing that really matters is how the light curve changes from the beginning to the end of the exposure. If you're missing some data in the middle, that'll be a second order effect.

barentsen:
Thanks @dfm. I like the idea a lot!

I'd be very happy to start by adding a simple `exposure_time` property (?) to the `TargetPixelFile`, `LightCurveFile` and `LightCurve` classes.  Lightkurve aims to provide an interface to Kepler/TESS data products, and those data products always have a single exposure time, so a property makes sense.

[...]

PR #724, opened 2020-04-10
Adding extra_columns attribute to LightCurve objects

dfm:
This proposal adds a more general interface for having extra columns of data in a light curve that should be propagated through copies, appends, and subscripts. The existing implementation is brittle and adding a new column (for example as a subclass) is hard. This pull request should reproduce the current behavior, but simplify the procedure for adding new columns. This would lay the ground work for #713.

Note: it's not ready to merge yet (needs some tests!) but I wanted to post it here for feedback in the meantime.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,7 @@
 1.10.dev (unreleased)
 ==================
 
+- Added the ``extra_columns`` attribute to ``LightCurve`` objects. [#724]
 
 
 
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -146,7 +146,8 @@ class LightCurve(object):
                                  "".format(len(self), len(other)))
             if np.any(self.time != other.time):
                 warnings.warn("Two LightCurve objects with inconsistent time "
-                              "values are being added.")
+                              "values are being added.",
+                              LightkurveWarning)
             newlc.flux = self.flux + other.flux
             newlc.flux_err = np.hypot(self.flux_err, other.flux_err)
         else:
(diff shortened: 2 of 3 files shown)


barentsen:
@dfm I like this PR a lot. Thank you!! :+1:

Strengths of this PR:
* it makes the LightCurve class hierarchy more generic and consistent through fairly minor changes;
* it does not alter existing behavior;
* none of the changes prevent us from adopting an `astropy.TimeSeries`-based table solution in the future, should we wish to go down that (more complicated) route.

**Question**: how will this PR address #713?  In #713, we identified that there are certain single-valued attributes (e.g. `exposure_time`, `sector`, `mission`, `camera`) which could be thought of as columns when different light curves are stitched together.  Merging this PR would not enable these attributes to be treated as columns (yet?) because `extra_columns` is a hard-coded class property.  This PR does enable columns to be added by sub-classing `LightCurve`, is that the idea?

dfm:
Thanks! And yes, that's exactly what I'm doing in another project. This feature makes it much easier to implement subclasses without reimplementing things like the append method. 

It's worth noting that this doesn't quite duplicate the existing behavior in append. Now, it'll throw a warning if any of the light curves are missing any extra columns and returns more sensible results. Previously, the results of append would actually be really funky if any of the lc objects didn't have all the same columns. You would end up with things like cadenceno not being the same length as time. This is only a hypothetical currently because TESS and Kepler light curves have all the same extra columns, but this won't always be true for subclasses.

barentsen:
Thanks again @dfm!

I plan to add a few commits here today to add a test for the new behavior of `append`, to make sure `warnings.warn` uses type `LightkurveWarning`, and to add a change log entry. Other than that, this should be good to merge!

barentsen:
Oops, I don't seem able to push to this branch!

dfm:
Are you sure? You should be able to!

barentsen:
@dfm You're right!  Epic fail on my part!

dfm:
More like EPIC amirite?

barentsen:
Yes. More like ... EPIC. You are right!

PR #732, opened 2020-05-15
Added support for scipy.sparse in DesignMatrix and RegressionCorrector to speed up regression on large, sparse matrices

christinahedges:
This is a refactor of #722, which adds the new `SparseDesignMatrix` and `SparseDesignMatrixCollection` classes. These subclass `DesignMatrix` and `DesignMatrixCollection`, but use the `scipy.sparse` to speed up the calculations in `RegressionCorrector` and reduce memory. 

@barentsen has pointed out that it might be best to use `SparseDesignMatrix` only, and remove `DesignMatrix`, however
1) `SparseDesignMatrix` is slower if the matrix is actually dense
2) `DesignMatrix` is slightly more user friendly, in that it works with numpy arrays, pandas dataframes and dictionaries.

### DIFF
--- a/lightkurve/correctors/designmatrix.py
+++ b/lightkurve/correctors/designmatrix.py
@@ -329,7 +329,7 @@ class DesignMatrix():
                                   prior_mu=self.prior_mu,
                                   prior_sigma=self.prior_sigma)
 
-    def join(self, matrix):
+    def collect(self, matrix):
         """ Join two designmatrices, return a design matrix collection """
         return DesignMatrixCollection([self, matrix])
 
@@ -680,7 +680,7 @@ class SparseDesignMatrix(DesignMatrix):
     def __repr__(self):
         return '{} SparseDesignMatrix {}'.format(self.name, self.shape)
 
-    def join(self, matrix):
+    def collect(self, matrix):
         """ Join two designmatrices, return a design matrix collection """
         return SparseDesignMatrixCollection([self, matrix])
 
--- a/lightkurve/correctors/designmatrix.py
+++ b/lightkurve/correctors/designmatrix.py
@@ -129,7 +129,7 @@ class DesignMatrix():
         """Returns a random sample from the prior distribution."""
         return np.random.normal(self.prior_mu, self.prior_sigma)
 
-    def split(self, row_indices):
+    def split(self, row_indices, inplace=False):
         """Returns a new `.DesignMatrix` with regressors split into multiple
         columns.
 
@@ -156,19 +156,29 @@ class DesignMatrix():
         # Where do the submatrices begin and end?
         lower_idx = np.append(0, row_indices)
         upper_idx = np.append(row_indices, len(self.df))
+
         dfs = []
         for idx, a, b in zip(range(len(lower_idx)), lower_idx, upper_idx):
             new_columns = dict(
                 ('{}'.format(val), '{}'.format(val) + ' {}'.format(idx + 1))
                 for val in list(self.df.columns))
             dfs.append(self.df[a:b].rename(columns=new_columns))
-        new_df = pd.concat(dfs, axis=1).fillna(0)
+
         prior_mu = np.hstack([self.prior_mu for idx in range(len(dfs))])
         prior_sigma = np.hstack([self.prior_sigma for idx in range(len(dfs))])
-        return DesignMatrix(new_df, name=self.name, prior_mu=prior_mu,
-                            prior_sigma=prior_sigma)
 
-    def standardize(self):
+        if inplace:
+            dm = self
+        else:
+            dm = self.copy()
+        dm.df = pd.concat(dfs, axis=1).fillna(0)
+        dm.columns = dm.df.columns
+        dm.X = dm.df.values
+        dm.prior_mu = prior_mu
+        dm.prior_sigma = prior_sigma
+        return dm
+
+    def standardize(self, inplace=False):
         """Returns a new `.DesignMatrix` in which the columns have been
         median-subtracted and sigma-divided.
 
@@ -189,6 +199,7 @@ class DesignMatrix():
         `.DesignMatrix`
             A new design matrix with median-subtracted & sigma-divided columns.
         """
+
         ar = np.asarray(np.copy(self.df))
         ar[ar == 0] = np.nan
         # If a column has zero standard deviation, it will not change!
@@ -197,7 +208,13 @@ class DesignMatrix():
         std = np.atleast_2d(np.nanstd(ar, axis=0)[~is_const])
         ar[:, ~is_const] = (ar[:, ~is_const] - median) / std
         new_df = pd.DataFrame(ar, columns=self.columns).fillna(0)
-        return DesignMatrix(new_df, name=self.name)
+        if inplace:
+            dm = self
+        else:
+            dm = self.copy()
+        dm.df = new_df
+        dm.X = dm.df.values
+        return dm
 
     def pca(self, nterms=6):
         """Returns a new `.DesignMatrix` with a smaller number of regressors.
@@ -226,7 +243,7 @@ class DesignMatrix():
         new_values, _, _ = pca(self.values, nterms, n_iter=10)
         return DesignMatrix(new_values, name=self.name)
 
-    def append_constant(self, prior_mu=0, prior_sigma=np.inf):
+    def append_constant(self, prior_mu=0, prior_sigma=np.inf, inplace=False):
         """Returns a new `.DesignMatrix` with a column of ones appended.
 
         Returns
@@ -235,12 +252,16 @@ class DesignMatrix():
             New design matrix with a column of ones appended. This column is
             named "offset".
         """
+        if inplace:
+            dm = self
+        else:
+            dm = self.copy()
         extra_df = pd.DataFrame(np.atleast_2d(np.ones(self.shape[0])).T, columns=['offset'])
-        new_df = pd.concat([self.df, extra_df], axis=1)
-        prior_mu = np.append(self.prior_mu, prior_mu)
-        prior_sigma = np.append(self.prior_sigma, prior_sigma)
-        return DesignMatrix(new_df, name=self.name,
-                            prior_mu=prior_mu, prior_sigma=prior_sigma)
+        dm.df = pd.concat([self.df, extra_df], axis=1)
+        dm.X = dm.df.values
+        dm.prior_mu = np.append(self.prior_mu, prior_mu)
+        dm.prior_sigma = np.append(self.prior_sigma, prior_sigma)
+        return dm
 
     def _validate(self, rank=True):
         """Helper function for validating."""
@@ -278,7 +299,7 @@ class DesignMatrix():
     @property
     def shape(self):
         """Tuple specifying the shape of the matrix as (n_rows, n_columns)."""
-        return self.df.shape
+        return self.X.shape
 
     @property
     def values(self):
@@ -292,26 +313,28 @@ class DesignMatrix():
         return '{} DesignMatrix {}'.format(self.name, self.shape)
 
     def to_sparse(self):
+        """ Convert a dense DesignMatrix object to a SparseDesignMatrix
+
+        The values of this design matrix will be converted to a
+        `scipy.sparse.csr_matrix`, which stores the values in a
+        lower memory matrix. This is not recommended for dense matrices.
+        """
         return SparseDesignMatrix(csr_matrix(self.values), name=self.name,
                             columns=self.columns, prior_mu=self.prior_mu,
                              prior_sigma=self.prior_sigma)
 
     def join(self, matrix):
         """ Join two designmatrices, return a design matrix collection """
-        return self.__class__([self, matrix])
+        return DesignMatrixCollection([self, matrix])
 
-    def multichoose(self):
-        """ Create products of each vector in the design matrix """
-        raise NotImplementedError
 
 
 class DesignMatrixCollection():
     """A set of design matrices."""
-    def __init__(self, matrices, warn=True):
+    def __init__(self, matrices):
         if np.any([issparse(m.X) for m in matrices]):
-            if warn:
-                warnings.warn(('Some matrices are `SparseDesignMatrix` objects. '
-                                'Sparse matrices will be converted to dense matrices.'), LightkurveWarning)
+            warnings.warn(('Some matrices are `SparseDesignMatrix` objects. '
+                            'Sparse matrices will be converted to dense matrices.'), LightkurveWarning)
             dense_matrices = []
             for m in matrices:
                 if isinstance(m, SparseDesignMatrix):
@@ -437,7 +460,7 @@ class DesignMatrixCollection():
                     ''.join(['\t{}\n'.format(i.__repr__()) for i in self])
 
 
-    def flatten(self, name=None):
+    def to_designmatrix(self, name=None):
         """Flatten a DesignMatrixCollection into a DesignMatrix"""
         if name is None:
             name = self.matrices[0].name
@@ -447,7 +470,7 @@ class DesignMatrixCollection():
 class SparseDesignMatrix(DesignMatrix):
     """A matrix of column vectors for use in linear regression.
 
-    This class is similar to the `lk.SparseDesignMatrix` class, but uses the
+    This class is similar to the `DesignMatrix` class, but uses the
     `scipy.sparse` library to improve speed in the case of sparse matrices.
 
     The purpose of this class is to provide a convenient method to interact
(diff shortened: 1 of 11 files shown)


lightkurve/correctors/designmatrix.py
```diff
@@ -264,16 +286,44 @@ def values(self):
...
+    def multichoose(self):
+        """ Create products of each vector in the design matrix """
+        raise NotImplementedError
+
 
 class DesignMatrixCollection():
     """A set of design matrices."""
-    def __init__(self, matrices):
-        self.matrices = matrices
+    def __init__(self, matrices, warn=True):
```
barentsen, line 310: The `warn` parameter should be removed.  We should stick to one of two options for controlling warnings:
* If a user is making a mistake and should fix their code, use `warnings.warn("message", LightkurveWarning)` and let them silence it using `catch_warnings()` if they want to.  As a general rule we should only emit warnings if we think the user HAS to fix their code (e.g. deprecation warning).  We shouldn't provide convenient options to silence these.
* To record messages to help diagnose what Lightkurve is doing and to help the user, use `log.debug()` or `log.info()`.  The user can turn these on or off using `lk.log.setLevel()`.
(Details: 
https://github.com/KeplerGO/lightkurve/issues/691#issuecomment-586016365)

I'd use `log.debug()`, unless you think this is really stupid, in which case always emit a warning.

lightkurve/correctors/designmatrix.py
```diff
@@ -379,44 +429,396 @@ def __getitem__(self, key):
...
-        [d._validate() for d in self]
+    def validate(self):
+        [d.validate() for d in self]
 
     def __repr__(self):
         return 'DesignMatrixCollection:\n' + \
                     ''.join(['\t{}\n'.format(i.__repr__()) for i in self])
 
 
+    def flatten(self, name=None):
```
barentsen, line 440: Could `to_designmatrix()` be a more intuitive name?

lightkurve/correctors/designmatrix.py
```diff
@@ -379,44 +429,396 @@ def __getitem__(self, key):
...
+        """Flatten a DesignMatrixCollection into a DesignMatrix"""
+        if name is None:
+            name = self.matrices[0].name
+        return self._child_class(self.X, columns=self.columns, prior_mu=self.prior_mu, prior_sigma=self.prior_sigma,
+                            name=name)
+
+class SparseDesignMatrix(DesignMatrix):
+    """A matrix of column vectors for use in linear regression.
+
+    This class is similar to the `lk.SparseDesignMatrix` class, but uses the
```
barentsen, line 450:
```suggestion
    This class is similar to the `DesignMatrix` class, but uses the
```

lightkurve/correctors/designmatrix.py
```diff
@@ -379,44 +429,396 @@ def __getitem__(self, key):
...
+        # nterms cannot be langer than the number of columns in the matrix
+        if nterms > self.shape[1]:
+            nterms = self.shape[1]
+        # We use `fbpca.pca` instead of `np.linalg.svd` because it is faster.
+        # Note that fbpca is randomized, and has n_iter=2 as default,
+        # we find this to be too few, and that n_iter=10 is still fast but
+        # produces more stable results.
+        from fbpca import pca  # local import because not used elsewhere
+        new_values, _, _ = pca(self.values, nterms, n_iter=10)
+        return SparseDesignMatrix(type(self.X)(new_values), name=self.name, **kwargs)
```
barentsen, line 612: A lot of code here is duplicated from `DesignMatrix.pca()`.

It would be better practice to call the parent method, e.g.:
```python
dense_dm = super().pca(nterms)
return SparseDesignMatrix(type(self.X)(dense_dm.X), name=self.name)
```

or perhaps

```python
return super().pca(nterms).to_sparse()
```

If that doesn't work, both methods could at least share the same helper function.

Also note: `SparseDesignMatrix.pca()` accepts `**kwargs` and `DesignMatrix.pca()` doesn't.  When overriding a method, changing the function signature should be avoided if possible.

lightkurve/correctors/designmatrix.py
```diff
@@ -379,44 +429,396 @@ def __getitem__(self, key):
...
+
+    def append_constant(self, prior_mu=0, prior_sigma=np.inf, inplace=False):
```
barentsen, line 614: The parent class does not have an `inplace` parameter.  We should try to be consistent *if* possible.

lightkurve/correctors/designmatrix.py
```diff
@@ -379,44 +429,396 @@ def __getitem__(self, key):
...
+    def __getitem__(self, key):
+        loc = np.where(np.asarray(self.columns) == key)[0]
+        if len(loc) == 0:
+            raise ValueError('No such column as `{}`.'.format(key))
+        return self.X[:, loc].toarray()
+
+    @property
+    def shape(self):
+        """Tuple specifying the shape of the matrix as (n_rows, n_columns)."""
+        return self.X.shape
```
barentsen, line 641: I think `DesignMatrix.shape` could adopt this implementation?  In that case, you won't have to override it here.

lightkurve/correctors/designmatrix.py
```diff
@@ -379,44 +429,396 @@ def __getitem__(self, key):
...
+
+    def __repr__(self):
+        return '{} SparseDesignMatrix {}'.format(self.name, self.shape)
+
+    def join(self, matrix):
+        """ Join two designmatrices, return a design matrix collection """
+        return SparseDesignMatrixCollection([self, matrix])
```
barentsen, line 648: `join` can be misunderstood for concatenation or cross product in this context.  Do we need this method?  Is it useful?

lightkurve/correctors/designmatrix.py
```diff
@@ -379,44 +429,396 @@ def __getitem__(self, key):
...
+
+    def to_dense(self):
+        return DesignMatrix(self.values, name=self.name,
+                            columns=self.columns, prior_mu=self.prior_mu,
+                             prior_sigma=self.prior_sigma)
+
+
+class SparseDesignMatrixCollection(DesignMatrixCollection):
+    """A set of design matrices."""
+    def __init__(self, matrices, warn=True):
```
barentsen, line 658: I'd get rid of `warn` here as well.

lightkurve/correctors/designmatrix.py
```diff
@@ -379,44 +429,396 @@ def __getitem__(self, key):
...
 ####################################################
 # Functions to create commonly-used design matrices.
 ####################################################
 
-def create_spline_matrix(x, n_knots=20, degree=3, name='spline',
-                         include_intercept=False):
-    """Returns a `.DesignMatrix` which models splines using `patsy.dmatrix`.
 
+@jit(nopython=True)
+def basis(x, degree, i, knots):
```
barentsen, line 712: Would this be a better name?

```suggestion
def spline_basis_vector(x, degree, i, knots):
```

lightkurve/correctors/designmatrix.py
```diff
@@ -379,44 +429,396 @@ def __getitem__(self, key):
...
+    degree : int
+        Degree of spline to calculate basis for
+    i : int
+        The index of the knot to calculate the basis for
+    knots : np.ndarray
+        Array of all knots
+    Returns
+    -------
+    B : np.ndarray
+        A vector of same length as x containing the spline basis for the ith knot
```
barentsen, line 727: Adding a short `Examples` section to the docstring could be really great for these spline functions.

lightkurve/correctors/designmatrix.py
```diff
@@ -231,28 +242,39 @@ def append_constant(self, prior_mu=0, prior_sigma=np.inf):
         return DesignMatrix(new_df, name=self.name,
                             prior_mu=prior_mu, prior_sigma=prior_sigma)
 
-    def _validate(self):
-        """Raises a `LightkurveWarning` if the matrix has a low rank."""
+    def _validate(self, rank=True):
```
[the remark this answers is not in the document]
christinahedges: Yes, we need to ignore the rank warning when we use sparse. Calculating the rank warning requires we evaluate the whole matrix (using scipy.sparse.toarray) which completely negates the value of using sparse matrices. I did look into a sparse way of calculating rank, but i think it can't be done.

lightkurve/correctors/designmatrix.py
```diff
@@ -264,16 +286,44 @@ def values(self):
...
         return '{} DesignMatrix {}'.format(self.name, self.shape)
 
+    def to_sparse(self):
+        return SparseDesignMatrix(csr_matrix(self.values), name=self.name,
+                            columns=self.columns, prior_mu=self.prior_mu,
+                             prior_sigma=self.prior_sigma)
+
+    def join(self, matrix):
+        """ Join two designmatrices, return a design matrix collection """
+        return self.__class__([self, matrix])
```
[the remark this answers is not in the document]
christinahedges: Sorry this was an under-development function that i never actually developed, I had an idea for doing a pandas-like join, but i think we don't need the functionality!

lightkurve/correctors/designmatrix.py
```diff
@@ -264,16 +286,44 @@ def values(self):
...
+
+    def multichoose(self):
+        """ Create products of each vector in the design matrix """
+        raise NotImplementedError
```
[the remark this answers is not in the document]
christinahedges: 👍

lightkurve/correctors/designmatrix.py
christinahedges: Sure! This is actually a hidden function I'd rather not have the user know about (it's hard to use) so I think I'll call it `_spline_basis_vector`

lightkurve/correctors/designmatrix.py
christinahedges: I'm leaving it as a warning. Mixing sparse and non-sparse matrices is always dangerous, but there are reasons that a user may want to do it. Thanks for the explanation!

lightkurve/correctors/designmatrix.py
christinahedges: The purpose of join was to reduce this line

`lk.SparseDesignMatrix([dm1, dm2])`
to
`dm1.join(dm2)`

lightkurve/correctors/designmatrix.py
christinahedges: It's hard to add an example here because this is a recursive function, what sort of example were you thinking?

lightkurve/correctors/designmatrix.py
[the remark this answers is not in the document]
christinahedges: The join I implemented there was wrong, it is supposed to be 
`lk.DesignMatrixCollection([self, matrix])`

It's a convenience function to make the "correct" DesignMatrixCollection object.

lightkurve/correctors/designmatrix.py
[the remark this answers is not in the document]
barentsen: That makes sense!  May be worth putting this in the docstring.  It would prevent people like me activating the rank warning in the future without thinking!

lightkurve/correctors/designmatrix.py
barentsen: Even though it's not a public function, I think it's important to add a few more words to the docstring, e.g. a link to the equation (e.g. Wikipedia) that's being implemented, or a short example showing what it returns and why.

lightkurve/correctors/designmatrix.py
barentsen: I wonder if calling `lk.SparseDesignMatrixCollection([dm1, dm2])` is actually better here, because it's not intuitive to me that `dm1.join(x)` returns a collection rather than just another design matrix?

lightkurve/correctors/tests/test_regressioncorrector.py
```diff
@@ -57,28 +57,29 @@ def test_sinusoid_noise():
     true_lc = LightCurve(time, true_flux, flux_err=0.1*np.ones(size))
     # Noisy light curve has a sinusoid single added
     noisy_lc = LightCurve(time, true_flux+noise, flux_err=true_lc.flux_err)
-    dm = DesignMatrix({'noise': noise,
+    design_matrix = DesignMatrix({'noise': noise,
                        'offset': np.ones(len(time))},
                       name='noise_model')
```
barentsen, line 62: :point_up:  It's just a test, but let's align the indents with the opening delimiters here -- makes the code :100:x more readable to me.

lightkurve/correctors/tests/test_tesspldcorrector.py
```diff
@@ -1,18 +1,22 @@
...
+import numpy as np
 
 from ... import open as lkopen
 from .. import TessPLDCorrector
 
 
-TESS_SIM = ("https://archive.stsci.edu/missions/tess/ete-6/tid/00/000"
-            "/004/176/tess2019128220341-0000000417699452-0016-s_tp.fits")
+TESS_DATA = ("https://mast.stsci.edu/api/v0.1/Download/file/?uri=mast:"
+            "TESS/product/tess2018319095959-s0005-0000000238201762-0125-s_tp.fits")
```
barentsen, line 11: Sounds good!!

lightkurve/correctors/designmatrix.py
```diff
@@ -3,17 +3,20 @@
 These classes are intended to make linear regression problems with a large
```
barentsen, line 3: :point_up: Let's update line 1 here.

lightkurve/correctors/designmatrix.py
```diff
@@ -224,56 +252,101 @@ def append_constant(self, prior_mu=0, prior_sigma=np.inf):
...
+    def to_sparse(self):
+        """ Convert a dense DesignMatrix object to a SparseDesignMatrix
+
+        The values of this design matrix will be converted to a
+        `scipy.sparse.csr_matrix`, which stores the values in a
+        lower memory matrix. This is not recommended for dense matrices.
+        """
+        return SparseDesignMatrix(csr_matrix(self.values), name=self.name,
+                            columns=self.columns, prior_mu=self.prior_mu,
+                             prior_sigma=self.prior_sigma)
```
barentsen, line 324: Let's align the indent with the opening delimiter here :point_up:, and perhaps only have one parameter on each line.

lightkurve/correctors/designmatrix.py
```diff
@@ -224,56 +252,101 @@ def append_constant(self, prior_mu=0, prior_sigma=np.inf):
...
+
+    def join(self, matrix):
+        """ Join two designmatrices, return a design matrix collection """
+        return DesignMatrixCollection([self, matrix])
+
+
 
 class DesignMatrixCollection():
     """A set of design matrices."""
```
barentsen, line 333: Let's take the opportunity to expand this docstring a tiny bit -- it will be prominent in the online docs.

lightkurve/correctors/designmatrix.py
```diff
@@ -224,56 +252,101 @@ def append_constant(self, prior_mu=0, prior_sigma=np.inf):
...
+                if isinstance(m, SparseDesignMatrix):
+                    dense_matrices.append(m.copy().to_dense())
+                else:
+                    dense_matrices.append(m)
+            self.matrices = dense_matrices
+        else:
+            self.matrices = matrices
+        self.X = np.hstack(tuple(m.X for m in self.matrices))
+        self._child_class = DesignMatrix
+        self.validate()
```
barentsen, line 349: It's fairly obvious to use now, but I'd add 2 or 3 lines of comments in this constructor along the lines of:

`# This collection is designed for dense matrices, so we raise a warning if a SparseDesignMatrix is passed`

and

`# If a sparse matrix is parsed, we convert it do a dense one for user convenience`

lightkurve/correctors/designmatrix.py
```diff
@@ -224,56 +252,101 @@ def append_constant(self, prior_mu=0, prior_sigma=np.inf):
...
+            dense_matrices = []
+            for m in matrices:
+                if isinstance(m, SparseDesignMatrix):
+                    dense_matrices.append(m.copy().to_dense())
+                else:
+                    dense_matrices.append(m)
+            self.matrices = dense_matrices
+        else:
+            self.matrices = matrices
+        self.X = np.hstack(tuple(m.X for m in self.matrices))
```
barentsen, line 347: Is this line necessary?  I ask for two reasons:
* I notice `self.X` is only accessed inside `to_designmatrix()`, so maybe `hstack` can just be called there.
* I notice `RegressionCorrector.X` is used to hold the `DesignMatrixCollection`, while here it is used to hold a stacked array.  We end up using `self.X.X` inside `correct()`, which is very confusing. Is there a way we can simplify this?

lightkurve/correctors/designmatrix.py
christinahedges: This has been changed to collect

PR #665, opened 2020-01-16
TPF objects can now be added, subtracted, multiplied and divided by numpy arrays

christinahedges:
It's useful to be able to do simple maths with TPF objects in order to e.g. remove background. I've added the capability to do that.

There are special rules depending on the shape of the object you try to add/subtract/divide/multiply. 

- If it's a point, it's add/subtract/divide/multiply-ed from all values.
- If it's a vector with **length time** , it's add/subtract/divide/multiply-ed from each frame.
- If it's a vector **NOT of length time**  it raises a value error
- If it's a 2D array with **shape npixels x mpixels** , it's add/subtract/divide/multiply-ed from each pixel.
- If it's a 2D array with **shape NOT npixels x mpixels** , it raises a value error
- If it's a 3D array with **shape time x npixels x mpixels** , it's add/subtract/divide/multiply-ed from each value.
- If it's a 3D array with **shape NOT time x npixels x mpixels** , it raises a value error

image

There are lovely accompanying tests.

PR #733, opened 2020-05-15
Drop support for Python 2

barentsen:
It has becoming increasingly difficult to support Python 2. For example, it is currently the cause of the unit test failures over in #732, and it is preventing #665 from getting merged.

This PR proposes to drop support for Python 2.

Consequences include:
* We will bump the Lightkurve version number to v2.0.
* We can now require `astropy>=3.1` and `astroquery>=0.3.10`.

### DIFF
--- a/.github/workflows/lightkurve-tests.yml
+++ b/.github/workflows/lightkurve-tests.yml
@@ -1,13 +1,12 @@
 # GitHub Actions workflow for Lightkurve's continuous integration.
 #
-# This file configures seven different jobs:
-# 1) pytest on Linux, Python 2.7.
-# 2) pytest on Linux, Python 3.7.
-# 3) pytest on Linux, Python 3.8, with --remote-data enabled (i.e. get data from MAST).
-# 4) pytest on Linux, Python 3.8, without installing optional dependencies.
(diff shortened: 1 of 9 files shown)


mrtommyb:
this should probably be an issue, rather than a pull request to get understanding and concurrence from all parties.

barentsen:
I like a PR for this because it helps us understand exactly what would change!

I'm happy to discuss concerns and consequences of course.  How should I go about getting concurrence?

christinahedges:
Down with Python 2. ⬇️ Praise Python 3 🙏 .

barentsen:
I released Lightkurve v1.11.0 on pypi and created a special branch for it (`v1.11.x`).  The special branch will allow us to continue maintaining Python 2 support (a.k.a Lightkurve v1) if the need arises.

I plan to go ahead and merge this PR into master, i.e. the `master` branch will now be the source of v2.x releases which support Python 3 only.

The most important justification for doing it is that Astropy dropped Python 2 support in Feb 2018, so we have been missing out on lots of sweet new stuff that's in Astropy.  (Same is true for scipy/numpy/etc.)

PR #734, opened 2020-05-16
Changed the x and y axis of the tpf interactive plotting 

Higgins00:
This is one solution to making the tpf.interact_sky() have the stars from the Gaia data appear in the correct locations on the image. 

If the desired result is to have the pixel row and column numbers not be centered then there is another solution. Simply subtract 0.5 from the x and y locations of the stars in dictionary for the Gaia data.

### DIFF
--- a/.github/workflows/greetings.yml
+++ b/.github/workflows/greetings.yml
@@ -9,5 +9,5 @@ jobs:
     - uses: actions/first-interaction@v1
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


barentsen:
Thanks so much for making the effort to open a PR, @Higgins00!

It looks like your proposed fix does indeed fix the problem.

## Before this PR

interact-before

## After this PR

interact-after

barentsen:
Oops, I messed up this PR because the the PR branch name is `master`.

I merged your commit into Lightkurve via the terminal.  Thanks again for the great fix! :+1:

PR #737, opened 2020-05-19
Replace references to header with get_header() in TargetPixelFile to remove warning

nksaunders:
With #736 merged, downloading or opening any `TargetPixelFile` gives the user a `DeprecationWarning` for the `header` property. This PR replaces all references to `self.header` with the new `self.get_header()` method.

```python
LightkurveWarning: `TargetPixelFile.header` is deprecated, please use `TargetPixelFile.get_header()` instead.
```

This deprecation warning is useful when a user calls `.header` directly, but confusing if it shows up unprompted.

Not sure if this is the right approach here, but it gets rid of the warning. @barentsen @christinahedges thoughts?

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -994,7 +994,7 @@ class TargetPixelFile(object):
             if label in hdu.header:
                 hdu.header[label] = fits.card.Undefined()
 
-        keys = np.asarray([k for k in self.header.keys()])
+        keys = np.asarray([k for k in self.get_header().keys()])
         if 'KEPLERID' in keys:
             hdu.header['KEPLERID'] = '{}{}'.format(hdu.header['KEPLERID'], '_CUTOUT')
         if 'TICID' in keys:
@@ -1112,7 +1112,7 @@ class KeplerTargetPixelFile(TargetPixelFile):
                                 bitmask=quality_bitmask)
 
         # check to make sure the correct filetype has been provided
-        filetype = detect_filetype(self.header)
+        filetype = detect_filetype(self.get_header())
         if filetype == 'TessTargetPixelFile':
             warnings.warn("A TESS data product is being opened using the "
                           "`KeplerTargetPixelFile` class. "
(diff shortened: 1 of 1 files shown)


barentsen:
Thanks @nksaunders, this looks good!

The tests seem to have difficulty with `tesscut`. Let's re-run...

barentsen:
The test named `test_search_tesscut_download` is consistently failing on GitHub, but I can't reproduce locally.  Can anyone else reproduce?

The command to try is:
```bash
pytest -k test_search_tesscut_download --remote-data
```

nksaunders:
@barentsen I can reproduce this error on my machine. TESSCut can identify an observation for this target, but fails to return the data when you try to download it.

```python
from lightkurve import search_tesscut

ra, dec = 30.578761, -83.210593
search_string = search_tesscut('{}, {}'.format(ra, dec), sector=[1, 12])
search_string
```
```
SearchResult containing 2 data products.

 #   observation        target_name      productFilename distance
--- -------------- --------------------- --------------- --------
  0  TESS Sector 1 30.578761, -83.210593         TESSCut      0.0
  1 TESS Sector 12 30.578761, -83.210593         TESSCut      0.0
```
```python
tpf = search_string[1].download()
```
```
WARNING: NoResultsWarning: No FFIs found for input target. Cutout not performed. [astroquery.mast.cutouts]
```

I can reproduce the same results in the TESSCut web API. 

Screen Shot 2020-05-19 at 11 43 37 AM

I think it's still working for you because you have this target cached—if you clear your local cache I bet it will fail. Looks like a problem on MAST's end, so hopefully this test starts working again on its own.

barentsen:
You genius!  Indeed, I can also reproduce the failing test after deleting `~/.lightkurve-cache/tesscut`.  I reproduced the error on `master` and it is thus not specific to this PR.

Thank you for posting the screenshot of the Tesscut interface.  This confirms that the failure is not likely be on Lightkurve's side.  I'll go ahead on merge this PR, and see what happens to the failing tests in the next 24 hours before making further investigations!

Thanks++!

PR #738, opened 2020-05-19
Added option to plot TPF background to `TargetPixelFile.plot()`

ojhall94:
For cases where there are strong background features (e.g. near the corners of TESS CCDs) in Target Pixel File data, it might be useful to be able to plot the TPF background flux using the Lightkurve infrastructure.

This PR changes the keyword arguments of `TargetPixelFile.plot()`, so that:

- `bkg=True` plots the TPF flux background _only_
- `add_bkg=True` plots the TPF flux with the background added (this functionality was previously under the `bkg=True` argument).

If `bkg=True` and `add_bkg=True` are passed, the `bkg` keyword is ignored.

Example usage:
image

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -716,7 +716,7 @@ class TargetPixelFile(object):
             return res, np.in1d(self.astropy_time.jd, res.epoch)
         return res
 
-    def plot(self, ax=None, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
+    def plot(self, ax=None, frame=0, cadenceno=None, bkg=False, add_bkg=False, aperture_mask=None,
              show_colorbar=True, mask_color='pink', title=None, style='lightkurve',
              **kwargs):
         """Plot the pixel data for a single frame (i.e. at a single time).
@@ -734,8 +734,11 @@ class TargetPixelFile(object):
         cadenceno : int, optional
             Alternatively, a cadence number can be provided.
             This argument has priority over frame number.
-        bkg : bool
+        bkg: bool
+            If True, plot the background pixel values only.
+        add_bkg : bool
             If True, background will be added to the pixel values.
+            Can not be True if 'bkg' is True.
         aperture_mask : ndarray or str
             Highlight pixels selected by aperture_mask.
         show_colorbar : bool
@@ -764,8 +767,10 @@ class TargetPixelFile(object):
                                  "must be in the range {}-{}.".format(
                                      cadenceno, self.cadenceno[0], self.cadenceno[-1]))
         try:
-            if bkg and np.any(np.isfinite(self.flux_bkg[frame])):
+            if add_bkg and np.any(np.isfinite(self.flux_bkg[frame])):
                 pflux = self.flux[frame] + self.flux_bkg[frame]
+            elif bkg and np.any(np.isfinite(self.flux_bkg[frame])):
+                pflux = self.flux_bkg[frame]
             else:
                 pflux = self.flux[frame]
         except IndexError:
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -767,26 +767,25 @@ class TargetPixelFile(object):
                                  "must be in the range {}-{}.".format(
                                      cadenceno, self.cadenceno[0], self.cadenceno[-1]))      
         try:
-            if column not in ('FLUX','FLUX_ERR','FLUX_BKG','FLUX_BKG_ERR',
-                                'COSMIC_RAYS','RAW_CNTS'):
-                raise ValueError("column must be one of the following: ('FLUX','FLUX_ERR',"
-                                 "'FLUX_BKG','FLUX_BKG_ERR','COSMIC_RAYS','RAW_CNTS')")
             if column == 'FLUX':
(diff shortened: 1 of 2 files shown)


barentsen:
(The two remote-data test runs are failing due to no fault of your own.  Please ignore.  I hope it'll get fixed shortly.)

ojhall94:
Great suggestion @barentsen, updated the PR. Added in the unit tests for the new functionality too. It still has an issue--- TESS TPFs store the COSMIC_RAYS data in a different extension of the FITS file. I'm guessing this can be alleviated in a clever way, but can't think of it right now. Will give another pass over with your comments.

barentsen:
Thanks @ojhall94.  This looks good!

Just one more request: to keep the function as flexible and future-proof as possible, I'd recommend removing the bit where you check the value of `column` against the hard-coded list of column names, and instead do something like this:

```python
try:
    data_to_plot = self.hdu[1].data[column]
except KeyError:
    raise ValueError(f"column {column} does not exist in this file")
```

This will also resolve the issue with `COSMIC_RAYS`, i.e. it'll work for Kepler and raise a ValueError for TESS, which is fine.

(You can of course keep mentioning the typically available column names in the docstring, that is fine!)

ojhall94:
Great suggestion, made the changes.

lightkurve/targetpixelfile.py
```diff
@@ -734,8 +734,11 @@ def plot(self, ax=None, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
         cadenceno : int, optional
             Alternatively, a cadence number can be provided.
             This argument has priority over frame number.
-        bkg : bool
+        bkg: bool
             If True, background will be added to the pixel values.
+        column :
```
barentsen, line 739: Make sure we specify the expected data type:
```suggestion
        column : str
```

lightkurve/targetpixelfile.py
```diff
@@ -762,22 +765,37 @@ def plot(self, ax=None, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
             except IndexError:
                 raise ValueError("cadenceno {} is out of bounds, "
                                  "must be in the range {}-{}.".format(
-                                     cadenceno, self.cadenceno[0], self.cadenceno[-1]))
+                                     cadenceno, self.cadenceno[0], self.cadenceno[-1]))
```
barentsen, line 768: Remove trailing whitespace:
```suggestion
                                     cadenceno, self.cadenceno[0], self.cadenceno[-1]))
```

lightkurve/targetpixelfile.py
```diff
@@ -762,22 +765,37 @@ def plot(self, ax=None, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
...
+
         with plt.style.context(style):
             if title is None:
                 title = 'Target ID: {}, Cadence: {}'.format(self.targetid, self.cadenceno[frame])
             img_extent = (self.column, self.column + self.shape[2],
                           self.row, self.row + self.shape[1])
-            ax = plot_image(pflux, ax=ax, title=title, extent=img_extent,
-                            show_colorbar=show_colorbar, **kwargs)
+            ax = plot_image(data_to_plot, ax=ax, title=title, extent=img_extent,
+                            show_colorbar=show_colorbar, clabel = clabels[column], **kwargs)
```
barentsen, line 798: Calling `clabel = clabels[column]` would crash if someone did call this function on a fancy TargetPixelFile that has an extra column we didn't expect.  A good solution here is to use `dict.get(key, default)`, which returns `default` if `key` is not available in the dictionary:

```suggestion
                            show_colorbar=show_colorbar, clabel = clabels.get(column, column), **kwargs)
```

lightkurve/targetpixelfile.py
```diff
@@ -762,22 +765,37 @@ def plot(self, ax=None, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
...
+                if bkg and np.any(np.isfinite(self.flux_bkg[frame])):
+                    data_to_plot = self.flux[frame] + self.flux_bkg[frame]
+                else:
+                    data_to_plot = self.flux[frame]
             else:
-                pflux = self.flux[frame]
+                data_to_plot = self.hdu[1].data[column][self.quality_mask][frame]
+        except KeyError:
+            raise ValueError("column must be one of the following: ('FLUX','FLUX_ERR',"
+                                "'FLUX_BKG','FLUX_BKG_ERR','COSMIC_RAYS','RAW_CNTS')")
```
barentsen, line 779: PEP8 recommends aligning the string with the opening delimiter for easier readability:
```suggestion
                             "'FLUX_BKG','FLUX_BKG_ERR','COSMIC_RAYS','RAW_CNTS')")
```

lightkurve/targetpixelfile.py
```diff
@@ -734,8 +734,11 @@ def plot(self, ax=None, frame=0, cadenceno=None, bkg=False, aperture_mask=None,
         cadenceno : int, optional
             Alternatively, a cadence number can be provided.
             This argument has priority over frame number.
-        bkg : bool
+        bkg: bool
             If True, background will be added to the pixel values.
```
barentsen, line 738: Let's be slightly more specific:
```suggestion
            If True and `column="FLUX"`, background will be added to the pixel values.
```

ojhall94:
Thanks for checking my code! :) Made the suggested fixes.

PR #739, opened 2020-05-20
Add support for community light curves (K2SFF & EVEREST)

barentsen:
This PR adds support for searching, downloading, and reading community light curves.  This PR only adds support for K2SFF and EVEREST, but it can easily be extended to support other products in the future.

## Demo

```python
import lightkurve as lk

# Read a K2SFF light curve file from its path
lc = lk.read("http://archive.stsci.edu/hlsps/k2sff/c16/212100000/00236/hlsp_k2sff_k2_lightcurve_212100236-c16_kepler_v1_llc.fits")

#  Retrieve a K2SFF light curve via search
lc = lk.search_lightcurvefile("GJ 9827", author="K2SFF").download()

#  Retrieve an EVEREST light curve via search
lc = lk.search_lightcurvefile("GJ 9827", author="EVEREST").download()

#  Retrieve all light curves from any author
lc = lk.search_lightcurvefile("GJ 9827", author="any").download_all()

#  Retrieve only the official data pipeline products
lc = lk.search_lightcurvefile("GJ 9827").download()
```

## Summary of changes

* Added `read_k2sff(path)` and `read_everest(path)` functions which return `KeplerLightCurve` objects given a K2SFF or EVEREST light curve fits file.
* Added the `read(path)` function which automatically detects the file type and calls the correct reader function.
* Deprecated `open(path)` in favor of `read(path)`.  The naming of `open()` had always been a problem because it collides with Python's built-in `open()` function.
* Added a new `io.py` module to hold all `read` functions.
* Added the `author` parameter to `search_lightcurvefile` and `search_targetpixelfile`.  This parameter is identical to the `provenance_name` parameter used by the MAST Portal API to identify the provenance of data products. The default is `author=('Kepler', 'K2', 'SPOC')`. This is what the official Kepler/K2/TESS products are registered under. 

## Notes

* To keep things simple, I propose we do not add special sub-classes of `LightCurve` or `LightCurveFile` for each community data product. That would quickly become impossible to maintain!  Instead, I propose we add ~10-line reader functions for community light curves which return an instance of `LightCurve`.
* We may eventually add these reader functions as readers to the astropy `TimeSeries` class, but it's easier to develop them inside Lightkurve first.
* Unfortunately I marked two tests with `@pytest.mark.xfail` (`test_search_tesscut_download` and `test_search_split_campaigns`). I believe they are failing due to errors on MAST's side.

Thoughts? Comments?


ojhall94:
Hi @barentsen, thanks for this awesome PR!

I'm having a play around with the `search_lightcurvefile()` function. I don't really understand what the difference is between the `mission` and `provenance_name` keywords. I'm guessing this will be alleviated by [barentsen/notebooks#6](https://github.com/barentsen/notebooks/issues/6). Is there a meaningful difference between the 'Kepler' and 'K2' provenances?

lightkurve/io.py
```diff
@@ -0,0 +1,243 @@
...
+from .lightcurve import KeplerLightCurve
+from .utils import validate_method, LightkurveWarning
+
+log = logging.getLogger(__name__)
+
+__all__ = ['read', 'open', 'read_k2sff']
+
+
+def open(path_or_url, **kwargs):
+    """DEPRECATED. Please use `lk.read()` instead.
```
christinahedges, line 16: Very happy with this!

lightkurve/io.py
```diff
@@ -0,0 +1,243 @@
...
+            log.debug("Detected filetype: '{}'.".format(filetype))
+    except OSError as e:
+        filetype = None
+        # Raise an explicit FileNotFoundError if file not found
+        if 'No such file' in str(e):
+            raise e
+
+    # Community-provided science products
+    if filetype == "K2SFF":
+        return read_k2sff(path_or_url, **kwargs)
```
christinahedges, line 76: If you can do this with `read`, do we ever want to expose these helper functions to the user? It might confuse people if they think they have to have a specific read function.

lightkurve/io.py
```diff
@@ -0,0 +1,243 @@
...
+        "flux_unit": "",  # SFF light curves are normalized
+        "cadenceno": f[ext].data['CADENCENO'],
+        "targetid": f[0].header["KEPLERID"],
+        "channel": f[0].header["CHANNEL"],
+        "campaign": f[0].header["CAMPAIGN"],
+        "mission": f[0].header["MISSION"],
+        "ra": f[0].header["RA_OBJ"],
+        "dec": f[0].header["DEC_OBJ"],
+        "label": f[0].header["OBJECT"]
+    }
```
christinahedges, line 125: Should we add something in this `lc` object to show it's not a pipeline product? e.g. changing the label to 
`{} [SFF]'.format(f[0].header["OBJECT"])`

lightkurve/io.py
```diff
@@ -0,0 +1,243 @@
...
+        "flux_err": f[1].data['FRAW_ERR'],
+        "cadenceno": f[1].data['CADN'],
+        "quality": f[1].data['QUALITY'],
+        "targetid": f[0].header["KEPLERID"],
+        "channel": f[0].header["CHANNEL"],
+        "campaign": f[0].header["CAMPAIGN"],
+        "mission": f[0].header["MISSION"],
+        "ra": f[0].header["RA_OBJ"],
+        "dec": f[0].header["DEC_OBJ"],
+        "label": f[0].header["OBJECT"]
```
christinahedges, line 161: Same as above?

lightkurve/search.py
```diff
@@ -63,24 +64,24 @@ def _add_columns(self):
...
-                        "constructor; the MAST API may have changed.")
+            log.debug("Unexpected data encountered in the ``SearchResult`` "
+                      "constructor; the MAST API may have changed.")
 
     def __repr__(self, html=False):
         out = 'SearchResult containing {} data products.'.format(len(self.table))
         if len(self.table) == 0:
             return out
-        columns = ['#', 'observation', 'target_name', 'productFilename', 'distance']
+        columns = ['#', 'observation', 'target_name', 'productFilename', 'provenance_name', 'distance']
```
christinahedges, line 84: provenance_name is quite ambiguous, it's possible to rename these columns if we wanted to. Should we rename it something a little more user friendly? The advantage to leaving it as is would be that it's consistent with MAST.

lightkurve/search.py
```diff
@@ -421,6 +424,10 @@ def search_targetpixelfile(target, radius=None, cadence='long',
         'long' or 'short'.
     mission : str, list of str
         'Kepler', 'K2', or 'TESS'. By default, all will be returned.
+    provenance_name : str, list of str
```
christinahedges, line 427: Can you add what community products are supported? Not all of the HLSPs are supported in this PR.

lightkurve/tests/test_io.py
```diff
@@ -0,0 +1,92 @@
...
+    try:
+        FileNotFoundError
+    except NameError:
+        FileNotFoundError = IOError
+    with pytest.raises(FileNotFoundError):
+        read("DOESNOTEXIST")
+
+
+@pytest.mark.remote_data
+def test_read_k2sff():
```
christinahedges, line 54: Do you want to save a shortened K2sff file for tests? Same with EVEREST?

barentsen:
Upon further discussion with @christinahedges, we think it may be more user-friendly to adopt parameter name `author` instead of `provenance_name`.  For example:

```python
search_lightcurvefile(target, author="K2SFF")
```

By default, `author` will default to the official data pipelines i.e. `('Kepler', 'K2', 'SPOC')`.  I envisage adding a user-friendly option `author="any"` or `include_community_products=True` which will turn off filtering by author, example:
```python
search_lightcurvefile(target, author="any")
```
or perhaps
```python
search_lightcurvefile(target, include_community_products=True)
```

Thoughts?

lightkurve/io.py
barentsen: Agreed!  I'm implementing your suggestion.

lightkurve/search.py
barentsen: Resolved: I renamed `provenance_name` to `author`.

lightkurve/search.py
barentsen: Done!

lightkurve/tests/test_io.py
barentsen: I'm going to use remote data for now, until the API settles down.

lightkurve/io.py
barentsen: I don't mind exposing these functions because (a) they advertise which products we support, and (b) it provides a convenient fall-back option if our file type detection fails.  We can focus on simply advertising `read` in tutorials and documentation however.

lightkurve/io.py
```diff
@@ -0,0 +1,243 @@
...
+def read_k2sff(path_or_url, ext="BESTAPER", **kwargs):
+    """Read a K2SFF light curve file.
+
+    More information: https://archive.stsci.edu/hlsp/k2sff
+
+    Parameters
+    ----------
+    path_or_url : str
+        Path or URL of a K2SFF light curve FITS file.
+    ext : str
```
nksaunders, line 99: Love this 👍👍 I was hoping this would be an option.

lightkurve/search.py
```diff
@@ -419,8 +407,12 @@ def search_targetpixelfile(target, radius=None, cadence='long',
         units of arcseconds.  If `None` then we default to 0.0001 arcsec.
     cadence : str
         'long' or 'short'.
-    mission : str, list of str
+    mission : str, tuple of str
         'Kepler', 'K2', or 'TESS'. By default, all will be returned.
+    author : str, tuple of str, or "any"
```
nksaunders, line 412: Passing `any` returns K2SC light curves, which don't seem to be supported. They will be downloaded but not fail when `read` is called on them. 

```
ValueError: Not recognized as a Kepler or TESS data product
```

We should either make it possible to open these in `lightkurve` or remove them from the list for now.

nksaunders:
Another issue I came across:

`search_lightcurvefile(target).download()` returns a `LightCurve` for HLSPs, but a `LightCurveFile` for K2 products. This will be a problem if a user tries to use `download_all()` on a search result containing both HLSPs and official pipeline products.

**Example:**
```python
sr = lk.search_lightcurvefile('k2-199', author=['EVEREST', 'K2SFF', 'K2'])
lcfc = sr.download_all()
lcfc
```
```
...
AttributeError: 'KeplerLightCurveFile' object has no attribute 'label'
```

Alternatively, if you only include HLSPs in your author query, you can successfully make a `LightCurveFileCollection` that is only populated by `LightCurve` objects. This is a separate issue that probably shouldn't be allowed, because it breaks if you try to use functions like `lcfc.plot`.

`download` should always return the same object type. I'm not sure what the best solution is here, since community pipeline products won't have the same attributes as a PDC `LightCurveFile`. This should definitely be addressed before merging, though.

barentsen:
@nksaunders Agreed we should try to support a few more HLSPs, including K2SC.

The fact that `download_all()` now returns both `LightCurve` and `LightCurveFile` objects is definitely a major issue.

My dream is to get rid of `LightCurveFile` altogether when we make `LightCurve` a sub-class of Astropy `TimeSeries` (because the `TimeSeries` class is designed to hold e.g. SAP_FLUX and PDCSAP_FLUX within the same object).  I'm going to try to prototype that first before continuing with this PR!

barentsen:
I modified this PR to take the `TimeSeries` integration into account, and to provide a helpful error message when K2SC or K2VARCAT light curves are requested. (I want to leave supporting them to a future PR.)

This should be good to merge.  Let me know if anyone objects!

ojhall94:
Not that it matters now that it's merged, but I've had a play around with the new PR and I love how it works! It's really easy to compare different corrections to a K2 target, which I personally like.

barentsen:
Thanks @ojhall94!

PR #743, opened 2020-05-24
Fixing the proper motion corrections in TPF Interact Sky function 

Higgins00:
In order to display the gaia star data in the correct location (xcoords,ycoords) of when the TESS image was taken, the proper motion must be accounted for before calling the tpf.wcs.all_world2pix(radecs, 0) function. This pull request just moves that proper motion correction to a position directly before where x and y coordinates are generated for the gaia source stars.


lightkurve/interact.py
```diff
@@ -235,12 +235,15 @@ def add_gaia_figure_elements(tpf, fig, magnitude_limit=18):
...
     radecs = np.vstack([result['RA_ICRS'], result['DE_ICRS']]).T
-    coords = tpf.wcs.all_world2pix(radecs, 0)
+    
     year = ((tpf.astropy_time[0].jd - 2457206.375) * u.day).to(u.year)
     pmra = ((np.nan_to_num(np.asarray(result.pmRA)) * u.milliarcsecond/u.year) * year).to(u.deg).value
     pmdec = ((np.nan_to_num(np.asarray(result.pmDE)) * u.milliarcsecond/u.year) * year).to(u.deg).value
     result.RA_ICRS += pmra
     result.DE_ICRS += pmdec
+    
+    coords = tpf.wcs.all_world2pix(radecs, 0)
```
barentsen, line 245: I think you'll also want to move line 237 to line 244?

barentsen:
Thank you for pointing this out @Higgins00! :+1: :+1:

I think this PR needs one additional fix, which is to move the line that says
```python
radecs = np.vstack([result['RA_ICRS'], result['DE_ICRS']]).T
```
to a place below where the proper motions are applied.  Do you want to try and fix this?

Second question: it will be important to test that our fix really works.  Would you be able to suggest a good test case or coordinate where the error is clearly visible before/after?  (e.g. how did you first discover the bug?)  Thanks!!

keatonb:
> I think this PR needs one additional fix, which is to move the line that says
> 
> ```python
> radecs = np.vstack([result['RA_ICRS'], result['DE_ICRS']]).T
> ```
> 
> to a place below where the proper motions are applied. Do you want to try and fix this?

I think instead of applying the proper motion correction directly to `result`, which is the content returned from Gaia DR2, the proper motion corrections should be applied to `radecs` (and this should maybe hold a copy of the ra and dec columns from `result`, not sure).  The propagated coordinates are useful for putting the points in the right locations on the image, but users might expect the unpropagated Gaia DR2 coordinates in the tooltip when they hover over a point.

barentsen:
I have incorporated this fix in https://github.com/KeplerGO/lightkurve/pull/755.

Thanks so much for reporting! :+1:

PR #744, opened 2020-05-26
Turn `LightCurve` into a sub-class of AstroPy `TimeSeries`

barentsen:
As part of Lightkurve v2.x, this PR investigates the consequences of turning `LightCurve` into a subclass of `astropy.time.TimeSeries` (itself a subclass of `astropy.table.Table`).

### Demo

https://colab.research.google.com/drive/1YA7Jni9ekzmOglg3qRs_tOnBXnFSJtjq?usp=sharing

### Why change the Lightkurve API?

The motivation of this change is that it would reduce the duplication of documentation and development efforts between the Lightkurve and AstroPy ecosystems.   For example, it would enable Lightkurve to benefit from generic tools developed for AstroPy's `TimeSeries` and `Table` objects (e.g. the [aas-timeseries](https://aas-timeseries.readthedocs.io) package).

At the same time, it presents an opportunity to fix a few shortcomings in the Lightkurve v1.x API.  Specifically, by sub-classing `LightCurve` from `TimeSeries` we would...
1. gain the ability of `LightCurve` objects to hold arbitrary numbers of additional columns (e.g. both SAP_FLUX and PDCSAP_FLUX would now be columns of a single light curve object);
2. gain the ability to remove Lightkurve's limited and confusing `LightCurveFile` objects in favor of the uniform i/o interface offered by `TimeSeries`, i.e. we'd have `LightCurve.read()` and `LightCurve.write()` methods similar to `Table.read()` and `Table.write()`;
3. gain the ability to utilize `astropy.time.Time` objects for the `time` column and `astropy.units.Quantity` for all other columns.

### What would `LightCurve` add on top of `TimeSeries`?

A `TimeSeries` object is a table which is guaranteed to have a special `time` column, which is always the first column and guaranteed to be an `astropy.time.Time` object.

`LightCurve` would extend this class by adding special `flux` and `flux_err` columns, which are always the second and third column.  These dedicated columns enable us to provide extra methods which are specific to the manipulation of brightness data.  For example, methods which are currently unique to `LightCurve` objects include `plot()`, `scatter()`, `errorbar()`, `normalize()`, `fill_gaps()`, `flatten()`, `remove_outliers()`, `remove_nans()`, `estimate_cdpp()`, `plot_river()`, `to_periodogram()`, `to_seismology()`, `query_solar_system_objects()`, `interact_bls()`, etc.  Some of these methods may eventually be migrated to `TimeSeries` if they are sufficiently generic and mature, but many Kepler/TESS-specific methods will likely remain exclusive to `LightCurve`.

Other differences include:
* `LightCurve` objects offer math operations, e.g. `lc1 + lc2` or `lc1 / lc2`.
* `LightCurve` objects will provide user-friendly access to metadata via attributes.  For example, `lc.meta['mission']` will be exposed as `lc.mission`.

### Will existing Lightkurve code break?

While we try to maintain a high degree of backwards compatibility with Lightkurve v1.x, it looks like nine API changes are unavoidable:
1. **lc.time is now always an Astropy Time object.**  Unfortunately, `Time` objects cannot be passed to many numpy/scipy functions, you typically have to pass `lc.time.value` instead.
2. **All other columns are now Astropy Quantity objects.** Most numpy/scipy functions can deal with these, but you will nevertheless have to use `lc.flux.value` at times.
3. **All metadata is now stored in the lc.meta dictionary**, including all FITS header values.  For convenience, we support attribute access, e.g. `lc.sector` resolves to `lc.meta['sector']`, but the use of the latter is encouraged because not every light curve will have a sector keyword.
4. **LightCurveFile classes have been removed.** Instead, files must now be loaded using `lightkurve.read(filename, flux_column="pdcsap_flux", quality_bitmask="default")`.
5. **lc.time can no longer contain NaN values**.  This is not allowed by `Time` objects. As a result, `lk.read()` will always ignore rows with `time=NaN`.  Other columns can still have NaNs however.
6. **lc.bin() now takes time_bin_size and time_bin_start arguments.** This is for consistency with `astropy.timeseries.aggregate_downsample`. This significantly alters the behavior of binning, because bins are now defined in time instead of by number of data points.
7. **lc.fold() now returns un-normalized phase values by default.**  This is for consistency with `TimeSeries.fold()`. For the same reason, the `t0` argument has been deprecated in favor of `epoch_time`, and the method now accepts the extra `epoch_phase`, `wrap_phase`, and `normalize_phase` arguments.
8. **lc.append() no longer supports inplace=True.**  This allows us to implement append via a simple call to `astropy.table.vstack`.
9. **Several attributes and methods have been deprecated.**  They still work but will yield a warning and may be removed in the future:
  * `lc.astropy_time` => `lc.time`
  * `lc.time_format` => `lc.time.format`
  * `lc.time_scale` => `lc.time.scale`
  * `lc.flux_quantity` => `lc.flux`
  * `lc.flux_unit` => `lc.flux.unit`
  * `lc.to_table()` => `lc`
  * `lc.to_timeseries()` => `lc`
  * `search_lightcurvefile()` => `search_lightcurve()`

### How do I try this proposed change?

You can install this PR branch via git as follows:
```python
git checkout -b geert-timeseries-refactor master
git pull https://github.com/barentsen/lightkurve.git timeseries-refactor
python setup.py install
```

Alternatively, you can install this branch using pip as follows:
```
pip install --quiet --upgrade git+https://github.com/barentsen/lightkurve.git@timeseries-refactor
```


ojhall94:
Wow Geert, this is a really massive PR. I've looked through the files changed and nothing faulty stuck out to me, but I find it easier to test these things in code anyway. I'll have play around with this PR later and catalogue the changes required to the `periodogram` and `seismology` modules as a result of this change.

ojhall94:
Alright, I've had a play around with the tutorial! I have some (hopefully constructive) comments:
- I couldn't find a way to access the FITS data the way we previously could through the `LightCurve` and `LightCurveFile` `hdu` property. Some of the metadata is carried over, but I can't seem to find how to e.g. access the KIC stellar properties, which I think are pretty useful in the context of using Lightkurve as a first-look tool.

- Now that PDCSAP_FLUX and SAP_FLUX are stored as columns, there is no way to plot them individually using Lightkurve's infrastructure, losing some of the functionality that was previously available through independent `LightCurve` objects for both fluxes. Ditto with `normalize`, `flatten` etc. applied to those properties.

-   In a similar vein, now that both `pdcsap_flux` and `flux` are columns in the `Timeseries` object, the data is duplicated. I don't know if this is an issue-- I didn't find it intuitive on the first look, but then I suppose it was duplicated in the original build as well.

- The `flux` contents being returned as an `Astropy.units.Quantity` raises the barrier to entry a little I think, requiring users to be familiar with the `Quantity` package before using Lightkurve. I think the way to mitigate this is to make sure it is featured early in the tutorials and documentation.

- I miss the pretty repr for the `LightCurve` object. Now that a `LightCurve` returns the repr for an astropy table, its inconsistent with other Lightkurve objects. We could have best of both worlds by hacking the repr a little so it returns the object type & identifier as well as the astropy table.

- Data downloaded through `search_lightcurvefile()` no longer works with the new `LightCurve` infrastructure, butI guess that is to be expected.

Cloning the branch and installing it locally doesn't get past cell 2 on the tutorial for me, I get the following error:

```python3
IORegistryError: No reader defined for format 'kepler' and class 'LightCurve'.

The available formats are:

   Format   Read Write Auto-identify
----------- ---- ----- -------------
kepler.fits  Yes    No            No
  tess.fits  Yes    No            No
```

Hope these are helpful!

barentsen:
Thanks for those valuable thoughts @ojhall94! :+1:

This prototype PR is definitely still extremely rough.  For example, I have not attempted to modify the `search` features yet, there is no TESS support, and I haven't even looked at the unit tests yet.  I've only really tried to prototype how the core `LightCurve` would change so far.

The good news is that I do think we'll be able to address each of your comments:

> I couldn't find a way to access the FITS data the way we previously could

This is a good comment.  At minimum we should be able to record the path of the file that was read in as part of the `meta` object, and perhaps even offer a convenience method to access or open that path as before  (e.g. `lc.hdu` could be an alias for `astropy.io.fits.open(lc.meta['path'])`, though we may not want to do this because we'd also support e.g. ascii and hdf5 files now).

> Now that PDCSAP_FLUX and SAP_FLUX are stored as columns, there is no way to plot them

My current thinking is that all the light curve operations will operate on the magic `flux` and `flux_err` columns.  Users will be able to select the contents of the flux column as part of the read/search functions, e.g. `read(path, flux_column="pdcsap_flux")`.  The choice can be revised at any time by executing e.g. `lc.flux = lc.pdcsap_flux`.  Would that work?

> In a similar vein, now that both pdcsap_flux and flux are columns in the Timeseries object, the data is duplicated. 

That's a good comment! My thinking is that we could have the `repr` make it very clear that `time`, `flux`, and `flux_err` are the magic columns (e.g. by putting them as the first three columns and bold-facing them?).  All other columns in a LightCurve object will be optional nice-to-haves.

> The flux contents being returned as an Astropy.units.Quantity raises the barrier 

That's a fair comment.  We'd have to decide that the benefits of using `Quantity` are worth the cost.  Fortunately `Quantity` objects behave as numpy arrays the majority of the time, and they are of course already in use across AstroPy.

> I miss the pretty repr for the LightCurve object.

Fair!  I'm thinking to prepend the Lightkurve-style repr to the TimeSeries-style repr, so we'd get the best of both worlds.  I do find that seeing the columns at a glance, astropy.Table-style, is nice.

> Data downloaded through search_lightcurvefile() no longer works

I've not attempted to modify the `search` functions yet, but I expect they will be fairly straightforward to adapt.

I do not expect the `periodogram` and `seismology` modules will need any significant changes, but I have not tried!  I'd very much encourage you to play around with this, though I expect it will be easier to evaluate this PR once it has matured for another few days.

Thanks++!

gully:
Neat!  Astropy Tables have additional support for masking, correct?  Does that functionality extend to `timeseries` and if so do you see any advantages/challenges/side-effects for leveraging mask-array features of Tables?

barentsen:
> Astropy Tables have additional support for masking, correct? Does that functionality extend to timeseries

@gully This is an excellent question!  Unfortunately masking does *not* yet work for columns that are `Quantity` objects (e.g. see [this astropy issue](https://github.com/astropy/astropy/issues/10119)).  

We'll have to choose between either having units *or* having masks attached to our columns.  My intuition is to use `Quantity` and let the `read` function deal with removing rows that should be masked by a quality mask (which is what we ~do in Lightkurve v1.x).  In this scenario, we'd still be able to adopt masks in the future once AstroPy starts supporting masked quantities in tables.

astrofrog:
I don't have time to review this at the moment but I'm very much in favor of the idea of subclassing from astropy's timeseries class! :clap: If there are limitations the become a problem, such as the masking of quantities, let's use this as an opportunity to improve the astropy class :smile:

barentsen:
@astrofrog Definitely! :+1

barentsen:
This PR is now fully functional and ready for testing and discussion! I don't plan any major changes other than cleaning up docstrings and tests. Please try it out and let me know if you think this drastic change is a good idea! 

### How to install this branch?

You can try this branch as follows:
```python
git checkout -b geert-timeseries-refactor master
git pull https://github.com/barentsen/lightkurve.git timeseries-refactor
python setup.py install
```

### Summary of API Changes

I tried to maintain a high degree of backwards compatibility with Lightkurve v1.x.  Nevertheless, approximately 10 API changes are necessary:
1. **LightCurveFile classes have been removed.** Instead, files must now be loaded using `lightkurve.read(filename, flux_column="pdcsap_flux", quality_bitmask="default")`.
2. **All metadata is now stored in the lc.meta dictionary**, including all FITS header values.  For convenience, we support attribute access, e.g. `lc.sector` resolves to `lc.meta['sector']`, but the use of the latter is encouraged because not every light curve will have a sector keyword.
3. **The LightCurve constructor no longer accepts positional arguments.**  For example, you are now required to instantiate a light curve using `LightCurve(time=t, flux=f)` instead of `Lightcurve(t, f)`.  This is to maintain consistency with the constructor of `TimeSeries`.
4. **lc.time is now always an Astropy Time object.**  Unfortunately, `Time` objects cannot be passed to many numpy/scipy functions, you typically have to pass `lc.time.value` instead.
5. **All other columns are now Astropy Quantity objects.** Most numpy/scipy functions can deal with these, but you will nevertheless have to use `lc.flux.value` at times.
6. **lc.time can no longer contain NaN values**.  This is not allowed by `Time` objects. As a result, `lk.read()` will always ignore rows with `time=NaN`.
7. **lc.bin() now takes time_bin_size and time_bin_start arguments.** This is for consistency with `astropy.timeseries.aggregate_downsample`. This significantly alters the behavior of binning, because bins are now defined in time instead of by number of data points.
8. **The t0 argument of lc.fold() has been deprecated in favor of epoch_time.**  This is for consistency with `TimeSeries.fold()`. The method now also accepts the extra `epoch_phase`, `wrap_phase`, and `normalize_phase` arguments.
9. **lc.append() no longer supports inplace=True.**  This allows us to implement append via a simple call to `astropy.table.vstack`.
10. **Several attributes and methods have been deprecated.**  They still work but will yield a warning and may be removed in the future:
  * `lc.astropy_time` => `lc.time`
  * `lc.time_format` => `lc.time.format`
  * `lc.time_scale` => `lc.time.scale`
  * `lc.flux_quantity` => `lc.flux`
  * `lc.flux_unit` => `lc.flux.unit`
  * `lc.to_table` => `lc`
  * `lc.to_timeseries` => `lc`
  * `search_lightcurvefile()` => `search_lightcurve()`

ojhall94:
Hi @barentsen! This refactor looks super thorough! My thoughts and comments, in no particular order:

>lc.time can no longer contain NaN values. 
- Does this mean that `remove_nans()` is deprecated?
- Theres a lot of lines in code that safely remove nan values before doing something. Imo its not worth the time to remove them, but thought I'd highlight.

- I think that with - or shortly after- this PR we should update the tutorials to make it clear that flux and time are now Astropy quantities--- for some people this package might be the first time interacting with things that aren't numpy arrays of data.

I'm having a weird bug where I can't seem to initiate a lightcurve object on my laptop--- but it works fine on Colab. 

```python3
target = 'EPIC 201367296'
lc = lk.search_lightcurve(target, cadence='short', mission='K2').download()

/usr/local/lib/python3.7/site-packages/astropy-4.1rc1-py3.7-macosx-10.14-x86_64.egg/astropy/io/registry.py in get_reader(data_format, data_class)
    450             "No reader defined for format '{}' and class '{}'.\n\nThe "
    451             "available formats are:\n\n{}".format(
--> 452                 data_format, data_class.__name__, format_table_str))
    453 
    454 

IORegistryError: No reader defined for format 'kepler' and class 'KeplerLightCurve'.

The available formats are:

   Format   Read Write Auto-identify
----------- ---- ----- -------------
kepler.fits  Yes    No            No
  tess.fits  Yes    No            No
```
I've uninstalled both Astropy and Lightkurve and installed both from scratch, but the issue persists. Bit at a loss with this one.

lightkurve/lightcurve.py
```diff
@@ -1216,25 +1142,27 @@ def _create_plot(self, method='plot', ax=None, normalize=False,
...
         # Default ylabel
         if ylabel is None:
-            if normalize or (self.flux_unit == u.dimensionless_unscaled):
+            if normalize or (self.flux.unit == u.dimensionless_unscaled):
                 ylabel = 'Normalized Flux'
-            elif self.flux_unit is None:
+            elif self.flux.unit is None:
                 ylabel = 'Flux'
             else:
                 ylabel = 'Flux [{}]'.format(self.flux_unit.to_string("latex_inline"))
```
ojhall94, line 1162: Deprecated `flux_unit`

barentsen: Thanks!  Fixed it!

barentsen:
> Will there still be a way to access HDU metadata (i.e. KIC temperature, logg) once lightcurve.hdu is deprecated?

The LightCurve object returned by `KeplerLightCurve.read(filename)` populates the `lc.meta` dictionary with all FITS header keywords, so you'll be able to access it that way.

Pro-tip: if you use `lc.meta.get("logg")` instead of `lc.meta["logg"]` it won't raise an exception if the keyword is missing, but return a configurable default.

I'd like to remove the `lc.hdu` shortcut because it is specific to FITS files, whereas light curves could be loaded from any file type, so I'd like to encourage the use of the `meta` dictionary.  You can of course access the hdu via `fits.open(lc.filename)`, and if there is a really good reason for it, we could retain `lc.hdu` as a shortcut for that.

> I'm having a weird bug where I can't seem to initiate a lightcurve object on my laptop--- but it works fine on Colab.
> IORegistryError: No reader defined for format 'kepler' and class 'KeplerLightCurve'.

Oh dear!  I'm unable to reproduce this, even with AstroPy 4.1rc1.  I may have to try it on a mac.

ojhall94:
Hi @barentsen, completely agree with your comments! Thanks for clarifying.

Not sure what to do about this error--- I haven't upgraded my Mac to Catalina yet, and I don't use Anaconda (just the Mac's own python installation). Could be a very specific issue to me, I'd be interested to see if it occurs on other peoples' macbooks.

rebekah9969:
Overall I think that this is really good! Well done. 
I think we need to warn the user of the changes and provide documentation of all API changes above. I also think a version should be kept without the changes such that the user can switch back it if they have issues. 

When I tried lcf.plot() for lcf = search_lightcurvefile('KIC 6922244', quarter=4).download() it didn't seem to work, any idea why? Asked as an issue as well.

When doing the above I also get the deprecated warning as expected but then if I switch to search_lightcurve I get the error "name 'search_lightcurve' is not defined".

Thanks

barentsen:
I intend to go ahead and merge this PR tomorrow.  Please raise your hand now if you object!

The only thing missing from this PR are the required updates to the tutorials.  The plan is to develop these over the next 2 months prior to releasing a proper v2.0 version.

KenMighell:
Will there be an easy way for users to use a previous version (say 1.11.1) 
because the new v2.x code will break existing scripts/applications?  

I independently came up with a fix to the half-pixel plot offset bug in tpf.plot.
The same solution appears in v1.11.1 but not v1.11.0.

barentsen:
Yes, users will be able to install a previous version using the standard pip-mechanism, e.g.:

    `pip install lightkurve==1.11.1`

KenMighell:
Hi Geert,

OK.  That is good to know!

Cheers,

-Ken

orionlee:
Note: one breaking change

- `LightCurveFile.get_header()`  / `LightCurveFile.header()` is gone. Use `LightCurve.meta` instead. The keys **are now lowercased**, however. E.g.,

```python
  # in v1.x,
  lightCurveFile.get_header()['SECTOR']

  # in v2.0, it would be
  # the key is in lowercase
  lightCurve.meta['sector']
```

dhomeier:
Hi @barentsen,

1. A few additional comments on the API change:

> 6\. **lc.bin() now takes time_bin_size and time_bin_start arguments.** This is for consistency with `astropy.timeseries.aggregate_downsample`. This significantly alters the behavior of binning, because bins are now defined in time instead of by number of data points.

Related to this it has an `n_bins` instead of the `bins` argument, which I think can easily lead to further confusion (at least did in my case), the reason being that `aggregate_downsample` always constructs a binned timeseries of length `n_bins * time_bin_size` even with just the default (1 d) for the latter available. I.e. the length of the returned time series will have little relation to that of the input. I think it would be more user-friendly to offer a functionality equivalent to the old `lc.bin()` or perhaps just keeping `bins` as an additional argument, so that the default bin size would be adapted as
`time_bin_size = (self.time[0] - time_bin_start) / bins`
Then of course one would need to define a priority scheme and appropriate warnings if 2 or more of `time_bin_size`, `bins`, `n_bins` are set...

The other argument that has no equivalent in `aggregate_downsample` at all is `binsize` as a constant number of data points. I think it would actually not be too hard to implement this by directly constructing a `BinnedTimeSeries` with `time_bin_start = self.time[::bins]`, but maybe the more important question is whether this binning method is really a better option for signal processing. So certainly lower priority.

2. My other main point is that this list of changes should be documented in an accessible location, as I only came upon this  partly by chance after a few days (I was not specifically searching either, but I think comments in a merged PR are a place easily forgotten).

orionlee:
I've done a small migration trial, and jotted down my experience:

https://github.com/orionlee/lightkurve/wiki/Lightkurve-2-Upgrade-Trial

PR #746, opened 2020-05-28
Update `PLDCorrector` using `RegressionCorrector` and `DesignMatrix`

nksaunders:
This pull request improves `PLDCorrector` by making it more modular and compatible with the `RegressionCorrector` and `DesignMatrix` classes. This is based primarily on the current `TessRegressionCorrector`, but seeks to generalize the functionality for Kepler/K2 or any timeseries data. 

**Major Changes:**
 - `PLDCorrector` is now a subclass of `RegressionCorrector`. 
 - I added a method to `DesignMatrix` called `multichoose`, which produces the pixel combinations for higher order PLD design matrices. This currently returns a `DesignMatrixCollection` with an entry for each order.

**Open Questions:**
- should the PLD design matrix be a single `DesignMatrix` object with all orders appended as columns, or a `DesignMatrixCollection` with a `DesignMatrix` object for each order?
 - should we have separate `KeplerPLDCorrector` and `TessPLDCorrector` classes, or make a single class with informed default parameters based on the mission?
 - will different missions require different diagnostic plots?

This PR is still very preliminary, but feedback is always welcome!


lightkurve/correctors/designmatrix.py
```diff
@@ -231,6 +232,26 @@ def append_constant(self, prior_mu=0, prior_sigma=np.inf):
...
+        -------
+        `.DesignMatrix`
+            New design matrix with products of columns appended as new columns.
+        """
+        # higher order PLD design matrices
+        new_dms = [self]
+        for i in range(2, order+1):
+            regressors = np.product(list(multichoose(self.values.T, order)), axis=1).T
+            prior_mu = prior_mu*np.ones(nterms)
+            prior_sigma = prior_sigma*np.ones(nterms)
```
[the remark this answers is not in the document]
nksaunders: `prior_mu` and `prior_sigma` still default to arrays of zeros and infs respectively as set by the default argument values in [the method definition](https://github.com/nksaunders/lightkurve/blob/pld-regression/lightkurve/correctors/designmatrix.py#L235). I multiply by an array of ones to match the number of columns. This has the same effect as [lines 53-56](https://github.com/nksaunders/lightkurve/blob/pld-regression/lightkurve/correctors/designmatrix.py#L53-L56) but is flexible to nonzero `prior_mu`.

Am I misunderstanding the expected syntax or shape of the priors?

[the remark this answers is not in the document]
barentsen: Oops, you're right, I misread these lines!

nksaunders:
This PR is now working and ready for feedback. I still need to fill in some documentation, but I wanted to get input before finalizing the functions and how they handle keywords.

### Overview

The `PLDCorrector` inherits `RegressionCorrector`, and has its own subclasses for `KeplerPLDCorrector` and `TessPLDCorrector`. The only differences between these mission-specific subclasses are the values of default arguments in the `correct` method.

The major difference in defaults is that `KeplerPLDCorrector` uses a Gaussian Process to fit long-term trends, while `TessPLDCorrector` defaults to a spline. I found that in the presence of K2 noise, the long-term trend was much better preserved when fit with a Gaussian Process.

Here is the inheritance tree for `RegressionCorrector` and the different mission-specific defaults.
```
RegressionCorrector
│
└── PLDCorrector
    │
    ├── KeplerPLDCorrector
    │   ├── pld_order=2
    │   ├── use_gp=True
    │   └── pixel_components=15
    │
    └── TessPLDCorrector
        ├── pld_order=1
        ├── use_gp=False
        └── pixel_components=3
```

Setting `use_gp` to `False` will instead use a spline to capture long-term trends, set by the default arguments in `create_design_matrix` and `correct`.

### Unresolved Issues

 - **Sparseness**: Implementation of the Gaussian Process required changing the covariance portion of the `RegressionCorrector._fit_coefficients` method. This implementation currently does not handle sparse matrices, and will require some adjustment.

 - **GP prediction speed**: In order to generate informative diagnostic plots, the GP prediction is evaluated in the `RegressionCorrector._create_diagnostic_lightcurves` method. When `use_gp` is set to `True` for a TESS short cadence target, this step takes a long time (~5 minutes). Should this step come with with a warning, or be omitted for short cadence targets entirely?

 - **Best defaults**: These defaults were found with trial and error for multiple targets. Any feedback about better defaults is welcome.

---

@barentsen @christinahedges @ojhall94 Please review this PR if you have a chance, or at least throw your favorite K2/TESS targets at it to see how it performs. Thanks!

nksaunders:
Additional thoughts after chatting with @barentsen:

The improvement from using a GP seems negligible in almost every case, while the work required to develop, test, and maintain the GP version is non-negligible. It's probably best right now to focus on getting the splines right, and leave GPs as a potential future exercise.

I'll split the GP version off onto a new branch, and simplify this PR to focus only on sPLD (spline PLD). It's cool having a `RegressionCorrector` implementation with built in Gaussian Process covariance support, so this still felt like a productive experiment!

ojhall94:
Hi Nick,

This PR looks great, I'll try and give it a test today. For what its worth, I agree with parking the GP implementation for a while. With the addition of accessing HLPs with Lightkurve, EVEREST data should be able to fill that void for a while too.

nksaunders:
Thanks for the feedback @ojhall94! 

On your second point: Uh oh, that diagnostic plot doesn't look great. 
- There shouldn't be that trade off between pixel series and background model, which was a silly mistake on my part and is fixed now.
- The diagnostic plot by default shows the corrected light curve with the trend removed (`restore_trend=False`) in the bottom panel. I agree that it's a little confusing, so I've changed this to have it plot the same version of the corrected light curve as the one that's returned. 

As for the first point, I'm wrestling with the best way to handle this. You have the option to call `tpf.to_corrector('pld')` or just `tpf.to_corrector()`, which will give you the correct version of PLD. 

I'm trying to come up with the best way to add logic to make sure `PLDCorrector(tpf)` works the same way. However, I can't figure out a clear way to return an instance of a specific subclass from within `PLDCorrector().__init__()`. I'll keep hacking on it, but @barentsen I'll probably need your input on this.

ojhall94:
Hey @nksaunders,

Thanks for the response! I've just had a look-- the diagnostic plot looks much better now, and I'm excited to go through the tutorial you're writing and learn more about this functionality.

I agree with what you've said about the `PLDCorrector`, that does make more sense. From having a play around with some data, the `repr` for all subclasses are the same, e.g.

```python3
ktpf = lk.search_targetpixelfile('Kepler 8')[4].download()
ktpf.to_corrector('pld')
>>>PLDCorrector (LC: KIC 6922244)
```

where in this case the PLDCorrector is really a `KeplerPLDCorrector` class. I'll leave it to you to decide what you think the appropriate repr is, just thought it was worth bringing up for consistency with the repr conventions for other subclasses.

nksaunders:
This PR is now ready for review! 

For some reason `test_periodogram -> test_bls` is failing only on Windows with some quantity error that I can't figure out, @barentsen could you take a look? All other tests are passing.

lightkurve/correctors/designmatrix.py
```diff
@@ -272,6 +272,24 @@ def append_constant(self, prior_mu=0, prior_sigma=np.inf, inplace=False):
         dm.prior_sigma = np.append(self.prior_sigma, prior_sigma)
         return dm
 
+    def mask(self, cadence_mask):
```
barentsen, line 275: Do you think we *need* to add this new method?  It doesn't appear to be used by any new code in this PR.

This function is somewhat inconsistent with LightCurve and TargetPixelFile objects, which can only be masked using rectangular brackets rather than a mask method, so my preference is not to introduce this method unless there is a compelling reason to do so?

nksaunders: You're right, this was used in the now-removed GP functions and was added for clarity. I'll remove it.

lightkurve/correctors/pldcorrector.py
```diff
@@ -85,52 +84,151 @@ class PLDCorrector(Corrector):
...
+
         self.tpf = tpf
+        if aperture_mask is None:
+            aperture_mask = tpf.create_threshold_mask(2)
+        self.lc = self.tpf.to_lightcurve(aperture_mask=aperture_mask)
 
-        self.flux = tpf.flux
-        self.flux_err = tpf.flux_err
-        self.time = tpf.time.value
+        super(PLDCorrector, self).__init__(lc=self.lc)
```
barentsen, line 95: Now that we only support Python 3, we can simply call `super()`, like this:
```suggestion
        super().__init__(lc=self.lc)
```

lightkurve/correctors/pldcorrector.py
```diff
@@ -85,52 +84,151 @@ class PLDCorrector(Corrector):
...
-    def correct(self, aperture_mask=None, cadence_mask=None, gp_timescale=30,
-                use_gp=True, pld_order=2, n_pca_terms=10, pld_aperture_mask=None):
-        r"""Returns a PLD systematics-corrected LightCurve.
+    def __repr__(self):
+        return 'PLDCorrector (LC: {})'.format(self.lc.label)
+
+    @property
+    def X(self):
+        return self.dm
+
```
barentsen, line 103: I recommend removing the `X` property because it does not appear to be used in the code, and the `RegressionCorrector` parent class already exposes this property both via `self.dmc` and `self.design_matrix_collection`.  There should only be one way to access the design matrix!
```suggestion
```

lightkurve/correctors/pldcorrector.py
```diff
@@ -85,52 +84,151 @@ class PLDCorrector(Corrector):
...
+        # Collect each matrix
+        dm_pixels = DesignMatrixCollection(all_pld).to_designmatrix(name='pixel_series')
+        dm_bkg = DesignMatrix(simple_bkg, name='background_model')
+        dm_spline = spline(self.lc.time.value, n_knots=spline_n_knots,
+                             degree=spline_degree).append_constant()
+        with warnings.catch_warnings():
+            warnings.simplefilter('ignore')
+            dm = DMC([dm_pixels, dm_bkg, dm_spline])
+
+        self.dm = dm
```
barentsen, line 203: I recommend removing this to avoid confusion, because `RegressionCorrector.correct()` already stores the design matrix as `self.design_matrix_collection`:
```suggestion
        self.dm = dm
```

lightkurve/correctors/pldcorrector.py
```diff
@@ -139,144 +237,284 @@ def correct(self, aperture_mask=None, cadence_mask=None, gp_timescale=30,
...
+                                           pixel_components=pixel_components,
+                                           spline_n_knots=spline_n_knots,
+                                           spline_degree=spline_degree,
+                                           sparse=sparse)
+
+
+        clc = super(PLDCorrector, self).correct(dm, **kwargs)
+        if restore_trend:
+            clc += (self.diagnostic_lightcurves['spline'] - np.median(self.diagnostic_lightcurves['spline'].flux))
+        self.dm = dm
```
barentsen, line 281: Likewise, I recommend removing this to avoid confusion, because `RegressionCorrector.correct()` already stores the design matrix as `self.design_matrix_collection` (and `self.dmc` is available as a shorthand alias):
```suggestion
```

lightkurve/correctors/pldcorrector.py
```diff
@@ -139,144 +237,284 @@ def correct(self, aperture_mask=None, cadence_mask=None, gp_timescale=30,
...
+    Returns
+    -------
+    clc : `.TessLightCurve`
+        Noise-corrected `.TessLightCurve`.
+    """
+
+    def __repr__(self):
+        return 'TessPLDCorrector (LC: {})'.format(self.lc.label)
+
+    def correct(self, dm=None, pld_order=1, pixel_components=3, background_mask=None,
```
barentsen, line 489: As far as I can tell, the parameters and defaults for `TessPLDCorrector.correct()` are identical to `PLDCorrector.correct()`.  It looks like `correct()` isn't being changed in any way compared to the method provided by the parent class, so we can simply remove this entire method from `TessPLDCorrector`?

lightkurve/correctors/pldcorrector.py
```diff
@@ -139,144 +237,284 @@ def correct(self, aperture_mask=None, cadence_mask=None, gp_timescale=30,
...
+    Returns
+    -------
+    clc : `.KeplerLightCurve`
+        Noise-corrected `.KeplerLightCurve`.
+    """
+
+    def __repr__(self):
+        return 'KeplerPLDCorrector (LC: {})'.format(self.lc.label)
+
+    def correct(self, dm=None, pld_order=3, pixel_components=15, background_mask=None,
```
barentsen, line 393: Can you add a few words to the docstring why `pld_order` and `pixel_components` have a different default for Kepler?  Are we certain that this is necessary?  Are these defaults not suitable for TESS?  Could there be a way to determine the defaults based on generic properties (e.g. amount of motion, number of pixels in TPF, number of cadences, ...) rather than mission name?

I ask these questions because it is way more expensive to maintain both `KeplerPLDCorrector` and `TessPLDCorrector` classes compared to a single `PLDCorrector` class, so we want to be absolutely certain that it is necessary to introduce this class hierarchy!

barentsen:
@nksaunders Would it be handy to add a `diagnose_mask()` function that shows the target and background masks?  (Perhaps simply as a two-panel plot utilizing tpf.plot() twice.)

nksaunders:
To arrive at the default values set for `pld_order` and `pixel_components`, which should be different for K2 and TESS, I performed a brute force grid search for a target from each mission. I chose targets which had noise properties and magnitudes characteristic of a typical target, which also included the presence of stellar variability and transiting exoplanets. This analysis is not detailed enough to truly embody the best defaults for any target, but it gives a good starting point. 

The code minimizes CDPP in the five dimensional parameter space. An example is found below. In testing, it's clear that the first two variables have the most dramatic impact on the precision of the corrected light curve, and there exist typical values that work well for both K2 and TESS.

```python
import lightkurve as lk
import numpy as np
from tqdm.notebook import tqdm
import warnings

tpf = lk.search_targetpixelfile('k2-199')[0].download()

pld_order_ = np.linspace(1, 3, 3)
pixel_components_ = np.linspace(2, 16, 8)
spline_n_knots_ = np.linspace(20, 220, 13)
spline_degree_ = np.linspace(3, 6, 4)
n_pca_terms_ = np.linspace(2, 20, 10)

param_space = np.zeros((len(pld_order_), len(pixel_components_), len(spline_n_knots_), 
                        len(spline_degree_), len(n_pca_terms_)))

for i0, pld_order in tqdm(enumerate(pld_order_)):
    for i1, pixel_components in enumerate(pixel_components_):
        for i2, spline_n_knots in enumerate(spline_n_knots_):
            for i3, spline_degree in enumerate(spline_degree_):
                for i4, n_pca_terms in enumerate(n_pca_terms_):
                    pld = tpf.to_corrector('pld')
                        warnings.simplefilter('ignore')
                        try:
                            clc = pld.correct(pld_order=int(pld_order),
                                              pixel_components=int(pixel_components),
                                              spline_n_knots=int(spline_n_knots),
                                              spline_degree=int(spline_degree),
                                              n_pca_terms=int(n_pca_terms))
                            cdpp = clc.estimate_cdpp()
                            param_space[i0, i1, i2, i3, i4] = cdpp
                        except:
                            param_space[i0, i1, i2, i3, i4] = 1e6

ideal_param_inds = np.where(param_space == np.min(param_space))
```

In the future, it may be beneficial to estimate spline values from properties of the light curve (i.e. number of data points or some function of time).

lightkurve/correctors/pldcorrector.py
```diff
@@ -85,52 +88,158 @@ class PLDCorrector(Corrector):
...
+        Returns
+        -------
+        dm : `.DesignMatrixCollection`
+            `.DesignMatrixCollection` containing pixel, background, and spline
+            components.
+        """
+
+        if background_mask is None:
+            # Default to pixels <1-sigma above the background
+            background_mask = ~self.tpf.create_threshold_mask(3, reference_pixel=None)
```
barentsen, line 158: The docstring and comment above this line say that we use threshold=1, but the code uses threshold=3.  Is there a good reason?  I'd say threshold=1 is a better background mask?
```suggestion
            background_mask = ~self.tpf.create_threshold_mask(1, reference_pixel=None)
```

lightkurve/correctors/pldcorrector.py
```diff
@@ -85,52 +88,158 @@ class PLDCorrector(Corrector):
...
+            If `None` or `all` are passed in, all pixels will be used.
+            If 'pipeline' is passed, the mask suggested by the official pipeline
+            will be returned.
+            If 'threshold' is passed, all pixels brighter than 3-sigma above
+            the median flux will be used.
+        spline_n_knots : int
+            Number of knots in spline.
+        spline_degree : int
+            Polynomial degree of spline.
         n_pca_terms : int
```
barentsen, line 141: `n_pca_terms` doesn't appear to be used at all, we're using the `pixel_components` argument instead, so all references to `n_pca_terms` should probably go.

lightkurve/correctors/pldcorrector.py
nksaunders: Yes, I was testing different values and forgot to change this back. Go ahead and commit this change!

barentsen:
@nksaunders I propose we name the argument `n_pca_terms` instead of `pixel_components` to increases the compatibility with the existing version of Lightkurve (where the same argument is called `n_pca_terms`).  Unless you think `pixel_components` is a much better name?

nksaunders:
> @nksaunders I propose we name the argument `n_pca_terms` instead of `pixel_components` to increases the compatibility with the existing version of Lightkurve (where the same argument is called `n_pca_terms`). Unless you think `pixel_components` is a much better name?

@barentsen I'm fine with keeping `n_pca_terms` for consistency. I think it's descriptive enough, and we only perform PCA on pixel components, so as long as we make that clear in the docstring this should be fine!

barentsen:
(I pushed one more commit to fix that `test_bls` failure.  Other than that, I won't change anything -- let me know if we can merge!)

nksaunders:
Awesome! These changes all look great. I'll squash and merge. Thanks for the help @barentsen!

barentsen:
Thanks @nksaunders! :+1:

PR #749, opened 2020-06-05
Allow TPF cutout() to find edges

astrobel:
Hi,

As discussed with @barentsen, I've made a quick fix to the TPF `cutout` function. To illustrate, this is the example TPF I'm working with:
image

When using cutout to grab a 3x3 pixel square centred on (137,870), it misses the edges:
image

With the change, cutout can now grab TPF edges, and doesn't seem to run into any problems when the centre is too far from the edges to be square.
image


barentsen:
Thanks @astrobel! :+1: :+1: 

Looks like the tests are failing due to an unrelated issue with the remote SkyBot service.

I verified this fix works. We can merge!

PR #750, opened 2020-06-05
Fix warning message when loading light curve files

christinahedges:
Currently search is throwing a warning because we changed `lc.header` to `lc.get_header()`

### DIFF
--- a/lightkurve/search.py
+++ b/lightkurve/search.py
@@ -1023,7 +1023,7 @@ def open(path_or_url, **kwargs):
     # pass header into `detect_filetype()`
     try:
         with fits.open(path_or_url) as temp:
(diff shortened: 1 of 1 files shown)


lightkurve/search.py
```diff
@@ -1023,7 +1023,7 @@ def open(path_or_url, **kwargs):
     # pass header into `detect_filetype()`
     try:
         with fits.open(path_or_url) as temp:
-            filetype = detect_filetype(temp[0].header)
+            filetype = detect_filetype(temp[0].get_header())
```
barentsen, line 1026: The tests are failing because `temp[0]` is a FITS HDU object here, i.e. the use of `.header` here is correct and not the cause of the warnings.

PR #765, opened 2020-06-24
Added `column` argument to `LightCurve` plotting functions

ojhall94:
Following @barentsen's recent PR #744, Lightkurve `LightCurve` objects are a subclass of Astropy Timeseries. This means that for a single `LighCurve`, we have access to a ton of time-dependent properties such as the PDCSAP flux, SAP flux, Momentum and PSF drift, you name it. 

This PR makes it a little easier to plot these different columns side-by-side without having to manipulate your `LightCurve` object. The new behaviour could be as follows:

```python3
lc = lightkurve.search_lightcurve('Kepler-8', quarter=4).download()
ax = lc.plot(normalize=True, label='PDCSAP')
lc.plot(column='sap_flux', ax=ax, normalize=True, label='SAP')
lc.plot(column='sap_bkg', ax=ax, normalize=True, label='Background')
```

In this example, the second line plots the PDCSAP flux, because by default `.plot()` plots what is in the `flux` column, which is by default is the PDCSAP flux. This produces the plot below.

image

In the case where the y-axis units are not flux, `.plot()` will show the name of the chosen `column` keyword along with the units (if applicable) on the y-axis. If you try to use the `.errorbar()` function to plot a property that doesn't have errors (such as `cadenceno` or `timecorr`) it will throw a ValueError.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1262,7 +1262,7 @@ class LightCurve(TimeSeries):
                 if np.any(~np.isnan(flux_err)):
                     ax.errorbar(x=self.time.value, y=flux.value, yerr=flux_err.value, **kwargs)
                 else:
-                    raise ValueError(f"Column `{column}` has no associated errors.")
+                    log.warning(f"Column `{column}` has no associated errors.")
             else:
                 ax.plot(self.time.value, flux.value, **kwargs)
             ax.set_xlabel(xlabel)
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1211,39 +1211,38 @@ class LightCurve(TimeSeries):
                 xlabel = 'Time [JD]'
             else:
                 xlabel = 'Time'
+
         # Default ylabel
         if ylabel is None:
             if "flux" in column:
-                    ylabel = "Flux"
+                ylabel = "Flux"
             else:
-                    ylabel = f"{column}"
+                ylabel = f"{column}"
             if normalize or self.meta.get("normalized"):
                 ylabel = "Normalized " + ylabel
             elif (self[column].unit) and (self[column].unit.to_string() != ''):
                 ylabel += f" [{self[column].unit.to_string('latex_inline')}]"
-            else: 
-                pass
+
         # Default legend label
         if ('label' not in kwargs):
             kwargs['label'] = self.meta.get('label')
-        # Create a copy with the appropriate flux and flux_err columns
-        lc_plot = self.copy()
-        lc_plot.flux = lc_plot[column]
-        # Check if column has associated uncertainties
-        if column in ['flux','sap_flux','sap_bkg','pdscap_flux',
-                        'psf_centr1','psf_centr2','mom_centr1','mom_centr2']:
-            lc_plot.flux_err = lc_plot['{}_err'.format(column)]
-            errors = True
-        else:
-            lc_plot.flux_err = np.full(len(lc_plot.flux), np.nan)
-            errors = False
+
+        flux = self[column]
+        try:
+            flux_err = self[f'{column}_err']
+        except KeyError:
+            flux_err = np.full(len(flux), np.nan)
 
         # Normalize the data if requested
         if normalize:
-            lc_normed = lc_plot.normalize()
+            if column == "flux":
+                lc_normed = self.normalize()
+            else:
+                lc_tmp = self.copy()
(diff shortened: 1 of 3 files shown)


ojhall94:
I think the `column` argument can now also be passed to the `LightCurveCollection.plot()` function, but is not explicitly defined as a kwarg in that function, and seems to mess with the y-axis label a little. Im sure we could fix that easily but thought I'd see what people think the behaviour should be first.

barentsen:
I like the example you posted!

I note that `pandas` plotting functions take `x` and `y` as arguments (e.g. `df.plot(y='sap_flux')`), but `column` seems like a more generic choice for us.  I can imagine some of our non-plotting functions take a `column` argument as well in the future.

ojhall94:
Didn't know that about `pandas`! That's neat. I think keeping the `plot()` function to plot everything as a function of time works well as a guiding tool for the user, and then the `pandas` plotting can be used for people going into a bit more depth. Maybe we could mention it in a tutorial somewhere.

lightkurve/lightcurve.py
```diff
@@ -1213,22 +1213,41 @@ def _create_plot(self, method='plot', ax=None, normalize=False,
...
+                else:
+                    ylabel = 'Flux [{}]'.format(self.flux.unit.to_string("latex_inline"))
             else:
-                ylabel = 'Flux [{}]'.format(self.flux.unit.to_string("latex_inline"))
+                if normalize:
+                    ylabel = 'Normalized {}'.format(column)
+                elif self[column].unit is None:
+                    ylabel = '{}'.format(column)
+                else:
+                    ylabel = '{} [{}]'.format(column, self[column].unit.to_string("latex_inline"))
```
barentsen, line 1229: Setting the default `ylabel` is becoming quite complicated.  I wonder if we could simplify roughly as follows:

```python
if ylabel is None:
    if "flux" in column:
        if normalize or self.meta.get("normalized"):
            ylabel = "Normalized Flux"
        else:
            ylabel = "Flux"
    else:
        ylabel = f"{column}"
    # Add unit if available
    if self[column].unit:
        ylabel += f" {self[column].unit.to_string("latex_inline")}"
```

barentsen:
@ojhall94 I left a few comments above!  This should be good to merge once they are addressed! :+1:

lightkurve/lightcurve.py
```diff
@@ -1186,7 +1186,7 @@ def query_solar_system_objects(self, cadence_mask='outliers', radius=None,
         return res
 
     def _create_plot(self, method='plot', ax=None, normalize=False,
-                     xlabel=None, ylabel=None, title='', style='lightkurve',
+                     xlabel=None, ylabel=None, column='flux', title='', style='lightkurve',
```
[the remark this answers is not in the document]
ojhall94: Good suggestion. I've moved it to the front and done the same in the docstrings for `scatter`, `errorbar`, `plot`.

lightkurve/lightcurve.py
ojhall94: That's a really elegant solution! I've combined our two snippets of code into something that I think covers all the bases.

ojhall94:
Hey @barentsen,

Made your suggested changes! It all works smoothly, as far as I can tell. There is a single line in the updated code that you might disagree with--- if you do, feel free to commit the change directly through Github before merging. I'm indifferent :)

lightkurve/lightcurve.py
```diff
@@ -1213,22 +1213,37 @@ def _create_plot(self, method='plot', ax=None, normalize=False,
...
-            elif self.flux.unit is None:
-                ylabel = 'Flux'
+            if "flux" in column:
+                    ylabel = "Flux"
             else:
-                ylabel = 'Flux [{}]'.format(self.flux.unit.to_string("latex_inline"))
+                    ylabel = f"{column}"
+            if normalize or self.meta.get("normalized"):
+                ylabel = "Normalized " + ylabel
+            elif (self[column].unit) and (self[column].unit.to_string() is not ''):
```
ojhall94, line 1222: @barentsen I added the `(self[column].unit.to_string() is not '')` check for the following reason: I wanted to continue having the unit printed in square brackets for consistency, but wanted to avoid an empty "[]" appearing for the unitless columns. Even though these columns have no units, their data still appear in astropy Quantities, and so always passed this if statement.

barentsen:
Thanks @ojhall94 :+1:

PR #766, opened 2020-06-24
The `seismology` module's stellar parameter estimators now pull Teff from the metadata (if available)

ojhall94:
Previously, the `seismology` sub-module's stellar estimators required a temperature input. This meant that users would have to go searching for a temperature value for the star they were interested in, which isn't always trivial. 

This PR has changed this so that `teff` is no longer a required keyword in the `seismology.estimate_radius()`, `.estimate_mass()` and `.estimate_logg()` functions. If the metadata contains a `teff` (from the Kepler Input Catalogue), it will use this temperature value automatically. Because KIC temperatures tend to not to always be on par with the most contemporary measure of temperature, using this default Teff will result in a LightkurveWarning.

There are some other things to note in this PR:
- If you pass a temperature keyword (as was required before) this overrides the metadata temperature.
- If there is no temperature in the metadata, and you don't pass a temperature kwarg the functions throw a `ValueError`.
- `LombScarglePeriodogram` objects now inherit the `.meta` property from their associated `LightCurve` object if they are created through the `from_lightcurve()` function.
- Added specific unit tests for calling the stellar estimator functions from the `seismology` sub-module.


lightkurve/seismology/core.py
```diff
@@ -602,11 +602,24 @@ def estimate_radius(self, teff, numax=None, deltanu=None):
         """
         numax = self._validate_numax(numax)
         deltanu = self._validate_deltanu(deltanu)
+        if teff == None:
```
barentsen, line 605: Nitpicking: as a general rule we have to use `is None`, because an object *could* be passed to `teff` that defines a weird `==` operator:
```suggestion
        if teff is None:
```

lightkurve/seismology/core.py
```diff
@@ -602,11 +602,24 @@ def estimate_radius(self, teff, numax=None, deltanu=None):
...
+            teff = self.periodogram.meta.get('teff')
+            if teff == None:
```
barentsen, line 607:
```suggestion
            if teff is None:
```

lightkurve/seismology/core.py
```diff
@@ -602,11 +602,24 @@ def estimate_radius(self, teff, numax=None, deltanu=None):
...
+                raise ValueError("No effective temperature in metadata."
+                                "Please include effective temperature in function call.")
```
barentsen, line 609: I wonder if this error message can be more specific about the argument names, for example:
```
You must provide an effective temperature argument (`teff`) to `estimate_radius`,
because the Periodogram object does not contain it in its meta data (i.e. `pg.meta['teff']` is missing).
```

lightkurve/seismology/core.py
```diff
@@ -620,11 +633,24 @@ def estimate_mass(self, teff, numax=None, deltanu=None):
         """
         numax = self._validate_numax(numax)
         deltanu = self._validate_deltanu(deltanu)
+        if teff == None:
```
barentsen, line 636:
```suggestion
        if teff is None:
```

lightkurve/seismology/core.py
```diff
@@ -620,11 +633,24 @@ def estimate_mass(self, teff, numax=None, deltanu=None):
...
+            teff = self.periodogram.meta.get('teff')
+            if teff == None:
```
barentsen, line 638:
```suggestion
            if teff is None:
```

lightkurve/seismology/core.py
```diff
@@ -637,6 +663,18 @@ def estimate_logg(self, teff, numax=None):
             Stellar surface gravity estimate.
         """
         numax = self._validate_numax(numax)
+        if teff == None:
```
barentsen, line 666:
```suggestion
        if teff is None:
```

lightkurve/seismology/core.py
```diff
@@ -637,6 +663,18 @@ def estimate_logg(self, teff, numax=None):
...
+            teff = self.periodogram.meta.get('teff')
+            if teff == None:
```
barentsen, line 668:
```suggestion
            if teff is None:
```

ojhall94:
Thanks for the excellent comments @barentsen! I've made the suggested changes, and duplicated the docstring on the `estimate_` stellar parameters functions between `core.py` and `stellar_estimators.py`.

barentsen:
@ojhall94 Oops, you accidentally added `CHANGES.rst~` to the PR.  This is probably a temporary file created by your text editor.  Can you

```
git rm CHANGES.rst~
git commit -m 'Remove temp file'
git push
````

File: lightkurve/seismology/core.py [hunk 8]
```diff
@@ -602,16 +643,70 @@ def estimate_radius(self, teff, numax=None, deltanu=None):
...
         deltanu = self._validate_deltanu(deltanu)
+        if teff is None:
+            teff = self.periodogram.meta.get('teff')
+            if teff is None:
+                raise ValueError("You must provide an effective temperature argument (`teff`) to `estimate_radius`,"
+                        "because the Periodogram object does not contain it in its meta data (i.e. `pg.meta['teff']` is missing")
+            else:
+                log.info("Using value for effective temperature from the Kepler Input Catalogue."
+                        "These temperatue values may sometimes differ significantly from modern estimates.",
+                        LightkurveWarning)
```
barentsen, line 654: Oops, we need to remove `LightkurveWarning` from all calls to `log.info()`.  The log functions don't take a warning type as an argument.

ojhall94:
That was careless of me-- sorry! Changes made.

PR #768, opened 2020-06-26
Move from_fits_images to TargetPixelFile class

burke86:
This is an initial pull request to move the `from_fits_images()` method into the generic `TargetPixelFile` class to allow it to be used with TESS TPFs.

One issue I have not figured out is a problem with Cutout2D:

```
sky_coord = SkyCoord.from_name('NGC 5548')
search_result = lk.search_tesscut(sky_coord)
tpf = search_result.download(cutout_size=20)
hdu_list = [fits.ImageHDU(tpf.flux[i],header=tpf.get_header()) for i in range(len(tpf.flux))]
tpf = tpf.from_fits_images(hdu_list, sky_coord)
```

>   0%|          | 0/1235 [00:00<?, ?it/s]

> ---------------------------------------------------------------------------
> ValueError                                Traceback (most recent call last)
> <ipython-input-10-03db36804fee> in <module>
> ----> 1 tpf = tpf.from_fits_images(hdu_list, sky_coord)
> 
> ~/.local/lib/python3.6/site-packages/lightkurve/targetpixelfile.py in from_fits_images(images, position, size, extension, target_id, hdu0_keywords, **kwargs)
>    1247             else:
>    1248                 cutout = Cutout2D(hdu.data, position, wcs=wcs_ref,
> -> 1249                                   size=size, mode='partial')
>    1250             factory.add_cadence(frameno=idx, flux=cutout.data, header=hdu.header)
>    1251 
> 
> ~/.local/lib/python3.6/site-packages/astropy/nddata/utils.py in __init__(self, data, position, size, wcs, mode, fill_value, copy)
>     649                 raise ValueError('wcs must be input if position is a '
>     650                                  'SkyCoord')
> --> 651             position = skycoord_to_pixel(position, wcs, mode='all')  # (x, y)
>     652 
>     653         if np.isscalar(size):
> 
> ~/.local/lib/python3.6/site-packages/astropy/wcs/utils.py in skycoord_to_pixel(coords, wcs, origin, mode)
>     532 
>     533     if wcs.naxis != 2:
> --> 534         raise ValueError("WCS should contain celestial component")
>     535 
>     536     # Check which frame the WCS uses
> 
> ValueError: WCS should contain celestial component

This is a mystery to me, as I think the TPF `wcs.naxis` should be 2.

Second point is if one wants to create a new TPF from scratch with `tpf.from_fits_images()`, a TPF must be already created. Perhaps we should implement an option in `TargetPixelFile`'s `__init__()` method that calls `tpf.from_fits_images()`? Or allow an empty `tpf = TESSTargetPixelFile()` object to be created? Let me know what you think!

Colin


barentsen:
@burke86 The WCS error is caused by the fact that the header being passed does not contain a WCS.  By passing `header=tpf.get_header()` the factory receives the header of first TPF extension, whereas the WCS information is contained in the *second* extension of the TPF file (i.e. `tpf.get_header(2)`).

To make matters worse, the second extension of a TPF file is a binary table.  The header contains a WCS for each column, using the special WCS naming scheme for this purpose.  In contrast, the `hdu_list` being created contains traditional FITS images instead of a binary table.

To translate the WCS keywords from one naming scheme to the other, we have to do something like this:

```
wcs_keyword_map = {'1CTYP5': 'CTYPE1',
                  '2CTYP5': 'CTYPE2',
                  '1CRPX5': 'CRPIX1',
                  '2CRPX5': 'CRPIX2',
                  '1CRVL5': 'CRVAL1',
                  '2CRVL5': 'CRVAL2',
                  '1CUNI5': 'CUNIT1',
                  '2CUNI5': 'CUNIT2',
                  '1CDLT5': 'CDELT1',
                  '2CDLT5': 'CDELT2',
                  '11PC5': 'PC1_1',
                  '12PC5': 'PC1_2',
                  '21PC5': 'PC2_1',
                  '22PC5': 'PC2_2',
                  '1CRV5P': 'CRVAL1P',
                  '2CRV5P': 'CRVAL2P'}

old_header = tpf.get_header(1)
new_header = fits.Header({(newkey, old_header[oldkey]) for oldkey, newkey in wcs_keyword_map.items()})
```

You can then use `hdu_list = [fits.ImageHDU(tpf.flux[i], header=new_header) for i in range(len(tpf.flux))]`.

(We should eventually modify `TargetPixelFileFactory` so that it doesn't require a WCS to work.)

We should eventually modify `TargetPixelFileFactory` so that it runs fine even when the images do not have a WCS.

lightkurve/targetpixelfile.py
```diff
@@ -1674,10 +1675,10 @@ def _make_primary_hdu(self, hdu0_keywords):
         # Override the defaults where necessary
         hdu.header['ORIGIN'] = "Unofficial data product"
         hdu.header['DATE'] = datetime.datetime.now().strftime("%Y-%m-%d")
-        hdu.header['TELESCOP'] = "Kepler"
-        hdu.header['CREATOR'] = "lightkurve.KeplerTargetPixelFileFactory"
+        hdu.header['CREATOR'] = "lightkurve.TargetPixelFileFactory"
         hdu.header['OBJECT'] = self.target_id
-        hdu.header['KEPLERID'] = self.target_id
+        if hdu.header['TELESCOP'].lower() == "kepler":
```
barentsen, line 1680: You'll want to use `hdu.header.get()` here as follows:
```suggestion
        if hdu.header.get('TELESCOP', default='').lower() == "kepler":
```

This way, you will get an empty string instead of an error when the `TELESCOP` keyword is missing.

burke86:
Thanks @barentsen! Agreed, it might be nice to not require a WCS or somehow do this translation under-the-hood.

I ran into an issue with `hdu.header.get('TELESCOP', default='')` returning `None` for some reason, so changed that line to `if hdu.header['TELESCOP'] is not None and hdu.header['TELESCOP'] == "kepler":` which is definitely not as clean :/

Also, I see now the meaning of staticmethod allows this usage:
```
from lightkurve.targetpixelfile import TessTargetPixelFile
tpf_new = TessTargetPixelFile.from_fits_images(hdu_list, sky_coord)
```
which is very nice. 👍

burke86:
Looks like the unit tests are failing because they always expect a Kepler TPF, but from_fits_images now will detect the whether to return a generic, TESS, or Kepler TPF from the header keys. I'll wait for your feedback @barentsen about whether to change the unit test or do it a different way. Thanks!

barentsen:
Thanks so much @burke86!

(Yep, static methods can be called without instantiating an object. You can think of them as separate functions.)

I can't look at this today but will provide feedback first thing tomorrow! :+1:

burke86:
Hi @barentsen just wanted to send a friendly reminder/bump about this. It would be great to know if any additional changes need to be made for this to be merged (see note about why unit tests are failing above). Thanks again!

barentsen:
@burke86 Oops, sorry I let this wait for so long!  I won't get to it today but will do tomorrow! :crossed_fingers:

burke86:
Hi @barentsen, this is a merge with the current master branch. I also changed the related tests to expect a generic TPF when appropriate.

barentsen (approved):
@burke86 Please accept my apologies for letting this PR escape my attention for so long.

You did a great job.  This PR is in a great shape and I believe it is ready to merge. :+1:

I promise I'll give your next PR a much faster response. This should have been a quick & easy merge on my part.

Thank you so much for contributing! 🙏

PR #770, opened 2020-07-02
fix bug in diff'ing astropy table

danhey:
Hi all, I'm enjoying the new timeseries functionality lightkurve has! However, I've run into some problems with it appearing to break the Periodogram functionality. The Python kernel appears to gets stuck when performing a np.diff on lc.time? This is a small patch to address the problem, which assumes that the input time is always days. I will likely rewrite a large portion of the periodogram module in the near future to make it more compatible with these updates.

### DIFF
--- a/lightkurve/periodogram.py
+++ b/lightkurve/periodogram.py
@@ -761,7 +761,7 @@ class LombScarglePeriodogram(Periodogram):
         time = lc.time.copy()
 
         # Approximate Nyquist Frequency and frequency bin width in terms of days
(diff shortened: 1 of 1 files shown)


danhey:
Closed to incorporate into #769

barentsen:
Thanks @danhey!  Right, I noticed that `Periodogram` features had slowed down a worrying amount.

Likely related: I found that using `np.isnan(time)` on a large Time object has occasionally frozen my computer!  We ought to open an AstroPy issue.  For now, working around it in Lightkurve is definitely a good idea.

PR #771, opened 2020-07-02
New function for TPF pixel analysis: inspect_pixels()

astrobel:
Hi all,

In this PR, I'd like to introduce the `inspect_pixels()` function for Lightkurve TPFs. Modelled after [this old script](https://github.com/astrobel/chancealignments2/blob/master/4_pixels.py), `inspect_pixels()` produces a plot displaying either the light curves in each pixel, or their associated periodograms. The plots are displayed such that the pixels align with what you'd see using `interact()`.

Here's an example plot, showing the normalised light curves for KIC 2437317, as a result of this line of code:

```tpf.inspect_pixels(normalized=True, periodogram=False, aperture_mask='threshold', threshold=2)```

image

Note also that the function plots an aperture mask, whether pipeline, threshold, or user-generated, and can take a threshold value as a kwarg.

There are still some things that need to be improved/tested, and I'd appreciate anyone's feedback & input:
- The pixel size to figure size ratio is fine for regular TPFs, but what about bright/saturated stars which have large TPFs?
- At the moment, the periodogram version takes a very long time to run. Any ideas for how to trim that down, or if it's possible?
- The normalized light curves often retain some long-term trends, making the periodograms slightly less useful at the moment.
- In my original version of this function, users have the option to limit the scale of the periodogram, so for example you might only want to see the first 10% of the frequency range up to the Nyquist frequency. Would people be interested in seeing that implemented here?

Thanks!

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -1230,7 +1230,7 @@ class TargetPixelFile(object):
                         else:
                             pixel_list.append(lc)
 
-mask = self._parse_aperture_mask(aperture_mask)
+        mask = self._parse_aperture_mask(aperture_mask)
 
         with plt.style.context(style):
             if title is None:
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -1310,9 +1310,9 @@ class TargetPixelFile(object):
                             gax.plot(x_vals, y_vals, 'w.', ms=0.5)
                     else:
                         if periodogram:
-                            gax.plot(x_vals, y_vals, 'w-', lw=0.5)
+                            gax.plot(x_vals, y_vals, 'k-', lw=0.5)
                         elif not periodogram:
-                            gax.plot(x_vals, y_vals, 'w.', ms=0.5)
+                            gax.plot(x_vals, y_vals, 'k.', ms=0.5)
                     
                     gax.set_xlim(x_axis_min, x_axis_max)
                     gax.set_ylim(y_axis_min, y_axis_max)
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -1208,13 +1208,18 @@ class TargetPixelFile(object):
             title = f'Target ID: {self.targetid}'
         if corrector_func is None:
             corrector_func = lambda x: x.remove_outliers()
+        if show_flux:
+            cmap = plt.get_cmap()
+            norm = plt.Normalize(vmin=np.nanmin(self.flux[0].value),
+                                 vmax=np.nanmax(self.flux[0].value))
         mask = self._parse_aperture_mask(aperture_mask)
 
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", category=(RuntimeWarning, LightkurveWarning))
 
             # get an aperture mask for each pixel
-            masks = np.zeros((self.shape[1]*self.shape[2], self.shape[1], self.shape[2]), dtype='bool')
+            masks = np.zeros((self.shape[1]*self.shape[2], self.shape[1], self.shape[2]),
+                             dtype='bool')
             for i in range(self.shape[1]*self.shape[2]):
                 masks[i][np.unravel_index(i, (self.shape[1], self.shape[2]))] = True
             
@@ -1234,16 +1239,12 @@ class TargetPixelFile(object):
                     else:
                         pixel_list.append(lc)
 
-        tpf_plot = plt.imshow(self.flux[0])
-        plt.close()
-
         with plt.style.context(style):
             fig = plt.figure()
-            if ax is None:
+            if ax is None:  # Configure axes if none is given
                 ax = plt.gca()
                 ax.get_xaxis().set_ticks([])
                 ax.get_yaxis().set_ticks([])
-
                 if periodogram:
                     ax.set(title=title, xlabel='Frequency', ylabel='Power')
                 else:
@@ -1258,32 +1259,20 @@ class TargetPixelFile(object):
                     if periodogram:
                         x_vals = pixel_list[k].frequency.value
                         y_vals = pixel_list[k].power.value
-                        x_axis_min = 0
-                        x_axis_max = np.nanmax(x_vals)
-                        y_axis_min = 0
-                        y_axis_max = max(np.nanmax(y_vals), 1e-99)
                     else:
                         x_vals = pixel_list[k].time.value
                         y_vals = pixel_list[k].flux.value
-                        y_val_range = np.nanmax(y_vals) - np.nanmin(y_vals)
-                        x_axis_min = np.nanmin(x_vals)
-                        x_axis_max = np.nanmax(x_vals)
-                        y_axis_min = np.nanmin(y_vals) - 0.05*max(y_val_range, 0.1)
-                        y_axis_max = np.nanmax(y_vals) + 0.05*max(y_val_range, 0.1)
 
-                    no_mask = False
-                    if len(np.unique(mask)) == 1:
-                        no_mask = True
-
-                    if mask[x,y] and not no_mask:
-                        with plt.rc_context(rc={"axes.linewidth": 2, "axes.edgecolor": 'red'}):
-                            gax = fig.add_subplot(gs[self.shape[1] - x - 1,y])
+                    # Highlight aperture mask in red
+                    if aperture_mask is not None and mask[x,y]:
+                        rc = {"axes.linewidth": 2, "axes.edgecolor": 'red'}
                     else:
-                        with plt.rc_context(rc={"axes.linewidth": 1}):
-                            gax = fig.add_subplot(gs[self.shape[1] - x - 1,y])
+                        rc = {"axes.linewidth": 1}
+                    with plt.rc_context(rc=rc):
+                        gax = fig.add_subplot(gs[self.shape[1] - x - 1, y])
 
                     if show_flux:
-                        gax.set_facecolor(tpf_plot.cmap(tpf_plot.norm(self.flux.value[0,x,y])))
+                        gax.set_facecolor(cmap(norm(self.flux.value[0,x,y])))
                         if periodogram:
                             gax.plot(x_vals, y_vals, 'w-', lw=0.5)
                         else:
@@ -1294,8 +1283,7 @@ class TargetPixelFile(object):
                         else:
                             gax.plot(x_vals, y_vals, 'k.', ms=0.5)
                     
-                    gax.set_xlim(x_axis_min, x_axis_max)
-                    gax.set_ylim(y_axis_min, y_axis_max)
+                    gax.margins(y=.1, x=0)
                     gax.set_xticklabels('')
                     gax.set_yticklabels('')
                     gax.set_xticks([])
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -1169,7 +1169,7 @@ class TargetPixelFile(object):
         return self.__class__(newfits)
 
 
-    def inspect_pixels(self, ax=None, normalize=False, periodogram=False, aperture_mask=None, show_flux=False, style='lightkurve', title=None, **kwargs):
+    def plot_pixels(self, ax=None, normalize=False, periodogram=False, aperture_mask=None, show_flux=False, style='lightkurve', title=None, **kwargs):
         """ Plot the light curves or associated periodograms for each pixel in one quarter
         Note that all values are autoscaled and axis labels are not provided.
         This utility is designed for by-eye inspection of signal morphology.
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -1169,7 +1169,7 @@ class TargetPixelFile(object):
         return self.__class__(newfits)
 
 
-    def inspect_pixels(self, ax=None, normalized=False, periodogram=False, aperture_mask=None, style='lightkurve', title=None, **kwargs):
+    def inspect_pixels(self, ax=None, normalize=False, periodogram=False, aperture_mask=None, style='lightkurve', title=None, **kwargs):
         """ Plot the light curves or associated periodograms for each pixel in one quarter
         Note that all values are autoscaled and axis labels are not provided.
         This utility is designed for by-eye inspection of signal morphology.
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -1219,12 +1219,12 @@ class TargetPixelFile(object):
                     except IndexError:
                         pixel_list.append(None)
                 else:
-                    if normalized == True:
+                    if normalize == True:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lightkurve/targetpixelfile.py [hunk 1]
```diff
@@ -1130,6 +1131,124 @@ def cutout(self, center=None, size=5):
...
+                            pixel_list.append(lc)
+
+        if aperture_mask == 'pipeline':
+            mask = self.pipeline_mask
+        elif aperture_mask == 'threshold':
+            mask = self.create_threshold_mask(**kwargs)
+        elif aperture_mask == 'all' or aperture_mask == None:
+            mask = np.zeros((self.shape[1], self.shape[2]), dtype='bool')
+        else:
+            mask = aperture_mask
```
barentsen, line 1202: Good news: the class has a hidden method called `_parse_aperture_mask` to take care of this for you, so I think you can simplify as follows:
```suggestion
mask = self._parse_aperture_mask(aperture_mask)
```

lightkurve/targetpixelfile.py
```diff
@@ -1130,6 +1131,124 @@ def cutout(self, center=None, size=5):
...
+
+        with plt.style.context(style):
+            if title is None:
+                title = 'Target ID: {}, Quarter: {}'.format(self.targetid, self.quarter)
```
barentsen, line 1206: K2 and TESS data don't have a quarter number, so I suggest showing only the target id:
```suggestion
                title = f'Target ID: {self.targetid}'
```
(Pro-tip: now that Lightkurve is Python 3 only, we can use f-strings as shown above, which are awesome!)

lightkurve/targetpixelfile.py
```diff
@@ -1130,6 +1131,124 @@ def cutout(self, center=None, size=5):
         return self.__class__(newfits)
 
 
+    def inspect_pixels(self, ax=None, normalized=False, periodogram=False, aperture_mask=None, style='lightkurve', title=None, **kwargs):
```
barentsen, line 1134: Nitpicking: let's name the argument `normalize` instead of `normalized`, for consistency with the `LightCurve.plot()` methods which also accept a `normalize` argument.
```suggestion
    def inspect_pixels(self, ax=None, normalize=False, periodogram=False, aperture_mask=None, style='lightkurve', title=None, **kwargs):
```

barentsen, line 1134: Regarding the name: I like the idea of having every method that creates a plot start with the prefix `plot_`, so they are easy to discover.  For example, we have `plot_river()` in LightCurve and `plot_diagnostics()` in RegressionCorrector.  Following this philosophy, I wonder if e.g.  `plot_pixels()`, `plot_pixelseries()`, or `plot_pixelgrid()` would be a good name for this feature? What do you think?

barentsen:
@astrobel  To add a basic unit test for this feature, you can start by adding the following to the bottom of `lightkurve/tests/test_targetpixelfile.py`:
```python
def test_plot_pixels():
    tpf = KeplerTargetPixelFile(filename_tpf_one_center)
    tpf.inspect_pixels()
    plt.close('all')
```

Then, simply type `pytest -k test_plot_pixels` to verify that it all works.  (More guidance on running the tests is available here: https://docs.lightkurve.org/about/testing.html )

(Next, you can imagine extending the test to use optional arguments etc.)

barentsen:
Just for inspiration: I noticed that @noraeisner uses the following plotting style as part of a similar plot produced by the [LATTE](https://github.com/noraeisner/LATTE) package:

image

lightkurve/targetpixelfile.py
astrobel: Good point! I'm in favour of keeping it simple, so how about `plot_pixels()`?

astrobel:
Thanks for all the feedback, @barentsen! I'm starting to work on tests now, so I'll keep you updated. I'm definitely going to pass kwargs to the periodogram.

As for running systematics removal on the pixels, it would certainly be useful for my purposes, and I'd be keen to hear if more people would like that as a feature! In that case, would we then remove normalization and replace it with this?

I love the pixel plot from LATTE. I'm going to look into implementing an option that allows users to toggle the colour scale background, as well as see if I can get the mask working as an outline instead of shading to accommodate that (whether or not the colour scale background is turned on).

ojhall94:
Hey @astrobel!

This PR is awesome. I've had a little play around with it and am having a strange bug--- the plot doesn't appear in the same style as it does in your (very pretty) example at the top of this PR, and the `normalized` kwarg doesnt' seem to affect the appearance of the plot. Not sure if this is a true bug or a result of this PR being a WIP, but looking at the code I would expect the plot to follow the MPLSTYLE.

I don't think any changes have happened to lightkurve `master` branch since you branched off for this PR--- if you rebase this branch onto master (or merge master into this branch) I think it should fix the slow periodogram issue, which will make everything easier to test.

image

astrobel:
Thanks @ojhall94! The plot's look is a good catch; that's both a bug and a WIP thing... I've been updating the plotting function to do some different stuff and it seems like when there's no aperture specified it treats every pixel as part of the aperture. Thankfully that's easily fixed!

(Though I can't work out why the y axis label seems to have disappeared... ?)

And `normalized` isn't working because a few hours ago I changed it to `normalize` as per [this commit](https://github.com/KeplerGO/lightkurve/pull/771/commits/86a31f3c24be0f78b79475997b5ee0ffd82df8b3).

astrobel:
Just a quick update on where I'm at with this PR! Following @barentsen's suggestion above (and inspired by @noraeisner's work, though I didn't look at the LATTE code, just the image), I've added an option to plot the flux (in frame zero - would people maybe prefer an average?) as a face colour to each pixel, in which case the flux/power will be plotted in white instead of black for visibility. I've also added a new-look aperture outline, which is the same whether or not the flux is plotted; plotting fluxes (but not power) in dots for legibility; and tidied up the figure size to make the pixels square. This is definitely going to make larger TPFs unwieldy, though, so I'd like to look more into scaling it!

image

The above plot is generated by running `tpf.plot_pixels(normalize=True, show_flux=True, aperture_mask='threshold')`. (Also note the new function name!)

I've also added the ability to pass kwargs to `to_periodogram`. Previously the kwargs were for specifying a threshold, but `_parse_aperture_mask` is unaware of changing the threshold by kwarg, so that keyword isn't needed anymore. Now, for example, you can use the `nyquist_factor` keyword to only plot up to a certain fraction of the Nyqust frequency:

image

... from `tpf.plot_pixels(periodogram=True, show_flux=True, aperture_mask='threshold', nyquist_factor=0.5)`.

I've also begun to implement some tests, which you can see in the latest commit.

danhey:
Hey Isabel, awesome work! I'm a huge fan of this PR. If I can put in my 2 cents, I think a cool feature would be to allow passing a corrector function which would apply to each light curve generated from the pixel. A good example of this is [here](https://github.com/KeplerGO/lightkurve/blob/master/lightkurve/collections.py#L101).

astrobel:
@danhey Corrector functions are here! Let me know if you (or anyone!) get a chance to play around with it, as I've only just tested some basic usage now.

lightkurve/tests/test_targetpixelfile.py
```diff
@@ -624,3 +624,16 @@ def test_get_header():
     # ``tpf.header`` is deprecated
     with pytest.warns(LightkurveDeprecationWarning, match='deprecated'):
         tpf.header
+
+def test_plot_pixels():
+    tpfs = [KeplerTargetPixelFile(filename_tpf_one_center), TessTargetPixelFile(filename_tpf_one_center)]
```
barentsen, line 629: The statement `TessTargetPixelFile(filename_tpf_one_center)` will yield this warning during the tests:
```LightkurveWarning: A Kepler data product is being opened using the `TessTargetPixelFile` class.```

While the warning could be silenced, it would be fine to simply run this test just for `KeplerTargetPixelFile(filename_tpf_one_center)`.  That will get rid of the warning without reducing the test coverage in any significant way.

File: lightkurve/targetpixelfile.py [hunk 5]
```diff
@@ -1130,6 +1131,163 @@ def cutout(self, center=None, size=5):
...
+                        y_axis_min = 0
+                        y_axis_max = np.nanmax(y_vals)
+                    elif periodogram == False:
+                        x_vals = pixel_list[k].time.value
+                        y_vals = pixel_list[k].flux.value
+                        y_val_range = np.nanmax(y_vals) - np.nanmin(y_vals)
+                        x_axis_min = np.nanmin(x_vals)
+                        x_axis_max = np.nanmax(x_vals)
+                        y_axis_min = np.nanmin(y_vals) - 0.05*y_val_range
+                        y_axis_max = np.nanmax(y_vals) + 0.05*y_val_range
```
barentsen, line 1254: The two lines above will indirectly trigger the following warning during the tests:
```
/home/gb/dev/lightkurve/lightkurve/targetpixelfile.py:1280: UserWarning: Attempting to set identical bottom == top == 1.0 results in singular transformations; automatically expanding.
    gax.set_ylim(y_axis_min, y_axis_max)
```

This is because the unit test uses a fake TPF (`filename_tpf_one_center`) in which the pixel values are constant over time (i.e. the pixel values in that file are always zero, except for the center pixel, which is always one).

It may not be completely unusual to have TPFs with pixels that are zero all the time, so the best workaround here may be something like this:
```suggestion
                        y_axis_min = np.nanmin(y_vals) - 0.05*max(y_val_range, 0.1)
                        y_axis_max = np.nanmax(y_vals) + 0.05*max(y_val_range, 0.1)
```

lightkurve/targetpixelfile.py
```diff
@@ -1130,6 +1131,163 @@ def cutout(self, center=None, size=5):
...
+        """ Plot the light curves or associated periodograms for each pixel in one quarter
+        Note that all values are autoscaled and axis labels are not provided.
+        This utility is designed for by-eye inspection of signal morphology.
+        
+        Parameters
+        ----------
+        ax : `~matplotlib.axes.Axes`
+            A matplotlib axes object to plot into. If no axes is provided,
+            a new one will be generated.
+        normalized : bool
```
barentsen, line 1144:
```suggestion
        normalize : bool
```

lightkurve/targetpixelfile.py
```diff
@@ -1130,6 +1131,163 @@ def cutout(self, center=None, size=5):
...
+        kwargs : dict
+            e.g. `threshold` to be passed to create_threshold_mask
+        
+        """
+
+        if style == 'lightkurve' or style is None:
+            style = MPLSTYLE 
+        
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
```
barentsen, line 1173: Complete nit-pick: it's always good to limit the types of warnings being ignored, so we don't accidentally silence e.g. an important deprecation warning:
```suggestion
            warnings.simplefilter("ignore", category=(RuntimeWarning, LightkurveWarning))
```

lightkurve/targetpixelfile.py
```diff
@@ -1130,6 +1131,163 @@ def cutout(self, center=None, size=5):
...
+            for i in range(self.shape[1]*self.shape[2]):
+                masks[i][np.unravel_index(i, (self.shape[1], self.shape[2]))] = True
+            
+            pixel_list = []
+
+            for j in range(self.shape[1]*self.shape[2]):
+
+                lc = self.to_lightcurve(aperture_mask=masks[j])
+
+                if normalize: # overrides corrector function
```
barentsen, line 1186: Let's change this into:
```suggestion
                if normalize or corrector_func is None: # overrides corrector function
```
...which would allow us to get rid of lines 1189-1191, which reduces code duplication! :+1:

lightkurve/targetpixelfile.py
```diff
@@ -1130,6 +1131,163 @@ def cutout(self, center=None, size=5):
...
+                    x, y = np.unravel_index(k, (self.shape[1], self.shape[2]))
+
+                    if periodogram:
+                        x_vals = pixel_list[k].frequency.value
+                        y_vals = pixel_list[k].power.value
+                        x_axis_min = 0
+                        x_axis_max = np.nanmax(x_vals)
+                        y_axis_min = 0
+                        y_axis_max = max(np.nanmax(y_vals),1e-99)
+                    elif not periodogram:
```
barentsen, line 1247: Nitpick: this can of course just be `else`:
```suggestion
                    else:
```

lightkurve/targetpixelfile.py
```diff
@@ -1130,6 +1131,163 @@ def cutout(self, center=None, size=5):
...
+                            gax = fig.add_subplot(gs[self.shape[1] - x - 1,y])
+                    else:
+                        with plt.rc_context(rc={"axes.linewidth":1}):
+                            gax = fig.add_subplot(gs[self.shape[1] - x - 1,y])
+
+                    if show_flux:
+                        gax.set_facecolor(tpf_plot.cmap(tpf_plot.norm(self.flux[0,x,y])))
+                        if periodogram:
+                            gax.plot(x_vals, y_vals, 'w-', lw=0.5)
+                        elif not periodogram:
```
barentsen, line 1271:
```suggestion
                        else:
```

barentsen:
@astrobel Oops, I am getting blank plots by default, unless I use `show_flux=True`.  Could it be that the plot is showing white markers on a white background by default?

Example:

pixelplot

astrobel:
> @astrobel Oops, I am getting blank plots by default, unless I use `show_flux=True`. Could it be that the plot is showing white markers on a white background by default?

Yes, looks like that was just a typo on my end! Should be fixed now.

barentsen:
Thanks again @astrobel!

I took the liberty to rebase your PR on the `master` branch, because there have been a few changes there which affect this PR. Most notably, `tpf.flux` is now an AstroPy quantity, which was causing the tests to fail.

In addition, I spotted a few more ways to simplify this function:
* I removed the logic related to setting the x and y limits.  I found that we can achieve identical behavior by calling `gax.margins(y=.1, x=0)`.  This tells matplotlib to  auto-scale the X limits to the min-max range, and the Y limits to the min-max range plus 10%.
* I removed the `normalize` option.  Given that we don't show Y axis ticks, and given that matplotlib auto-scales the Y limits on each subplot, I found that setting `normalize` to either True or False didn't actually change the plot.
* I noticed that a second helper plot was created using `imshow` for the purpose of being able to map flux values to a correct color when `show_flux=True`:
```
tpf_plot = plt.imshow(self.flux[0].value)
plt.close()
```
I simplified this by removing this helper plot, and instead create the desired colormap and normalization function directly, which is faster:
```
cmap = plt.get_cmap()
norm = plt.Normalize(vmin=np.nanmin(self.flux[0].value), vmax=np.nanmax(self.flux[0].value))
```

Let me know if you have any concerns about the changes I made.  Otherwise, I'm happy for this to be merged!

astrobel:
Hi @barentsen, thanks for all these changes! Everything looks good to me, and I'm happy for it to be merged too!

barentsen:
Thanks again @astrobel! :+1: :+1: :+1:

PR #778, opened 2020-07-15
modified one line in targetpixelfile.py

KenMighell:
[lightkurve_bug_report_2020JUL15.ipynb.txt](https://github.com/KeplerGO/lightkurve/files/4928294/lightkurve_bug_report_2020JUL15.ipynb.txt)

Modified one line in targetpixelfile.py:

ra, dec = w.wcs_pix2world(X.ravel(), Y.ravel(), 1)

should be changed to 

ra, dec = w.wcs_pix2world(X.ravel(), Y.ravel(), 0)

The attached Python3 notebook shows the problem with the existing code.  Essentially wcs_pix2world
needs to be called with an origins value of zero and not one when one is using NUMPY (see the attached
astropy documentation snippet in PNG format).

astropy wcs wcs_pix2world_documentation

### DIFF
--- a/lightkurve/targetpixelfile.py~
+++ b/lightkurve/targetpixelfile.py~
@@ -363,7 +363,7 @@ class TargetPixelFile(object):
              np.atleast_3d(pos_corr2_pix).transpose([1, 2, 0]))
 
         # Pass through WCS
-        ra, dec = w.wcs_pix2world(X.ravel(), Y.ravel(), 0)  # KJM: was ...,1)
+        ra, dec = w.wcs_pix2world(X.ravel(), Y.ravel(), 1)
         ra = ra.reshape((pos_corr1_pix.shape[0], self.shape[1], self.shape[2]))
         dec = dec.reshape((pos_corr2_pix.shape[0], self.shape[1], self.shape[2]))
         ra, dec = ra[self.quality_mask], dec[self.quality_mask]
--- a/requirements.txt~
+++ b/requirements.txt~
@@ -1,5 +1,5 @@
 numpy>=1.11
-astropy>=4.0
+astropy>=3.2
 scipy>=0.19.0,!=1.4.0,!=1.4.1
 matplotlib>=1.5.3
 astroquery>=0.3.10


KenMighell:
A conflict with the requirments.txt file was reported.  This pull request did not touch that file.

KenMighell:
When using tpf.get_coordinates with the current code, the right ascension and declination value are off by one row and one column.  A practical effect of this bug is that the target object does not fall on the correct location within an overlay of the aperture on a survey image (e.g., 2MASS-J).

barentsen:
@KenMighell Thank you for testing.

The origin to be used in these conversions must match that of the Kepler pipeline, which adopts the FITS standard i.e. origin=1.  

While Numpy uses a different zero-based origin, it has no bearing here because what matters is the origin convention that was used to produce the WCS.

This can be verified using the following test:

```python
import lightkurve as lk
import math
tpf = lk.search_targetpixelfile("Kepler-10")[10].download()
col, row = tpf.wcs.all_world2pix(tpf.hdu[1].header['1CRVL5'],tpf.hdu[1].header['2CRVL5'], 1)
assert math.isclose(col, tpf.hdu[1].header['1CRPX5'])
assert math.isclose(row, tpf.hdu[1].header['2CRPX5'])
```

In contrast, this test fails for origin=0.

Does that make sense?

KenMighell:
Hi Geert,

GitHub sent your email to mighell@noao.edu which I very 
soon will
no longer be able to access.  kmighell@seti.org or 
kenneth.j.mighell@nasa.gov are preferred email addresses.

Your question: Does that make sense?

Yes.  If you are using the position of (1CRPX5,2CRPX5)
which come from keywords in the FITS header. Using NUMPY
zero-offest coordinates, the position of 
(1CRPX5-1,2CRPX5-1)
gives the correct ra and dec values when origin=0 is used.

In tpf.get_coordinates() the lower left pixel has a 
location
of (0,0):

X, Y = np.meshgrid(np.arange(self.shape[2]), 
np.arange(self.shape[1]))

which is correct for NUMPY arrays.

Today I submitted a new pull request that includes a 
Python3
function which detects bad ra and dec values when origin=1 
is
used in tpf.get_coordinates().  As matters now stand, when 
one
uses the ra and dec values created by 
tpf.get_coordinates() to
create an overlay of a TPF aperture on a 2MASS-J survey 
image,
the target will be incorrectly seen one row and one column 
away
from the true position.

As an alternative to changing the origin value in the 
pix2world
call, one should be able to keep the call as is *but* add 
one to
*all* elements in the X and Y arrays. That way, the input 
coordinates
are FITS-compliant and the use of origin=1 is appropriate.
Mind you, lightkurve (numpy) users must still use 
zero-offset
coordinates in order to access the proper pixels given, 
for
example, in tpf.flux.

As a sanity check, ds9 gives the ra and dec positions 
produced by
using origin=0 with X running from 0.. and Y running from 
0..

Now I get to ask you: Does that make sense?  :-)

In my opinion, the discussion of the use of the origin 
parameter
in the astropy.wcs documentation could significantly 
improved.

Please call my apartment in Santa Clara (669-230-4085) if 
you
would like to discuss this further.

Cheers,

-Ken

PR #780, opened 2020-07-16
Add check on uniform frequency spacing to ACF2D seismology estimators (Fixed #774)

### LINKED ISSUE #774: array size mismatch when using seismology.estimate_numax()
#### Problem description

For some combinations of values for the keyword arguments window_width and spacing supplied to the acf2d function, estimate_numax() fails with an array size mismatch error. This seems to be because the preallocated array size for the output of utils.autocorrelate does not actually match the size of the array returned by it.

In particular, this seems to happen for red giants observed at short cadence. While I can get the acf2d method to work by masking out frequencies above an (artificial) maximum frequency, this seems like a more subtle issue: just because the code doesn't break doesn't mean it's necessarily not otherwise faulty.

#### Example

```python
import numpy as np
import lightkurve as lk

def get_PG():
        
    data = np.loadtxt("PSD_all.pow")
    ν = data[..., 0] * u.uHz
    P = u.Quantity(data[..., -1])
    
#     m = ν.value < 500
    m = slice(None, None)
    
    return lk.Periodogram(ν[m], P[m])

pg = get_PG()
seis = pg.flatten().to_seismology()
seis.estimate_numax(window_width=25, spacing=1)
```

For the provided power spectrum (see here: [PSD_all.zip](https://github.com/KeplerGO/lightkurve/files/4882335/PSD_all.zip)), this fails with spacing < 1.49945671858..., which I don't have a good explanation for:

```
<ipython-input-78-76ccb8c22ce8> in <module>
----> 1 seis.estimate_numax(window_width=25, spacing=1.499456718580)

~/.local/lib/python3.8/site-packages/lightkurve/seismology/core.py in estimate_numax(self, method, **kwargs)
    544         if method == "acf2d":
    545             from .numax_estimators import estimate_numax_acf2d
--> 546             result = estimate_numax_acf2d(self.periodogram, **kwargs)
    547         self.numax = result
    548         return result

~/.local/lib/python3.8/site-packages/lightkurve/seismology/numax_estimators.py in estimate_numax_acf2d(periodogram, numaxs, window_width, spacing)
    150     for idx, numax in enumerate(numaxs):
    151         acf = utils.autocorrelate(periodogram, numax, window_width=window_width, frequency_spacing=fs)      #Return the acf at this numax
--> 152         acf2d[:, idx] = acf                                     #Store the 2D acf
    153         metric[idx] = (np.sum(np.abs(acf)) - 1 ) / len(acf)  #Store the max acf power normalised by the length
    154 

ValueError: could not broadcast input array from shape (554) into shape (584)
```

(the line numbers don't match up because I added a bunch of print statements to see what's going on)

#### Expected behavior

It should not return a ValueError.

#### Environment

-  Linux (Arch)
-  lightkurve version 1.11.1
-  installed with pip

ojhall94:
I've added a check when `estimate_numax()` or `estimate_deltanu()` are called that throws a `ValueError` if the underlying periodogram does not have uniformly spaced frequencies. An identical check was already in place in other `periodogram` functions that required uniform spacing, and not including it here was an oversight (so I'm glad it's now fixed!).

### DIFF
--- a/lightkurve/seismology/core.py
+++ b/lightkurve/seismology/core.py
@@ -544,8 +544,6 @@ class Seismology(object):
         if method == "acf2d":
             from .numax_estimators import estimate_numax_acf2d
 
-
-            print("CODE IS HERE")
             fs = np.mean(np.diff(self.periodogram.frequency))
             if not np.isclose(np.median(np.diff(self.periodogram.frequency.value)), fs.value):
                 raise ValueError("the ACF 2D method requires that the periodogram "
--- a/lightkurve/seismology/core.py
+++ b/lightkurve/seismology/core.py
@@ -592,7 +592,7 @@ class Seismology(object):
 
             result = estimate_deltanu_acf2d(self.periodogram, numax=numax)
 
-        self.deltanu = resultq
+        self.deltanu = result
         return result
 
     def diagnose_deltanu(self, deltanu=None):
--- a/lightkurve/seismology/core.py
+++ b/lightkurve/seismology/core.py
@@ -545,7 +545,7 @@ class Seismology(object):
             from .numax_estimators import estimate_numax_acf2d
 
             fs = np.mean(np.diff(self.periodogram.frequency))
-            if not np.isclose(np.median(np.diff(self.periodogram.append=frequency.value)), fs.value):
+            if not np.isclose(np.median(np.diff(self.periodogram.frequency.value)), fs.value):
                 raise ValueError("the ACF 2D method requires that the periodogram "
                                  "has a grid of uniformly spaced frequencies.")
 
--- a/lightkurve/seismology/core.py
+++ b/lightkurve/seismology/core.py
@@ -547,7 +547,7 @@ class Seismology(object):
             # Detect whether the frequency grid is evenly-spaced
             # by verifying that the first differences are all equal
             freqdiff = np.diff(self.periodogram.frequency.value)
-            if not np.allclose(freqdiff[0], freqdiff)
+            if not np.allclose(freqdiff[0], freqdiff):
                 raise ValueError("the ACF 2D method requires that the periodogram "
                                  "has a grid of uniformly spaced frequencies.")
 
@@ -590,7 +590,7 @@ class Seismology(object):
             # Detect whether the frequency grid is evenly-spaced
             # by verifying that the first differences are all equal
             freqdiff = np.diff(self.periodogram.frequency.value)
-            if not np.allclose(freqdiff[0], freqdiff)
+            if not np.allclose(freqdiff[0], freqdiff):
                 raise ValueError("the ACF 2D method requires that the periodogram "
                                  "has a grid of uniformly spaced frequencies.")
 
--- a/lightkurve/seismology/core.py
+++ b/lightkurve/seismology/core.py
@@ -544,6 +544,8 @@ class Seismology(object):
         if method == "acf2d":
             from .numax_estimators import estimate_numax_acf2d
 
+
+            print("CODE IS HERE")
             fs = np.mean(np.diff(self.periodogram.frequency))
             if not np.isclose(np.median(np.diff(self.periodogram.frequency.value)), fs.value):
                 raise ValueError("the ACF 2D method requires that the periodogram "
--- a/lightkurve/seismology/core.py
+++ b/lightkurve/seismology/core.py
@@ -543,14 +543,6 @@ class Seismology(object):
         method = validate_method(method, supported_methods=["acf2d"])
         if method == "acf2d":
             from .numax_estimators import estimate_numax_acf2d
-
-            # Detect whether the frequency grid is evenly-spaced
-            # by verifying that the first differences are all equal
-            freqdiff = np.diff(self.periodogram.frequency.value)
-            if not np.allclose(freqdiff[0], freqdiff):
-                raise ValueError("the ACF 2D method requires that the periodogram "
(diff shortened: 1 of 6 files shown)


lightkurve/seismology/core.py
```diff
@@ -578,9 +584,15 @@ def estimate_deltanu(self, method='acf2d', numax=None):
...
+
+            fs = np.mean(np.diff(self.periodogram.frequency))
+            if not np.isclose(np.median(np.diff(self.periodogram.frequency.value)), fs.value):
+                raise ValueError("the ACF 2D method requires that the periodogram "
+                                 "has a grid of uniformly spaced frequencies.")
+
             result = estimate_deltanu_acf2d(self.periodogram, numax=numax)
 
-        self.deltanu = result
+        self.deltanu = resultq
```
barentsen, line 595: I guess this is a typo -- the unit tests fail here with `E       NameError: name 'resultq' is not defined`.

lightkurve/seismology/core.py
```diff
@@ -543,6 +543,12 @@ def estimate_numax(self, method="acf2d", **kwargs):
         method = validate_method(method, supported_methods=["acf2d"])
         if method == "acf2d":
             from .numax_estimators import estimate_numax_acf2d
+
+            fs = np.mean(np.diff(self.periodogram.frequency))
+            if not np.isclose(np.median(np.diff(self.periodogram.frequency.value)), fs.value):
```
barentsen, line 548: I'm confused by this, what is the logic behind comparing
`np.mean(np.diff(self.periodogram.frequency)).value`
against
`np.median(np.diff(self.periodogram.frequency.value))` ?

barentsen:
Thanks @ojhall94!  This looks like a good thing to change, but I have two questions about the implementation, see above :point_up: .

lightkurve/seismology/core.py
ojhall94: Careless of me-- fixed!

lightkurve/seismology/core.py
ojhall94: If the frequencies are uniformly spaced, the first differences of the periodogram frequency should all be (near) identical. In this scenario, the mean and the median of the first differences should also be near identical. If the frequencies are not evenly spaced, the mean and the median of the differences will begin to diverge.

I'm not sure if this is the most robust way of checking this. Basically it just needs a simple operation that checks whether there is uniformity of values in an array--- you might have a more elegant solution!

barentsen: Ahh, I got distracted by the repeated call to `np.diff` and the different location of `.value`, which made me think something more mysterious was happening.  If you want to use this solution, it would be best to document and simplify the code as follows:
```python
# Detect whether the frequency grid is evenly-spaced
# by comparing the median and mean of the first differences
freqdiff = np.diff(self.periodogram.frequency.value)
if not np.isclose(np.median(freqdiff), np.mean(freqdiff)):
```
*However* I think you can further simplify this as follows:
```python
# Detect whether the frequency grid is evenly-spaced
# by verifying that the differences are all equal
freqdiff = np.diff(self.periodogram.frequency.value)
if not np.allclose(freqdiff[0], freqdiff):
```

Does that make sense?

To quickly run the test locally after making this change, you can type `pytest lightkurve/seismology/tests/test_butler.py --remote-data`.

lightkurve/seismology/tests/test_butler.py
```diff
@@ -70,6 +70,14 @@ def test_estimate_numax_basics():
     nmxday = u.Quantity(true_numax*u.microhertz, 1/u.day)
     assert(np.isclose(nmxday, numax, atol=.1*nmxday))
 
+    # Assert numax estimator fails when frequqencies are not uniform
+    f, p, true_numax, _ = generate_test_spectrum()
+    f += np.random.uniform(size=len(f))
+    snr = SNRPeriodogram(f*u.microhertz, u.Quantity(p, None))
+
+    with pytest.raises(ValueError):
+        numax = snr.to_seismology().estimate_numax()
```
barentsen, line 79: A ValueError could be raised for a number of reasons, so I recommend checking that we really get the exception we want here.  This can be done as follows: 
```suggestion
    with pytest.raises(ValueError) as exc:
        numax = snr.to_seismology().estimate_numax()
    assert "uniformly spaced" in str(exc)
```
(You'll also want to do that in the second test below.)

barentsen:
@ojhall94 When you get a chance, I left two additional suggestions (see threads above :point_up:)

lightkurve/seismology/tests/test_butler.py
ojhall94: Thanks! I had to use `str(exc.value)`, not sure if that's a version thing or not, but it works :)

lightkurve/seismology/core.py
ojhall94: I knew you'd have a more elegant solution! I've checked and tested this works!

barentsen:
@ojhall94 I made three more changes:

* Added a `Periodogram._is_evenly_spaced()` helper method to check whether the frequency grid is evenly spaced.  This test was being made in three places now, so the helper method avoids duplication of the slightly non-trivial logic.
* Moved calling the check from `Seismology.estimate_numax()` into the `estimate_numax_acf2d()` helper function.  It wouldn't make sense for one to raise the error and not the other.   Bonus: it keeps all the logic related to the ACF-based estimation method in that helper function, including error handling.   I made the identical change for `estimate_deltanu()` of course.
* Added an entry to `CHANGES.rst`.

This should now be good to merge if the tests pass!

PR #789, opened 2020-07-28
Clarify masking in `RegressionCorrector`

nksaunders:
Currently, `RegressionCorrector` lumps the user-input `cadence_mask` together with the outlier mask. This makes the diagnostic plot a little confusing, because cadences included in the `cadence_mask` will be labeled as outliers. This super simple PR introduces a new parameter of `RegressionCorrector` called `user_mask` (or `input_mask`?) to differentiate between the two distinct sets of cadences.

### Before:
```python
tpf = lk.search_targetpixelfile('K2-199')[0].download()
pld = tpf.to_corrector('pld')
pld.correct(cadence_mask=transit_mask)
pld.diagnose()
```
before

### After:
```python
tpf = lk.search_targetpixelfile('K2-199')[0].download()
pld = tpf.to_corrector('pld')
pld.correct(cadence_mask=transit_mask)
pld.diagnose()
```
after

The `cadence_mask` parameter remains unchanged and it will still include both outliers and masked points, so this change really only applies in the diagnostic plotting step.

### DIFF
--- a/lightkurve/correctors/pldcorrector.py
+++ b/lightkurve/correctors/pldcorrector.py
@@ -397,8 +397,8 @@ class PLDCorrector(RegressionCorrector):
                                             s=10, label='Outliers', ax=ax)
             if self.user_mask is not None:
                 clc[~self.user_mask].scatter(normalize=False, c='dodgerblue',
-                                                marker='x', s=10, label='Masked',
-                                                ax=ax)
+                                             marker='x', s=10, label='Masked',
+                                             ax=ax)
             clc.plot(normalize=False, label='Corrected', ax=ax, c='k')
         return axs
 
--- a/lightkurve/correctors/pldcorrector.py
+++ b/lightkurve/correctors/pldcorrector.py
@@ -393,12 +393,11 @@ class PLDCorrector(RegressionCorrector):
             # Plot final corrected light curve with outliers marked
             ax = axs[2]
             self.lc.plot(ax=ax, normalize=False, alpha=0.2, label='Original')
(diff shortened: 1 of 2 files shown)


barentsen:
Thanks @nksaunders. This improves the plot! :+1:

Just one request: at some point in the future we will be confused as to why `self.cadence_mask` does not contain the `cadence_mask` value passed into the constructor.  Can I propose we use `self.cadence_mask` to hold the user-provided mask, and `self.outlier_mask` to hold the class-generated outliers?

I pushed a commit which explains what I mean.

:arrow_right:  Would you be willing to review if my commit this still does the correct thing?  If so, then please go ahead and merge if the tests pass!

nksaunders:
Thanks @barentsen, this looks good. I was trying to figure out the simplest way to make this change and I like the solution you came up with. 

I made a tiny syntax tweak but otherwise I think it's done. If the tests pass I'll squash and merge!

ojhall94:
Nitpick: Not sure if this is the intended behaviour, but I just ran @nksaunders "Performing PLD with the PLDCorrector in Lightkurve" tutorial which uses these diagnostic plots, and the legend seems to have changed from the example above:

image

I think the tilde in the "~cadence_mask" label might confuse some users-- its probably just fine to call it the cadence mask, I think, or "Masked cadences" instead.

barentsen:
@ojhall94 This was intentional on my part. The API provides access to `cadence_mask`, so I figure it's less confusing to use the exact same word in the diagnostic plot.

Admittedly these plots need explanation. In addition to providing this in docstrings and tutorials, I wonder if we could even print clarifying information using a verbose option.

PR #798, opened 2020-08-07
Periodogram now automatically removes nans from lightcurve

ojhall94:
Following a discussion had on the development of a tutorial, I've updated the `LightCurve` and `Periodogram` classes so that the check for NaN values in the light curve is now checked in the `LightCurve.to_periodogram()` function, instead of the `Periodogram.from_lightcurve()` function. If there are NaN values present when `to_periodogram()` is called, the function removes them and logs a warning informing the user of what's happening.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1604,8 +1604,6 @@ class LightCurve(TimeSeries):
             from . import LombScarglePeriodogram
 
             if (~np.isfinite(self.flux)).any():
-                log.warning('Lightcurve contains NaN values.'
-                            'These are removed before creating the periodogram.')
                 return LombScarglePeriodogram.from_lightcurve(lc=self.remove_nans(), **kwargs)
             else:
                 return LombScarglePeriodogram.from_lightcurve(lc=self, **kwargs)
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1602,7 +1602,13 @@ class LightCurve(TimeSeries):
             return BoxLeastSquaresPeriodogram.from_lightcurve(lc=self, **kwargs)
         else:
             from . import LombScarglePeriodogram
-            return LombScarglePeriodogram.from_lightcurve(lc=self, **kwargs)
+
+            if (~np.isfinite(self.flux)).any():
+                log.warning('Lightcurve contains NaN values.'
+                            'These are removed before creating the periodogram.')
+                return LombScarglePeriodogram.from_lightcurve(lc=self.remove_nans(), **kwargs)
+            else:
+                return LombScarglePeriodogram.from_lightcurve(lc=self, **kwargs)
 
     def to_seismology(self, **kwargs):
         """Returns a `~lightkurve.seismology.Seismology` object for estimating
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1602,11 +1602,7 @@ class LightCurve(TimeSeries):
             return BoxLeastSquaresPeriodogram.from_lightcurve(lc=self, **kwargs)
         else:
             from . import LombScarglePeriodogram
-
-            if (~np.isfinite(self.flux)).any():
-                return LombScarglePeriodogram.from_lightcurve(lc=self.remove_nans(), **kwargs)
-            else:
-                return LombScarglePeriodogram.from_lightcurve(lc=self, **kwargs)
+            return LombScarglePeriodogram.from_lightcurve(lc=self, **kwargs)
 
     def to_seismology(self, **kwargs):
         """Returns a `~lightkurve.seismology.Seismology` object for estimating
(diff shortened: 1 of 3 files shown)


ojhall94:
To be honest having played around with it a little I dont think the warning is necessary--- the removal of NaNs is just absorbed into the process.

lightkurve/lightcurve.py
```diff
@@ -1602,7 +1602,13 @@ def to_periodogram(self, method="lombscargle", **kwargs):
             return BoxLeastSquaresPeriodogram.from_lightcurve(lc=self, **kwargs)
         else:
             from . import LombScarglePeriodogram
-            return LombScarglePeriodogram.from_lightcurve(lc=self, **kwargs)
+
+            if (~np.isfinite(self.flux)).any():
+                log.warning('Lightcurve contains NaN values.'
+                            'These are removed before creating the periodogram.')
```
dhomeier:
```suggestion
                log.info('Lightcurve contains NaN values.'
                            'These are removed before creating the periodogram.')
```
I would suggest not to remove the message entirely, going from previous behaviour of raising an exception to silently "correcting" the same input without any feedback to the user. See general discussion for alternatives.

lightkurve/periodogram.py
```diff
@@ -767,10 +767,6 @@ def from_lightcurve(lc, minimum_frequency=None, maximum_frequency=None,
             raise ValueError('You have input keyword arguments for both frequency and period. '
                              'Please only use one.')
 
-        if (~np.isfinite(lc.flux)).any():
-            raise ValueError('Lightcurve contains NaN values. Use lc.remove_nans()'
-                             ' to remove NaN values from a LightCurve.')
-
```
dhomeier, line 773: As the NaNs are already filtered out in the caller, there is no real need to remove the exception here. And worse, calling `LombScarglePeriodogram.from_lightcurve()` directly will now return an invalid periodogram without warning:
```
>>> pg = lk.periodogram.LombScarglePeriodogram.from_lightcurve(lc.normalize(unit='ppm'))
>>> pg.power
<Quantity [nan, nan, nan, ..., nan, nan, nan] ppm>
```

lightkurve/tests/test_periodogram.py
```diff
@@ -330,13 +330,6 @@ def test_error_messages():
     with pytest.raises(ValueError) as err:
         lc.to_periodogram(frequency=np.arange(10), period=np.arange(10))
 
-    # Don't accept NaNs
-    with pytest.raises(ValueError) as err:
-        lc_with_nans = lc.copy()
-        lc_with_nans.flux[0] = np.nan
-        lc_with_nans.to_periodogram()
-    assert('Lightcurve contains NaN values.' in err.value.args[0])
-
```
dhomeier, line 339: Complementary to the previous comment, I don't have a good feeling about removing a test for invalid input without substitute. One option would be (with the exception kept in place) to replace the check by
`LombScarglePeriodogram.from_lightcurve(lc_with_nans)`
but this probably needs some further discussion if we are willing to have the two methods `to_periodogram` /  `from_lightcurve` behave differently on NaN input.

dhomeier (changes requested):
Overall this looks good and certainly helpful for curating input data.
But as indicated in the inline comments, I am a bit concerned that this makes too drastic modifications of the input data without any user notification. But I don't have much practical expertise; does the NaN removal pop up so frequently that even the warnings become annoying? I noted that `TimeSeries` sets NaNs quite abundantly for missing data, empty bins etc...
I think one option would be to reduce the warning level as suggested, another to add a kwarg to enable `remove_nans` (possibly even defaulting to `True`), yet another, as also sketched, keeping `LombScarglePeriodogram..from_lightcurve` as a NaN-aware method – though that different behaviour could also cause confusion. Or a combination of the three...
And I guess this deserves a Changelog entry.

barentsen:
I agree that:
* removing NaNs should probably happen in the `from_lightcurve` function so the behavior is consistent regardless of how the periodogram is created;
* I would at most use `log.debug` to log NaNs being removed -- it is not worth a noisy warning in my opinion.

ojhall94:
Hey all,

I've made the changes proposed above & discussed in the telecon-- `from_lightcurve` makes the checks now, which is in line with the `BLSPeriodogram` class as well.

lightkurve/periodogram.py
```diff
@@ -725,6 +725,10 @@ def from_lightcurve(lc, minimum_frequency=None, maximum_frequency=None,
         """
         # Input validation
         normalization = validate_method(normalization, ['psd', 'amplitude'])
+        if (~np.isfinite(lc.flux)).any():
```
barentsen, line 728: `remove_nans` will only remove nans, not infinite values, so I think we should do this?
```suggestion
        if np.isnan(lc.flux).any():
```

lightkurve/periodogram.py
```diff
@@ -725,6 +725,10 @@ def from_lightcurve(lc, minimum_frequency=None, maximum_frequency=None,
...
+            lc = lc.remove_nans()
+            log.debug('Lightcurve contains NaN values.'
+            'These are removed before creating the periodogram.')
```
barentsen, line 731: PEP8 indentation:
```suggestion
            log.debug('Lightcurve contains NaN values.'
                      'These are removed before creating the periodogram.')
```

PR #799, opened 2020-08-10
Move `DesignMatrix` validation to `RegressionCorrector.correct` to avoid confusing warning

### LINKED ISSUE #792: Warning when running Regression Corrector Object
```
tpf_2min = lk.search_targetpixelfile(target, mission='TESS', sector=15).download()

aper = tpf_2min.pipeline_mask
sap_lc = tpf_2min.to_lightcurve()


dm = lk.DesignMatrix(tpf_2min.flux[:, ~aper], name='pixels').pca(5).append_constant()


reg = lk.RegressionCorrector(sap_lc)
corrected_lc = reg.correct(dm)
ax = sap_lc.errorbar(label='Raw light curve')
corrected_lc.errorbar(ax=ax, label='Corrected light curve');
```

Produces the warning  

LightkurveWarning: The design matrix has low rank (4) compared to the number of columns (114), which suggests that the matrix contains duplicate or correlated columns. This may prevent the regression from succeeding. Consider reducing the dimensionality by calling the `pca()` method.
  LightkurveWarning)

Any insight into this warning would be helpful.

nksaunders:
Fixes #792. (thanks for pointing this out @rebekah9969!)

Super simple PR to move the `DesignMatrix` validation step to the `RegressionCorrector.correct` method. This also allows us to simplify the `PLDCorrector` by removing the `warnings.catch_warnings()` call.

### DIFF
--- a/lightkurve/correctors/regressioncorrector.py
+++ b/lightkurve/correctors/regressioncorrector.py
@@ -209,6 +209,9 @@ class RegressionCorrector(Corrector):
         design_matrix_collection.validate()
         self.design_matrix_collection = design_matrix_collection
 
+        # Validate the design matrix. Emits a warning if the matrix has low rank.
+        self.design_matrix_collection.validate()
+
         if cadence_mask is None:
             self.cadence_mask = np.ones(len(self.lc.time), bool)
         else:
--- a/lightkurve/correctors/regressioncorrector.py
+++ b/lightkurve/correctors/regressioncorrector.py
@@ -206,11 +206,10 @@ class RegressionCorrector(Corrector):
                 design_matrix_collection = SparseDesignMatrixCollection([design_matrix_collection])
             elif isinstance(design_matrix_collection, DesignMatrix):
                 design_matrix_collection = DesignMatrixCollection([design_matrix_collection])
-        design_matrix_collection.validate()
-        self.design_matrix_collection = design_matrix_collection
 
         # Validate the design matrix. Emits a warning if the matrix has low rank.
-        self.design_matrix_collection.validate()
+        design_matrix_collection.validate()
+        self.design_matrix_collection = design_matrix_collection
 
         if cadence_mask is None:
             self.cadence_mask = np.ones(len(self.lc.time), bool)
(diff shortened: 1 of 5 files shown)


lightkurve/correctors/regressioncorrector.py
```diff
@@ -209,6 +209,9 @@ def correct(self, design_matrix_collection, cadence_mask=None, sigma=5,
         design_matrix_collection.validate()
         self.design_matrix_collection = design_matrix_collection
 
+        # Validate the design matrix. Emits a warning if the matrix has low rank.
+        self.design_matrix_collection.validate()
```
barentsen, line 213: This seems identical to what happens on line 209 -- is it necessary?

nksaunders: 🤦🏻‍♂️ whoops, that is unnecessary. Fixed.

barentsen:
I'm concerned that making the entire `DesignMatrix.validate()` step optional isn't quite the right fix, because the `DesignMatrix._validate()` helper method includes the following tests which shouldn't be optional:

```python
        if self.prior_mu is not None:
            if len(self.prior_mu) != self.shape[1]:
                raise ValueError('`prior_mu` must have shape {}'
                                 ''.format(self.shape[1]))
        if self.prior_sigma is not None:
            if len(self.prior_sigma) != self.shape[1]:
                raise ValueError('`prior_sigma` must have shape {}'
                                 ''.format(self.shape[1]))
            if np.any(np.asarray(self.prior_sigma) <= 0):
                raise ValueError('`prior_sigma` values cannot be smaller than '
                                 'or equal to zero')
```

I think we may want to leave the call to the `validate` method as-is, and instead move the low-rank check out of `validate` and into a new helper method which is only called by `correct()` as and when necessary?

nksaunders:
I considered that, but I don't think this change makes `validate` optional in practice because these tests are still performed when the design matrix is used by the corrector. It will still check the dimensions of priors and raise appropriate errors if there is a problem, but I like the flexibility of being able to modify the design matrix (including priors) without vocal warnings before calling `correct`.

In my experience playing around with `RegressionCorrector`, it makes sense to validate the design matrix if it's being used, as opposed to validating each iteration as you construct the full design matrix. But if you think this will cause issues for users, I'd be happy with an intermediate validation step without the low-rank check.

PR #807, opened 2020-08-25
Ensure `download()` passes kwargs to the reader functions

barentsen:
The new LightCurve reader functions contain a number of optional keyword arguments.  For example, the TESS LightCurve reader allows the `flux_column` parameter to be passed to determine which column should be used to populate the `flux` column.

At present the `download()` method of the `SearchResult` class does not pass kwargs through to these reader functions.  This PR changes this, to allow e.g. the following use case:
```python
lc = lk.search_lightcurve("Pi Men", sector=12).download(flux_column='sap_flux')
```

(previously, `flux_column` could not be passed in the example above)

### DIFF
--- a/lightkurve/search.py
+++ b/lightkurve/search.py
@@ -116,7 +116,7 @@ class SearchResult(object):
         """Returns an array of dec values for targets in search"""
         return self.table['s_dec'].data.data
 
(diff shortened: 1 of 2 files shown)


jcsmithhere:
Just in time coding...

PR #808, opened 2020-08-25
Introduce simple `create_transit_mask` method

nksaunders:
This addition is in response to discussion about [this notebook](https://github.com/barentsen/notebooks/issues/1).

### Summary
It would be useful to have a function to create a transit mask with known parameters without the user needing to interact with a `BLSPeriodogram` object. This PR introduces the `create_transit_mask` method for `lightkurve.LightCurve` objects. It takes required arguments for `period`, `duration`, and `transit_time`. 

#### Choices I made that people might have strong opinions about
* I made this a method of `LightCurve`, which makes the most sense to me, though it could theoretically also be a method of `TargetPixelFile`, or just a standalone function in `utils` that takes a time array as well as these parameters.
* I wrapped `BLSPeriodogram` to allow the mask to be generated from a transit model, and to and reduce code repetition.
* At the beginning of the method, I create a copy of the `LightCurve`, and set the `flux` and `flux_err` that were `NaN` to something very small (`1e-10` instead of `0`) so that `to_periodogram` works without removing data points. This ensures that you can create a transit mask for a target even if you haven't removed `NaN`s. I added this because my output mask was a different size than expected and the error was uninformative. The alternative is to throw an exception unless the user calls `remove_nans` before creating their transit mask, but that feels restrictive. 
* `period`, `duration`, and `transit_time` can be passed in as lists or np.arrays so you can easily make multi-planet masks.

### Examples
```python
# single-planet mask
mask = lc.create_transit_mask(period=5.0, duration=0.5, transit_time=0.)

# multi-planet mask
mask = lc.create_transit_mask(period=[5.0, 10.0, 15.0], 
                              duration=[0.5, 0.75, 1.0], 
                              transit_time=[0., 1., 2.])
```

### Unrelated issue
While developing this, I found that every time I tried to create the `BLSPeriodogram` object, I got this error:
```
TypeError: Unsupported operand type(s) for ufunc add: 'Time,Quantity'
```
...but running the cell again worked fine. @barentsen @ojhall94 is this a known issue?

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1920,27 +1920,41 @@ class LightCurve(TimeSeries):
         return ax
 
     def create_transit_mask(self, period, transit_time, duration):
-        """Returns a np.array that is False during transits and True everywhere else.
+        """Returns a boolean array that is ``True`` during transits and
+        ``False`` elsewhere.
 
-        This function is flexible to multi-planet systems, allowing users to pass
-        in transit parameters as arrays or lists.
+        This method supports multi-planet systems by allowing ``period``,
+        ``transit_time``, and ``duration`` to be array-like lists of parameters.
 
         Parameters
         ----------
         period : float, or array-like
-            Period of the transits.
+            Period(s) of the transits.
         duration : float, or array-like
-            Duration of the transits.
-        transit_time : float, or array-like
-            Transit midpoint of the transits.
+            Duration(s) of the transits.
+        transit_time : `~astropy.time.Time`, float, or array-like
+            Transit midpoint(s) of the transits.
 
         Returns
         -------
-        transit_mask : np.array of Bool
-            Mask that removes transits. Mask is True where there are no transits.
-        """
+        transit_mask : np.array of bool
+            Mask that flags transits. Mask is ``True`` where there are transits.
 
-        # Make all parameters arrays
+        Examples
+        --------
+        You can create a transit mask for a single-planet system as follows::
+
+            >>> import lightkurve as lk
+            >>> lc = lk.LightCurve({'time': [1, 2, 3, 4, 5], 'flux': [1, 1, 1, 1, 1]})
+            >>> lc.create_transit_mask(transit_time=2., period=2., duration=0.1)
+            array([False,  True, False,  True, False])
+
+        The method accepts lists of parameters to support multi-planet systems::
+
+            >>> lc.create_transit_mask(transit_time=[2., 3.], period=[2., 10.], duration=[0.1, 0.1])
+            array([False,  True,  True,  True, False])
+        """
+        # Ensure all parameters are 1D-arrays
         period = np.atleast_1d(period)
         duration = np.atleast_1d(duration)
         transit_time = np.atleast_1d(transit_time)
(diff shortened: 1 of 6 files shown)


barentsen:
I was thinking this could be a simple self-contained function without a dependency on BLS, because the logic is straightforward, i.e. I was dreaming about something more simple like this (inspired by `BoxLeastSquares.model()`):

```python
def create_transit_mask(self, period, duration, transit_time):
    transit_mask = np.empty(len(self), dtype=bool)
    transit_mask[:] = False
    hp = period / 2.
    in_transit = np.abs((lc.time - transit_time + hp) % period - hp) < 0.5*duration
    transit_mask[in_transit] = True
    return transit_mask
```

You are of course absolutely right that avoiding code duplication is important, and this duplicates some logic currently available in AstroPy's `BoxLeastSquares.model()`, however I'm worried that the overhead and complexity of implementing this function via `BLSPeriodogram` is making it slow and error-prone.

(My prototype is of course incomplete, it needs input validation, support for multiple planets, and tests.  I love the clever way you support multiple planets in the PR -- we could use recursion in the same way with the snippet above.)

Thoughts?

barentsen:
@nksaunders Are you willing to modify the implementation to directly use the logic above, or is that a silly idea?

nksaunders:
@barentsen sure, I'll give this a go! I agree that we should probably minimize complexity for a simple function like this. Speed and error-aversion are the priority here. Stay tuned!

lightkurve/lightcurve.py
```diff
@@ -1919,6 +1919,66 @@ def plot_river(self, period, epoch_time=None, ax=None, bin_points=1,
...
+            raise ValueError("period, duration, and transit_time must have "
+                             "the same number of values.")
+
+        # Create empty cadence mask
+        transit_mask = np.empty(len(self), dtype=bool)
+        transit_mask[:] = True
+
+        # Create a transit mask for a single planet
+        if isinstance(period, (float, int)):
+            in_transit = self._get_masked_cadences(period, duration, transit_time)
```
barentsen, line 1956: I recommend ensuring that period/duration/transit_time is always an array (e.g. using np.atleast_1d at the start of a function), and treating the single-planet case as a "multi-planet system with number of planets = 1".  Put differently, I recommend getting rid of lines 1954-1959 and always treat period/duration/transit as a list of size 1 or longer.

lightkurve/lightcurve.py
```diff
@@ -1919,6 +1919,66 @@ def plot_river(self, period, epoch_time=None, ax=None, bin_points=1,
...
+        elif isinstance(period, (list, np.ndarray)):
+            # Create mask for first planet
+            in_transit = self._get_masked_cadences(period[0], duration[0],
+                                                   transit_time[0])
+            # Iterate through other planets
+            for i,p in enumerate(period[1:]):
+                n = i+1
+                new_in_transit = self._get_masked_cadences(p, duration[n],
+                                                           transit_time[n])
+                in_transit = np.logical_or(in_transit, new_in_transit)
```
barentsen, line 1968: Lies 1960-1968 seem more complicated than need be.  I think they can be replaced by:
```python
for per, dur, tt in zip(period, duration, transit_time):
    in_transit |= self._get_masked_cadences(period, duration, transit_time)
```

nksaunders:
> I changed `create_transit_mask` to return `True` when in-transit. I did this because `aperture_mask` is `True` for aperture pixels, and `cadence_mask` in RegressionCorrector is `True` for cadences being used, so for consistency I would expect a transit mask to be `True` during transit. What do you think?

@barentsen I agree with this, however if we make this change I think we also need to change the output of `BLSPeriodogram.get_transit_mask`, which is why I returned the inverse originally. All transit masks we create should have the same behavior. 

This is a minor change, and otherwise I think this is ready!

barentsen:
Thank you @nksaunders!

PR #811, opened 2020-08-25
docfix - query_system_object()'s cadence_mask

### LINKED ISSUE #809: query_solar_system_objects documentation / api - cadence_mask
#### Problem description
The issue concerns the documentation for `query_solar_system_objects()` and possibly some API tweaks.

https://docs.lightkurve.org/api/lightkurve.lightcurve.LightCurve.html#lightkurve.lightcurve.LightCurve.query_solar_system_objects

It states that `cadence_mask` must be str or bool.

But `cadence_mask` could also be a Boolean `numpy.ndarray`, flagging the cadences to be searched.

The same consideration is also applicable to `TargetPixelFile`
1. If the intention is to support arbitrary `cadence_mask`, the documentation needs to be updated.
2. If (1) is true, we might want to consider to support any array-like structure rather than `numpy.ndarray`, to be more in parity with the rest of the API.

#### Example

```python
import lightkurve


# For a given lightkurve, check if there are solar system objects
# in the given time frame, e.g., around a possible transit.
# The codes work on lightkurve v1.11.
df_sso = lc.query_solar_system_objects(cadence_mask=np.logical_and(lc.time >= 2014.1, lc.time <= 2014.9))

```

#### Environment

-  lightkurve version: 1.11

### LINKED ISSUE #810: Lightcurve.query_solar_system_objects() documentation - location
#### Problem description

In Lightcurve.query_solar_system_objects(), there is an optional `location` argument. It'd be better if the documentation states the default is `Lightcurve.mission`

The documentation was missing in `v1.11`. Here I am referring to what's in the latest `master` v2.0a2.

-  lightkurve version (e.g. 1.0b6): latest `master`, `2.0a2`

orionlee:
Closes #809 and closes #810

- The initial commits only changes the docstring in `LightCurve.query_system_object()`, assuming we stick with the existing behavior requiring `ndarray`. If agreed, similar changes will be added to `TargetPixelFile`

- I cannot verify the doc output. Somehow in my local build, the generated documentation does not have `query_system_object()` ( nor similarly relatively method `plot_river()`).

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1167,12 +1167,12 @@ class LightCurve(TimeSeries):
         sigma : optional, float
             If `cadence_mask` is set to `"outlier"`, `sigma` will be used to identify
             outliers.
-        location : str
-            Spacecraft location. Options include `'kepler'` and `'tess'`.
+        location : optional, str
+            Spacecraft location. Options include `'kepler'` and `'tess'`. Default: `self.mission`
         cache : optional, bool
             If True will cache the search result in the astropy cache. Set to False
             to request the search again.
-        return_mask: bool
+        return_mask: optional, bool
             If True will return a boolean mask in time alongside the result
 
         Returns
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -4,6 +4,7 @@ import datetime
 import logging
 import warnings
 
+import collections
 from typing import Iterable
 
 import numpy as np
@@ -1159,7 +1160,7 @@ class LightCurve(TimeSeries):
         cadence_mask : str, or boolean ndarray with length of self.time
             mask in time to select which frames or points should be searched for SSOs.
             Default "outliers" will search for SSOs at points that are `sigma` from the mean.
-            "all" will search all cadences. Alternatively, pass a boolean ndarray with values of "True"
+            "all" will search all cadences. Alternatively, pass a boolean array with values of "True"
             for times to search for SSOs.
         radius : optional, float
             Radius in degrees to search for bodies. If None, will search for
@@ -1192,12 +1193,23 @@ class LightCurve(TimeSeries):
             if not hasattr(self, '{}'.format(attr)):
                 raise ValueError('Input does not have a `{}` attribute.'.format(attr))
 
+        sequence_type = None
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", category=DeprecationWarning)
+            # use the deprecated collections.Sequence to be compatible with python 2.7
+            sequence_type = collections.Sequence
+
         # Validate `cadence_mask`
         if isinstance(cadence_mask, str):
             if cadence_mask == 'outliers':
                 cadence_mask = self.remove_outliers(sigma=sigma, return_mask=True)[1]
             elif cadence_mask == 'all':
                 cadence_mask = np.ones(len(self.time)).astype(bool)
+        elif isinstance(cadence_mask, sequence_type):
+            cadence_mask = np.array(cadence_mask)
+        elif isinstance(cadence_mask, (bool)):
+            # for boundary case of a single element tuple, e.g., (True)
+            cadence_mask = np.array([cadence_mask])
         elif not isinstance(cadence_mask, np.ndarray):
             raise ValueError('the `cadence_mask` argument is missing or invalid')
         # Avoid searching times with NaN flux; this is necessary because e.g.
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1203,6 +1203,8 @@ class LightCurve(TimeSeries):
                 cadence_mask = self.remove_outliers(sigma=sigma, return_mask=True)[1]
             elif cadence_mask == 'all':
                 cadence_mask = np.ones(len(self.time)).astype(bool)
+            else:
+                raise ValueError('invalid `cadence_mask` string argument')
         elif isinstance(cadence_mask, collections.abc.Sequence):
             cadence_mask = np.array(cadence_mask)
         elif isinstance(cadence_mask, (bool)):
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1156,7 +1156,7 @@ class LightCurve(TimeSeries):
 
         Parameters
         ----------
-        cadence_mask : str, or boolean ndarray with length of self.time
+        cadence_mask : str, or boolean array with length of self.time
             mask in time to select which frames or points should be searched for SSOs.
             Default "outliers" will search for SSOs at points that are `sigma` from the mean.
             "all" will search all cadences. Alternatively, pass a boolean array with values of "True"
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -834,7 +834,7 @@ class LightCurve(TimeSeries):
                 ntime.append(t)
             ntime = np.asarray(ntime, float)
             in_original = np.in1d(ntime, lc.time.value)
-        
+
         # Fill in time points
         newdata['time'] = Time(ntime, format=lc.time.format, scale=lc.time.scale)
         f = np.zeros(len(ntime))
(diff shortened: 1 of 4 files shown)


lightkurve/lightcurve.py
```diff
@@ -1181,6 +1181,12 @@ def query_solar_system_objects(self, cadence_mask='outliers', radius=None,
             DataFrame object which lists the Solar System objects in frames
             that were identified to contain SSOs.  Returns `None` if no objects
             were found.
+
+        Examples
+        --------
+        Find if there are SSOs affecting the lightcurve for the given time frame:
+
+            >>> df_sso = lc.query_solar_system_objects(cadence_mask=np.logical_and(lc.time >= 2014.1, lc.time <= 2014.9))
```
barentsen, line 1189: By default our continuous testing will try to evaluate this example, which in this case is failing because `lc` is not defined.  We either have to expand the example a little, or otherwise add `# doctest: +SKIP` behind the line to make the testing skip the line:

```suggestion
            >>> df_sso = lc.query_solar_system_objects(cadence_mask=(lc.time.value >= 2014.1) & (lc.time.value <= 2014.9))  # doctest: +SKIP
```

In the suggested change, I also changed `lc.time` into `lc.time.value` which is necessary now that `lc.time` is an AstroPy Time object as of Lightkurve v2.x.

orionlee: fixed.

orionlee:
- now `cadence_mask` accepts array like objects.
- If the changes look good, I will apply similar changes to `TargetPixelFile`

lightkurve/lightcurve.py
```diff
@@ -1181,17 +1181,39 @@ def query_solar_system_objects(self, cadence_mask='outliers', radius=None,
...
         """
         for attr in ['ra', 'dec']:
             if not hasattr(self, '{}'.format(attr)):
                 raise ValueError('Input does not have a `{}` attribute.'.format(attr))
 
+        sequence_type = None
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", category=DeprecationWarning)
+            # use the deprecated collections.Sequence to be compatible with python 2.7
+            sequence_type = collections.Sequence
```
barentsen, line 1204: As of Lightkurve v2.x we no longer support Python 2, so it's not necessary to use a deprecated feature here, i.e. we can use a Python 3 solution here.  Is there a better/non-deprecated way to do this in Python 3?

orionlee: done, with `collections.abc.Sequence`

barentsen:
@orionlee This looks good to me!  Thanks so much!  Do you want to go ahead and make the same fixes in `targetpixelfile.py` before we merge?

orionlee:
Analogous changes are now applied to `TargetPixelFile`.

A few issues: 
1. In `TargetPixelFile` docstring, before the PR, 
  a. the notes says the search cone is 15 spacecraft pixels. 
  b. The `radius` argument says "within 5 pixels of all pixels in the TPF."
  c. I think it is meant to be "within 5 pixels of all pixels in the TPF," and update the doc as such.
2. I think there is a bug in the default `TargetPixelFile` default radius calculation, and will post a separate comment about it.
3. For `cadence_mask` processing, in `LightCurve` case, it will remove the times where flux is `NaN`. But in `TargetPixelFile` case, it does not do so. If it's not intentional, `NaN` filtering should be added to `TargetPixelFile` variant as well.
4. I was tempted to do some refactoring to extract common pre-processing logic used by the two variants, e.g., cadence_mask processing, but decided it's probably better to do it in another PR if agreed.

orionlee:
Possible bug in default radius calculation for TargetPixelFile,

https://github.com/KeplerGO/lightkurve/blob/631d6a39f05104b3a04a53f04228f669c41f7410/lightkurve/targetpixelfile.py#L857

I think it should be:

```
            radius = (2**0.5*(pixel_scale * (np.max(self.shape[1:])) + 5))*u.arcsecond.to(u.deg)
```

Take the example of current a 11x11 pixel TESS TPF as an example,
1. the current logic effectively add 5 arcsecond to the diagonal length of the TPF. The `+ 5` is applied after pixel scaling
2. I think the intent is probably to get the diagonal length of (TPF length + 5 pixel), scaling to degree.

barentsen:
Thanks @orionlee!

>    c. I think it is meant to be "within 5 pixels of all pixels in the TPF," and update the doc as such.

Looking at the implementation, I believe you are correct!

>    I think there is a bug in the default TargetPixelFile default radius calculation, and will post a separate comment about it.

Thank you!

> For cadence_mask processing, in LightCurve case, it will remove the times where flux is NaN (see source). However, in TargetPixelFile case, it does not do so. If it's not intentional, NaN filtering should be added to TargetPixelFile variant as well.

To my knowledge this is not intentional, so this would be another good change!

>    I was tempted to do some refactoring to extract common pre-processing logic used by the two variants, e.g., cadence_mask processing, but decided it's probably better to do it in another PR if agreed.

I am happy to merge the current PR as-is because all the changes look excellent.  I will let you decide if you want to add any more improvements to this PR, or open a new one.  Let me know!

(Thanks++.  This is great stuff!)

barentsen:
> I think the intent is probably to get the diagonal length of (TPF length + 5 pixel), scaling to degree.

Yes, I agree!

orionlee:
> > For cadence_mask processing, in LightCurve case, it will remove the times where flux is NaN (see source). However, in TargetPixelFile case, it does not do so. If it's not intentional, NaN filtering should be added to TargetPixelFile variant as well.
> 
> To my knowledge this is not intentional, so this would be another good change!
> 

The reason I hesitated to change right away is that for `TargetPixelFile` case, the lightcurve used to determine outliers is a default lightcurve generated by the package (using pipeline mask), rather than something the user use / access directly. So I wonder if `NaN` filtering is intentionally left out because the lightcurve may or may not reflect what the user is actually interested in.

orionlee:
I pushed the bugfix for default radius calculation and NaN filtering on `cadence_mask`.

From my point of view, this PR is done, pending any further review changes.

I might do some refactoring as a separate PR, time permitting.

barentsen:
> The reason I hesitated to change right away is that for TargetPixelFile case, the lightcurve used to determine outliers is a default lightcurve generated by the package (using pipeline mask), rather than something the user use / access directly. So I wonder if NaN filtering is intentionally left out because the lightcurve may or may not reflect what the user is actually interested in.

Hmm, right.  In the `TargetPixelFile` case, we would only want to ignore a cadence if *all* the pixels in that cadence are NaN, but such cadences are typically removed by the `quality_bitmask` feature already, so this wouldn't be an important feature, unless we find a good example where it matters!

lightkurve/targetpixelfile.py
```diff
@@ -837,10 +848,18 @@ def query_solar_system_objects(self, cadence_mask='outliers', radius=None,
                     aper = 'all'
                 lc = self.to_lightcurve(aperture_mask=aper)
                 cadence_mask = lc.remove_outliers(sigma=sigma, return_mask=True)[1]
-
-            if cadence_mask == 'all':
+                # Avoid searching times with NaN flux; this is necessary because e.g.
+                # `remove_outliers` includes NaNs in its mask.
+                cadence_mask &= ~np.isnan(lc.flux)
```
barentsen, line 853: Interesting.  Does the mask returned by `remove_outliers` tend to flag cadences where `lc.flux` is NaN?

lightkurve/tests/test_lightcurve.py
```diff
@@ -969,17 +969,26 @@ def test_bin_issue705():
         lc.bin(binsize=15)
 
 
-@pytest.mark.xfail  # As of June 2020 the SkyBot service is returning MySQL errors
```
barentsen, line 972: Thanks for re-activating this test.  I'm happy to see SkyBot works again!

lightkurve/targetpixelfile.py
orionlee: I honestly don't know. The logic and comments are straight from `LightCurve` variant. Maybe Christina will have a better idea?

barentsen: I'm happy to go with it for now, I don't think it will harm!

orionlee:
I'm done . The PR as-is includes the 'NaN' filtering for TargetPixelFile outliers candence_mask case. If it's determined not needed, they can be backed out relatively easilty as it's the last commit.

barentsen:
Merging! Thanks @orionlee! :+1:

orionlee:
@barentsen BTW, do you have any ideas why in my local build, the generated documentation misses some methods, including `query_solar_system_objects()`?

barentsen:
@orionlee  For some reason I have not investigated yet, some of the auto-generated files in docs do not get regenerated automatically unless you explicitly delete them.

The following should fix this:

```bash
rm lightkurve/docs/source/api/lightkurve.*
```

....then run `make html` again in the docs directory, and it should all be there! :crossed_fingers:

PR #814, opened 2020-08-25
Improve the aperture mask plotting style (feedback welcome!)

barentsen:
The style we currently use to visualize aperture masks using `tpf.plot(aperture_mask=mask)` is not very clear because it uses faint pink transparency on top of blue/yellow pixels.  I would like to propose to change it to a less ambiguous red hatching style.  See the example below!

Is anyone against this change?

### Current style:

aper-plot-before

### Proposed new style

aper-plot-after

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -38,6 +38,9 @@ lightkurve.targetpixelfile
 - Modified ``to_lightcurve()`` to default to using ``aperture_mask='threshold'``
   if the 'pipeline' mask is missing or empty, e.g. for TESSCut files. [#812]
 
+- Modified `plot()` to use a more clear hatched style when visualizing the
+  aperture mask on top of pixel data. [#814]
+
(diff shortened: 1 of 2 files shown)


mrtommyb:
👍

astrobel:
I think it's a great idea to change the hatching style! Would it be possible to switch to diagonal lines or crosses for the shading? This style would be hard to look at for people who suffer from trypophobia (fear of clusters of dots/holes).

barentsen:
I think I just become trypophobic by entering the term into Google Images, so I'm happy to change it.

Shall we go with this instead?

aper-2

astrobel:
Looks great!

kcolon:
Yes! Hatched looks good.

barentsen:
Thanks everyone!  I went ahead and merged this change, but happy to revise in the future if someone can think of a better style!

PR #815, opened 2020-08-26
filter collection by TESS sector

### LINKED ISSUE #621: Convenience methods for accessing a lightcurve by sector
#### Problem description
It'd be more productive if the package provides a way access a `LightCurveFile` among in a `LightCurveFileCollection` by sector. It could make some codes more readable.

#### Example

```python
import lightkurve as lk

# get the lightcurves for a given target
lcf_coll = lk.search.search_lightcurvefile(target='TIC232612416', mission='TESS').download_all()

# proposed new method
lc14 = lcf_coll.get_by_sector(14)

lc14.PDCSAP_FLUX.plot()
```

The new method, tentatively `get_by_sector(sector_num)`, returns the `LightCurveFile` of a given sector. 

The behavior when there are multiple `LightCurveFile` instances with the same sector: throw exception, or maybe return the first one with warnings.

#### Environment

-  lightkurve version (e.g. 1.0b6): 1.3.0

orionlee:
Closes #621 

The commits are a first cut for review.

From API users' perspective, he/she can then filter a LightCurve / TargetPixelFile Collection by  the following.

``` python

# Assuming lc_col has 3 LightCurves, in sectors 24, 25 and 26.

> filtered = lc_col[lc_col.sector == 25]
LightCurveCollection of 1 objects: # the one for sector 25

> filtered = lc_col[np.in1d(lc_col.sector, (25, 26))]
LightCurveCollection of 2 objects: # the one for sector 25, 26

# To get the lightcurve of one sector
> lc_col[lc_col.sector == 25][0]

```

Notes:
- The return results are LightCurve / TargetPixelFile Collection, so users can use the usual APIs e.g., stitch, plot, etc.
- Implementation-wise, I have to add TESS-specific `sector` property to `Collection` base class. And Kepler-specific ones `campaign` and `quarter` will be needed as well. It's a bit ugly, but I don't know how we can get around it, unless we have new TESS / Kepler-specific collections (4 additional classes), something like:

```python
class TessLightCurveCollection(LightCurveCollection, TessPropertiesMixin):
   ...
```

### DIFF
--- a/lightkurve/collections.py
+++ b/lightkurve/collections.py
@@ -38,7 +38,7 @@ class Collection(object):
         return len(self.data)
 
     def __getitem__(self, indexOrMask):
-        if (isinstance(indexOrMask, (int, slice))):
+        if (isinstance(indexOrMask, (int, np.integer, slice))):
             return self.data[indexOrMask]
         else:
             # assume indexOrMask is array like, e.g., np.ndarray, collections.abc.Sequence, etc.
--- a/lightkurve/collections.py
+++ b/lightkurve/collections.py
@@ -33,7 +33,7 @@ class Collection(object):
     --------
     Filter a collection by boolean array indexing.
 
-        >>> lcc_filtered = lcc[(lcc.sector == 13) & (lcc.sector <= 19)]  # doctest: +SKIP
+        >>> lcc_filtered = lcc[(lcc.sector >= 13) & (lcc.sector <= 19)]  # doctest: +SKIP
         >>> lc22 = lcc[lcc.sector == 22][0]  # doctest: +SKIP
 
     Filter a collection by integer array indexing to get the object at index 0 and 2.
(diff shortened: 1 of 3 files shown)


lightkurve/collections.py
```diff
@@ -85,6 +98,12 @@ def __repr__(self):
             result += '\n'
         return result
 
+    @property
+    def sector(self):
+        """(TESS-specific) the sectors of the lightcurves / target pixel files
+        """
+        return np.array([lc.sector for lc in self.data])
```
barentsen, line 105: I recommend implementing this as follows:
```suggestion
        return np.array([lc.meta.get('sector') for lc in self.data])
```
....so that we return an array of `None` if `lc.sector` does not exist.

barentsen: Ah, scrap that!  `self.data` may be a `TargetPixelFile` here, which does not have a `meta` dictionary yet, so it's fine the way you have it.

barentsen:
I would be fine with this change!

Indeed, we'll want to support `campaign` and `quarter` as well for consistency.  This isn't ideal because we want the `Collection` class to be a generic container for data from any mission, but Kepler/TESS are the focus of Lightkurve, so it seems fine to have convenience hooks for hose.

I don't think introducing TESS / Kepler-specific collection objects would be a good solution, because `Collection` should be able to hold a mix of both Kepler and TESS (or other) data.  For example, I envisage `LightCurveCollection.stitch()` becoming a robust way to combine a Kepler and a TESS light curve from the same object together, though that hasn't been tested/implemented in earnest at this time.

orionlee:
Ok, so I'll use the commits as the basis, flush out the edge cases, add tests, etc.

The next sets of commits won't come right away, probably in a few days to a week.

barentsen:
> The next sets of commits won't come right away, probably in a few days to a week.

That is absolutely fine of course.  Thanks so much for working on this! :+1:

orionlee:
With the latest commits, 
- filter a collection by TESS sector code complete, including tests and boundary conditions.
- support is also added to filter by int array-like, e.g., `lcc[[0, 3]]` will return a new collection with lightcurves in indices `[0]` and `[3]`. It's not really needed in the use cases, but it makes the API behaves the same way that `np.ndarray` does, reducing potential surprises.

Open:
- I assume Lightkurve 2 requires python 3.6+ (used `f""` to construct some string messages)  If not, please let me know.

Remaining Issues:
- [ ] documentation: add docstring, and include `Collection` to the API doc to be produced.
- [ ] support Kepler's counterpart: `campaign` and `quarter`

orionlee:
@barentsen The latest commits contain all the intended changes, including support for `quarter` and `campaign`. It's ready for review.

Open:
- I added some documentation about indexing in `Collection` class, but `Collection` itself is not included in the apidoc. The docstring is not included in the `LightCurveCollection`or `TargetPixelFileCollection` either. Thoughts?

- I assume Lightkurve 2 requires python 3.6+ (used `f""` to construct some string messages)  If not, please let me know.

barentsen:
Thanks so much @orionlee! I am short on time in the next 24 hours, but will get back to this asap.

(The windows build failure is not your fault because it is also failing on master, caused by a stack overflow, tho I have no idea what's causing it at this time.)

orionlee:
rebased against the latest `master`. Ready to be reviewed / merged.

barentsen (approved):
Thanks @orionlee!  I am very impressed with this PR.  I am going to play with it locally when I have more time later tonight, but I fully expect to merge it before going to bed! :+1:

PR #823, opened 2020-09-02
bugfix - TessLightCurveFile compatibility API

orionlee:
The fix also resolves the pytest failure on windows with python 3.8.5 seen in CI.

The stack overthrow comes from the following test:

```python
@pytest.mark.xfail  # LightCurveFile was removed in Lightkurve v2.x
def test_get_header():
    """Test the basic functionality of ``tpf.get_header()``"""
    lcf = TessLightCurveFile(filename_tess_custom)
    assert lcf.get_header()['CREATOR'] == lcf.get_keyword("CREATOR")
    assert lcf.get_header(ext=2)['EXTNAME'] == "APERTURE"
    # ``tpf.header`` is deprecated
    with pytest.warns(LightkurveWarning, match='deprecated'):
        lcf.header()
```

Without the fix, the string `filename_tess_custom` is treated as the actual lightcurve data  (time) rather than the filename of the data. Somehow it got into an infinite recursion in creating the lightcurve object with python 3.8.5 on windows (Normally it should raise `AttributeError: 'str' object has no attribute 'keys'`, then ignored by `xfail`). 
Python 3.7.3 on windows doesn't result in infinite recursion.

### DIFF
--- a/lightkurve/lightcurvefile.py
+++ b/lightkurve/lightcurvefile.py
@@ -7,4 +7,4 @@ from . import LightCurve, KeplerLightCurve, TessLightCurve
 
 LightCurveFile = LightCurve
 KeplerLightCurveFile = KeplerLightCurve.read
-TessLightCurveFile = TessLightCurve
+TessLightCurveFile = TessLightCurve.read
--- a/lightkurve/tests/test_lightcurve.py
+++ b/lightkurve/tests/test_lightcurve.py
@@ -143,7 +143,7 @@ def test_KeplerLightCurveFile(path, mission):
                          ['hardest', 'hard', 'default', None,
                           1, 100, 2096639])
 def test_TessLightCurveFile(quality_bitmask):
-    lc = TessLightCurveFile.read(TESS_SIM, quality_bitmask=quality_bitmask, flux_column="sap_flux")
+    lc = TessLightCurveFile(TESS_SIM, quality_bitmask=quality_bitmask, flux_column="sap_flux")
     hdu = pyfits.open(TESS_SIM)
 
     assert lc.mission == 'TESS'


orionlee:
BTW, on my local environment (python 3.8.5, Anaconda on Windows), `test_lightcurve.py` also raised some errors `LinAlgError`.

They are not reproduced in CI environment, nor my local python 3.7.3 environment so it's probably something peculiar with the environment. But if someone happens to know the reason I'd appreciate it.

---

A few tests failed with `LinAlgError`
```
base_dev_py38\lib\site-packages\scipy\linalg\basic.py:1221: LinAlgError
================================================ short test summary info ================================================
FAILED test_lightcurve.py::test_lightcurve_scatter - numpy.linalg.LinAlgError: SVD did not converge in Linear Least Square
```

The failing tests are: `test_cdpp`, `test_cdpp_tabby`, and `test_lightcurve_scatter`.

barentsen:
BRILLIANT!!

barentsen:
Again, thanks so much @orionlee! :+1:

Yep, this seems to resolve the issue.  It would have taken me a long time to debug because I don't currently have access to a Windows machine. Your PR saved me hours for sure, even though the bug was quite silly. I'm very grateful!

I'm puzzled by the `LinAlgError` raised by `test_cdpp`, `test_cdpp_tabby`, and `test_lightcurve_scatter`. One thing those methods have in common is that they all call `flatten()`, which in turn calls `scipy.signal.savgol_filter`.  I don't expect that function to raise an `SVD did not converge in Linear Least Squares` error message because I don't think it uses SVD or Least Squares.

To further debug this issue, it would be good to see the full stacktrace so we can see exactly which line in the code raised the error.  If you happen to have that stacktrace, feel free to open it as a new issue!

In meanwhile, I'll go ahead and merge this PR. Thanks so much!!

orionlee:
@barentsen For `LinAlgError`, it's been resolved. Thanks for thinking about it.

---

Details: After I setup a new python 3.8 + Lightkurve dev environment using `pip install -e .`, some of the installed packages have weird issues (I believe it's `numpy` or `scipy`).

I fixed it by re-installing `scipy`, `numpy` using `conda`.

The `LinAlgError` itself was quite weird. I was able to isolate and reproduce the offending codes. 
Somehow with the problematic packages, it still ran okay standalone, but the same code raised `LinAlgError` in a pytest

Isolated offending codes. It ran okay in a standalone python script, but if the code are put as a test function, say, in `test_lightcurve.py` and run in `pytest`, it raised `LinAlgError`.

```
import numpy as np
from scipy import signal

# emulate what's done in test_lightcurve.py::test_cdpp
vals = np.ones(200)
trsig = signal.savgol_filter(x=vals,
                            window_length=101,
                            polyorder=2)
print('savgol: ', trsig)
```

barentsen:
@orionlee Interesting!!  I guess your experiment shows that this is a `scipy` issue rather than a `lightkurve` bug, so I'm happy and relieved :smile:

orionlee:
For my environment, `pip` would install the latest `scipy` (1.5.2); while `conda` would report 1.5.2 is not compatible, and install 1.5.0 instead.
```
Specifications:
  - scipy==1.5.2 -> python[version='>=3.6,<3.7.0a0|>=3.7,<3.8.0a0']

Your python: defaults/win-64::python==3.8.5=he1778fa_0
```

---

I'm tempted to suggest a change in [dev version install instructions](https://docs.lightkurve.org/about/install.html#installing-the-development-version):

```
# For windows, first use conda to install the dependencies
> conda install --file=requirements.txt 
> pip install -e .
```

But it isn't quite simple, as `requirements.txt` isn't guaranteed to be installable by `conda`. E.g., the current `requirements.txt` lists `astropy>=4.1rc1`, not yet available on `conda`.

barentsen:
@orionlee My impression is that `conda` is becoming less important now that all important Python packages are providing "binary wheels" on pypi, i.e. `pip` no longer tries to compile e.g. numpy or scipy but instead installs a binary blob just like conda does.  (At least this is the case on Linux, I haven't checked if our dependencies all have binary wheels for Windows available.)

Because `pip install -e .` effectively calls `pip install -r requirements.txt`, and because the latter has improved a lot, I'd be inclined to start recommend using pip over conda... but again, that's only based on my positive experience with pip on Linux!

barentsen:
> pip install -e .

We should change this to `python -m pip install -e .` by the way, cf. https://snarky.ca/why-you-should-use-python-m-pip/

orionlee:
1. In my specific case, I suspect `pip` installed a binary wheel that doesn't completely works but `conda` identified and selected a different version. But I heard `pip`'s upcoming release will improve its conflict detection so we can probably keep our fingers crossed.

2. For changing instructions to `python -m pip install -e .` : if so, do we need to spell out that people need to substitute `python` with the actual python interpreter to be used, e.g, `python3`?

barentsen:
> do we need to spell out that people need to substitute python with the actual python interpreter to be used, e.g, python3?

Yes, I like that!

While I think this specifically addresses the common issue on OSX where `python` often refers to system Python, I think it is fine to change `python` into `python3` because e.g. Anaconda on Linux also provides a python3 executable by default.

Let's do it! :+1:

PR #672, opened 2020-01-28
[WIP] Tess CBV corrector

jcsmithhere:
Greetings!

This is a first rough version of a CBVCorrector for LightKurve! This is definitely a **WORK IN PROGRESS** but enough functionality is implemented to demonstrate some basic methods, and get some feedback. 

The main tools created so far are:

1. **get_cbvs** -- retrieves the requested Kepler/K2/TESS Cotrending Basis Vectors from MAST and creates a CotrendingBasisVectors object.

2. **CotrendingBasisVectors** -- Stores Cotrending Basis Vectors (CBVs) for the Kepler/K2/TESS missions.
 
3. **CBVCorrector** -- class to perform a correction on a LightCurve object utilizing the CBVs via the RegressionCorrector class. Right now it simply does Least-Squares correction; that will change!

Attached is a PDF of a Jupyter notebook demonstrating functionality. (Apparently you can't attach Jupyter notebooks to pull requests so converted to PDF)
[CBVCorrector Demonstrator.pdf](https://github.com/KeplerGO/lightkurve/files/4120250/CBVCorrector.Demonstrator.pdf)

The more I look at this the more I realize I am not creating a "CBV Corrector" but a wrapper for RegressionCorrector that happens to be able to load in the CBVs, along with other design matrices. In time, the class will allow for sophisticated optimization techniques. So, perhaps it should not be called "CBVCorrector" but a more generic name, perhaps "OptimizationCorrector" or something to emphasize the optimization techniques to be used, not just that it uses the CBVs. ...Or, just move it all into RegressionCorrector. I welcome feedback on this!

There is a lot more that can be done:

1. Finish code documentation
2. Write Unit Tests
3. Develop Goodness Metrics to assess performance
4. Allow user to specify signal bands they care about for creating goodness metrics
5. Use goodness metrics as a prior regularizing term
6. Add other prior options
7. Add L1 Norm (Lasso) and L2 Norm (Ridge Regression) regularizers
8. Add methods to band split the light curves and perform correction in specific bands
9. ...And even more!

This won't all be completed anytime soon. I will begin by trying to develop some goodness metrics to assess over- and under-fitting, based on what is calculated in PDC.

PR #826, opened 2020-09-02
CBVCorrector Part I: Retrieving the CBVs

jcsmithhere:
Here is the first of two PRs for the CBVCorrector. This is the code to retrieve the Kepler/K2/TESS CBVs from MAST and convert them into a DesignMatrix for use with cotrending against light curves. A new tutorial has been created called '04-how-to-use-cbvs.ipynb' which should explain the functionality. I reference a second, not yet created, tutorial to use the cbvcorrector class, which happens in the second PR. I am not certain if I properly set up the links for the new tutorial.

As is, the tools herein can be used on their own but it's best we also include the cbvcorrector class before releasing to the public.

This is the actual PR after completing WIP #672, which will be closed.


barentsen:
Thans @jcsmithhere!  Reviewing this is at the top of my list today!

Just noting here that the Windows unit test is failing due to an unrelated issue which has since been resolved (#823), so the tests look good.

lightkurve/correctors/cbvcorrector.py
```diff
@@ -1,258 +1,946 @@
...
+
+        # Here we instantiate the actual object
+        return self(data=dataTbl, time=cbvTime, **kwargs)
 
-        if self.lc.mission == 'Kepler':
-            quarter = 'q{:02}'.format(self.lc.quarter)
+
+    @property
+    def mission(self):
+        return self.meta['MISSION']
```
barentsen, line 480: I recommend using `meta.get(key, default=None)`, which will return `default` if `key` is missing from the dictionary:
```suggestion
        return self.meta.get('MISSION')
```

lightkurve/correctors/cbvcorrector.py
```diff
@@ -1,258 +1,946 @@
...
+        # @property methods do not error
+        if data is not None:
+            if not data.meta.__contains__('MISSION'):
+                data.meta['MISSION'] = None
+            if not data.meta.__contains__('CBV_TYPE'):
+                data.meta['CBV_TYPE'] = None
+            if not data.meta.__contains__('MODULE'):
+                data.meta['MODULE'] = None
+            if not data.meta.__contains__('OUTPUT'):
+                data.meta['OUTPUT'] = None
```
barentsen, line 404: I recommend removing lines 393-404 and instead change the property methods such that they return `self.meta.get(key, default=None)` instead of `self.meta[key]`.  The former returns `None` and the latter raises an exception if `key` is missing.  It means we can have less code :smiley:

lightkurve/correctors/cbvcorrector.py
```diff
@@ -1,258 +1,946 @@
...
-        # keeps the ones for which 'href' ends with 'fits'
-        # this might slow things down in case the user wants to fit 1e3 stars
-        soup = BeautifulSoup(requests.get(self.cbv_base_url).text, 'html.parser')
-        cbv_files = [fn['href'] for fn in soup.find_all('a') if fn['href'].endswith('fits')]
+        assert module > 1 and module < 85, 'Invalid module number'
+        assert output > 0 and output < 5, 'Invalid output number'
+
+        # Get the mission: Kepler or K2
+        # Sadly, the HDU does not explicitly say if this is Kepler or K2 CBVs.
+        if HDU['PRIMARY'].header.__contains__('QUARTER'):
```
barentsen, line 438: The `__contains__` method implements Python's `in` operator, so this is exactly the same but more Pythonic:
```suggestion
        if 'QUARTER' in HDU['PRIMARY'].header:
```

lightkurve/correctors/cbvcorrector.py
```diff
@@ -1,258 +1,946 @@
...
+            mission = 'Kepler'
+        elif HDU['PRIMARY'].header.__contains__('CAMPAIGN'):
```
barentsen, line 440:
```suggestion
        elif 'CAMPAIGN' in HDU['PRIMARY'].header:
```

lightkurve/correctors/cbvcorrector.py
```diff
@@ -1,258 +1,946 @@
...
+        
+    @property
+    def output(self):
+        return self.meta['OUTPUT']
+
+    @output.setter
+    def output(self, output):
+        self.meta['OUTPUT'] = output
+        
+    def __repr__(self):
```
barentsen, line 538: Time-permitting, I recommend implementing the custom repr by overriding `_base_repr_()` method instead.  This method is called by the repr methods of the `TimeSeries` parent class. You could copy/paste `LightCurve._base_repr_()` to see how it can be overriden.

barentsen (changes requested):
Thanks @jcsmithhere!  This is great stuff! :+1:

Thank you for including the tutorial notebook, which provided an excellent and helpful overview!

I expect we can merge this PR very quickly!  I left a few nitpicky comments in the code (see below), but most of them will be trivial to resolve.

In addition, I have a few suggestions related to the API for your consideration:

* I recommend using function names `download_kepler_cbvs()` and `download_tess_cbvs()` instead of `get_kepler_cbvs()` and `get_tess_cbvs()`.  I'm suggesting this because it makes it more explicit that the function will require internet access.  It also makes the name more consistent with the `download()` method we provide in the search module.

* I expected `get_tess_cbvs()` to provide a way to specify the cadence, e.g. `cadence='20'` or `cadence='ffi'` as a parameter. I'm guessing you didn't implement this yet because 2-min CBVs are the only ones available for TESS right now.  That's fine, of course!  I just wanted to check: do you think we'll easily be able to add e.g. a `cadence` keyword to these functions in the future?  (Did Kepler have separate short cadence CBVs, or did we always interpolate?)

* In the notebook, I notice that `cbvs_short.align(lc_short)` does not align the cadences and instead returns a message saying `"There are cadences in the light curve that are not in the CBVs. NO SYNCHRONIZATION OCCURED"`.  I think this should raise an exception rather than a warning because the alignment failed.  (It does work when `trim_lc=True` is passed.)  I think this API can be made more user-friendly however, see next comment...

* I wonder if we want to make `cbvs.align(lc)` more similar to `cbs.interpolate(lc)` such that both methods always return a new `CotrendingBasisVectors` object that contains the exact same cadences as `lc`.  The only difference between `align` and `interpolate` could be that the latter interpolates missing cadences, whereas the former would populate them with NaNs.  I think this might make both functions easier to explain and use. One could easily swap out `align` for `interpolate` and vice versa in this case. (Perhaps there could even be a single method that does both, i.e. `align(interpolate=True/False)`.)
...but maybe you have a good reason for the different behavior of both functions?

Finally, I have two minor suggestions for the tutorial notebook:

* Consider adding one sentence explaining when someone might prefer single-scale vs multi-scale, i.e. why are both options available.  I keep forgetting =)

* In the notebook, you could consider calling `cbv_designmatrix.plot()` for a fun visual representation of the design matrix.  Optional, of course!

That's it for now.  Let me know if you have any questions!  I can't wait to merge!! :+1:

jcsmithhere:
Hello @barentsen, Thanks for your comments! I have no issues almost everything you mention and should be able to make your recommended changes pretty quickly. Regarding combining the `align` and `interpolate` functions. My logic here is `align` is a harmless function because it does not create any new CBV cadence values, it simply strips out unrequested cadences. But `interpolate` is potentially very dangerous. Interpolating a CBV can cause artifacts. We could use an `interpolate` flag for the `align` method and just have a very strong warning in the tutorial for the users. I'll think about it.

jcsmithhere:
There is on other topic which I thought you might comment on. For `cbv_type` I use camelCase: `('SingleScale', 'MultiScale', 'Spike')`. I know this is not PEP8 and thought you would mention so. I use CamelCase because that is the way I have always expressed the naming in PDC. I thought about changing them to `('single_scale', 'multi_scale', 'spike')` but since you didn't mention anything perhaps I'll just keep them the way it is.

lightkurve/correctors/cbvcorrector.py
```diff
@@ -1,258 +1,946 @@
...
+            if not data.meta.__contains__('CCD'):
+                data.meta['CCD'] = None
+
+
+        # Initialize attributes common to all CotrendingBasisVector classes
+        super(TessCotrendingBasisVectors, self).__init__(data=data,
+                time=time, **kwargs)
+
+    @classmethod
+    def from_HDU(self, HDU=None, cbv_type=None, band=None, **kwargs):
```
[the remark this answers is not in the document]
jcsmithhere: No problem!

lightkurve/correctors/cbvcorrector.py
```diff
@@ -1,258 +1,946 @@
...
-                ax.set_title('Kepler CBVs (Module : {}, Output : {}, Quarter : {})'
-                             ''.format(module, output, self.lc.quarter))
-            elif self.lc.mission == 'K2':
-                ax.set_title('K2 CBVs (Module : {}, Output : {}, Campaign : {})'
-                             ''.format(module, output, self.lc.campaign))
-            ax.grid(':', alpha=0.3)
-            ax.legend()
-        return ax
+def get_tess_cbvs (sector=None, camera=None,
+        CCD=None, cbv_type='SingleScale', band=None):
```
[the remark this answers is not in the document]
jcsmithhere: changed.

lightkurve/correctors/cbvcorrector.py
```diff
@@ -1,258 +1,946 @@
...
+                if len(cbv_matrix) == 0:
+                    cbv_matrix =  np.array(self['VECTOR_{}'.format(idx)])[...,None]
+                else:
+                    cbv_matrix = np.hstack((cbv_matrix, 
+                        np.array(self['VECTOR_{}'.format(idx)])[...,None]))
+                cbv_names.append('VECTOR_{}'.format(idx))
+
+        return DesignMatrix(cbv_matrix, columns=cbv_names, name=name)
+
+    def plot_cbvs(self, cbv_indices='ALL', ax=None):
```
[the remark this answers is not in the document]
jcsmithhere: Changed.

lightkurve/correctors/cbvcorrector.py
```diff
@@ -1,258 +1,946 @@
...
+
+from ..lightcurve import LightCurve
+from ..utils import channel_to_module_output, validate_method
 
 log = logging.getLogger(__name__)
 
-__all__ = ['KeplerCBVCorrector']
+# For Kepler/K2/TESS max number of stored CBVs has always been 16
+# Nevertheless, set to a ridiculously high number to future-proof the code
+MAX_NUMBER_CBVS = 99
```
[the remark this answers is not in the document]
jcsmithhere: `MAX_NUMBER_CBVS` is vestigial from an earlier version of the code. `MAX_NUMBER_BANDS` is still used by the cbvcorrector class but I think I know how to not require it.

lightkurve/correctors/cbvcorrector.py
jcsmithhere: changed.

lightkurve/correctors/cbvcorrector.py
jcsmithhere: Removed.

lightkurve/correctors/cbvcorrector.py
jcsmithhere: Changed.

lightkurve/correctors/cbvcorrector.py
jcsmithhere: Changed.

barentsen:
@jcsmithhere Thanks for the quick edits.  Ping me when you think you are finished and I will have another careful look!

Here are quick replies on your recent two messages:

> There is on other topic which I thought you might comment on. For cbv_type I use camelCase: ('SingleScale', 'MultiScale', 'Spike')

I didn't comment on it because these are parameter values rather than variable/function names, so we have the option to treat them in a case-insensitive way. Lightkurve already ignores case for many string parameters across the API. I'm fine with using CamelCase as the default for now because it matches the PDC data products, and we can still implement case-insensitive behavior in the future if we think it benefits user-friendliness.  (I guess my review focused on API decisions that will be expensive to modify in the future, and this isn't one.)

> My logic here is align is a harmless function because it does not create any new CBV cadence values, it simply strips out unrequested cadences.

A design choice we've tried to pursue across the API, to the extent possible, is to consistently have `object.verb()`-style methods return a modified copy of `object`.  This makes the API fun to use because (i) users can guess and remember return values more easily, and (ii) methods can be chained, e.g. `cbvs.align(lc).plot()`.

Having `cbvs.align()` return a light curve in addition to a modified CBV object makes it a bit less intuitive to use.  We can certainly do it this way if it significantly reduces complexity elsewhere in the API, but I expect that `CBVCorrector.correct()` may be a more natural place to take care of trimming light curves if we want that as an option?

jcsmithhere:
Regarding the align/interpolate method question. I think it is best we keep them as separate functions. `align` uses the cadence numbers but `interpolate` uses the cadences times. `align` is safe because no manipulation occurs, we are simply trimming cadences to the light curve. `interpolate` generates new values at new cadences.

But, I see your point about having `align` return only a new CBV object. So, I will eliminate the `trim_lc` option and simply returns NaNs where no CBV values exist for the light curve.

barentsen:
@jcsmithhere I'm fine with that! 👍

jcsmithhere:
Hi @barentsen, I made all the requested changes (except _base_repr_()) so I think the PR is good to go!

`align` is now slow due to it using `add_row`. There is no `add_rows` (plural) so I have to add a bunch of rows to the table in series. There's probably a faster way to do this but everything appears to work so I don't want to wait any longer. If I have time later I can speed it up.

barentsen (approved):
Thanks @jcsmithhere. This looks good to merge. :+1:  We can continue with the part 2!

(I pushed an extra commit to added an entry to `CHANGES.rst` to record this PR and alter the white space in a few docstrings to make sure the online docs will render correctly.)

PR #827, opened 2020-09-03
Remove NaNs of centroids

ckm3:
For the default situation, the centroids are coming from raw fits file, but some of them may contain NaNs (e.g., ktwo211083408-c04_lc). It would be better to remove NaNs of centroids before the sff correction.

### DIFF
--- a/lightkurve/correctors/sffcorrector.py
+++ b/lightkurve/correctors/sffcorrector.py
@@ -125,7 +125,7 @@ class SFFCorrector(RegressionCorrector):
                           LightkurveWarning)
 
         if centroid_col is None:
-            self.lc = self.lc[~np.isnan(self.lc.centroid_col)]
+            self.lc = self.lc.remove_nans(column='centroid_col')
             centroid_col = self.lc.centroid_col
         if centroid_row is None:
             self.lc = self.lc[~np.isnan(self.lc.centroid_row)]
--- a/lightkurve/correctors/sffcorrector.py
+++ b/lightkurve/correctors/sffcorrector.py
@@ -125,8 +125,10 @@ class SFFCorrector(RegressionCorrector):
                           LightkurveWarning)
 
         if centroid_col is None:
+            self.lc = self.lc[~np.isnan(self.lc.centroid_col)]
             centroid_col = self.lc.centroid_col
         if centroid_row is None:
+            self.lc = self.lc[~np.isnan(self.lc.centroid_row)]
             centroid_row = self.lc.centroid_row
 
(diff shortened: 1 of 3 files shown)


barentsen:
@ckm3 Thank you for opening this issue! :+1:

You are correct! Although I have not encountered this before, it looks like this target contains many good-quality cadences with NaN centroids:
```python
In [1]: import lightkurve as lk 
   ...: import numpy as np 
   ...: lc = lk.search_lightcurve("EPIC 211083408").download() 
   ...: np.isnan(lc.centroid_col).sum()                                         
Out[1]: 59
```

As a result, SFF correction fails:
```python
In [2]: lc.remove_nans().to_corrector("sff").correct()                          
<ipython-input-2-deee05a49869> in <module>
----> 1 lc.remove_nans().to_corrector("sff").correct()

~/dev/lightkurve/lightkurve/correctors/sffcorrector.py in correct(self, centroid_col, centroid_row, windows, bins, timescale, breakindex, degree, restore_trend, additional_design_matrix, polyorder, sparse, **kwargs)
    131 
    132         if np.any([~np.isfinite(centroid_row), ~np.isfinite(centroid_col)]):
--> 133             raise ValueError('Centroids contain NaN values.')
    134 
    135         self.window_points = _get_window_points(centroid_col, centroid_row,

ValueError: Centroids contain NaN values.
```

lightkurve/correctors/sffcorrector.py
```diff
@@ -125,8 +125,10 @@ def correct(self, centroid_col=None, centroid_row=None, windows=20, bins=5,
                           LightkurveWarning)
 
         if centroid_col is None:
+            self.lc = self.lc[~np.isnan(self.lc.centroid_col)]
```
barentsen, line 128: Just to make this slightly more elegant, I propose changing it into this:
```suggestion
            self.lc = self.lc.remove_nans(column='centroid_col')
```

lightkurve/correctors/sffcorrector.py
```diff
@@ -125,8 +125,10 @@ def correct(self, centroid_col=None, centroid_row=None, windows=20, bins=5,
...
             centroid_col = self.lc.centroid_col
         if centroid_row is None:
+            self.lc = self.lc[~np.isnan(self.lc.centroid_row)]
```
barentsen, line 131:
```suggestion
            self.lc = self.lc.remove_nans(column='centroid_row')
```

barentsen:
@ckm3 Thanks again for opening this issue!  I took the liberty to modify your PR slightly to add a regression test and update the change log.  It looks ready to merge!

This fix will be part of the next v2.0 release of Lightkurve.

Thanks again! :+1:

PR #829, opened 2020-09-03
Fix memory seepage in TargetPixelFile.__getitem__

dhomeier:
Proposed fix for https://github.com/KeplerGO/lightkurve/issues/330#issuecomment-686618489

`TargetPixelFile.__getitem__` was essentially creating a full deepcopy of its associated HDUList, but on replacing the `data` part with the indexed range, the rest of the HDU's data seemed to keep lurking around in the murky regions of the FITS data model.
Trying to solve the issue by instead creating a new `BinTableHDU` from just the selected index range.

Notes:
1. Perhaps the test should also check the data in `tpf.flux`; do we have a test file that is not all zeros?

2. A single index still returns `tpf[i].shape == (i, N, M)`, rather than `(N, M)`, which might be confusing, but perhaps this is intended.

3. Since pip also gives `astropy >= 3.1` as dependency, it looks like this could be fairly easily backported to 1.11.x. Waiting on Milestone assignment for the changelog.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -47,6 +47,9 @@ lightkurve.targetpixelfile
 - Modified ``_parse_aperture_mask()`` to ensure that masks composed of integer
   or floats are always converted to booleans. [#694]
 
+- Fixed a bug in ``TargetPixelFile.__getitem__()`` which caused a substantial
+  memory leak on indexing or slicing a tpf. [#829]
+
 lightkurve.search
 ^^^^^^^^^^^^^^^^^
 
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -9,7 +9,7 @@ import logging
 import collections
 
 from astropy.io import fits
-from astropy.io.fits import Undefined
+from astropy.io.fits import Undefined, BinTableHDU
 from astropy.nddata import Cutout2D
 from astropy.table import Table
 from astropy.wcs import WCS
@@ -94,8 +94,8 @@ class TargetPixelFile(object):
             warnings.simplefilter('ignore', UserWarning)
             # AstroPy added `HDUList.copy()` in v3.1, but we don't want to make
             # v3.1 a minimum requirement yet, so we copy in a funny way.
-            copy = fits.HDUList([myhdu.copy() for myhdu in self.hdu])
-            copy[1].data = copy[1].data[selected_idx]
+            copy = self.hdu.copy()
+            copy[1] = BinTableHDU(data=self.hdu[1].data[selected_idx], header=self.hdu[1].header)
         return self.__class__(copy, quality_bitmask=self.quality_bitmask, targetid=self.targetid)
 
     def __len__(self):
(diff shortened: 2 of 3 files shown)


barentsen:
Wow!  Thanks so much @dhomeier!  TESS will soon start releasing 20-second cadence Target Pixel Files, which will take up a large amount of memory, so this is a very timely and important improvement!

To answer your questions:

> 1. Perhaps the test should also check the data in tpf.flux; do we have a test file that is not all zeros?

We recently added three synthetically-generated TPFs in `tests/data/synthetic`, which are being used in `tests/test_synthetic_data.py` to e.g. see if we can recover a sinusoid signal after applying systematics removal.  They may also be useful here, though I'd equally be happy to merge this PR with the tests you already added!


 > 2. A single index still returns tpf[i].shape == (i, N, M), rather than (N, M), which might be confusing, but perhaps this is intended.

Interesting. I don't recall this being questioned or discussed before. I guess this is inconsistent with `numpy` so I'd be fine with changing in.  It surprises me because e.g. `self.hdu[1].data[0]`` does appear to return a two-dimensional array, so it's not readily obvious to me how we'd fix this!

> 3. Since pip also gives astropy >= 3.1 as dependency, it looks like this could be fairly easily backported to 1.11.x. Waiting on Milestone assignment for the changelog.

For the sake of reducing work and keeping v1.x stable I'm only back-porting critical bugs.  I'm not sure this bug is critical because few people have complained about it, so for now I'm happy to just put it in master!

Thanks so much!  This is awesome!  I'm happy to merge this PR as-is, but let me know if you still want to change anything!

dhomeier:
> Interesting. I don't recall this being questioned or discussed before. I guess this is inconsistent with `numpy` so I'd be fine with changing it. It surprises me because e.g. `self.hdu[1].data[0]`` does appear to return a two-dimensional array, so it's not readily obvious to me how we'd fix this!
> 
I stumbled over it when comparing the plot function to explicitly plotting `plt.imshow(tpf[i].flux.value)`, so it could be confusing, but I guess it is not normally supposed to be called that way. And one can always use `plt.imshow(tpf.flux[i].value)` instead.
I think the dimension is brought back in with `self.hdu[1].data['FLUX'][self.quality_mask]`, so this might be fixed by checking if `self.quality_mask` has dimension 1, but this would also need to somehow remember how the slice was originally created so that
```
tpf[0].flux.shape == (N, M)
tpf[0:1].flux.shape == (1, N, M)
```

dhomeier:
> > 1. Perhaps the test should also check the data in tpf.flux; do we have a test file that is not all zeros?
> 
> We recently added three synthetically-generated TPFs in `tests/data/synthetic`, which are being used in `tests/test_synthetic_data.py` to e.g. see if we can recover a sinusoid signal after applying systematics removal. They may also be useful here, though I'd equally be happy to merge this PR with the tests you already added!
> 
That would work, but `test-tpf-non-zero-center.fits` also seems to have non-constant flux, so I've now integrated the tests into the older `test_tpf_slicing`.

barentsen:
> given all the places self.quality_mask is applied I think there is too much risk of breaking something

Agreed!  If we think it is worth changing this behavior (which is not obvious yet), we'll want to do it in a separate PR.

barentsen (approved):
Thanks so much @dhomeier.  This is going to make a lot of users' lives easier. 

I am tempted to backport it to v1.11, but may give it a few days to see if we discover anything else.

Issue #740, opened 2020-05-21
Speed up the tests using `pytest-vcr` and `vcrpy`


barentsen:
## Problem
Lightkurve has a lot of unit tests which access the data archive at MAST.  This causes a number of problems:
* our tests are slow;
* many of our tests require internet access;
* tests often fail due to intermittent issues at MAST rather than Lightkurve issues;
* we add an unnecessary load on MAST's services.

## Current workaround
We have flagged the tests which require internet access with the `@pytest.mark.remote_data` decorator.  This flag enables us to skip these tests by default unless we explicitly run them using the `pytest --remote-data` flag.

## Better solution?
A better solution may be to use the `vcrpy` and `pytest-vcr` packages.  The `vcrpy` package appears to allow HTTP requests and responses to be captured and stored in small files that can be checked into the git repository.  In combination with `pytest-vcr`, this should enable us to run most remote_data tests offline.  (Though we'd still want to refresh the vcr files once in a while to check that the MAST API hasn't changed.)

This may be a fun project for a beginner to investigate.  I'd particularly encourage someone to open a PR demonstrating how `vcrpy` may be used to speed up a single test, which should help us understand whether this is a good solution!

Please leave a message below if you start working on this!  Anyone is welcome to try!

barentsen:
Documentation about our current test infrastructure is available here: https://docs.lightkurve.org/about/testing.html

texadactyl:
When running tests on multi-core and/or multi-cpu, are you taking advantage of pytest multithreading of independent tests?
E.g. py.test -n 8 # for a quiet machine with 4 cores rated at 8 threads

texadactyl:
Almost forgot:  Multithreading requires installing pytest-xdist.

texadactyl:
E.g. 
```
from os import chdir
from pytest import main
import lightkurve as lk
path = lk.__path__[0]
chdir(path)
main(['--remote-data', '-rf', '--durations=10', '-n', '2']) # running 2 threads

```

texadactyl:
A simpler approach might be to prep downloaded data in the cache once.  Subsequently, whenever pytest is run, --remote-data tests will make use of the cache.

E.g. the above main program could be enhanced as indicated by the comment(#).

texadactyl:
I see that your URLs are spread out and especially dense in test_search.py:

- ./lightkurve/prf/tests/test_prfmodel.py
- ./lightkurve/prf/tests/test_tpfmodel.py
[...]

PR #833, opened 2020-09-09
to_lightcurve(): implement 'default' aperture_mask

### LINKED ISSUE #791: `tpf.to_lightcurve()` always returns an empty light curve for TESScut data
I have been playing around with 
"Performing Pixel Level Decorrelation with the PLDCorrector in lightkurve" tutorial in colab and have been applying it to TESS data. I have noticed several potential issues - I apologize in advance if these are not issue or bugs, but simple mistakes made on my end. 

I first tried to run the task via creating a cut out from FFI data and then passing  through to PLDcorrector, but this brought up several issues. 

The code I used is the following, 

```
target = 'KIC 8462852' 
tpf = lk.search_tesscut(target, sector=15).download(cutout_size=(10, 10))
pld = lk.PLDCorrector(tpf)
corrected_lc = pld.correct()
raw_lc = tpf.to_lightcurve()
ax = raw_lc.normalize().scatter(c='r', alpha=0.5, label='Raw Flux')
corrected_lc.normalize().remove_outliers().scatter(ax=ax, label='PLD-corrected Flux');
```
This brings up a warning 
```
Warning: aperture mask contains zero pixels.
```

if I run
```
%debug
```
I get 

```
     13         raise NotImplementedError("This is an abstract base class.")
     14 
     15     def diagnose(self):
     16         """Returns plots which elucidate the most recent call to `correct()`."""
---> 17         raise NotImplementedError("This is an abstract base class.")
```
The light curve produced does not look correct, with no raw flux.
Now if I switch to using a tpf from MAST instead so...

```
tpf = lk.search_targetpixelfile(target, sector=15).download()
```

and run through in the same way. I get no warning when creating the light curves, however when I try to run 

```
pld.diagnose();
```

I get the following output 

```
NotImplementedError                       Traceback (most recent call last)
<ipython-input-8-48eda9cd8b1c> in <module>
----> 1 pld.diagnose();

~/Desktop/TESS_LK/lightkurve/docs/source/lightkurve/lightkurve/correctors/corrector.py in diagnose(self)
     15     def diagnose(self):
     16         """Returns plots which elucidate the most recent call to `correct()`."""
---> 17         raise NotImplementedError("This is an abstract base class.")

NotImplementedError: This is an abstract base class.
```

orionlee:
- refinement over PR #812 , which fixes #791. New behavior:
  when `aperture_mask == 'default'` : use `pipeline`, with `threshold` as fallback
- see https://github.com/KeplerGO/lightkurve/pull/812#discussion_r483087154

---
TODOs:
- [x] Update docstring and CHANGES.rst


orionlee:
In updating the `test_targetpixelfile.py`, I found a few things that I'd like to discuss:
1. `test_missing_pipeline_mask()`: if the intent is to strictly test cases a TPF with no pipeline mask, 
the updated `test_to_lightcurve()` in this PR covers the case and the test can be removed. If the intent is to also explicitly test with data from TESSCut, the test should be marked as `remote_data`.
2. `test_load_bad_file()` : it loads testing data remotely but it does not appear to be the purpose of the test. It'd make test run faster if we store the test data locally instead.
3. `test_wcs_tabby()`: same issue as `test_load_bad_file()`, but if it's indeed desirable to test with remote data, then the test should be marked as `remote_data`.
4. `test_tess_simulation()`: same as `test_load_bad_file()`

barentsen:
Thanks so much for this PR @orionlee.  This looks good to merge!

Here are my replies to your questions: 

> 1. test_missing_pipeline_mask(): if the intent is to strictly test cases a TPF with no pipeline mask,
the updated test_to_lightcurve() in this PR covers the case and the test can be removed. If the intent is to also explicitly test with data from TESSCut, the test should be marked as remote_data.

I think it's good to explicitly include a test with data from TESSCut, just to regularly check that the service hasn't changed in a way that breaks our aperture mask behavior.

You are absolutely right: it misses the `remote_data` decorator! :fearful:   I pushed an extra commit a few moments ago to add it.

> 2. test_load_bad_file() : it loads testing data remotely but it does not appear to be the purpose of the test. It'd make test run faster if we store the test data locally instead.

We do use the same file (`TABBY_Q8`) in a few places in both `test_lightcurve.py` and `test_correctors.py` as well.  Because Lightkurve opens FITS files using `astropy.io.fits.open()` behind the scenes, the file should only be downloaded once and be cached using Astropy's caching mechanism.  I see this behavior on Linux:

```
>>> from astropy.io import fits
>>> url = "https://archive.stsci.edu/missions/kepler/target_pixel_files/0084/008462852/kplr008462852-2011073133259_lpd-targ.fits.gz"
>>> f = fits.open(url)
>>> f.filename()
/home/gb/.astropy/cache/download/url/c0163e868cd2f54c67125d4ecd3aeb9f/contents
```

I do agree that it would be good to use small local files for as many tests as possible.  On the other hand, I do want to keep both the repo and the source distribution small, so it's a trade-off.  Some day I'm hoping it can be streamlined via #740.

> 3. test_wcs_tabby(): same issue as test_load_bad_file(), but if it's indeed desirable to test with remote data, then the test should be marked as remote_data.

The same file (TABBY_TPF) is also used in `test_lightcurve.py`, `test_interact.py`, and `test_correctors.py`, so it should be in cache.

You're right, it was missing the `remote_data` decorator!  I also added this one in the extra commit I pushed.

> 4. test_tess_simulation(): same as test_load_bad_file()

I believe we added that test before launch, i.e. before real TESS data was available.  It could likely be removed, but it does seems to include a regression test for #236 which we should make sure is moved elsewhere.  I'll leave changing it to the day we do a more comprehensive review and stream-lining of the tests.

barentsen (approved):

orionlee:
> Because Lightkurve opens FITS files using astropy.io.fits.open() behind the scenes, the file should only be downloaded once and be cached using Astropy's caching mechanism. 

No wonder the test does not slow down as much as I expect!

The behavior could be surprising to developers in some cases, as it won't be apparent whether a test run involves loading data over network or not.

But I suppose it's okay for the tests in CI: presumably a new environment is created so that for each CI test run, it will test both loading data over network and loading data from local cache. (Unless the CI setup does some clever caching such that astropy cache is also saved for subsequent runs)

PR #835, opened 2020-09-09
doc tweak: state `python -m pip` rather than just `pip`

### LINKED ISSUE #834: Doc refinement: suggest python -m pip
> pip install -e .

We should change this to `python -m pip install -e .` by the way, cf. https://snarky.ca/why-you-should-use-python-m-pip/

_Originally posted by @barentsen in https://github.com/KeplerGO/lightkurve/pull/823#issuecomment-687528569_

orionlee:
Closes #834 

The main callout of `python` is in `install.rst`

image

I am not quite convinced it's clear, but feel if we say anything more, it'd be somewhat lengthy explanation on where python is, with various combination of OSes and python environment.

### DIFF
--- a/docs/source/about/contributing.rst
+++ b/docs/source/about/contributing.rst
@@ -39,7 +39,7 @@ In brief, the steps are as follows:
 .. code-block:: bash
 
     $ cd lightkurve
-    $ pip install -e .
+    $ python -m pip install -e .
 
 4. Add the KeplerGO remote to your GitHub enviroment:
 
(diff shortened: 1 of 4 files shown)


barentsen (approved):
Thanks so much @orionlee.  This is a great fix.

I am starting to see `python -m pip` recommended in other docs as well (e.g. the `emcee` package), so it may be fairly universal soon.  It is certainly best to leave details on Python environments and installation issues to other resources on the web, so I think this PR is perfect! 

Thank you!! :+1: :+1:

PR #841, opened 2020-09-16
bugfix: ensure LightCurve.time can be reassigned

### LINKED ISSUE #839: LightCurve.time cannot be reassigned
#### Problem description
In trying to change a LightCurve's `time` to a relative time, Lightkurve raises `AttributeError`.

#### Example

```python
import lightkurve as lk
lcf_coll = lk.search.search_lightcurvefile("TIC286923464", mission='TESS').download_all()
lc = lcf_coll[0]

# Set the lc's time relative to sector start time so that the plot shows relative time
lc = lcf_coll[0]
time_relative = lc.time - lc.meta['tstart']
lc.time_orig = lc.time
lc.time = time_relative  # Error is raised
lc.scatter()
```

The error:
```
<ipython-input-40-4e9941faed62> in <module>
      2 time_relative = lc.time - lc.meta['tstart']
      3 lc.time_orig = lc.time
----> 4 lc.time = time_relative
      5 
      6 # time_relative[1000:1010]

c:\dev\lightkurve\lightkurve\lightcurve.py in __setattr__(self, name, value, **kwargs)
    228         if ('columns' in self.__dict__) and (name in self.__dict__['columns']):
    229             if not isinstance(value, Quantity):
--> 230                 value = Quantity(value, dtype=value.dtype)
    231             self.replace_column(name, value)
    232         elif ('_meta' in self.__dict__) and (name in self.__dict__['_meta']):

C:\pkg\_winNonPortables\Anaconda3\envs\base_dev\lib\site-packages\astropy\time\core.py in __getattr__(self, attr)
   1732         else:
   1733             # Should raise AttributeError
-> 1734             return self.__getattribute__(attr)
   1735 
   1736     @override__dir__

AttributeError: 'Time' object has no attribute 'dtype'
```

#### Environment

-  lightkurve version (e.g. 1.0b6): `master` with commits on Sep 14, 2020, v2.0a2 based

orionlee:
Closes #839

### DIFF
--- a/lightkurve/tests/test_lightcurve.py
+++ b/lightkurve/tests/test_lightcurve.py
@@ -1075,6 +1075,24 @@ def test_mixed_instantiation():
     LightCurve(time=[1,2,3], flux=[1,2,3], data={'flux_err': [3,4,5]})
 
 
+def test_assignment_time():
+    """Ensure time property can be reassigned"""
+    lc = KeplerLightCurve(time=Time([1,2,3], scale='tdb', format='bkjd'), flux=[4,5,6], flux_err=[7,8,9])
+    time_adjusted = lc.time - 0.5
+    lc.time = time_adjusted
+    assert_array_equal(lc.time, time_adjusted)
+
+    # case the input is not given format / scale, ensure default format / scale is applied
+    time_adjusted_raw = [11., 12., 13.]
+    lc.time = time_adjusted_raw
+    assert_array_equal(lc.time, Time(time_adjusted_raw, scale='tdb', format='bkjd'))
+
+    # case the input is scalar
+    time_adjusted_raw_singular = 21
+    lc.time = time_adjusted_raw_singular
+    assert_array_equal(lc.time, Time(time_adjusted_raw_singular, scale='tdb', format='bkjd'))
+
+
 def test_create_transit_mask():
     """Test for `LightCurve.create_transit_mask()`."""
     # Set planet parameters
--- a/lightkurve/tests/test_lightcurve.py
+++ b/lightkurve/tests/test_lightcurve.py
@@ -1087,10 +1087,9 @@ def test_assignment_time():
     lc.time = time_adjusted_raw
     assert_array_equal(lc.time, Time(time_adjusted_raw, scale='tdb', format='bkjd'))
 
-    # case the input is scalar
-    time_adjusted_raw_singular = 21
-    lc.time = time_adjusted_raw_singular
-    assert_array_equal(lc.time, Time(time_adjusted_raw_singular, scale='tdb', format='bkjd'))
+    # case the input is scalar, it'd be broadcasted to the existing time's length
+    lc.time = 21
+    assert_array_equal(lc.time, Time([21, 21, 21], scale='tdb', format='bkjd'))
 
 
 def test_create_transit_mask():
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -225,7 +225,9 @@ class LightCurve(TimeSeries):
 
     def __setattr__(self, name, value, **kwargs):
         """To get copied, attributes have to be stored in the meta dictionary!"""
-        if ('columns' in self.__dict__) and (name in self.__dict__['columns']):
+        if (name == 'time'):
+            self['time'] = value # astropy will convert value to Time if needed
+        elif ('columns' in self.__dict__) and (name in self.__dict__['columns']):
             if not isinstance(value, Quantity):
                 value = Quantity(value, dtype=value.dtype)
             self.replace_column(name, value)


lightkurve/tests/test_lightcurve.py
```diff
@@ -1075,6 +1075,24 @@ def test_mixed_instantiation():
...
+    assert_array_equal(lc.time, time_adjusted)
+
+    # case the input is not given format / scale, ensure default format / scale is applied
+    time_adjusted_raw = [11., 12., 13.]
+    lc.time = time_adjusted_raw
+    assert_array_equal(lc.time, Time(time_adjusted_raw, scale='tdb', format='bkjd'))
+
+    # case the input is scalar
+    time_adjusted_raw_singular = 21
+    lc.time = time_adjusted_raw_singular
```
barentsen, line 1092: Interesting!  Is this equivalent to assigning [21, 21, 21]?

barentsen (approved):
Thanks @orionlee!  This looks like a perfect fix!  I particularly like the extensive unit test. :+1: :+1:

You mention WIP above.  Were you thinking of changing anything?  If not, I'm happy to merge!

barentsen:
(The `3.8-remote-data` test failure is due to a transient failure of a web service and is not related to the changes in this PR.)

lightkurve/tests/test_lightcurve.py
orionlee: hmm, you're correct it is `[21,21,21]`. I thought it'd be `[21]`

I don't understand why the assertion passed:
```
assert_array_equal(lc.time, Time(time_adjusted_raw_singular, scale='tdb', format='bkjd'))
```

orionlee: I changed the test to make it explicit that it's expecting `[21, 21, 21]`

orionlee:
It's ready to be merged.

barentsen:
@orionlee This is funny... this passes:
```
assert_array_equal(Time(1, scale='tdb', format='bkjd'), Time([1, 1, 1], scale='tdb', format='bkjd'))
```
this FAILS:
```
assert_array_equal(Time([1], scale='tdb', format='bkjd'), Time([1, 1, 1], scale='tdb', format='bkjd'))
```

Likewise, this passes:
```
assert_array_equal(np.array([1, 1, 1]), np.array(1))
```
this FAILS:
```
assert_array_equal(np.array([1, 1, 1]), np.array([1]))
```

I wonder if this is intentional or a bug.  I don't understand!  Either way, it doesn't seem to be a Lightkurve/Astropy issue.

orionlee:
One unrelated issue I noticed that probably do not warrant changes, but I am noting for just in case:

The logic currently does conversion to `Quantity` if some value is assigned to a predefined column, e.g., `flux`, but no conversion will happen if it's assigned as a new property.

```python
    flux_adjusted = np.array([7., 8., 9.])  # some routine that adjusts the flux, unitless
    lc.flux = flux_adjusted                 # <-- converted to Quantity as flux is a predefined column
    lc.flux_adjusted = flux_adjusted        # <-- NOT converted to Quantity
```

It's probably convenient for typical use cases, but it could be surprising if someone wants to add a new column/property, e.g., some adjusted flux to be visualized / compared side-by-side.

barentsen:
I guess we only support the `lc['flux_adjusted'] = ...` syntax for creating *new* columns for now.  I think this is fine because `pandas` shows the same behavior, and I'm in favor of replicating `pandas` whenever possible for consistency.

Here is what `pandas` does:

```python
>>> import pandas as pd
>>> df = pd.DataFrame({'a':[1,2,3], 'b':[4,5,6]})
>>> df.a = [7, 8, 9]  # allowed
>>> df.c = [7, 8, 9]  # not allowed
<ipython-input-8-fc15db4051f9>:1: UserWarning: Pandas doesn't allow columns to be created via a new attribute name - see https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access
```

orionlee:
> `lc['flux_adjusted'] = ...` syntax for creating new columns for now

Hmm, the column created with `lc['flux_adjusted'] = ...` has the type `Column` rather than `Quantity`. Functionally they are similar but they can't be used interchangeably, e.g., to access raw value, one uses `.value` for `Quantity`, and `.data` for `Column`.

I don't have a concrete suggestion per se, but the current behavior seem to be surprising when one wants to add new column / property (the distinction is blurry in the current API).

orionlee:
BTW, regarding `assert_array_equal`, I now believe the behavior is probably intentional.

1. `assert_array_equal` seems to broadcast a scalar to an array before comparison. So the following passes:
```python
    assert_array_equal([1, 1], 1)
    assert_array_equal(1, [1, 1])
```

2. for astropy `Time`, there is also scalar vs array difference
```python
    t_s = Time(1, format='jd')   # a scalar
    t_v = Time([1], format='jd') # a vector
    # t_s is a scalar
    assert(t_s.ndim == 0)
    assert(t_s.shape == ())
    assert(t_s == t_v)    # <-- this also passes, I'm not sure if it's also intentional.
    
```

3. So in my original test code, it compared a vector Time with a scalar Time.  `assert_array_equal` does the broadcast magic so the assert passed, not quite my original intention.

orionlee:
@barentsen This PR is ready to be merged, in case it gets lost in the side conversation.

barentsen:
> Hmm, the column created with lc['flux_adjusted'] = ... has the type Column rather than Quantity

That's indeed not great!  We should probably make it consistent, though it has the drawback of moving ever further away from `LightCurve` being a fairly standard `TimeSeries` object.

orionlee:
> > Hmm, the column created with lc['flux_adjusted'] = ... has the type Column rather than Quantity
> 
> That's indeed not great! We should probably make it consistent, though it has the drawback of moving ever further away from `LightCurve` being a fairly standard `TimeSeries` object.

What is the rationale of using `Quantity` rather than `Column` for the `.flux`, etc.?

barentsen:
> What is the rationale of using Quantity rather than Column for the .flux, etc.?

I can't remember the reason! (I should have added a comment in the code!)

We should probably take it out.

As is, operations like `lc.flux = [1, 2, 3]` crash anyway because a list does not have a `dtype`.

I'd have to play with it a bit more to explore the use cases, but I'm starting to get convinced we should take it out and let the default Table behavior deal with it!

orionlee:
Regarding `Quantity` vs `Column`: it turns out that once a column is assigned an unit, e.g., `lc['detrended'].unit = u.pct`, the column becomes `Quantity`.

The magic comes from underlying astropy [`QTable`](https://docs.astropy.org/en/latest/api/astropy.table.QTable.html#astropy.table.QTable).

Issue #831, opened 2020-09-08
Add support for filtering TESS products on exposure length


barentsen:
### Problem description

Currently, all TESS SPOC data products returned by `search_lightcurve()` and `search_targetpixelfile()` are 2-minute cadence products, because that is the only observing mode for which SPOC TPFs and LCs were archived during the prime mission.

This will change very soon! The extended mission includes a new, "fast", 20-second cadence mode. In addition, the SPOC and others will likely start releasing FFI-based TPF and LC products for both the prime (30-minute cadence) and extended mission (10-minute cadence) as HLSPs in the near future.

The search functions currently include a `cadence='long'` and `cadence='short'` option to filter Kepler and K2 data products by exposure length, but this argument is currently ignored for TESS data products.

### Solution

I propose to add support for filtering TESS data products as follows:
* When a numerical value is passed to `cadence`, it will be used to filter the `t_exptime` value in the MAST Portal API.
* When a string is passed to `cadence`, it will be parsed as follows:
  * `cadence='fast'` will map to `t_exptime < 60`
  * `cadence='short'` will map to `60 <= t_exptime < 300`
  * `cadence='long'` or `cadence='ffi'` will map to `t_exptime >= 300`.

Comments or feedback on this proposal are welcome!

### Caveats

* The exact integration time for TESS is slightly shorter than 60, 120, 600, or 1800 seconds, however these are the values we should use because [the MAST portal docs](https://mast.stsci.edu/api/v0/_c_a_o_mfields.html) defines `t_exptime` as the *total exposure length* rather than the *exact integration time*.  This is good because users will want to filter on exposure length, not the exact integration time!
* For Kepler and TESS FFI images, I believe the MAST portal accidentally populated `t_exptime` with *integration time* (e.g. `t_exptime=1425.599379` for FFI images), even though *exposure length* is consistently used for TPF and LC products (e.g. `t_exptime=1800` for Kepler long cadence products).

orionlee:
1. Should such change be made in v2.0 branch or v1.11.x ? The data products from TESS extended mission should be in soon, and this change might be needed sooner than v2.0 time frame.
2. It might be clearer to users if we list out the data products that could return depending on the `cadence`.
3. The current default value for `cadence` is `long`, which will not make sense for TESS.

---
[*] A sample table on the impact of `cadence` on the data products returned for TESS

|   | `fast` | `short` | `long` |
| --- | --- | --- | --- |
[...]

PR #844, opened 2020-09-18
Add support for the new TESS cadence modes to the search functions

barentsen:
This PR implements the changes described in #831.

It is a Work In Progress (WIP) because I am waiting for the new data products to become available at MAST before adding unit tests & merging.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -62,6 +62,14 @@ lightkurve.targetpixelfile
 lightkurve.search
 ^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


barentsen:
Note to self: before merging, please add tests for AU Mic, which is a high-profile 20-second target in Sector 27.

barentsen:
Update: it looks like TP/LC files will only be ingested into MAST tomorrow, so I'm holding off testing and merging until at least then!

barentsen:
The first S27 light curves are arriving in MAST now.  Here is what the new search behavior looks like:

Screen Shot 2020-09-24 at 07 24 59

PR #847, opened 2020-09-19
Make `PLDCorrector` more robust

barentsen:
The `PLDCorrector` does not appear to work out of the box for most TESS targets right now.  I'm opening this PR to share some ideas with @nksaunders as to how we might simplify `PLDCorrector` to make it more robust for random TESS targets.

Changes I explored in this PR are:
* Changed the default `pca_components` value from 7 to 3 for TESS targets, to make the default more robust against overfitting.
* Changed the default `pld_aperture_mask` to `'background'` for TESS targets, to help prevent overfitting the stellar signal.
* Modified the default value of `spline_n_knots` to be `len(lc) / 50` so that light curves with more data points automatically have more spline knots.
* The PLD and spline coefficients were not being regularized by the priors at all, so I modified their `prior_sigma` to be equal to `median(flux) / pca_components`.  This is a fairly weak constraint but I believe it helps prevent the fit from going crazy.
* Removed the `background` design matrix and `background_aperture_mask` argument for now. The PLD coefficients were often degenerate with the background model coefficient.
* Modified the `pld.diagnose()` method so that the labels show the CDPP values before/after correction.  Also changed the Y axis limits so that all three panels use the same limits, which makes comparisons easier.
* Commented out `regressors = np.array([r / f for r, f in zip(regressors, self.lc.flux.value)])` for now because it makes TESS corrections significantly worse, as far as I can tell.

Going forward, what we urgently need are a set of unit tests which verify PLD detrending against a set of CBV-corrected Kepler/K2/TESS light curves from MAST, so that we can automatically verify whether the default PLD settings do a reasonable job at all times.  It is too complicated to verify this by hand!

## Before this PR (significant overfitting)

Screen Shot 2020-09-18 at 18 30 54

## After this PR (no more overfitting)

Screen Shot 2020-09-18 at 18 29 43

### DIFF
--- a/lightkurve/correctors/designmatrix.py
+++ b/lightkurve/correctors/designmatrix.py
@@ -269,6 +269,7 @@ class DesignMatrix():
             dm = self.copy()
         extra_df = pd.DataFrame(np.atleast_2d(np.ones(self.shape[0])).T, columns=['offset'])
         dm.df = pd.concat([self.df, extra_df], axis=1)
+        dm.columns = list(dm.df.columns)
         dm.prior_mu = np.append(self.prior_mu, prior_mu)
         dm.prior_sigma = np.append(self.prior_sigma, prior_sigma)
         return dm
--- a/lightkurve/correctors/pldcorrector.py
+++ b/lightkurve/correctors/pldcorrector.py
@@ -115,7 +115,9 @@ class PLDCorrector(RegressionCorrector):
         return 'PLDCorrector (ID: {})'.format(self.lc.label)
 
     def create_design_matrix(self, pld_order=3, pca_components=16, pld_aperture_mask=None,
-                             spline_n_knots=None, spline_degree=3, sparse=False):
+                             background_aperture_mask='background', spline_n_knots=None,
+                             spline_degree=3, normalize_background_pixels=None,
+                             sparse=False):
         """Returns a `.DesignMatrixCollection` containing a `DesignMatrix` object
         for the background regressors, the PLD pixel component regressors, and
         the spline regressors.
@@ -165,6 +167,8 @@ class PLDCorrector(RegressionCorrector):
         # Validate the inputs
         pld_aperture_mask = self.tpf._parse_aperture_mask(pld_aperture_mask)
         self.pld_aperture_mask = pld_aperture_mask
+        background_aperture_mask = self.tpf._parse_aperture_mask(background_aperture_mask)
+        self.background_aperture_mask = background_aperture_mask
 
         if spline_n_knots is None:
             # Default to a spline per 50 data points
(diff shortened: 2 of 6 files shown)


lightkurve/correctors/designmatrix.py
```diff
@@ -269,6 +269,7 @@ def append_constant(self, prior_mu=0, prior_sigma=np.inf, inplace=False):
             dm = self.copy()
         extra_df = pd.DataFrame(np.atleast_2d(np.ones(self.shape[0])).T, columns=['offset'])
         dm.df = pd.concat([self.df, extra_df], axis=1)
+        dm.columns = list(dm.df.columns)
```
barentsen, line 272: (This is a small bug in DesignMatrix; `dm.columns` was not updated after calling `dm.append_constant()`)

jcsmithhere:
In my CBVCorrector tutorial I am using an example of how the CBVCorrector goodness metrics, CBVs and regularization can be used in conjunction with a background pixel level correction to generate an excellent looking light curve. I am providing an example where a pixel level correction on its own does indeed dramatically overfit. 

We should coordinate our tutorials before releasing 2.0. I should have my CBVCorrector PR ready very soon to show you all what I am doing.

barentsen:
@jcsmithhere That sounds fab! :+1:

nksaunders:
I've done some experimentation to find the best approach to avoid overfitting for both K2 and TESS, specifically with reference to normalization, aperture masks, and priors. Below is a description of the changes I propose in a [different PR](https://github.com/barentsen/lightkurve/pull/7).

Note that since writing this, `on_target_pix` has simply become `pld_aperture_mask` and `off_target_pix` is now `background_aperture_mask`.

## Normalization

The normalization step is separated into `on_target_pix` and `off_target_pix` for three important reasons:
- For K2, we want to normalize out the time-varying signal that is not due to motion noise, as motion noise is the dominant source we need to remove
- For both K2 and TESS, we want to normalize out the time-varying *stellar* signal from on-target pixels to ensure the PLD vectors don't overfit and remove stellar signal
- When we normalize out the stellar signal, we don't want to divide an off-target pixel by an on-target pixel to avoid injecting stellar signal into the background

I typically use K2-199 as a benchmark for these changes, as it is dramatically affected by K2 noise and has easily visible stellar variability. Here I show three cases for K2-199.

### 1. No flux normalization

download-2

While this initially looks like a good correction qualitatively, and the CDPP is very low, it's important to notice in the middle panel that the pixel series follows the same trend as the spline, fitting out the stellar rotation signal. We don't want that. This is a case of **overfitting** by PLD.

### 2. Only on-target normalization

This is what we want to do for TESS, and trying it on your favorite TESS target will demonstrate that it works well (for instance, WR40 maintains a clear pulsation signal; see section 4 for more details). This will prevent PLD from fitting out stellar signal by normalizing it out of on-target regressors for K2 targets, and should have no effect on TESS by default as there is no overlap between the `aperture_mask` and `pld_aperture_mask` (though it would serve the same purpose as for K2 were someone to pass a custom `pld_aperture_mask`). 

download-1

This has resolved the primary issue present in case 1, and the pixel series no longer fits out the stellar rotation. However, the scatter of the corrected light curve is much more dramatic and the CDPP has spiked. I believe this is because normalizing the on-target pixel regressors (which should be the dominant regressors in the model for K2) down-weights them relative to the background pixel regressors. This remains consistent with EVEREST, which flux-normalizes every pixel (though we've gone back and forth about whether background pixels should be normalized by a light curve that includes on-target pixels, which I believe they should not). This takes us to the third case.

### 3. On-target and off-target normalization

Here, I show the result of separately normalizing on-target and off-target pixels.

download

The CDPP is greatly improved by normalizing the background, and the pixel series still does not fit out the stellar signal. I believe this is the ideal case for K2.

### 4. Justification for not normalizing off-target pixels for TESS

The dominant source of noise in TESS FFI cutouts is the scattered light background. This high-amplitude signal should **not** be normalized out from each cadence, as it is a high-amplitude, time-varying, additive signal that should be captured by the pixel design matrix in the absence of a background design matrix. 

Here, I show two summary plots for WR40, a star with distinct stellar variability. First, the case where we treat TESS and K2 identically in the creation of the design matrix, and flux-normalize the background pixel regressors.

download-4

This is a terrible correction for two reasons:
- The pulsation signal of WR40 has been completely fit out by PLD
- The background scattered light signal is still present in the corrected light curve

Let's try treating this separately from K2, and not normalize the off-target pixel regressors.

download-5

Much better! The pixel series has isolated the background scattered light, and the pulsations of WR40 remain in the final corrected light curve.

## Priors

For the issue of priors, I found that with correct normalization, this value is not as important as long as it is finite and reasonable. That being said, I made this change after realizing that in the case of dramatic overfitting (i.e. case 1), setting prior sigma to something lower, and dependent on the standard deviation, helped prevent the pixel components from completely fitting out the stellar signal. There will be cases in which the default aperture masks are not ideal, and I believe this will help us prevent worse overfitting in those cases.

I'll demonstrate this with an example. Here, I show the correction in case 1 (with a prior of 10σ) versus the correction with a looser prior (median flux).

download-7

The tighter prior appears to have prevented worse overfitting in this case of poor design matrix construction. While this is not terribly demonstrative of the issues we will run into regularly, I think it will be helpful in cases where PLD would otherwise overfit. 

As a final example, I'll show the difference in performance when applied to a correctly normalized design matrix.

download-9

While both of these look like good corrections, and the CDPP is almost identical, the "hockey stick" feature is more pronounced in the correction with the looser prior. This is a very common issue for K2 corrections, and it appears that the tighter prior helps prevent it from blowing up at the beginning and end of the light curve.

I also believe that when normalizing the regressors, the standard deviation makes more sense as a prescription for prior sigma. In my tests, it seems that using 10σ instead of median has no dramatic impact on how well the correction performs, but can prevent overfitting in specific problem cases.

nksaunders (approved):
This looks great! I'm happy for this PR to be merged as is.

barentsen:
Thanks @nksaunders. Merging!

Issue #846, opened 2020-09-19
Lightkurve 2 API - normalize data not in flux column


orionlee:
#### Problem description
With Lightkurve2, if one wants to normalize data that is not `LightCurve.flux`, one has to jump through a few hoops, comparing to with Lightkurve 1.

#### Example

To normalize `SAP_FLUX`, in Lightkurve1, one can do it in 1 line

```python
    lc_normalized = lc.SAP_FLUX.normalize(unit='percent')
```

In Lightkurve2, one has to do it in a few steps (unless he/she uses the deprecated `.SAP_FLUX`)
```python
    lc_sap = lc.copy()
    lc_sap['flux] = lc_sap['sap_flux']
    lc_sap['sap_flux_err'] = lc_sap['sap_flux_err']
    lc_normalized = lc_sap.normalize(unit='percent')
```

#### Suggestion
Make `normalize`  takes an optional `column` attribute:

```python
    lc_normalized = lc.normalize(column='sap_flux', unit='percent')
```

The method gives a warning if the `{column}_err` does not exist.

barentsen:
Agreed!  `lc.normalize()` could take an optional `column` argument.

Note: if a user wants to work with `SAP_FLUX`, my recommendation is to use
```python
lc = lk.read(filename, flux_column='sap_flux')`
```
or
```python
lk.search(...).download(flux_column='sap_flux')
```

This will automatically populate `flux` and `flux_err` with `sap_flux` and `sap_flux_err`.

orionlee:
hmm, the having `flux_column` in `read` / `download` makes it less necessary to have `column` in `lc.normalize()` 

The `flux_column` in `read` / `download` is not yet (publicly) documented, however.

barentsen:
Right!  I think adding a `column` parameter to `lc.normalize` could be a useful enhancement, e.g. we also added `lc.plot(column=...)`, but I don't think it is necessary to go ahead with v2.0.

Absolutely correct that the `flux_column` option is not well documented as this time.  Lots of work left to do re docs!

PR #850, opened 2020-09-21
docfix: document flux_column for read / download ; also cleanup quality mask

orionlee:
Partially address #846,  by documenting `flux_column` in read / download so that `SAP_FLUX` can be normalized.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -2218,6 +2218,29 @@ class KeplerLightCurve(LightCurve):
 
     @classmethod
     def read(cls, *args, **kwargs):
+        """Returns a `KeplerLightCurve` by reading the given file.
+
+        Parameters
+        ----------
+        filename : str
+            Local path or remote url of a Kepler light curve FITS file.
+        flux_column : str, optional
+            The column in the FITS file to be read as `flux`. Defaults to 'pdcsap_flux'.
+            Typically 'pdcsap_flux' or 'sap_flux'.
+        quality_bitmask : str or int, optional
+            Bitmask (integer) which identifies the quality flag bitmask that should
+            be used to mask out bad cadences. If a string is passed, it has the
+            following meaning:
+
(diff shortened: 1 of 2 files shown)


orionlee:
In `KeplerLightCurve.read()`, it defaults to `Kepler` format, does that mean if one tries to read, say, a K2SFF lightcurve, one has to specify the `format` explicitly?

In download codepath, it seems to be able to detect different filetype variants and switch accordingly.

https://github.com/KeplerGO/lightkurve/blob/v2.0b1/lightkurve/io/read.py#L65

barentsen (approved):
Thanks @orionlee. This looks like an excellent docs improvement. :+1:

barentsen:
> In KeplerLightCurve.read(), it defaults to Kepler format, does that mean if one tries to read, say, a K2SFF lightcurve, one has to specify the format explicitly?

Correct.  I suppose it would be a useful enhancement to automatically detect the file format inside `KeplerLightCurve.read()` as well.

Both the docstring and file format detection were/are missing from those class methods because I've wanted to advertise the generic `lk.read()` function as the main interface to open a path, but of course that shouldn't stop us from making the `LightCurve.read()` methods as user-friendly as possible.

orionlee:
> I've wanted to advertise the generic `lk.read()` function as the main interface to open a path.

I didn't realize `lk.read()` is meant to be a public API - currently it is not listed. :-(

We should probably then properly expose and document `lk.read`.

orionlee:
I tried to expose `lk.read()` in api doc, but don't know sphinx enough to make it so - I can only expose the actual `lk.io.read()`.

barentsen:
@orionlee I've been working on a few tutorials which will feature it, and before releasing v2.0, I intend to revisit the API docs page to make sure all those key functions are prominently listed.  (Don't let that stop you from making the awesome improvements you made in this PR!  They benefit the API page directly, of course!)

PR #851, opened 2020-09-22
Doc flux column parameter to lk.read

orionlee:
Partially address #846 and follow-up on #850 

- Fix the docstring of `lk.read()`, the intended primary read api method.
- exposes `lk.read` in apidoc. See screenshot below.

image


barentsen (approved):
Brilliant!  Perfect PR!  💯 👍

PR #855, opened 2020-09-24
CBVCorrector Part II: The CBVCorrector class

jcsmithhere:
Here is the second PR for CBVCorrector. The accompanying tutorial (04-how-to-use-cbvcorrector.ipynb) should explain how to use the new class. I tried to make the class as versatile as I can but still making it simple enough in default mode for a novice user.

Key features:

1. Automatically downloads relevant CBVs from MAST.
2. Several fitting methods all utilizing L1-Norm or L2-Norm regularization.  
3. Ability to use an external DesignMatrix to perform a joint fit with the CBVs.
4. Over-fitting and under-fitting goodness metrics to use as a Loss Function to optimize the fit.
5. Some good plotting and diagnostic tools.
6. A good tutorial (I hope!).
7. I tried to be thorough with the code comments!

There are several improvements I can think of but I better get this PR in now, see what people think, then think about how to make the improvements. 

Some future improvements:

1. We should probably coordinate with PLDCorrector development to make both methods as compatible as possible.
2. Add in ability to set band-passes for measuring over- and under-fitting
3. Ability to optimize fit with CDPP
4. Move the over and under-fitting goodness metrics outside CBVCorrector because in principle they can be used with any correction method.


jcsmithhere:
Forgot about the new dependency: sklearn. That needs to be added in.

barentsen:
Thanks @jcsmithhere! This looks awesome! :+1:

I'll try to complete a full review asap!  Here's a first quick item of feedback which may sound like a big change but is probably easy to solve:

I note you have three correction methods: `correct_ElasticNet`, `correct_optimizer`, and `correct_gaussian_prior`.  To make the interface as user-friendly as possible, do you think we can pick one of these as the "default" method and rename that method `correct`?

I ask because the abstract `Corrector` class stipulates that Corrector classes must provide a `correct()` and a `diagnose()` method.  This adds a nice level of consistency across Lightkurve's correction classes (it lets users swap different correctors in and out).

Let's say we decide that the gaussian prior method ought to be the default (just for example), there are two ways we could make the two alternative methods easily accessible:

### Option A: Provide a merged `correct()` method

If you think it is possible to merge the three docstrings together in a single `correct()` docstring (e.g. because there is sufficient overlap in the keyword arguments), and if a single `diagnose()` plot is sufficient to diagnose all three methods, then we could add a merged function like this:
```python
def correct(method="gaussian_prior", ...):
    """Some merged version of the three docstrings."""
     if method == "gaussian_prior":
         correct_gaussian_prior(...)
     elif method == "elasticnet":
         correct_elasticnet(...)
     elif method == 'optimizer':
         correct_optimizer(...)
```

### Option B: Add `MethodNameCBVCorrector` sub-classes

If merging the docstrings makes everything too complicated, or if each method would benefit from having a customized `dianose()` plot, we should consider adding separate `ElasticNetCBVCorrector` and `OptimizerCBVCorrector` sub-classes, which could be simple classes which inherit most functions from the default `CBVCorrector` and only override the `correct()` and `diagnose()` methods.

Does that make sense?  Do you have a preference?  Or is there a reason we shouldn't be recommending a default method to users?

(If we think `ElasticNet` ought to be the default `CBVCorrector.correct()` method, then please go ahead and add scikit-learn to `requirements.txt` file in the root directory, so we can make the tests pass.  If you think it ought to be an optional or less-common alternative, we may want to make it an optional dependency.  I can help with that.)

jcsmithhere:
I purposely did not overload `correct()` because I want the superlcass `correct()` method accessible. In my tutorial right at the end I discuss how the  `RegressionCorrector.correct()` method is still accessible if one wishes to fine tune the priors on the `design_matrix_collection`.

So there are actually 4 `correct*` methods available. Would you prefer I have only a single `correct()` method with the correct(method="gaussian_prior", ...)` option like you give above? To me, that just seems more confusing to the user than having different method names for each. 

My unit tests call the ElasticNet method so doesn't that mean sklearn cannot be optional? I don't know how optional dependencies work.

I really think Nick, you and myself should have a discussion some time to coordinate development between PLDCorrector and CBVCorrector. The two methods should really be used in conjunction, like I show in my tutorial section titled "Using the Goodness Metrics and CBVCorrector with other Design Matrices". I think what we really might want is a CotrendingBasisVectors class, a PixelLevelBasisVectors class then a third class called something like RobustCorrector that uses the goodness metrics, CDPP and other metrics to regularize a joint fit between the CBVs and PLDs, plus what other models one wishes to use.

barentsen:
@jcsmithhere Stay tuned!  I'm looking carefully at the code and your excellent/detailed tutorial now.

jcsmithhere:
Hi @barentsen! CBVCorrector is now set up to use the external metrics module. My unit tests and tutorials all work. It looks like sklearn is a failed dependency. I am not sure how to resolve that. Is it simply I add scikit-learn to the `requirements.txt` file?

The only other outstanding question is regarding the `correct_*` methods. As I explained in a previous comment, I am hesitant to role all four `correct*` methods into a single overloaded method called `correct`. To me that seems to be more complex than having separate methods given their usage and arguments are different. `correct_optimizer` in particular is distinct from the other methods because it is a numerical optimizer, not an analytical solver. At the very least, that one should be a separate method. Perhaps the other three could be  combined but to me that might confuse users.

I can see some potential improvements but I think those should be in a separate PR. The current PR has been open for a long time.

barentsen:
Thanks @jcsmithhere!  I expect to get to this later today!

barentsen:
Thanks for working on this @jcsmithhere! 

> Is it simply I add scikit-learn to the requirements.txt file?

The dependencies are now defined in the `pyproject.toml` file, which recently became the standard way for Python packages to define dependencies.  I want ahead and added it for you to see if it makes the tests pass! (cf. e4d31b24fcc05db7750a8e57d7741170310276b2)

barentsen:
> As I explained in a previous comment, I am hesitant to roll all four correct* methods into a single overloaded method called correct. 

I am ok with a `Corrector` class offering multiple alternative correct methods, but I would really like there to be a default `correct()` method so that we have a consistent and user-friendly API across all the classes that implement the abstract `Corrector` interface.

Quick fix: can I suggest we rename `correct_optimizer()` into `correct()`?

I make this suggestion for three reasons:
* It would allows us to promote `correct_optimizer()` as the default and recommended way to apply CBV correction.  This feels right to me because it is most similar to what the official pipeline does. We should of course have lots of warning bells to explain that optimization is slow and that there is no default method which works perfect all the time. On the other hand, making this the default would really emphasize that systematics correction involves making complicated trade-offs, which is a good thing.
* It would allow us to explain that `correct_elasticnet()` and `correct_gaussian_prior()` are alternative methods for users who need the correction to run much faster, or who wish to dig deeper in the details of the correction.
* I understand that redefining `correct()` takes away the ability to run the parent `RegressionCorrector.correct()` method.  I'm happy to add a `CBVCorrector.correct_regression()` method to expose this parent method, though I'd recommend users who wish to tune the design matrix to interact with `RegressionCorrector` directly.

If we rename `correct_optimizer()` => `correct()`, I would be happy to merge!

barentsen:
(Note: the `pytest-windows` check is failing for reasons that are not related to this PR, so the tests look good!)

jcsmithhere:
Hi @barentsen, I get your argument about making `correct_optimizer()` the default. My main concern is `correct_optimizer()` is a numerical optimizer. Whereas the others, including `RegressionCorrector.correct()`are analytical. Are you OK having `correct()` sometimes being either analytical and numerical?

What I can do is rename `correct_optimizer()` => `correct()` and then create a `correct_RegressionCorrector()` as a way to access the superclass `RegressionCorrector.correct()` method.

I will work on this today.

jcsmithhere:
OK, @barentsen. I think we might be ready to merge.

barentsen:
Thanks @jcsmithhere!  This is looking good!!

I'm happy to go ahead, rebase on master, and merge.  Would you be upset if I changed the case of these two method names:
* `correct_ElasticNet` -> `correct_elasticnet`
* `correct_RegressionCorrector` -> `correct_regressioncorrector`

...and this class name:
* `minTargetsError` -> `MinTargetsError`

...this would make the PR consistent with the "functions_are_always_lowercase" and "ClassesAreAlwaysCamelCase" convention in use across Lightkurve (and prescribed by PEP8).

It does of course make sense to use camel case for ElasticNet and RegressionCorrector because they refer to class names, however we have precedents across Lightkurve where we didn't follow this logic (e.g. `to_lightcurve()` instead of `to_LightCurve()`), so I think it's better to maintain this tradition for API consistency.

jcsmithhere:
Hi @barentsen, your changes look good to me!

barentsen (approved):

PR #856, opened 2020-09-24
Improve behavior of interact() on fast-cadence TESS TPFs

barentsen:
## Problem

By default, `TargetPixelFile.interact()` has a limitation of `max_cadences=30000` to avoid freezing a user's computer.  This leads to a poor user experience with the new TESS fast-cadence (20s) data:

Screen Shot 2020-09-24 at 11 54 10

## Solution

This PR improves the default behavior in three ways:
* The default value for `max_cadences` is increased to 200000 so it doesn't crash out of the box with TESS fast-cadence data.
* A user-friendly warning (rather than error) is shown if the number of cadences exceeds 30000.
* The error message shown has been made much more user-friendly.

Example new warning message:

Screen Shot 2020-09-24 at 12 18 05

Example new error message:

Screen Shot 2020-09-24 at 12 18 25

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,7 +54,7 @@ lightkurve.targetpixelfile
   memory leak on indexing or slicing a tpf. [#829]
 
 - Modified ``interact()`` to use ``max_cadences=200000`` by default to allow
-  it to be used on fast-cadence TESS data.
+  it to be used on fast-cadence TESS data. [#856]
 
 lightkurve.search
 ^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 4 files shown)


mrtommyb:
Looks like an excellent fix. running with 200,000 cadences provides poor performance. But in my view, poor performance is better than failure.

barentsen:
Agreed!

rebekah9969:
This looks great Geert and thank you for doing it. A future suggestion would be adding in a specified time frame such that the user can look at the object around their date of interest.

barentsen:
@rebekah9969 Good news, we already support this masking.  For example, to inspect the TPF near BTJD=2050, the user can do this:
```python
tpf[(tpf.time.value > 2049.5) & (tpf.time.value < 2050.5)].interact()
```

PR #857, opened 2020-09-25
Allow many flux inputs to from_fits_images

burke86:
This is an initial PR to allow flux_err, flux_bkg, etc. inputs to the `from_fits_images` function in addition to flux (the only option allowed previously). This enhances the flexibility of `from_fits_images` at risk of making it more complicated. Some code could probably be improved. Feedback is welcome!


barentsen:
@burke86 This looks pretty good to me at first sight! :+1:

Are you still working on it, or would you like a full review?  (I'm not sure I'll have any comments to be honest, but I'd check more carefully.)

Just curious: is there a good use case for making full-blown TPFs like this?  I guess it enables this feature to make better cut-outs from TESS FFIs?  (just guessing)

burke86:
Hi @barentsen, I think it is ready for review. Thanks for your quick reply!

Yes I am using this to go from TESS FFIs, performing some custom image processing on the FFIs (difference imaging in particular requires a nearly full-sized FFI to get multiple reference stars + custom background subtraction), to TargetPixelFile objects which can then be analyzed using lightkurve. Basically I am building the TPFs myself from our FFIs. We have had success detrending TESS images this way to look at variability in extended sources, with NGC 4395 being the best example so far!

barentsen (approved):
@burke86 Ha! My guess was correct! 🧙 

I had a careful second look and it all looks great.  I'm even going to give you a gold star for using `enumerate`: ⭐.

Two minor things are missin:
* We need an entry in `CHANGES.rst` (I can take care of that)
* There are no tests for these new arguments yet. I guess I'm ok with that because this isn't a critical feature of Lightkurve yet, and the existing `test_tpf_factory` at least confirms that we didn't break previous behavior.

I'm happy to merge! Thanks++! :+1:

PR #860, opened 2020-09-28
Fix BLSPeriodogram apidoc ; change duration defaults

### LINKED ISSUE #859: BLS BoxLeastSquaresPeriodogram 's default yields inaccurate results when transit duration is short
#### Problem description

`BoxLeastSquaresPeriodogram` 's `duration` parameter has a lightkurve-specified default of `0.25` (6 hours).

- The result could be quite inaccurate when the actual transit duration is significantly shorter.
- It's also inconsistent with the default in `interact_bls()`, which is `0.05` (1.2 hours)

#### Expected behavior
It'd be great if the default should give more accurate results.

#### Proposal
Change the default `duration` to `[0.05, 0.10, 0.15, 0.20, 0.25, 0.33]` , which is closer to the range typical planetary transit duration.

#### Environment

-  lightkurve version (e.g. 1.0b6): 1.11.2

orionlee:
1. Add apidoc for `BLSPeriodogram`, mainly to explain the parameters, especially those with defaults created by lightkurve (rather than just relying on underlying astropy).
2. Close #859 (`duration` defaults)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -107,6 +107,9 @@ lightkurve.periodogram
 - Modified ``create_transit_mask`` method to return ``True`` during transits and
   ``False`` elsewhere for consistent mask syntax. [#808]
 
+- Modified ``BoxLeastSquaresPeriodogram`` to use ``duration=[0.05, 0.10, 0.15, 0.20, 0.25, 0.33]``
+  by default, which yields more accurate results (albeit slower). [#859, #860]
+
 
 
 1.11.3 (2020-10-06)
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -1715,9 +1715,9 @@ class LightCurve(TimeSeries):
         power spectrum object.
 
         This method will call either
-        `lightkurve.periodogram.LombScarglePeriodogram.from_lightcurve()` or
-        `lightkurve.periodogram.BoxLeastSquaresPeriodogram.from_lightcurve()`,
-        which in turn wrap `astropy.stats.LombScargle` and `astropy.stats.BoxLeastSquares`.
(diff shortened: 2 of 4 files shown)


barentsen:
Oops, it looks like the tests are genuinely failing with this change.  I can look into this more closely tomorrow.

orionlee:
hmm, I'm not quite sure what we should do. 

For the synthetic lightcurve used in the failing test, it seems that BLS algorithm does yield a second peak at period 0.666 day (the correct period is 2 days) when a short duration (0.05, the correct duration is 0.10) is included.

image

Specifically, 4 periods gave out identical power. The periodogram reports the shortest one.

image


I played with the parameters a bit, it seems that when the duration including values between 0.03 and 0.05 , it'll find a peak with 0.666 day period.

orionlee:
`interact_bls()` , with a default duration of `0.05` day, also yields the incorrect 0.666 period for the synthetic lightcurve.

image

orionlee:
The latest commit tweaked the synthetic lightcurve to make it more realistic.
The old one has only 1 data point during each dip. The new one is more granular so that it has a few data points during each dip.

image

Generally I'd feel tweaking the test data to make a test pass is cheating. But I think in this case it's justified because the test data is atypical that it does make the underlying BLS stumble.

I'd think the main point is whether the new default duration `[0.05, 0.10, 0.15, 0.20, 0.25, 0.33]` is preferable over the existing one `0.25` day.

barentsen (approved):
Thanks @orionlee!  I'm very happy with these changes and improvements! :+1:

> Generally I'd feel tweaking the test data to make a test pass is cheating. But I think in this case it's justified

Agreed, I think it's justified!

barentsen:
I pushed 7cb8cd4 to add an entry in CHANGES.rst to document the change to the default duration.  Will merge when tests pass! :+1:

PR #821, opened 2020-09-02
make TargetPixelFile.meta read-only

orionlee:
- avoid potential inconsistency if underlying hdu header is modified.

See the discussion in: 
https://github.com/KeplerGO/lightkurve/commit/b8f0f7993a28b6b46b28cc96973e2bc111ed6e3a

PR #864, opened 2020-10-05
Attribute access for LightCurve meta / columns ; meta keys' case

### LINKED ISSUE #845: Lightkurve2 API - make LightCurve API less surprising
I did a small exercise, comparing the codes using Lightkurve2 `LightCurve` vs plain astropy `TimeSeries` in doing some lightcurve processing / plotting:
  https://gist.github.com/orionlee/633a42d34a8e0b6f166a85e35afa27d2

#### Observations
In current Lightkurve2 alpha, `LightCurve` object provides a set of shortcuts that 
1. expose underlying column data (e.g., `time`, `flux`, `quality` etc.) as attributes: e.g., `lc.quality`
2. expose `meta` items (e.g., `TSTART`, `SECTOR` in the underlying FITS header) as attributes: e.g., `lc.tstart`.

While they are quite convenient, it could be confusing when one wants to attach additional data to the object, e.g., detrended flux, relative time, etc. E.g., if one wants to add detrended flux to a LightCurve object by using `lc.flux_detrended`, it won't be treated as an additional column. 

Not treating as an column could lead to some behavior difference that could be surprising:
1. the data can't be used as a column in helpers such as `plot(column=...)`
2. When assigning data to a column, e.g., `lc['flux'] = flux_adjusted`, the values of `flux_adjusted` are copied to the `lc['flux']` column. Subsequent changes to `flux_adjusted` won't affect `lc['flux']` (and vice versa). In contrast, when data to an attribute, the data is not copied.

---

#### Suggested changes:
1. Eliminate attribute shortcuts that hide underlying meta or columns. Or provide read-only access, rather than read-write. 
  - astropy `TimeSeries` provides no such shortcut, with the lone exception of a read-only `.time`
  - when one tries to attach some data to LightCurve object, it will be clear what the code does
     - add / update columns, e.g. `lc['flux_detrended'] = ...`
     - add / update metadata, e.g., `lc.meta['tstart'] = ...`
     - add / update attributes is still possible

2. If (1) is adopted, consider to drop the automatic lowercasing of metadata key, e.g., accessing `TSTART` will become `lc.meta['TSTART']`
  - astropy `TimeSeries` does not do such conversion either.
  - For people familiar with Lightkurve 1 or FITS, the conversion could be a magic that surprises them.
  - I presume the motivation for automatic lowercasing is to make accessing them via attributes more natural (e.g., `lc.tstart`). With adoption of (1), either eliminating such attributes or make them read-only, I think the automatic lowercasing is no longer needed.

orionlee:
Closes #845
- [x] Access columns / meta values via attributes ; emits warnings when user creates a new attribute (pandas behavior)
- [x] Document attribute access behavior and reference it in the warning.
- [x] retain meta key's case (i.e., uppercase for TESS / Kepler headers)
- [x] `TargetPixelFile` meta key's case

### DIFF
--- a/lightkurve/interact.py
+++ b/lightkurve/interact.py
@@ -148,7 +148,7 @@ def make_lightcurve_figure_elements(lc, lc_source, ylim_func=None):
     step_renderer : GlyphRenderer
     vertical_line : Span
     """
-    mission = lc.meta.get('mission')
+    mission = lc.meta.get('MISSION')
     if mission == 'K2':
         title = "Lightcurve for {} (K2 C{})".format(
             lc.label, lc.campaign)
@@ -551,7 +551,7 @@ def show_interact_widget(tpf, notebook_url='localhost:8888',
         aperture_mask = np.zeros(tpf.flux.shape[1:]).astype(bool)
         aperture_mask[0, 0] = True
 
-    lc.meta['aperture_mask'] = aperture_mask
+    lc.meta['APERTURE_MASK'] = aperture_mask
 
     if transform_func is not None:
         lc = transform_func(lc)
@@ -617,7 +617,7 @@ def show_interact_widget(tpf, notebook_url='localhost:8888',
             selected_indices = np.array(selected_pixel_indices)
             selected_mask = np.isin(pixel_index_array, selected_indices)
             lc_new = tpf.to_lightcurve(aperture_mask=selected_mask)
-            lc_new.meta['aperture_mask'] = selected_mask
+            lc_new.meta['APERTURE_MASK'] = selected_mask
             if transform_func is not None:
                 lc_transformed = transform_func(lc_new)
                 if (len(lc_transformed) != len(lc_new)):
@@ -626,7 +626,7 @@ def show_interact_widget(tpf, notebook_url='localhost:8888',
                             'Skipping the transformation...', LightkurveWarning)
                 else:
                     lc_new = lc_transformed
-                    lc_new.meta['aperture_mask'] = selected_mask
+                    lc_new.meta['APERTURE_MASK'] = selected_mask
             return lc_new
 
         def update_upon_pixel_selection(attr, old, new):
@@ -685,10 +685,10 @@ def show_interact_widget(tpf, notebook_url='localhost:8888',
                                                     transform_func=transform_func)
                 lc_new.to_fits(exported_filename, overwrite=True,
                                flux_column_name='SAP_FLUX',
-                               aperture_mask=lc_new.meta['aperture_mask'].astype(np.int),
+                               aperture_mask=lc_new.meta['APERTURE_MASK'].astype(np.int),
                                SOURCE='lightkurve interact',
                                NOTE='custom mask',
-                               MASKNPIX=np.nansum(lc_new.meta['aperture_mask']))
+                               MASKNPIX=np.nansum(lc_new.meta['APERTURE_MASK']))
                 if message_on_save.text == " ":
                     text = '<font color="black"><i>Saved file {} </i></font>'
                     message_on_save.text = text.format(exported_filename)
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -81,6 +81,10 @@ class LightCurve(TimeSeries):
     Note that you *cannot* create a new column using the attribute interface. If you do so,
     a new attribute is created instead, and a warning is raised.
 
+    If you do create such attributes on purpose, please note that the attributes are not carried
+    over when the lightcurve object is copied, or a new lightcurve object is derived
+    based on a copy, e.g., ``normalize()``.
+
 
     Examples
     --------
@@ -273,7 +277,8 @@ class LightCurve(TimeSeries):
             to_set_as_attr = True
         if to_set_as_attr:
             if name not in self.__dict__ and not name.startswith("_") and not self._new_attributes_relax:
-                warnings.warn(("Lightkurve doesn't allow columns or meta values to be created via a new attribute name"
+                warnings.warn(("Lightkurve doesn't allow columns or meta values to be created via a new attribute name."
+                               "A new attribute is created. It will not be carried over when the object is copied."
                                " - see https://docs.lightkurve.org/api/lightkurve.lightcurve.LightCurve.html"),
                               UserWarning, stacklevel=2)
             super().__setattr__(name, value, **kwargs)
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -234,6 +234,8 @@ class LightCurve(TimeSeries):
         elif ('_meta' in self.__dict__) and (name in self.__dict__['_meta']):
             self.__dict__['_meta'][name] = value
         else:
+            warnings.warn("Lightkurve doesn't allow columns or meta values to be created via a new attribute name",
+                          UserWarning)
             super().__setattr__(name, value, **kwargs)
 
     def _base_repr_(self, html=False, descr_vals=None, **kwargs):
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -110,6 +110,12 @@ class LightCurve(TimeSeries):
     _default_time_format = "jd"
     _default_time_scale = "tdb"
 
+    # To emulate pandas, we do not support creating new columns or meta data
+    # fields via attribute assignment, and raise a warning in __setattr__ when
+    # a new attribute is created.  We need to relax this warning during the
+    # initial construction of the object using `_new_attributes_relax`.
+    _new_attributes_relax = True
+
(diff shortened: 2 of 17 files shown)


lightkurve/interact.py
```diff
@@ -551,7 +551,7 @@ def show_interact_widget(tpf, notebook_url='localhost:8888',
         aperture_mask = np.zeros(tpf.flux.shape[1:]).astype(bool)
         aperture_mask[0, 0] = True
 
-    lc.meta['aperture_mask'] = aperture_mask
+    lc.meta['APERTURE_MASK'] = aperture_mask
```
barentsen, line 555: Note: a small number of meta data keywords, such as `label` and `aperture_mask`, don't come from the FITS header but are instead set by lightkurve methods.  I wonder if we should adopt lowercase or uppercase for those.  I guess uppercase for everything makes things more consistent.

orionlee:
The PR is at a point that is ready for review.

Open: `TargetPixelFile` also make meta key lowercased. To keep things consistent, we probably should change the case in this PR as well.

barentsen:
@orionlee This looks good at first sight!  My only remark is the comment I left above, but I think it's probably better to do everything uppercase (as you did).

Yep, we'll have to modify `targetpixelfile.py`, including the `extract_aperture_photometry` methods which populate some LightCurve meta data.

barentsen:
(PS: I'm pretty sure you can ignore the failing osx test.  Looks like a transient numerical issue.)

lightkurve/interact.py
orionlee: I feel keeping all the meta keys uppercase would be easier for users.

In the current PR, it does let meta to have both uppercase and non-uppercase keys (it's a `dict`), a test is added to show how we resolve such cases if someone access it via attributes.

I considered going further, emitting a warning if someone sets meta with a a non uppercase keys, but it might be overly prescriptive (and extra codes needed to replace the standard `dict` with one that checks key cases).

orionlee:
I've updated `TargetPixelFile.meta` so that the the key retains original case. 
I also made it read-only, subsuming PR #821.

One thing I noticed is that for the underlying `hdu.header`, an astropy fits header instance, the keys are case-insensitive. In other words,

```
   assert(tpf.hdu[0].header['MISSION'] == tpf.hdu[0].header['mission'])
```

astropy fits header: keys are case insensitive, with the keys get normalized to uppercase.
I wonder if we should follow suit.

orionlee:
The warning logic has an issue that is showstopper.

The `LightCurve` class, and the parent astropy `TimeSerires` / `Table` classes create real attributes internally. As it is, tons of warnings will be emitted during typical usage, rendering them practically useless.

```python
>>> lc = lk.LightCurve(time=[1, 2, 3, 4], flux=[0.98, 1.02, 1.03, 0.97])
C:\dev\lightkurve\lightkurve\lightcurve.py:115: UserWarning: Lightkurve doesn't allow columns or meta values to be created via a new attribute name - see https://docs.lightkurve.org/api/lightkurve.lightcurve.LightCurve.html
  self._required_columns_relax = True
C:\dev\lightkurve\lightkurve\lightcurve.py:162: UserWarning: Lightkurve doesn't allow columns or meta values to be created via a new attribute name - see https://docs.lightkurve.org/api/lightkurve.lightcurve.LightCurve.html
  self._required_columns)
C:\pkg\_winNonPortables\Anaconda3\envs\base_dev\lib\site-packages\astropy\table\table.py:1773: UserWarning: Lightkurve doesn't allow columns or meta values to be created via a new attribute name - see https://docs.lightkurve.org/api/lightkurve.lightcurve.LightCurve.html
  self._masked = masked
C:\pkg\_winNonPortables\Anaconda3\envs\base_dev\lib\site-packages\astropy\table\table.py:1777: UserWarning: Lightkurve doesn't allow columns or meta values to be created via a new attribute name - see https://docs.lightkurve.org/api/lightkurve.lightcurve.LightCurve.html
  self._column_class = self.MaskedColumn if self._masked else self.Column
...
```

orionlee:
@barentsen I think we have to give up emitting a warning when users create a new attribute. Thoughts?

Approaches considered to workaround the problem that attributes are created internally.
- Use `object.__setattr__` to bypass the check: it'd work for `Lightkurve` codes, but not for the astropy `TimeSeries` codes,
- do not emit warning for internal attributes (starts with `_`): it alleviates the problem, but it doesn't solve the problems for public attributes created by `TimeSeries`.
- We could possibly enumerate all attributes created by `TimeSeries`, bake them into the code so that they are ignored. But it'd be a ongoing headache to maintain such lists.

barentsen:
Thanks for digging into this @orionlee!

We may be able to draw inspiration from the `_required_columns_relax` attribute, which TimeSeries uses to circumvent required column restrictions during the initial creation of the object.

I played around with this idea and pushed a new commit 03570c6 which implements a similar `_new_attributes_relax` mechanism, to avoid raising the attribute assignment warnings during object construction.

I verified that running `pytest --remote-data` no longer encounters any attribute creation warning, so I think this may be a successful solution.  What do you think?

(PS: While debugging, I added the options `-W error --pdb` to pytest, which turns warnings into exceptions and immediately opens the pdb debugger upon an exception.  That made it super easy to inspect where and why a warning was raised.)

orionlee:
I considered the approach of relaxing attribute assignment check during object creation. My main concern is that the base `TimeSeries` codes could create new attributes from any other methods as well after the object is created.

Creating new attributes after object creation probably doesn't happen frequently, if at all; but we have no sure way to know. 

If we adopt such approach, we might have to live with occasional warnings when users call certain `TimeSeries` methods (directly or indirectly). The occurrence could change from from one version of astropy to another too.

barentsen:
@orionlee Agreed, that's a fair concern, though I haven't found examples yet of Table methods assigning attributes (other than private attributes starting with an underscore).

It sounds like we'll have to make a trade-off between having the warning on one hand, and the risk of raising it inadvertently on the other.  I think I'm fine with having the warning, but I'll have a look through AstroPy's table methods to see if there are any that would trigger the warning inadvertently.

barentsen:
I looked through `astropy/table/table.py` and `astropy/table/operations.py` and did not spot any methods which create new table attributes (other than the constructor).

orionlee:
I suppose we can keep our fingers crossed, and hope future version of TimeSeries won't add new public attributes after object creation.

Has there been any talk in TimeSeries of possibly adopting pandas-like behavior regarding attribute access?

orionlee:
On my end, the PR is ready to be merged, pending any review feedback.

I just recalled there is a separate concern on the types of values accepted to be a column [*]. I'll discuss it separately (and a separate PR if needed).

[*] In the current logic, a new column accept any array-like values, while updating an existing column, it requires a `Quantity` or a typed array (and converts it to `Quantity` internally).

barentsen:
> Has there been any talk in TimeSeries of possibly adopting pandas-like behavior regarding attribute access?

Not that I'm aware of!  I'm not able to find a discussion on this topic in Astropy's GitHub issue tracker.  I asked the question on AstroPy's #table slack channel just now.

lightkurve/lightcurve.py
```diff
@@ -62,6 +62,30 @@ class LightCurve(TimeSeries):
...
+    >>> lc.flux = lc.flux * 1.05  # update the values of a column.   # doctest: +SKIP
+
+    In case the given name is both a column name and a key in ``meta``, the column will be returned.
+
+    Note that you *cannot* create a new column using the attribute interface. If you do so,
+    a new attribute is created instead, and a warning is raised.
+
+    If you do create such attributes on purpose, please note that the attributes are not carried
+    over when the lightcurve object is copied, or a new lightcurve object is derived
+    based on a copy, e.g., ``normalize()``.
```
barentsen, line 86: I love this docstring!  Thank you!!

lightkurve/tests/test_lightcurve.py
```diff
@@ -1092,6 +1092,107 @@ def test_assignment_time():
...
+    lc = LightCurve(time=Time([1, 2, 3], scale='tdb', format='jd'), flux=[4, 5, 6] * u_e_s)
+
+    # case the name is present both as a column name and a meta key: column is returned
+    val_of_col = [5, 6, 7]
+    val_of_meta_key = 'value'
+    lc['foo'] = val_of_col
+    lc.meta['FOO'] = val_of_meta_key
+    assert_array_equal(lc.foo, val_of_col) # lc.foo refers to the column
+
+
```
barentsen, line 1195: I love these tests! :+1:

barentsen (approved):
Thanks so much @orionlee!  I looked through all the changes again, and they all look good.  I intend to go ahead and merge!

Thank you!!

barentsen:
@orionlee There is a thread about it on the AstroPy issue tracker now!  https://github.com/astropy/astropy/pull/10809

Issue #863, opened 2020-10-05
Issue with lightkurve plots in Google Colab


avanderburg:
#### Problem description
Plots from lightkurve do not show up, and matplotlib plots do not show up after lightkurve has been imported (unless matplotlib is imported and used first). 

#### Example
I have created two colabs showing the problem and workaround here: 
Showing the issue: https://colab.research.google.com/drive/10sT-3kK08qERkPDkEgKMxqKA5yfVVvyB?usp=sharing
Showing the workaround: https://colab.research.google.com/drive/130GWUVJ29IahvpdXNHjJcu-_xYEphz5f?usp=sharing

#### Expected behavior
I expected lightkurve plots to show up in Colab without having to import matplotlib and plotting something first. 

#### Environment

-  platform (e.g. Linux, OSX, Windows): Google Colab
-  lightkurve version (e.g. 1.0b6): 1.11.2
-  installation method (e.g. pip, conda, source): pip

PS: I have a bunch of new students, and all of them are using lightkurve to access Kepler and TESS data. It's a great way to introduce them to the datasets and analysis!

avanderburg:
Actually this looks like the same issue DFM noticed - if so disregard!

barentsen:
Thanks @avanderburg!  Yep, this looks like the same issue as #861.  I'm happy you have a temporary workaround, and hope to have a fix available later today!

barentsen:
@avanderburg Thanks again for reporting this issue.  I appreciate your time!

I released a new version of Lightkurve on pypi tonight (v1.11.3) which fixes the issue.

I verified that the Google Colab notebook you shared shows the missing plots now (see screenshot below). 

Don't hesitate to open new issues for any other problems you encounter. Thanks++!

### Screenshot confirming the fix

<img width="808" alt="Screen Shot 2020-10-06 at 22 20 13" src="https://user-images.githubusercontent.com/817669/95290480-340b5400-0822-11eb-96e7-d2760f07d4ab.png">

Issue #861, opened 2020-10-03
Issue with %matplotlib inline in jupyter notebooks


dfm:
#### Problem description

In a few scenarios I'm finding that importing lightkurve in a jupyter notebook after running `%matplotlib inline` changes the matlotlib settings such that plots don't show up.

#### Example

A full example can be seen here: https://colab.research.google.com/drive/1SrqUIP5nN3_2GBxqphFLtak4ZoFV6tH-?usp=sharing but it's not isolated to Colab environments because I'm also getting it on GitHub actions.

Essentially: 

```python
%matplotlib inline
import lightkurve as lk
import matplotlib.pyplot as plt
plt.plot(...)
```

doesn't show a plot, but 

```python
import lightkurve as lk
%matplotlib inline
import matplotlib.pyplot as plt
plt.plot(...)
```

does.

#### Expected behavior

I would not expect importing lightkurve to affect the behavior of matplotlib.

#### Environment

-  platform (e.g. Linux, OSX, Windows): Ubuntu, Google Colab, perhaps others?
-  lightkurve version (e.g. 1.0b6): 1.11.2
-  installation method (e.g. pip, conda, source): pip

barentsen:
The behavior is caused by this section of our `__init__.py`:

https://github.com/KeplerGO/lightkurve/blob/bb5e5d23a985a4cad3b3d4db6ce8d4c94d31f66a/lightkurve/__init__.py#L30-L36

I believe the reason for messing with the backend in this way was to make the plotting functions run out of the box on GitHub Actions and other headless Linux environments.  This is clearly a bad idea and should be fixed!

barentsen:
Good news!  This issue has been fixed by #865 and is included in the new v1.11.3 and v2.0b3 releases.

I was not smart enough to come up with an automated unit test for this bug, but here is a screenshot to show that importing lightkurve no longer prohibits matplotlib plots from showing up in the Colab notebook you posted:

<img width="660" alt="Screen Shot 2020-10-06 at 20 00 12" src="https://user-images.githubusercontent.com/817669/95282559-99554a00-080e-11eb-9d92-8d854ef08435.png">


Thanks again for taking the effort to report this bug @dfm. I am grateful for your time! 🙏

dfm:
Woohoo! Thanks!!

PR #865, opened 2020-10-06
Fix plotting in Google Colab

barentsen:
At present, Lightkurve's `__init__` changes the matplotlib back-end if a headless Linux display is detected:

https://github.com/KeplerGO/lightkurve/blob/ea53c81f3d7617441a02288ed84c016e8ef80ceb/lightkurve/__init__.py#L30-L36

We did this to enable our plotting methods to be tested by continuous integration services (Travis, Azure, GitHub Actions).

Unfortunately, two recent issues (#861 & #863) demonstrate that this is a bad idea, e.g. because it messes with the behavior of matplotlib on Google Colab.

This PR removes the offending piece of code and will explore a different way to make GitHub Actions pass.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -109,6 +109,13 @@ lightkurve.periodogram
 
 
 
+1.11.3 (2020-10-06)
+===================
+
+- Fixed inline plots not appearing in Jupyter Notebooks and Google Colab. [#865]
+
+
+
 1.11.2 (2020-08-28)
 ===================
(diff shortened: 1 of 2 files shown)


barentsen:
All the tests pass, so I believe that messing with the matplotlib backend was only required to make Travis or Azure-pipelines pass.  We switched to Github Actions a while ago, and should probably have removed this logic when we did!

PR #869, opened 2020-10-07
Bugfix lightcurve attribute access

orionlee:
This is a follow up on PR #864 , addressing issue #845

It occurred to me while I was taking a walk. 

For updating via attribute interface in the (really edge) case that the name exists as a real attribute, a column name, and a meta key.

I think it should update the real attribute. Even without the PR, the existing codes already chooses the real attribute in the get case. This fix brings the parity to the update / set case.


barentsen (approved):
Yes, this makes sense!  Thanks so much for thinking this through so carefully.  I agree with this change! :+1:

PR #871, opened 2020-10-07
bugfix Lightcurve column update - consistent result across column / attribute API

### LINKED ISSUE #866: Lightcurve2 inconsistent behavior in replacing a column
#### Problem description

When updating / replacing a a column, the behavior is different depending whether the replacement is done via column interface (`lc['flux'] = ...`) or attribute interface (`lc.flux = ...`)

|  Input type | Column interface | Attribute interface | `TimeSeries` |
| --- | --- | --- | --- |
|  `Quantity` | `Quantity` | `Quantity` | `Quantity` |
|  `list` | `Column` | `AttributeError` | `Column` |
|  `ndarray` | `Column` | `Quantity` | `Column` |

---

**Proposal:** piggyback on base `TimeSeries` behavior. If we do want to enforce additional behavior, it should be done for both attribute and column interfaces.

#### Example

```python
>>> import lightkurve as lk
>>> from astropy import units as u
>>> import numpy as np
>>>
>>> u_e_s = u.electron / u.second
>>> lc = lk.LightCurve(time=[1, 2, 3], flux=[0.98, 1.02, 1.03] * u_e_s)
>>>
>>>
>>> # replacing existing column using column interface
>>> lc['flux'] = np.array([1., 2., 3.])  # same as TimeSeries: a unitless Column object
>>> print(type(lc['flux']),  ' unit: ', lc['flux'].unit)
<class 'astropy.table.column.Column'>  unit:  None
>>>
>>> # replacing existing column using attribute interface
>>> lc.flux = np.array([1., 2., 3.])  # different from TimeSeries: a Quantity object with no unit
>>> print(type(lc['flux']),  ' unit: ', lc['flux'].unit)
<class 'astropy.units.quantity.Quantity'>  unit:
>>>
>>> lc['flux'] = [1., 2., 3.]   # same as TimeSeries: a unitless Column object
>>> print(type(lc['flux']),  ' unit: ', lc['flux'].unit)
<class 'astropy.table.column.Column'>  unit:  None
>>>
>>> lc.flux = [1., 2., 3.]   # different from TimeSeries: an error is raised. TimeSeries will accept it as a unitless Column object
  File "<stdin>", line 1, in <module>
  File "C:\dev\lightkurve\lightkurve\lightcurve.py", line 267, in __setattr__
    value = Quantity(value, dtype=value.dtype)
AttributeError: 'list' object has no attribute 'dtype'
>>>```

#### Environment

-  lightkurve version (e.g. 1.0b6): `master`, lightkurve2.0b2 based

### PR DESCRIPTION
**Pull request by orionlee**
Closes #866

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -265,8 +265,6 @@ class LightCurve(TimeSeries):
         elif (name == 'time'):
             self['time'] = value # astropy will convert value to Time if needed
         elif ('columns' in self.__dict__) and (name in self.__dict__['columns']):
-            if not isinstance(value, Quantity):
-                value = Quantity(value, dtype=value.dtype)
             self.replace_column(name, value)
         elif ('_meta' in self.__dict__):
             if (name in self.__dict__['_meta']):
--- a/lightkurve/tests/test_lightcurve.py
+++ b/lightkurve/tests/test_lightcurve.py
@@ -1142,6 +1142,26 @@ def test_attr_access_columns():
         lc.foo = 'bar2'
     assert len(warn_record) == 0
 
+u_e_s = u.electron / u.second
+
+
+@pytest.mark.parametrize("new_col_val", [([2, 3, 4] * u_e_s),   # Quantity
+                                         (np.array([2, 3, 4])), # ndarray
+                                         ([2, 3, 4])            # list
+                                         ])
+def test_attr_access_columns_consistent_update(new_col_val):
+    """Replace/Update a column: ensure consistent behavior across column API and attribute API"""
+
+    lc1 = LightCurve(time=Time([1, 2, 3], scale='tdb', format='jd'), flux=[4, 5, 6] * u_e_s)
+    lc1['flux'] = new_col_val
+
+    lc2 = LightCurve(time=Time([1, 2, 3], scale='tdb', format='jd'), flux=[4, 5, 6] * u_e_s)
+    lc2.flux = new_col_val
+
+    # ensure the result type is the same,
+    # irrespective whether the update is done via column API or attribute API
+    assert(type(lc1['flux']) is type(lc2['flux']))
+
 
 def test_attr_access_meta():
     """Test accessing meta values as attributes"""

### CODE REVIEW

**Comment by barentsen** [comment]
@orionlee Thanks for this!!

I played around with this, and it looks like the original reason for having `value = Quantity(value, dtype=value.dtype)` in `__setattr__` was to make sure that we could rely on being able to call `.value` on any column without having to check whether it is a Quantity.

For example, `lc.flux.value` appears 19 times in the code base (and `.value` is likely also accessed for other columns).  For example, `lc.plot()` accesses `flux.value`, so with this PR we get the following failure after updating a column with an array:
```python
import lightkurve as lk
import numpy as np
lc = lk.LightCurve(time=[1,2,3], flux=[4,5,6])
lc.plot()  # this works
lc.flux = np.array([7,8,9])
lc.plot()  # this fails
```

The reason the first call to `lc.plot()` does work in the example boave is that `LightCurve.__init__` currently forces all columns to be Quantity objects, here:

https://github.com/KeplerGO/lightkurve/blob/1d37eebc51e299b86e9f64dbb4921f54d8cdedbd/lightkurve/lightcurve.py#L232-L235

I guess we have two options:
1. Modify Lightkurve to check whether a column is indeed a Quantity before accessing its `.value` attribute, i.e. tolerate non-Quantity objects.
2. Figure out how to consistently convert lists and arrays into Quantity objects when a column is updated, both via the attribute and the column interface.

I think I'd favor option #2 because it'll make LightCurve objects easier to work with for both users and the package.  We already know how to enforce Quantity via the attribute interface, we'd just have to figure how to do it for the column interface.

orionlee:
I'm inclined to believe forcing columns to be of `Quantity` is better for Lightkurve users, mainly because the `Column` api and `Quantity` api isn't quite compatible: for accessing data, `Column` uses `.data` while `Quantity` uses `.value` .

If we commit to such a change, chances it could be implemented by overriding various add / replace columns methods (there are at least 4 upon a cursory glance).

One question that makes me pause though: why does the underlying astropy `TimeSeries` / `QTable` choose not to enforce such behavior (always using `Quantity`) [*]? Are there use cases that a `Column` rather than `Quantity` is needed that we have not considered? If so, are they applicable in Lightkurve's case?


[*] `TimeSeries` / `QTable` behavior: unitless column data stays as `Column` type, until `.unit` is assigned, then the column is changed to `Quantity`. The closest I found related to the motivation / use case is in [Quantity and QTable](https://docs.astropy.org/en/stable/table/mixin_columns.html#quantity-and-qtable).

barentsen:
> One question that makes me pause though: why does the underlying astropy TimeSeries / QTable choose not to enforce such behavior (always using Quantity) 

That's an excellent question!!  I did some digging in the AstroPy issue tracker and found two comments which explain the reasoning behind why QTable columns are not always a Quantity:

https://github.com/astropy/astropy/issues/4497#issuecomment-172088736
https://github.com/astropy/astropy/issues/4497#issuecomment-172119288

TL;DR non-numbers cannot be Quantity objects, so restricting a QTable to only have Quantity columns means we couldn't have columns containing strings etc.

This leads me to conclude that LightCurve objects should certainly support both `Column` and `Quantity` columns.

Going forward, I think we'll want to adapt Lightkurve to allow columns to be of unitless and of type `Column`. This means replacing `column.value` with `_to_unitless(column)` in a bunch of places. (Except for `time` which is enforced by `TimeSeries` to always be a `Time` object.)

We do have the option to give `flux` and `flux_err` a special status and enforce that they are always a Quantity.  It's not readily obvious how to implement that in an easy way however, whereas replacing `flux.value` with `_to_unitless(flux)` across the code may be fairly easy.
 
Any thoughts?

orionlee:
Ah, non-numeric data! 
- Something along the line of `_to_unitless()` should work. (1)
- I think making `flux` special (that requires to be `Quantity`) could lead to other edge cases from user perspective, e.g., if someone processes the flux and somehow end up with a unitless one that he/she wants to assign as `flux`.
- A bonus usability feature could be that if an existing `Quantity` column is replaced with unitless value, `Lightkurve` emits a warning. It would help to help developers to catch potential bugs.

(1) If `Column` and `Quantity` at least provide the same API to access raw data, we would not need the extra `_to_unitless()`. Or do they deliberately keep the API method different for some reasons?

barentsen:
> If Column and Quantity at least provide the same API to access raw data, we would not need the extra _to_unitless(). Or do they deliberately keep the API method different for some reasons?

Excellent question!  I just asked this exact question in the #table channel of the AstroPy slack. Stay tuned...

barentsen:
Good news: it looks like calling `np.asarray(Column)` or `np.asarray(Quantity)` has the same effect as calling `Column.data` and `Quantity.value`.  That offers another workaround!

orionlee:
I was hoping there'd some work to make the two API more uniform, e.g., add `.value` to `Column` as an alias to `.data`

Anyway, in our case, I suggest we should still have a `_to_unitless()` (that calls `np.asarray()`), as  it'd make the intent more explicit.

Regarding the actual changes, should be added to this PR or in a separate PR?
- This PR as-is has a consistent API. It doesn't support `plot()` when one assigns unitless `flux`.
- I can do make the actual changes next week.

barentsen:
> I was hoping there'd some work to make the two API more uniform, e.g., add .value to Column as an alias to .data

True!  There is an AstroPy issue open about this now: https://github.com/astropy/astropy/issues/10859

We'll want to go ahead with the `_to_unitless()` solution, because even if AstroPy API becomes more uniform, it would likely only be included in AstroPy v4.3 which is at least 6-12 months away.

> Regarding the actual changes, should be added to this PR or in a separate PR?

I guess we can go ahead and merge this change now, because it reflects the behavior we want.  I'll open a new issue re the need for `_to_unitless()`.

barentsen:
I opened #879 to keep track of the outstanding issue related to `.value`.  Thanks @orionlee!!

PR #878, opened 2020-10-12
tpf.interact() - option to show background mask

### LINKED ISSUE #877: tpf.interact() enhancement - support showing background_mask
#### Problem description

`TargetPixelFile.interact()` currently marks the aperture pixels with white lines. It'd be great if it has an option to mark the background mask pixels too.

#### Example

As a proposal, the background mask pixels are marked with dark gray lines, if the user specifies so in `interact()` arguments.
image

#### Environment

-  lightkurve version (e.g. 1.0b6): 1.11

orionlee:
Closes #877 

Currently it is a proof of concept. Hardcoded to show background mask.
If we agree to move forward, the codes will be tweaked to make it an option of `TargetPixelFile.interact()`

### DIFF
--- a/lightkurve/interact.py
+++ b/lightkurve/interact.py
@@ -392,6 +392,12 @@ def make_tpf_figure_elements(tpf, tpf_source, pedestal=None, fiducial_frame=None
         fig.rect('xx', 'yy', 1, 1, source=tpf_source, fill_color='gray',
                 fill_alpha=0.4, line_color='white')
 
+    show_background_mask = True
+    if show_background_mask and getattr(tpf, 'background_mask', None) is not None:
+        bkg_source = prepare_tpf_datasource(tpf, tpf.background_mask)
+        fig.rect('xx', 'yy', 1, 1, source=bkg_source, fill_color='gray',
+                 fill_alpha=0.0, line_color='#4f4f4f')
+
     # Configure the stretch slider and its callback function
     if scale == 'log':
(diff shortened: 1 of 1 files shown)


barentsen:
Thanks for trying this out @orionlee!

My thoughts:

A key feature of `interact()` is the ability to change the aperture mask by selecting pixels in the right hand side of the plot, which automatically updates the light curve on the left.

If we decide to also show the background mask on the right hand side, a user might reasonably expect that this background mask is used to background-correct the light curve.  That seems difficult to implement, e.g. I don't see an obvious way to let a user change the selection of both foreground vs background pixels without making the widget much more complicated.  Do you have any thoughts on this, @orionlee?

(ps: one can currently do `tpf.interact(aperture_mask=tpf.background_mask)`, but that of course shows the background mask *instead* of the foreground mask)

orionlee:
> If we decide to also show the background mask on the right hand side, a user might reasonably expect that this background mask is used to background-correct the light curve. That seems difficult to implement, e.g. I don't see an obvious way to let a user change the selection of both foreground vs background pixels without making the widget much more complicated. 

That's a good point. It could get confusing. Even the existing aperture mask selection is already a tad cryptic: it took me some search and trial-and-error to find out the way to deselect a pixel.

As you said, there is an existing alternative (`aperature_mask=tpf.background_mask`) to achieve similar results.

I'd be happy to drop this and close the issue.

PR #882, opened 2020-10-20
LightCurve handle unitless column

### LINKED ISSUE #879: Use `_to_unitless()` instead of `.value` to obtain unitless arrays
Over in #871 we realized that `LightCurve` objects must allow columns to be of type `Column` in addition to `Quantity` and `Time`.  This is inherent to AstroPy `QTable` objects which `LightCurve` sub-classes.  For example, string values can only be represented by a `Column` object.

This is important because `Column` objects do not have a `.value` attribute.  Several parts of the code (e.g. `lc.plot()`) currently make the assumption that e.g. `lc.flux.value` will always work.

Fix needed: we should replace occurrences of `column.value` with a more generic `_to_unitless(column)` function which returns `.value` if `column` is a `Quantity` or `Time`, and returns `.data` in the case of `Column`.

orionlee:
Closes #879 , #883

Steps:
- [x] make `LightCurve` plotting functions handle unitless flux
- [x] Other applicable `LightCurve` functions, including
   - [x] `flatten()`
   - [x] handle non-numeric columns in `copy` (used by `fill_gap()`)
   - [x] `fill_gap()`
   - [x] `plot_river()`
- [ ] functions outside `lightcurve.py`, including
  - [x] `interact.py` (for `tpf.interact()`)
  - [x] `cbvcorrector.py`
  - [ ] `pldcorrector.py`
  - [ ] `regressioncorrector.py`

Out of scope:
- `interact_bls.py` related: deferred to PR #852
- TargetPixelFile, e.g., `tpf.flux.value` - they might not need to be changed as tpf seems to come from FITS file only, always carrying units.

### DIFF
--- a/lightkurve/correctors/cbvcorrector.py
+++ b/lightkurve/correctors/cbvcorrector.py
@@ -19,7 +19,7 @@ from scipy.interpolate import PchipInterpolator
 from .designmatrix import DesignMatrix, DesignMatrixCollection
 from .. import MPLSTYLE
 from ..lightcurve import LightCurve
-from ..utils import channel_to_module_output, validate_method
+from ..utils import channel_to_module_output, validate_method, _to_unitless
 
 log = logging.getLogger(__name__)
 
@@ -27,7 +27,7 @@ __all__ = ['CotrendingBasisVectors', 'KeplerCotrendingBasisVectors',
            'TessCotrendingBasisVectors', 'download_kepler_cbvs', 'download_tess_cbvs']
 
 #*******************************************************************************
-# Cotrending Basis Vectors Classes and Functions 
+# Cotrending Basis Vectors Classes and Functions
 
 class CotrendingBasisVectors(TimeSeries):
     """
@@ -43,9 +43,9 @@ class CotrendingBasisVectors(TimeSeries):
     Instantiate multiple objects to store multiple set of CBVS, for example, to
     save each of the three multi-scale bands in TESS.
 
-    CotrendingBasisVectors calls the standard __init__ from 
+    CotrendingBasisVectors calls the standard __init__ from
     astropy.timeseries.TimeSeries
-    
+
     Parameters
     ----------
     data : `~astropy.table.Table`
@@ -55,13 +55,13 @@ class CotrendingBasisVectors(TimeSeries):
         If 'GAP' is not given then it is filled with all False.
         If 'CADENCENO' is not given then it is filled with np.arange(nCadences)
     time : `~astropy.time.Time`
-        Time values. 
+        Time values.
     **kwargs : dict
         Additional keyword arguments are passed to `~astropy.table.QTable`.
 
     Attributes
     ----------
-    cadenceno       : int array-like 
+    cadenceno       : int array-like
         Cadence indices
     time            : flaot array-like
         CBV cadence times
@@ -136,7 +136,7 @@ class CotrendingBasisVectors(TimeSeries):
         Parameters
         ----------
         cbv_indices : list of ints
-            List of CBV vectors to use. 1-based indexing! 
+            List of CBV vectors to use. 1-based indexing!
             {'all' => Use all}
         name : str
             A Name for the DesignMatrix
@@ -153,7 +153,7 @@ class CotrendingBasisVectors(TimeSeries):
 
         if (isinstance(cbv_indices, str) and (cbv_indices == 'all')):
             cbv_indices = self.cbv_indices
-                    
+
         cbv_names = []
         cbv_matrix = np.array([])
         for idx in cbv_indices:
@@ -163,7 +163,7 @@ class CotrendingBasisVectors(TimeSeries):
                 if len(cbv_matrix) == 0:
                     cbv_matrix =  np.array(self['VECTOR_{}'.format(idx)])[...,None]
                 else:
-                    cbv_matrix = np.hstack((cbv_matrix, 
+                    cbv_matrix = np.hstack((cbv_matrix,
                         np.array(self['VECTOR_{}'.format(idx)])[...,None]))
                 cbv_names.append('VECTOR_{}'.format(idx))
 
@@ -253,7 +253,7 @@ class CotrendingBasisVectors(TimeSeries):
     def align(self, lc):
         """Aligns the CBVs to a light curve. The lightCurve object might not
         have the same cadences as the CBVs. This will trim the CBVs to be
-        aligned with the light curve. 
+        aligned with the light curve.
 
         This method will use the cadence number (lc.cadenceno) to
         perform the synchronization. Only cadence numbers that exist in both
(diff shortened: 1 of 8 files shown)


orionlee:
Feedback needed before I proceed further.
- The commits so far make LightCurve plotting functions (`plot`, `scatter`, etc.) handle unitless flux.

Feedback needed - So far I make the code to convert columns and `Time` to unitless using the `_to_unitless()` wrapper. I am unsure if we should use  `_to_unitless()` on Time.
1. on the one hand, it could make the overall codes more uniform. 
2. on the other hand, it is not necessary. Practically, it would introduces more lines / files of changes that one need to review, especially when considering changes needed outside `lightcurve.py` (the correctors, etc.)

barentsen:
@orionlee I think we can rely on the fact that `lc.time` is always a `Time` object.  This is enforced by the `TimeSeries` parent class. 

So, I think it would be fine to keep using `lc.time.value` instead of `_to_unitless(time)`.  The time column really is "special" so I think it's ok to treat it in a different way.

barentsen:
(And having to change less code is always a good thing, of course!)

orionlee:
Feedback needed - The fix for #883 (make `copy` works for non-numeric columns) removed forced conversion to `Quantity` during `__init__()`

Because of it, some columns become type `Column` by default. For typical TESS LC files, `cadenceno` and `quality` become type `Column`. 

Question: should we consider to make them `Quantity` in typical use cases (reading a FIT file or supplied to `LightCurve` constructor)?
- irrespective of the decision, Lightkurve internal codes need to be prepared to handle `Quantity` and `Column`. So there is no effect from that perspective.
- an argument for making them be `Quantity` in typical cases is that for Lightkurve users will not need to deal with the difference between `Column` and `Quantity` in typical uses.

barentsen:
@orionlee My initial reaction is that both `cadenceno` or `quality` are really unitless numbers, so perhaps it does make more sense to have them be `Column` in the first place, but perhaps I'm overlooking a situation where it will come to bite us.

Fortunately, I *think* we can always modify the reader functions in `lightkurve.io` if we wanted to enforce making a certain column a `Quantity`, which makes me think that it should be ok to remove the forced conversion in the constructor.

Does that make sense?

orionlee:
I'm in favor forcing `.cadenceno` and `quality` to `Quantity` the lightcurve comes from a FIT file [*].

My concern is if we leave `.cadenceno` and `quality` as `Column` in typical usage, we'd increase Lightkurve users' burden in using the API, making them less productive. 

Once users are used to V2 API, they would learn the typical way to access column data, e.g., accessing raw data with `.value` when needed. But if the user need to access `quality` or `cadenceno`, they could hit a surprise that the API is different. 

The primary difference is accessing raw values (`.value` vs `.data`). But there could be also other differences that we have not yet considered.

E.g., I learned the hard way the for `Time`, `np.min(lc.time)` works but is slow, and should be replaced with `lc.time.min()`. While this particular case does not apply to `Quantity`, I don't know if other annoyance could creep up, and might be different depending on whether it's `Quantity` or `Column`.

[*] There might be other similar cases that we need to check, e.g., `tpf.to_lightcurve()`.

barentsen:
@orionlee This sounds like a good plan.

I just verified that if we remove the forced Column-to-Quantity conversion in `LightCurve.__init__()`, and instead add the following line inside the `read_generic_lightcurve()` function in `lightkurve/io/generic.py`:
```python
tab['quality'].unit = u.dimensionless_unscaled
```
...then the quality column automatically becomes a `Quantity` object instead of `Column`.

So, it looks like we can indeed make users' lives easier by ensuring all numeric columns read in from light curve files or created via `to_lightcurve()` are Quantities, even if the constructor does not enforce this.

Going forward, it looks like AstroPy is happy to add a `.value` attribute to the `Column` class  (cf. https://github.com/astropy/astropy/issues/10859#issuecomment-708731285) and it also looks like the issues with e.g. `np.min(lc.time)` is close to being resolved (cf. https://github.com/astropy/astropy/pull/10889).  It may be a while before those improvements are available in stable AstroPy releases however, so I do think it is a good idea to ensure e.g. `quality` and `cadenceno` are of type `Quality` for now.

Thanks for thinking about this issue!! 🙏

orionlee:
1. The `cadenceno` as `Quantity` issue is tracked in #886 .
2. it occurred to me that while `Column` type is needed to handle non-numeric columns, we can possibly force all numeric columns as `Quantity`.

What made me hesitate is I wonder why astropy `QTable` does not enforce so. Having said that, astropy's rationale may or may not be applicable to our case.

Sample codes to check if a column is of numeric type:
```python
ts = astropy.timeseries.TimeSeries(time=astropy.time.Time([1, 2], format='jd'))
ts['flux'] = [4.1, 5.2]
dtype = ts['flux'].dtype
print(dtype, np.issubdtype(dtype, np.number))

ts['cadenceno'] = [1, 2]
dtype = ts['cadenceno'].dtype
print(dtype, np.issubdtype(dtype, np.number))

ts['col1'] = ['a', 'b']
dtype = ts['col1'].dtype
print(dtype, np.issubdtype(dtype, np.number))

ts['col2'] = [1, 'b']
type = ts['col2'].dtype
print(dtype, np.issubdtype(dtype, np.number))

```

Output:
```python
float64 True
int32 True
<U1 False
<U1 False
```

barentsen:
I had a look at the changes in this PR so far and they all look good!

I agree that we don't need to modify the use of `tpf.flux`.  Right now the `TargetPixelFile` class is not designed to be created in any other way than by reading in a FITS file, which guarantees the fact that `tpf.flux` will always be a Quantity.

Ping me when the PR is ready for final review! 👍

orionlee:
What do you think about the idea that `LightCurve` objects force all numeric columns to be `Quantity`?

See: https://github.com/KeplerGO/lightkurve/pull/882#issuecomment-713829822

If we adopt such behavior, this PR will be replaced.

barentsen:
> What do you think about the idea that LightCurve objects force all numeric columns to be Quantity

Oops, sorry I missed your important comment earlier.

I guess our goal is to stay as close to the Table/TimeSeries API as possible, albeit making changes where we feel the benefits outweigh the costs.

So, what are the benefits and costs associated with enforcing all numeric columns to be `Quantity`?

Benefits:
* We can avoid having to use `_to_unitless`.  This is a significant benefit because it means we don't have to worry about making sure that all future modifications of Lightkurve use `_to_unitless`.  It also means users are less likely to run into the column-vs-quantity confusion.

Costs:
* We slightly alter the Table API.  When I investigated this topic in the AstroPy issue tracker, I found good reasons to understand why not *every* column can be a Quantity (e.g. Quantity objects do not support strings), but I did not encounter good reasons as to why numeric columns shouldn't always be Quantity.

My conclusion: I think the benefits outweigh the costs.  I very much likely the idea of enforcing numeric columns to be Quantity.

However: do you think it is easy to implement this behavior for the case of adding a new column, or updating an existing one?

orionlee:
- I want to double check, in case enforcing `Quantity` on all numeric columns will be problematic for some use cases haven't been considered. So it sounds like we are all right on that front.
- In terms of implementation, I suspect it won't be complicated: overriding all add/replace column/columns methods (3 in total).

barentsen:
> I want to double check, in case enforcing Quantity on all numeric columns will be problematic for some use cases haven't been considered.

I don't expect problems, because we have effectively tested this behavior for the past several months by converting *all* columns (which is effectively the same as converting only the numeric ones, because kepler/tess products typically only have numeric columns).

barentsen:
I will close this issue because I believe it has been resolved by merging #889.

Thanks so much for working on this @orionlee! 🙏

Issue #883, opened 2020-10-20
Lightkurve2 - lc.copy() failed when there are non-numeric columns


orionlee:
#### Problem description
`lc.copy()` failed when there are non-numeric columns. It should work.

#### Example

```python
import lightkurve as lk
lc = lk.LightCurve(time=[1,2,3])
lc['col1'] = ['a', 'b', 'c']
lc_copy = lc.copy()
```

Error:
```python
...
c:\dev\lightkurve\lightkurve\lightcurve.py in __init__(self, data, time, flux, flux_err, *args, **kwargs)
    240         for col in self.columns:
    241             if not isinstance(self[col], (Quantity, Time)):
--> 242                 self.replace_column(col, Quantity(self[col], dtype=self[col].dtype))
    243 
    244         # Ensure flux and flux_err have the same units
...
TypeError: The value must be a valid Python or Numpy numeric type.
```

#### Environment

-  lightkurve version (e.g. 1.0b6): `master`, `2.0b3` based
-  installation method (e.g. pip, conda, source): source

barentsen:
After merging #889, I verified that the failing code snippet runs without exception now:
```python
import lightkurve as lk
lc = lk.LightCurve(time=[1,2,3])
lc['col1'] = ['a', 'b', 'c']
lc_copy = lc.copy()
```

A regression test covering this example was added as part of the new `test_support_non_numeric_columns` test which was added via #889, so I think we can close this issue.

Thanks @orionlee!! :+1:

PR #889, opened 2020-10-26
Force all numeric columns as Quantity

### LINKED ISSUE #879: Use `_to_unitless()` instead of `.value` to obtain unitless arrays

orionlee:
Closes #879, #883 
Replaces  #882 

Notes:
- `QTable`'s codes pointed out that for a column of `MaskedColumn` type, it'll drop the mask. If we force all numeric columns as `Quantity`, we effectively do not support `MaskedColumn` for any numeric type.
  - I believe it is acceptable, as the essential columns such as `flux` have the same behavior.
- the implementation overrides a `QTable` non-public method `_convert_col_for_table()`. 
  - I'm a tad uneasy, as it can be changed more easily that breaks our codes. But I think it's unlikely. 
  - I think it's still cleaner than overriding multiple public methods.

### DIFF
--- a/lightkurve/lightcurve.py
+++ b/lightkurve/lightcurve.py
@@ -14,7 +14,7 @@ from matplotlib import pyplot as plt
 from copy import deepcopy
 
 from astropy.stats import sigma_clip
-from astropy.table import Table
+from astropy.table import Table, Column
 from astropy.io import fits
 from astropy.time import Time, TimeDelta
 from astropy import units as u
@@ -36,7 +36,23 @@ __all__ = ['LightCurve', 'KeplerLightCurve', 'TessLightCurve', 'FoldedLightCurve
 log = logging.getLogger(__name__)
 
 
-class LightCurve(TimeSeries):
+class QTimeSeries(TimeSeries):
+    def _convert_col_for_table(self, col):
+        """Force all numeric columns as Quantity, not just those with units.
+
+           Override grandparent ``QTable`` behavior.
+        """
+        col = super()._convert_col_for_table(col)
+        if (isinstance(col, Column) and getattr(col, 'unit', None) is None and
+                np.issubdtype(getattr(col, 'dtype', None), np.number)):
+            # the logic is based on those in the grandparent QTable
+            qcol = Quantity(col.data, u.dimensionless_unscaled, dtype=col.dtype, copy=False)
+            qcol.info = col.info
+            col = qcol
+        return col
+
+
+class LightCurve(QTimeSeries):
     """Class to hold time series brightness data for an astronomical object.
 
     Compared to the generic `~astropy.timeseries.TimeSeries` class, `LightCurve`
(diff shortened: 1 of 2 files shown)


lightkurve/lightcurve.py
```diff
@@ -36,7 +36,23 @@
...
+           Override grandparent ``QTable`` behavior.
+        """
+        col = super()._convert_col_for_table(col)
+        if (isinstance(col, Column) and getattr(col, 'unit', None) is None and
+                np.issubdtype(getattr(col, 'dtype', None), np.number)):
+            # the logic is based on those in the grandparent QTable
+            qcol = Quantity(col.data, u.dimensionless_unscaled, dtype=col.dtype, copy=False)
+            qcol.info = col.info
+            col = qcol
+        return col
```
barentsen, line 52: This looks like a very clean and short solution! :+1:

lightkurve/tests/test_lightcurve.py
```diff
@@ -1299,3 +1313,21 @@ def test_fill_gaps_after_normalization():
...
+
+
+@pytest.mark.parametrize("new_col_val", [([2, 3, 4] * u_e_s),   # Quantity
+                                         (np.array([2, 3, 4])), # ndarray
+                                         ([2, 3, 4])            # list
+                                         ])
+def test_force_numeric_columns_as_quantity(new_col_val):
+    lc = LightCurve(time=[1, 2, 3])
+    lc['col1'] = new_col_val
+    assert isinstance(lc['col1'], Quantity)
```
barentsen, line 1325: Note: the tests are failing here with `NameError: name 'Quantity' is not defined`.  This will be easy to fix!

barentsen:
> we effectively do not support MaskedColumn for any numeric type.
> I believe it is acceptable, as the essential columns such as flux have the same behavior.

Agreed.  I believe efforts are being made on the AstroPy issue tracker to add support for masking in QTable i.e. Quantity columns, so hopefully we will be able to use this in the future. (It would be great to have masking so that we can make the quality-code based masking of LightCurve objects more explicit.)

As you point out, not having masking is fine for now, because if it doesn't work for the `flux` column then it might as well not work for any column.

barentsen:
This PR looks good to me!  We'll just have to fix the tiny `NameError` in the test and this should be good to go. :+1:

Thanks so much!!

lightkurve/tests/test_lightcurve.py
orionlee: Fixed. It's embarrassing - I failed to notice the test error in my local build after the last round of test tweak.

barentsen: No worries!  That happens to me all the time!  That's why we have GitHub CI to protect us from ourselves 😅

barentsen:
All looks good now!  I intend to go ahead and merge this PR in the next hour! :+1:

PR #891, opened 2020-10-28
Bugfix for #884: do not allow `stitch()` to change Quantity data type

### LINKED ISSUE #884: KeplerLightCurve.to_fits() doesn't work with Lightkurve.read() for stitched data.
The code I've used is as follows, and the error is below:

```python3
import lightkurve
sr = lk.search_lightcurve('KIC 1435467', mission='Kepler', cadence='long')
lc = sr[2:5].download_all().stitch()
lc.to_fits('test_llc.fits', overwrite=True)

reread = lk.read('test_llc.fits')

TypeError                                 Traceback (most recent call last)
/Users/Anaconda/anaconda3/lib/python3.8/site-packages/astropy/timeseries/sampled.py in read(self, filename, time_column, time_format, time_scale, format, *args, **kwargs)
    365             # First we try the readers defined for the BinnedTimeSeries class
--> 366             return super().read(filename, format=format, *args, **kwargs)
    367 

/Users/Anaconda/anaconda3/lib/python3.8/site-packages/astropy/table/connect.py in __call__(self, *args, **kwargs)
     59 
---> 60         out = registry.read(cls, *args, **kwargs)
     61 

/Users/Anaconda/anaconda3/lib/python3.8/site-packages/astropy/io/registry.py in read(cls, format, cache, *args, **kwargs)
    519         reader = get_reader(format, cls)
--> 520         data = reader(*args, **kwargs)
    521 

/Users/Anaconda/anaconda3/lib/python3.8/site-packages/lightkurve/io/kepler.py in read_kepler_lightcurve(filename, flux_column, quality_bitmask)
     42     # of masking. Details: https://github.com/astropy/astropy/issues/10119
---> 43     quality_mask = KeplerQualityFlags.create_quality_mask(
     44                                 quality_array=lc['sap_quality'],

/Users/Anaconda/anaconda3/lib/python3.8/site-packages/lightkurve/utils.py in create_quality_mask(cls, quality_array, bitmask)
    100         # The bitmask is applied using the bitwise AND operator
--> 101         quality_mask = (quality_array & bitmask) == 0
    102         # Log the quality masking as info or warning

TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

<ipython-input-14-99b1bd25d84d> in <module>
      4 lc.to_fits('test_llc.fits', overwrite=True)
      5 
----> 6 reread = lk.read('test_llc.fits')

/Users/Anaconda/anaconda3/lib/python3.8/site-packages/lightkurve/io/read.py in read(path_or_url, **kwargs)
     88 
     89     if filetype == "KeplerLightCurve":
---> 90         return KeplerLightCurve.read(path_or_url, format='kepler', **kwargs)
     91     elif filetype == "TessLightCurve":
     92         return TessLightCurve.read(path_or_url, format='tess', **kwargs)

/Users/Anaconda/anaconda3/lib/python3.8/site-packages/lightkurve/lightcurve.py in read(cls, *args, **kwargs)
   2301         if kwargs.get("format") is None:
   2302             kwargs['format'] = "kepler"
-> 2303         return super().read(*args, **kwargs)
   2304 
   2305     def to_fits(self, path=None, overwrite=False, flux_column_name='FLUX',

/Users/Anaconda/anaconda3/lib/python3.8/site-packages/astropy/timeseries/sampled.py in read(self, filename, time_column, time_format, time_scale, format, *args, **kwargs)
    371 
    372             if time_column is None:
--> 373                 raise ValueError("``time_column`` should be provided since the default Table readers are being used.")
    374 
    375             table = Table.read(filename, format=format, *args, **kwargs)

ValueError: ``time_column`` should be provided since the default Table readers are being used.
```

This error doesn't occur when reading in the saved `.fits` file for a single quarter alone.

I'm using MacOS 10.15.6, Astropy version 4.1rc2 and Lightkurve 2.0b3, installed from the repository.

barentsen:
The `LightCurveCollection.stitch()` is unexpectedly changing the data type of integer `Quantity` columns from `int` into `float`.  This is because the underlying `astropy.table.vstack` operation displays this behavior (see https://github.com/astropy/astropy/issues/10958)

For example:
Screen Shot 2020-10-28 at 10 33 37

This PR implements a workaround for this behavior in an attempt to fix #884:

Screen Shot 2020-10-28 at 10 34 36

However, perhaps an alternative solution would be to always use `Column` instead of `Quantity` for integer columns. (/cc @orionlee)

### DIFF
--- a/lightkurve/collections.py
+++ b/lightkurve/collections.py
@@ -139,7 +139,17 @@ class LightCurveCollection(Collection):
             corrector_func = lambda x: x
         lcs = [corrector_func(lc) for lc in self]
         # Need `join_type='inner'` until AstroPy supports masked Quantities
-        return vstack(lcs, join_type='inner', metadata_conflicts='silent')
+        stack = vstack(lcs, join_type='inner', metadata_conflicts='silent')
+
+        # Temporary workaround for https://github.com/astropy/astropy/issues/10958
+        # AstroPy `vstack` converts integer Quantity columns to floating point,
+        # e.g. the `cadenceno` and `quality` columns.  This causes issues with
+        # the __repr__ and the parsing of quality values.
+        for col in stack.colnames:
+            if hasattr(stack[col], 'dtype') and stack[col].dtype != lcs[0][col].dtype:
+                stack[col] = stack[col].astype(lcs[0][col].dtype)
+
+        return stack
 
     def plot(self, ax=None, offset=0., **kwargs) -> matplotlib.axes.Axes:
         """Plots all light curves in the collection on a single plot.
--- a/lightkurve/collections.py
+++ b/lightkurve/collections.py
@@ -8,6 +8,7 @@ import numpy as np
 
 from astropy.table import vstack
 from astropy.utils.decorators import deprecated
+from astropy.units import Quantity
 
 from . import MPLSTYLE
 from .targetpixelfile import TargetPixelFile
@@ -146,7 +147,7 @@ class LightCurveCollection(Collection):
         # e.g. the `cadenceno` and `quality` columns.  This causes issues with
         # the __repr__ and the parsing of quality values.
         for col in stack.colnames:
-            if hasattr(stack[col], 'dtype') and stack[col].dtype != lcs[0][col].dtype:
+            if isinstance(stack[col], Quantity) and stack[col].dtype != lcs[0][col].dtype:
                 stack[col] = stack[col].astype(lcs[0][col].dtype)
 
         return stack
--- a/lightkurve/tests/test_collections.py
+++ b/lightkurve/tests/test_collections.py
@@ -5,6 +5,7 @@ import numpy as np
 from numpy.testing import assert_array_equal
 
 from ..lightcurve import LightCurve
+from ..search import search_lightcurve
 from ..targetpixelfile import KeplerTargetPixelFile
 from ..collections import LightCurveCollection, TargetPixelFileCollection
 
@@ -106,3 +107,12 @@ def test_tpfcollection_plot():
     coll = TargetPixelFileCollection([tpf])
     coll.plot()
     plt.close('all')
+
+
+@pytest.mark.remote_data
+def test_stitch_repr():
+    """Regression test for #884."""
+    lc = search_lightcurve("Pi Men", mission='TESS', sector=1).download()
+    # The line below used to raise `ValueError: Unable to parse format string
+    # "{:10d}" for entry "70445.0" in column "cadenceno"`
+    LightCurveCollection((lc,lc)).stitch().__repr__()


orionlee:
I am unsure about this. I suppose we have to ask if `Quantity` is meant to work for integers.
- It was raised in https://github.com/astropy/astropy/pull/10950#issuecomment-717531619 that there might be reasons that one wants to keep dtype as float for `Quantity`.
- if it is meant to work for integers, but has not been previously used, I suppose we have to ask if we're going to bite the bullet and effectively hash out the rough edges.

I feel we have 2 choices:
1. keep the recent behavior of forcing numeric columns as `Quantity`, and work out the bugs / workarounds. This path is viable only if it's determined `Quantity` is meant to work for integers.
2. Retain TimeSeries default behavior. For `Lightkurve`, that'd mean `cadenceno` and `quality` are `Column` type.
   a. in this case, it'd make life easier if `.value` is added to `Column`. But my understanding is that such change can't be released any time soon.
3. Arguably, we could possibly still force all float type columns as `Quantity`. I don't know how useful it'd be, however. It won't change anything for the default TESS (and probably Kepler) columns. `Lightkurve` users would still need to reckon with `Column` type.

barentsen:
@orionlee I think we can implement option #2 in your list as follows:

1. We go ahead and add the `.value` attribute to `Column` inside AstroPy.  Over in AstroPy issue #10859 the maintainer of `astropy.table` seems OK with this.
2. As a temporary workaround, we adapt `QTimeSeries._convert_col_for_table` to make sure all non-Quantity columns in a LightCurve object are of type `QColumn` rather than `Column`, defined as follows:
```
class QColumn(Column):

    @property
    def value(self):
        return self.data
```
3. Whenever `column.value` is available in a mainstream release of AstroPy, we can remove both `QTimeSeries` and `QColumn` from Lightkurve and make that mainstream release a minimum dependency.

orionlee:
@barentsen Your suggestion of keeping TimeSeries behavior while adding a `QColumn` temporarily sounds promising. 1 note and 1 potential concern.
1. We also need a `QMaskedColumn`, TESS cadenceno and quality are both read as MaskedColumn
2. the potential concern: if `Lightkurve` add a column with their own custom `Column` subclass, then we probably shouldn't wrap around it. If so, the following codes could still break, even though it seems to be highly unlikely in practice.
```python
  lc.flux = UserCustomColumn(...)
  lc.scatter() # the flux column has no .value, being a custom Column
```

orionlee:
The concern for custom `Column` class in conjunction with plotting is alleviated.
- I just checked, and matplotlib can handle both `Quantity` and `Column`, without using raw data.

```
# plot Quantity directly
ax.scatter(lc1.time.value, lc1.flux)

# plot Column directly
ax.scatter(lc1.time.value, Column(lc1.flux.value, unit=lc1.flux.unit, name='flux'))
```

- I don't know if such support requires specific version of matplotlib / astropy. The versions I tried is `3.1.0` / `4.1rc1` respectively

- `Lightkurve`'s plotting codes, e.g., `LightCurve._create_plot()`, still needs to be modified to take advantage of it.

barentsen:
Closing this PR as the issue has been resolved by #895, which includes this PR's unit test! 🎉 

Thanks @orionlee!

PR #893, opened 2020-11-02
bugfix: tpf.wcs when expected headers do not exist

### LINKED ISSUE #892: TargetPixelFile WCS extraction possibly failed when header is missing
The `tpf.wcs` would raise `KeyError` if the requested key (`oldkey`) is missing in the TPF (line 378 below):

https://github.com/KeplerGO/lightkurve/blob/6265d26655f2bc44c5803d6e12501a3a9490d424/lightkurve/targetpixelfile.py#L376-L380

I believe the line is meant to be:
```python
        if (self.hdu[1].header.get(oldkey, None) is not None):
```

Affected version: `master`, Lightkurve 1 (it has the same code upon inspection)

orionlee:
Closes #892

I can't quite find an actual TPF that would trigger the problem (tried a recent TESSCut, tpf from sector 28 and sector 1). 
- Good news: the bug probably doesn't affect any typical usage.
- Bad news: I can't quite create a test for it. I first stumbled upon it when using another package (that copied the codes from lightkurve), but didn't quite have the tpf file that caused the error anymore (the package didn't save the tpf).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -59,6 +59,9 @@ lightkurve.targetpixelfile
 - Modified `TargetPixelFactory` to support creating TESS Target Pixel Files
   and to enable it to populate all data columns. [#768, #857]
 
+- Fixed a bug in ``TargetPixelFile.wcs`` which caused it to raise Error if
+the tpf does not contain expected WCS keywords in the header. [#892]
+
 lightkurve.search
 ^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


barentsen (approved):
Thanks @orionlee.  I am happy to merge this because `header.get()` really is the correct way to deal with potentially-missing keys.  The use of `.wcs` is tested in a number of places in `test_targetpixelfile.py`, so we can be fairly confident that this change does not break existing behavior.  :+1:

PR #905, opened 2020-11-10
interact_bls enhancement: wheel zoom and normalize 

### LINKED ISSUE #854: interact_bls() enhancement: wheel_zoom and normalize the LC
#### Problem description

I'd like to propose to minor tweaks to `interact_bls()`
1. Support wheel zoom: right now if one zooms in too much, one can only reset to zoom out.

2. Always normalize the lightcurve: Right now if the end users do not normalize the lightcurve before calling `interact_bls`, the resulting graph lightcurve plots aren't quite helpful, because the modeled lightcurve is always normalized. One can't really see both the lightcurve and the model easily.

image

#### Environment

-  lightkurve version (e.g. 1.0b6): 1.11

orionlee:
Closes #854 

Two minor enhancement:
1. Normalize the lightcurve if needed so that its scale matches the generated transit model LC.
2. Added zoom by scrolling mouse wheel for all 3 plots, e.g.
image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,11 @@ lightkurve.lightcurve
 - Added a ``column`` parameter to ``LightCurve.remove_nans()`` to enable
   cadences to be removed which contain NaN values in a specific column. [#828]
 
+- ``interact_bls()``: added the support zoom by scrolling mouse wheel. [#854]
+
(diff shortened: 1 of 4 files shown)


barentsen (approved):
Excellent!  Thanks @orionlee!! :+1:

PR #906, opened 2020-11-11
tpf.plot_pixels() misc. enhancement

### LINKED ISSUE #894: Lightkurve2 tpf.plot_pixels() misc. enhancement
#### Problem description
I've been using `Lightkurve2`'s `tpf.plot_pixels()`, and would suggest the following minor enhancement:

1. Include the time range, row /column numbers in the default plot output
2. when using `show_flux=True`, use black dots when the background is light (yellow or green) so that they are easier to spot.
3. ~~add a optional parameter, `pixel_size_inch`, so that~~ users can control the size of the diagram (currently it's hardcoded as 1.5 inch) - it can be done when `plot_pixels()` honors the optional `ax` parameter. The codes currently do not yet honor it.

Example plot with (1) and (2) adopted (from Nora Eisner's LATTE):

image

#### Environment

-  lightkurve version (e.g. 1.0b6): `master` (lightkurve2 beta)

orionlee:
Closes #894 

- [x] Include the time range
- [x] Include row /column numbers 
- [ ] when `show_flux=True`, make scatters dots easier to be spotted when the background is light (yellow/green)
- [x] honor `ax` parameter, use it (and its predefined size) to output the plot

--- 
Example plot so far:

image

When `ax` is given, from a figure with `figsize=(8,8)`
image

### DIFF
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -1552,15 +1552,20 @@ class TargetPixelFile(object):
                         pixel_list.append(lc)
 
         with plt.style.context(style):
-            fig = plt.figure()
-            if ax is None:  # Configure axes if none is given
+            if ax is None:
+                fig = plt.figure()
                 ax = plt.gca()
-                ax.get_xaxis().set_ticks([])
-                ax.get_yaxis().set_ticks([])
-                if periodogram:
-                    ax.set(title=title, xlabel='Frequency / Column (pixel)', ylabel='Power / Row (pixel)')
-                else:
-                    ax.set(title=title, xlabel='Time / Column (pixel)', ylabel='Flux / Row (pixel)')
+                set_size = True
+            else:
+                fig = ax.get_figure()
+                set_size = False
+
+            ax.get_xaxis().set_ticks([])
+            ax.get_yaxis().set_ticks([])
+            if periodogram:
+                ax.set(title=title, xlabel='Frequency / Column (pixel)', ylabel='Power / Row (pixel)')
+            else:
+                ax.set(title=title, xlabel='Time / Column (pixel)', ylabel='Flux / Row (pixel)')
 
             gs = gridspec.GridSpec(self.shape[1], self.shape[2], wspace=0.01, hspace=0.01)
 
@@ -1608,7 +1613,8 @@ class TargetPixelFile(object):
                     if x == self.shape[1] - 1 and y == 0:  # upper left
                         gax.set_ylabel(f'{self.row + self.shape[1] - 1}')
 
-            fig.set_size_inches((y*1.5, x*1.5))
+            if set_size:  # use default size when caller does not supply ax
+                fig.set_size_inches((y*1.5, x*1.5))
 
         return ax
 
--- a/lightkurve/targetpixelfile.py
+++ b/lightkurve/targetpixelfile.py
@@ -47,6 +47,20 @@ __all__ = ['KeplerTargetPixelFile', 'TessTargetPixelFile']
 log = logging.getLogger(__name__)
 
 
+# OPEN: consider to move to utils and
+# consolidate with the helper in lightcurve.py (for time label)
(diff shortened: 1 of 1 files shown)


orionlee:
For making scatters dots easier to be spotted when the background is light (yellow/green) when `show_flux=True`. I tried a few different approaches with example output below.

@barentsen @astrobel Thoughts?

<details>
<summary>white-with-black-outline markers trial</summary>
Given how dense the dots are, the end result looks like gray dots instead.

For example, if I change the plot from the current
```python
   # current style: dot with white color
  #  marker='.', color=markercolor, ms=0.5, lw=0)
   # white-with-black-outline markers trial
   marker='o', color='black', ms=1.5, lw=0, markeredgecolor='white', markeredgewidth='0.5')
```

the plot looks like the following (with 1.3 day of a typical TESS 2-minute cadence TPF):
image

I tried tweaking the markersize, and markeredgewith, and end up with different gray colors.
</details>

---
<details>
<summary>Increase marker size</summary>

Another option is to make the markers bigger. The resulting plots are easier to be spotted on yellow background.

The following plot is made by increasing markersize from the current `0.5` to `1.5`, with 3 days of TESS data.

image

One possible downside is when the dots are even more dense(e.g., increase the time window for the plot or make the plot smaller), the dots would overwhelm the plots

image

</details>

---

<details>
<summary>Use black dots in lightgreen / yellow background</summary>

Example plots:
image

For periodogram (line plots): 
image


</details>

barentsen:
@orionlee Thank you for trying out different approaches!   They all definitely improve the contrast.

It occurs to me that white dots do seem to work well for the vast majority of the default `viridis` color map, right up until where the color map turns yellow.  Would it make sense to solve our contrast problem by truncating `viridis` to avoid using yellow?

A hacky way to implement this would be to modify lines 1510-1511 in targetpixelfile.py to say e.g.

```python
 norm = plt.Normalize(vmin=np.nanmin(self.flux[0].value), 
                      vmax=np.nanmax(self.flux[0].value)*1.1) 
```

instead of

```python
 norm = plt.Normalize(vmin=np.nanmin(self.flux[0].value), 
                      vmax=np.nanmax(self.flux[0].value)) 
```

astrobel:
Thanks for these examples, @orionlee! It seems to me that white with black outlines is too visually confusing. I'm in favour of truncating `viridis`, as per @barentsen's suggestion, and in that case I would suggest the middle option, increased marker size. I'm hesitant to go for changing the markers to black on lighter colours as it draws the eye, and implies some additional significance to the light curves/periodograms in those pixels.

orionlee:
Here is what the same plot looks with hacking the colormap (1.3 day of data), retaining current markersize (0.5).

image

It does improve the contrast to the point that is legible.
---
Concerns:
1. It's possibly surprising / confusing to users, as it breaks the convention of the colormap. Assuming the users visualize the TPF with other methods, e.g., `plot`, `interact()`, etc., they would be used to seeing the brightest pixel (often the target) as yellow. The colormap hack would break the convention and be confusing.
2. The dots are more legible, but still a tad difficult (at least to my eyes).
  i. the issue affect the brightest pixels. In typical cases, that'd be what an user is interested in, as it's where the target is.
  ii. the problem is more severe when users zoom in to a small time window such that the dots would be less dense. 

Example of when I zoom in the plot to 0.5 day window:

image

barentsen:
Thanks for trying it out @orionlee! I agree with what you found & wrote.

I'd be fine with either of the three solutions you identified. Perhaps my vote is on "white-with-black-outline markers" because it seems to be the most robust and easy (?) solution, but I'm happy to go with any of them.

barentsen:
Although we didn't come to a conclusion on how to improve the scatter points, the other improvements in this PR are very valuable, so I intend to merge them tomorrow unless there is objection!

barentsen:
I went ahead and merged this PR so it can be part of the v2.0.0 release -- I hope we can make the necessary changes to the scatter style soon.

PR #913, opened 2020-11-25
Add support for the QLP & SPOC FFI light curve products

### LINKED ISSUE #912: HLSP TESS light curves are not found by search_lightcurvefile
The lightkurve search command does not find some TOIs like 980.01, 1187.01, 2173.01... is there a reason why?

Thank you

barentsen:
Fixes #912.

Both the MIT Quicklook Pipeline (QLP) and the Ames pipeline (SPOC) recently started delivering FFI-based light curve products and target pixel files as High Level Science Products (HLSP) to the data archive at MAST.  It is important for Lightkurve to support searching and reading these valuable products!

The following changes are required:
* Going forward, the `search_lightcurve` function will now return MAST portal products with filename suffix "lc.fits".  Previously it was looking for the string "Light Curve" appearing in the product description field, but this is not the case for these HLSPs.
* The `search_targetpixelfile` function will now return products with filename suffix "tp.fits" or "targ.fits.gz".  In the past it was looking for the string "Target Pixel" in the description field.

Note: to search for SPOC FFI light curves, users will have to pass `author="TESS-SPOC"`, whereas short cadence light curves will be known under `author="SPOC"`. Demo:

Screen Shot 2020-11-25 at 12 30 03

This is a Work In Progress (WIP) because I have yet to test downloading and opening these products.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -90,6 +90,8 @@ lightkurve.search
   such as 'KIC 5112705' or 'TIC 261136679', only return products known under
   those names, unless a search radius is specified. [#796]
 
+- Added support for the QLP and SPOC FFI HLSP light curve products.
+
 lightkurve.correctors
 ^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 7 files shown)


barentsen:
Update: it is now possible to search and download both SPOC and QLP light curves based on FFI data!

Screenshot:
Screen Shot 2020-11-25 at 17 12 39

Issue #914, opened 2020-12-01
Obtaining cached files via `SearchResult.download_all()` is unexpectedly slow


barentsen:
I find that loading 10 cached light curves using `download_all` is very slow.

The first call to `download_all` is of course very slow (25.7s) because 10 light curves have to be downloaded.  For example:

```python
>>> import lightkurve as lk
>>> search = lk.search_lightcurve("Pi Men", sector=1, radius=6000, limit=10)
>>> %time collection = search.download_all()
CPU times: user 4.41 s, sys: 344 ms, total: 4.75 s
Wall time: 25.7 s
```

This is entirely expected.  However, subsequent calls should be blazing fast because the files are now cached on disk.  Instead, they still take 7-8 seconds of wall time to run, which is unexpected:

```python
>>> %time collection = search.download_all()
CPU times: user 3.49 s, sys: 103 ms, total: 3.59 s
Wall time: 6.86 s
>>> %time collection = search.download_all()
CPU times: user 4.55 s, sys: 169 ms, total: 4.72 s
Wall time: 8.71 s
```

Some detailed profiling is required to understand why obtaining *cached* files using `download_all()` is so slow!

barentsen:
I line-profiled `download_all()` and concluded that the majority of the time appears to be lost inside the function that reads and parses a generic light curve FITS file, i.e. `lightkurve.io.generic.read_generic_lightcurve()`.  In other words, the slow performance does not appear to be due to the download caching mechanism.

We can likely optimize this function:

```python
>>> %lprun -f lk.io.generic.read_generic_lightcurve search[0].download()
Timer unit: 1e-06 s

Total time: 0.701839 s
/Users/gb/dev/lightkurve/lightkurve/io/generic.py
Function: read_generic_lightcurve at line 16

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    16                                           def read_generic_lightcurve(filename, flux_column,
    17                                                                       quality_column='quality',
    18                                                                       cadenceno_column='cadenceno',
    19                                                                       centroid_col_column='mom_centr1',
    20                                                                       centroid_row_column='mom_centr2',
    21                                                                       time_format=None,
    22                                                                       ext=1):
    23                                               """Generic helper function to convert a Kepler ot TESS light curve file
    24                                               into a generic `LightCurve` object.
    25                                               """
    26         1          5.0      5.0      0.0      if isinstance(filename, fits.HDUList):
[...]

PR #915, opened 2020-12-01
Speed up opening cached light curve files by a factor ~5x

barentsen:
As reported in #914, obtaining light curves via `download_all()` was surprisingly slow, even what all the files are available in the local cache.

Using `line_profiler`, I determined the primary culprit to be the fact that `astroquery.mast.Observations.download_products()` always submits a HTTP request to MAST to determine the length of the file prior to checking if the file already exists in the local cache.  This PR modifies `download()` and `download_all()` to skip that request.

This PR also includes a few easy speed-ups inside`lightkurve.io.generic.read_generic_lightcurve()`.

As a result, the time required to open 10 light curves (cf. example in #914) dropped from ~8.5s to ~1.5s, which is a ~5x improvement:

**Before:**
```python
>>> %time collection = search.download_all()
CPU times: user 4.55 s, sys: 169 ms, total: 4.72 s
Wall time: 8.71 s
```

**After:**
```python
>>> %time collection = search.download_all()
CPU times: user 1.26 s, sys: 100 ms, total: 1.36 s
Wall time: 1.56 s
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -93,6 +93,9 @@ lightkurve.search
 - Added support for searching and reading QLP and SPOC Full Frame Image (FFI)
   light curves available as High Level Science Products from MAST. [#913]
 
+- Improved the performance of `download()` operations by checking if a file
+  exists in local cache prior to contacting MAST.
+
 lightkurve.correctors
 ^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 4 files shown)


barentsen:
I expect to merge this PR after fixing the Windows CI failure and adding a unit test.

PR #929, opened 2021-01-12
Ensure SPOC FFI products show up in search operations by default

barentsen:
This PR proposes to show TESS FFI-based data products produced by the NASA SPOC pipeline by default in the result of search operations.   It was already possible to show these products by passing `author="TESS-SPOC"` or `author="any"` to the search functions, but this PR adds TESS-SPOC to the list of authors that are shown by default.  See examples below.

An outstanding question is whether or not *all* HLSP products should be shown by default.  I'd welcome any and all opinions on this!  There is a trade-off between making sure all options are visible on one hand, and not overwhelming users on the other hand.

## Current behavior

Screen Shot 2021-01-12 at 9 18 49 AM

## New behavior after this PR

Screen Shot 2021-01-12 at 9 18 26 AM

## Open question: should all HLSPs be shown by default?

Many more products could be shown by showing all products by default, at the risk of overwhelming the user.  What is the right thing to do?

Screen Shot 2021-01-12 at 9 24 42 AM

### DIFF
--- a/lightkurve/search.py
+++ b/lightkurve/search.py
@@ -430,7 +430,7 @@ class SearchResult(object):
 @cached
 def search_targetpixelfile(target, radius=None, cadence=None,
                            mission=('Kepler', 'K2', 'TESS'),
-                           author=('Kepler', 'K2', 'SPOC'),
+                           author=('Kepler', 'K2', 'SPOC', 'TESS-SPOC'),
                            quarter=None, month=None, campaign=None, sector=None,
                            limit=None):
     """Searches the `public data archive at MAST <https://archive.stsci.edu>`_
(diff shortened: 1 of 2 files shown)


mrtommyb:
I think they should probably be shown by default, because this is the only way to get lightcurves for many targets. However, we should certainly document how to only get the specific data a user wants

kcolon:
I agree with Tom - all should be shown by default since there could be cases where QLP exists but SPOC does not, or vice versa, or another TBD HLSP has light curves where others do not. I can see it could be overwhelming but can also prevent people from missing light curves that do exist?

barentsen:
@mrtommyb @kcolon Thank you for sharing your thoughts! 🙏   I will go ahead and create an alternative PR which implements the "show all light curves by default" strategy.  I think I agree with making that the default behavior, but I'll implement and test it first to be sure. Stay tuned!

barentsen:
I opened #933 to change the behavior such that *all* TESS and Kepler data products available at MAST will show up by default.

An open question is how the products ought to be sorted, i.e. who gets to appear on top.  I'd appreciate any input on that.

In meanwhile, this PR can be closed without being merged.  Thanks all for the feedback!

Issue #924, opened 2020-12-23
Feature request: aperture photometry with medians


avanderburg:
#### Problem description
I would like to request a feature that I don't think currently exists in lightkurve. In the "extract_aperture_photometry" method, I would like for there to be an option to take the median value of each pixel within the aperture, rather than summing the pixel values within the aperture. This will mainly be useful when extracting sky background light curves, and (for example) an asteroid passes nearby the star. When the background is calculated with a sum, that can artificially increase the measured background, and introduce transit-like dips into the target light curve. But a median will usually avoid this problem. I hope this is the right place to submit this request! 

https://docs.lightkurve.org/api/lightkurve.targetpixelfile.TessTargetPixelFile.html#lightkurve.targetpixelfile.TessTargetPixelFile.extract_aperture_photometry

#### Example
N/A

I think all that would need to be added would be a line here (if a certain keyword is given) to replace np.nansum with np.nanmedian 
https://github.com/KeplerGO/lightkurve/blob/9b36c395d1524127cc94e83d775575a7ab229301/lightkurve/targetpixelfile.py#L795

#### Expected behavior
N/A

#### Environment
N/A

barentsen:
Thank you for the request @avanderburg! :pray:

I think this is an excellent idea!  Stay tuned...

rebekah9969:
Hello @avanderburg 

Thank you for your inquiry. We have now incorporated a flux_method function into Lightkurve version 2.0.
This allows the user to specify if they want the sum, median, or mean flux of an aperture. 
You would use it in a method similar to that shown below, 

`lc =  tpf.to_lightcurve(aperture_mask=aper, flux_method="mean")`

The default is set to sum the flux in the aperture if you did not want to specify the method. 

Thank you for request.

avanderburg:
Thanks so much!!

barentsen:
Thanks again for the suggestion @avanderburg! 🙏 

This feature is part of the 2.0beta5 release of Lightkurve.  You can try it out using `pip install lightkurve --pre --upgrade` (the `--pre` flag is essential to get a beta version).

There are a bunch of other changes in 2.0 however.  These will be summarized in a What's New guide which I expect to be available next week, so you may prefer to wait for that. 

Thanks again!

PR #932, opened 2021-01-20
new median flux function

rebekah9969:
I have added a new variable for using the median flux instead of the sum in an aperture - addressing issue #924


barentsen:
Thanks for this PR @rebekah9969!  It is looking great!!

When making a change to the API, I always ask myself: "Does this change make future changes easier or harder?".  In this case, it is useful to consider whether we may want to add support for even more ways to combine the flux in an aperture in the future.

For example, one may want to...
* sum the flux across all pixels (`flux_method="sum"`) -- this is the current default;
* take the median pixel value (`flux_method="median"`);
* take the mean pixel value (`flux_method="mean"`);
* apply some custom function which rejects outliers (e.g. `flux_method=my_custom_function`).

For this reason, I wonder if it is better to introduce a more generic `flux_method` keyword rather than the `flux_median` keyword.  To get started, `flux_method` could be documented to accepted the string values (`"sum"`, `"median"`, or `"mean"`).  It would only require a fairly minor modification of your PR to make this work. What do you think?

rebekah9969:
I have introduced a new flux_mode replacing flux_median -  targetpixel code was updated along with the test

lightkurve/targetpixelfile.py
```diff
@@ -1740,7 +1762,7 @@ def mission(self):
         """'Kepler' or 'K2'. ('MISSION' header keyword)"""
         return self.get_keyword('MISSION')
 
-    def extract_aperture_photometry(self, aperture_mask='default', centroid_method='moments'):
+    def extract_aperture_photometry(self, aperture_mask='default', flux_mode='default', centroid_method='moments'):
```
barentsen: I recommend making `flux_mode="sum"` the default.  This would make the default behavior explicit, and avoids having two options for the same behavior (i.e. `flux_mode="sum"` and `flux_mode="default"` would do the same).

lightkurve/tests/test_targetpixelfile.py
```diff
@@ -734,3 +735,20 @@ def test_estimate_background():
...
+
+def test_fluxmode():
+    """This should verify the median flux use in an aperture"""
+    tpf = read(filename_fluxmode)
+    aper = np.zeros(tpf.shape[1:], dtype=bool)
+    aper[4:8, 4:7] = True
+    lc_n = tpf.extract_aperture_photometry(aperture_mask=aper, centroid_method='moments')
+    lc_sum = tpf.extract_aperture_photometry(aperture_mask=aper, flux_median="sum", centroid_method='moments')
+    lc_med = tpf.extract_aperture_photometry(aperture_mask=aper, flux_median="median", centroid_method='moments')
+    lc_mean = tpf.extract_aperture_photometry(aperture_mask=aper, flux_median="mean", centroid_method='moments')
```
barentsen: You'll want to replace `flux_median` with `flux_method` in the lines above.  The tests are likely failing because of this.

rebekah9969:
Changed flux_mode to flux_method for constancy

lightkurve/targetpixelfile.py
```diff
@@ -1740,7 +1762,7 @@ def mission(self):
         """'Kepler' or 'K2'. ('MISSION' header keyword)"""
         return self.get_keyword('MISSION')
 
-    def extract_aperture_photometry(self, aperture_mask='default', centroid_method='moments'):
+    def extract_aperture_photometry(self, aperture_mask='default', flux_method='default', centroid_method='moments'):
```
barentsen, line 1765: I recommend making `flux_method="sum"` the default on this line.  This makes the default behavior explicit, and removes duplicate behavior (i.e. right now `flux_method="sum"` and `flux_method="default"` would do the same thing, better to just offer one way).

lightkurve/targetpixelfile.py
```diff
@@ -2240,7 +2263,7 @@ def ccd(self):
     def mission(self):
         return 'TESS'
 
-    def extract_aperture_photometry(self, aperture_mask='default', centroid_method='moments'):
+    def extract_aperture_photometry(self, aperture_mask='default', flux_method='default', centroid_method='moments'):
```
barentsen, line 2266: I'd change `flux_method='default'` into `flux_method='sum'` here as well.

lightkurve/targetpixelfile.py
barentsen: Note: in the docstring below, we are going to want to add an entry for "flux_method" in the "Parameters" section. The entry should mention that three strings can be passsed, i.e. 'sum', 'median', and 'mean'.

lightkurve/targetpixelfile.py
```diff
@@ -792,7 +792,19 @@ def _aperture_photometry(self, aperture_mask, centroid_method='moments'):
...
+        elif flux_method == "median":
+            flux = np.nanmedian(self.flux[:, apmask], axis=1)
+            
+        elif flux_method == "mean":
+            flux = np.nanmean(self.flux[:, apmask], axis=1)
+            
+        #Want default flux to be sum if not specified
+        #In the future we may wish to add a user specified function 
+        else:
+            flux = np.nansum(self.flux[:, apmask], axis=1)
```
barentsen, line 807: If any other value is passed to `flux_method`, the user is using the function incorrectly. We should let them know by raising an error.  I recommend doing so by changing line 807 into this:
```suggestion
            raise ValueError("`flux_method` must be one of 'sum', 'median', or 'mean'.")
```

lightkurve/targetpixelfile.py
```diff
@@ -776,7 +776,7 @@ def _estimate_centroids_via_quadratic(self, aperture_mask):
         row_centr = Quantity(row_centr, unit='pixel')
         return col_centr, row_centr
 
-    def _aperture_photometry(self, aperture_mask, centroid_method='moments'):
+    def _aperture_photometry(self, aperture_mask, flux_method, centroid_method='moments'):
```
barentsen, line 779: For backwards compatibility, it is best to also assign a default in the helper function as follows:
```suggestion
    def _aperture_photometry(self, aperture_mask, flux_method="sum", centroid_method='moments'):
```

barentsen (changes requested):

lightkurve/targetpixelfile.py
barentsen: This is outdated -- sorry.

lightkurve/tests/test_targetpixelfile.py
barentsen: This is outdated -- ignore.

barentsen:
@rebekah9969 I left 5 minor comments above which should be easy to change.

One final request: would you be willing to update the `CHANGES.rst` file in the root folder of the repository to describe this change?  If you open that file it will be fairly self-explanatory to see how to record a change like this.

lightkurve/targetpixelfile.py
rebekah9969: Done

lightkurve/tests/test_targetpixelfile.py
rebekah9969: Done

lightkurve/targetpixelfile.py
rebekah9969: First part is complete.
Where are the docstrings for me to edit.

lightkurve/targetpixelfile.py
rebekah9969: Done

lightkurve/targetpixelfile.py
barentsen: The docstring is the block of comments that immediately follow this line, i.e. immediately below "def extract_aperture_photometry(...)".  There is a section in that block of comments called "Parameters" which we need to update if we add parameters.

lightkurve/targetpixelfile.py
rebekah9969: Done

lightkurve/targetpixelfile.py
rebekah9969: Done

rebekah9969:
Made adjustments to the code based on comments.

lightkurve/targetpixelfile.py
rebekah9969: updated.

rebekah9969:
Made changes to CHANGES.rst file  to reflect new flux_method variable.

rebekah9969:
Changed file use in test to reduce size of files

rebekah9969:
simplified test

rebekah9969:
> @rebekah9969 I left 5 minor comments above which should be easy to change.
> 
> One final request: would you be willing to update the `CHANGES.rst` file in the root folder of the repository to describe this change? If you open that file it will be fairly self-explanatory to see how to record a change like this.

Done

barentsen (approved):
@rebekah9969 Congratulations on finishing this PR complete with unit test, docs update, and changelog entry.  That's awesome!  Merging now! :+1:

Issue #918, opened 2020-12-03
Add support for additional TESS HLSP light curves: DIAMANTE, PATHOS, CDIPS, TASOC


barentsen:
This is a follow-up of #912.

Several excellent sets of TESS community light curves are now available at MAST as High Level Science Products.  We already support searching and opening SPOC and QLP products via Lightkurve's `search_lightcurve` function, but it would be very valuable to also support the following recent HLSPs which provide light curve files:
* DIAMANTE
* PATHOS
* CDIPS
* TASOC

The following steps are needed to add support for any of these
1. Add a custom reader function for each product type as `lightkurve/io/*NAME*.py`.
2. Register the custom reader in `lightkurve/io/__init__.py`.
3. Update `lightkurve/io/detect.py` and `lightkurve/io/read.py` to detect the products correctly.
4. Add unit tests.

Help is welcome in the form of PRs.  One for each product, please!

stephtdouglas:
I'm starting to work with the CDIPS and PATHOS light curves, so I can try to add the code for both of those, if nobody else is working on them yet

barentsen:
@stephtdouglas Nobody has started working on this yet, and it would be really cool to add this to Lightkurve.  I would be delighted to review and merge a PR if you were interested in preparing one! 👍 

The existing readers located under `lightkurve/io/xxxxx.py` should give you a pretty good example of what the readers might looks like (assuming they are FITS files as well?)  Let me know if you have questions!

stephtdouglas:
@barentsen I have the PATHOS version done, with unit tests written (https://github.com/stephtdouglas/lightkurve/tree/pathos_reader), but I'm having trouble running pytest on my local machine for some reason. If I open a pull request, will that trigger continuous integration testing?

barentsen:
@stephtdouglas  Yep!  When you open a PR, the tests will run and you will see the results online.  You're welcome to use that mechanism before completing work on a branch.  In that case, people often put "[WIP]" in the PR title, i.e. "Work In Progress".

Alternatively, I'm happy to help debug pytest if you copy/paste the error.

stephtdouglas:
[...]

PR #935, opened 2021-01-27
PATHOS HLSP reader

stephtdouglas:
This partially addresses #918 , and adds code to search for and read TESS PATHOS HLSP light curves. 

1. Adds a custom reader function as lightkurve/io/pathos.py.
2. Registers the custom reader in lightkurve/io/__init__.py.
3. Updates lightkurve/io/detect.py and lightkurve/io/read.py to detect PATHOS products correctly.
4. Adds unit tests.

I'm submitting the PR now to run the unit tests (I'm having some pytest issues on my local machine)

### DIFF
--- a/lightkurve/io/detect.py
+++ b/lightkurve/io/detect.py
@@ -48,7 +48,7 @@ def detect_filetype(hdulist: HDUList) -> str:
     # The 'pathos' name doesn't stay in the filename if when we use fits.open
     # to download a file, so we have to check for all the important columns
     # This will cause problems if another HLSP has the exact same colnames...
-    if all(x in f[1].columns.names for x in ["PSF_FLUX_RAW", "PSF_FLUX_COR", "AP4_FLUX_RAW", "AP4_FLUX_COR", "SKY_LOCAL"]):
+    if all(x in hdulist[1].columns.names for x in ["PSF_FLUX_RAW", "PSF_FLUX_COR", "AP4_FLUX_RAW", "AP4_FLUX_COR", "SKY_LOCAL"]):
         return "PATHOS"
 
     # Is it a K2VARCAT file?
--- a/lightkurve/io/detect.py
+++ b/lightkurve/io/detect.py
@@ -45,7 +45,7 @@ def detect_filetype(hdulist: HDUList) -> str:
 
     # Is it a PATHOS TESS light curve?
     # cf. http://archive.stsci.edu/hlsp/pathos
-    if "pathos" in hdulist[0].header.get('origin', '').lower():
+    if "pathos" in (hdulist.filename() or ""):
         return "PATHOS"
 
     # Is it a K2VARCAT file?
--- a/lightkurve/io/detect.py
+++ b/lightkurve/io/detect.py
@@ -22,6 +22,7 @@ def detect_filetype(hdulist: HDUList) -> str:
         * `'K2SC'`
         * `'K2VARCAT'`
         * `'QLP'`
+        * `'PATHOS'`
 
     If the data product cannot be detected, `None` will be returned.
 
@@ -36,10 +37,16 @@ def detect_filetype(hdulist: HDUList) -> str:
         A string describing the detected filetype. If the filetype is not
         recognized, `None` will be returned.
     """
+
     # Is it a MIT/QLP TESS FFI Quicklook Pipeline light curve?
     # cf. http://archive.stsci.edu/hlsp/qlp
     if "mit/qlp" in hdulist[0].header.get('origin', '').lower():
         return "QLP"
+
+    # Is it a PATHOS TESS light curve?
+    # cf. http://archive.stsci.edu/hlsp/pathos
+    if "pathos" in hdulist[0].header.get('origin', '').lower():
+        return "PATHOS"
 
     # Is it a K2VARCAT file?
     # There are no self-identifying keywords in the header, so go by filename.
--- a/lightkurve/io/detect.py
+++ b/lightkurve/io/detect.py
@@ -45,7 +45,10 @@ def detect_filetype(hdulist: HDUList) -> str:
 
     # Is it a PATHOS TESS light curve?
     # cf. http://archive.stsci.edu/hlsp/pathos
-    if "pathos" in (hdulist.filename() or ""):
+    # The 'pathos' name doesn't stay in the filename if when we use fits.open
+    # to download a file, so we have to check for all the important columns
+    # This will cause problems if another HLSP has the exact same colnames...
+    if all(x in f[1].columns.names for x in ["PSF_FLUX_RAW", "PSF_FLUX_COR", "AP4_FLUX_RAW", "AP4_FLUX_COR", "SKY_LOCAL"]):
         return "PATHOS"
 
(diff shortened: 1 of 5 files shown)


lightkurve/io/tests/test_pathos.py
```diff
@@ -0,0 +1,51 @@
+import pytest
+
+from astropy.io import fits
+import numpy as np
+from numpy.testing import assert_array_equal
+
+from ... import search_lightcurve
+from ..pathos import read_pathos_lightcurve
+from .detect import detect_filetype
```
barentsen, line 9: Oops, you'll want to add a dot in front of detect to drop down a directory:
```suggestion
from ..detect import detect_filetype
```

lightkurve/io/detect.py
```diff
@@ -36,11 +37,17 @@ def detect_filetype(hdulist: HDUList) -> str:
...
     """
+
     # Is it a MIT/QLP TESS FFI Quicklook Pipeline light curve?
     # cf. http://archive.stsci.edu/hlsp/qlp
     if "mit/qlp" in hdulist[0].header.get('origin', '').lower():
         return "QLP"
 
+    # Is it a PATHOS TESS light curve?
+    # cf. http://archive.stsci.edu/hlsp/pathos
+    if "pathos" in (hdulist.filename() or ""):
```
barentsen, line 48: Oops, it looks like we can't rely on `pathos` appearing in the filename when a file is opened using astropy's `fits.open()`, because astropy renames the file using a hash.  For example, `hdulist.filename()` is equal to `"/Users/gb/.astropy/cache/download/url/bff14286987df5550091e2ad8c8dae9b/contents"` when I run the `test_detect_pathos` test on my system. 

This makes life difficult because there are no obvious header keywords in PATHOS light curves that we can use to detect the product.  My best suggestion is to identify the product based on the fingerprint of column names?  For example, we could try to detect the product as follows:
```suggestion
    if all(x in f[1].columns.names for x in ["PSF_FLUX_RAW", "PSF_FLUX_COR", "AP4_FLUX_RAW", "AP4_FLUX_COR", "SKY_LOCAL"]):
```

We'd have to cross our little fingers that no other product will ever use the same combination of column names, but it's the only solution I can think of.

(If we do it this way, we'll want to explain this problem in a little comment.)

lightkurve/io/pathos.py
```diff
@@ -0,0 +1,60 @@
...
+                                 flux_column=flux_column.lower(),
+                                 time_format='btjd',
+                                 quality_column="DQUALITY")
+
+    # Filter out poor-quality data
+    # NOTE: Unfortunately Astropy Table masking does not yet work for columns
+    # that are Quantity objects, so for now we remove poor-quality data instead
+    # of masking. Details: https://github.com/astropy/astropy/issues/10119
+    quality_mask = TessQualityFlags.create_quality_mask(
+                                quality_array=lc['quality'],
```
barentsen, line 49: You'll want to change this to use the `dquality` column:
```suggestion
                                quality_array=lc['dquality'],
```

lightkurve/io/tests/test_pathos.py
```diff
@@ -0,0 +1,51 @@
...
+    # Verify different extensions
+    fluxes = []
+
+    exts = ['PSF_FLUX_RAW', 'PSF_FLUX_COR']
+    exts.extend([f'AP{ap}_FLUX_RAW' for ap in [1,2,3,4]])
+    exts.extend([f'AP{ap}_FLUX_COR' for ap in [1,2,3,4]])
+
+    for ext in exts:
+        lc = read_pathos_lightcurve(url, flux_column=ext)
+        assert type(lc).__name__ == "TESSLightCurve"
```
barentsen, line 34: This minor typo will make the test fail:
```suggestion
        assert type(lc).__name__ == "TessLightCurve"
```

lightkurve/io/tests/test_pathos.py
```diff
@@ -0,0 +1,51 @@
...
+        assert not np.array_equal(fluxes[i] , fluxes[i+1])
+
+@pytest.mark.remote_data
+def test_search_pathos():
+    """Can we search and download a PATHOS light curve?"""
+    search = search_lightcurve("TIC 93270923", author="PATHOS", sector=8)
+    assert len(search) == 1
+    assert search.table["author"][0] == "PATHOS"
+    lc = search.download()
+    assert type(lc).__name__ == "TESSLightCurve"
```
barentsen, line 50: This typo will make the test fail as well:
```suggestion
    assert type(lc).__name__ == "TessLightCurve"
```

lightkurve/io/tests/test_pathos.py
```diff
@@ -0,0 +1,51 @@
...
+
+    exts = ['PSF_FLUX_RAW', 'PSF_FLUX_COR']
+    exts.extend([f'AP{ap}_FLUX_RAW' for ap in [1,2,3,4]])
+    exts.extend([f'AP{ap}_FLUX_COR' for ap in [1,2,3,4]])
+
+    for ext in exts:
+        lc = read_pathos_lightcurve(url, flux_column=ext)
+        assert type(lc).__name__ == "TESSLightCurve"
+        # Are `time` and `flux` consistent with the FITS file?
+        assert_array_equal(f[1].data['TIME'], lc.time.value)
```
barentsen, line 36: You'll have to apply the quality mask for this test to pass:
```suggestion
        assert_array_equal(f[1].data['TIME'][lc.meta['QUALITY_MASK']], lc.time.value)
```

lightkurve/io/tests/test_pathos.py
```diff
@@ -0,0 +1,51 @@
...
+        assert_array_equal(f[1].data[ext], lc.flux.value)
```
barentsen, line 37: Same here:
```suggestion
        assert_array_equal(f[1].data[ext][lc.meta['QUALITY_MASK']], lc.flux.value)
```

barentsen (changes requested):
Thanks @stephtdouglas!!  You did a great job adding support for PATHOS.  It looks like the tests are failing for a few complicated/tricky reasons.  I took the freedom to go ahead and leave comments with the likely fixes that are needed.  I hope that helps!

lightkurve/io/tests/test_pathos.py
stephtdouglas: @barentsen thanks! I was really stuck on these tbh. I'll work on incorporating the edits tomorrow.

stephtdouglas:
@barentsen As best I can tell, the failing tests that are left aren't related to my new code - there seem to be some tests on other parts of lightkurve that fail (but aren't resulting in a failed check), and the Windows test that's failing seems to be a larger issue with files on Windows (e.g., https://github.com/pytest-dev/pytest/issues/14, https://github.com/pywinauto/pywinauto/issues/858). It's affecting all of the data download tests in that subdirectory. 

So I think this is ready to merge, unless you spot any other issues or want to try to find a workaround for the file error.

barentsen:
I tend to release either Lightkurve v2.0rc1 and/or v2.0.0 to pypi later this week, which will include this feature.

Issue #927, opened 2021-01-12
New lightkurve API can't bin folded, user generated data


christinahedges:
I'm trying to fold and bin a user generated light curve

The following code produces an error message that is hard to debug
```python
lk.LightCurve(np.arange(10), np.ones(10)).fold(1).bin(3)
```
produces
```python
ValueError: Input values did not match the format class jd:
TypeError: for jd class, input should be doubles, string, or Decimal, and second values are only allowed for doubles.
```
#### Environment

-  platform (e.g. Linux, OSX, Windows): OSX
-  lightkurve version (e.g. 1.0b6): 2.0b4
-  installation method (e.g. pip, conda, source): source

barentsen:
Thanks so much for reporting @christinahedges!  This looks like a real bug.

I am able to reproduce the crash, though I see a different error message:
```python
>>> lk.LightCurve(np.arange(10), np.ones(10)).fold(1).bin(3)
IndexError: index -1 is out of bounds for axis 0 with size 0
```

It looks like the error is triggered by the behavior of this edge case in AstroPy:
```python
>>> from astropy.time import Time
>>> from astropy import units as u
>>> from astropy.timeseries import aggregate_downsample, TimeSeries
>>> ts = TimeSeries(time=Time([0], format='jd'))
>>> aggregate_downsample(ts, time_bin_size=1*u.day)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-84-44ec4dc9a4a0> in <module>
      4 
      5 ts = TimeSeries(time=Time([0], format='jd'))
----> 6 aggregate_downsample(ts, time_bin_size=1*u.day)

~/Library/Caches/pypoetry/virtualenvs/lightkurve-HqDeGDfc-py3.8/lib/python3.8/site-packages/astropy/timeseries/downsample.py in aggregate_downsample(time_series, time_bin_size, time_bin_start, n_bins, aggregate_func)
    100 
    101     # Create new binned time series
--> 102     binned = BinnedTimeSeries(time_bin_start=bins[:-1], time_bin_end=bins[-1])
    103 
    104     # Determine rows where values are defined

~/Library/Caches/pypoetry/virtualenvs/lightkurve-HqDeGDfc-py3.8/lib/python3.8/site-packages/astropy/timeseries/binned.py in __init__(self, data, time_bin_start, time_bin_end, time_bin_size, n_bins, **kwargs)
    160                     times = time_bin_start.copy()
[...]

PR #953, opened 2021-02-18
:bug: fixing issue #927 when binning a folded light curve

jorgemarpa:
Here's an easy fix to this issue (#927)
This works when binning folded and non-folded light curves, for the cases when `time_bin_start` is passed to `bin()` as a `float` or as a `Time`/`TimeDelta` class for non-folded/folded light curves. 
Also added a simple testing function.


barentsen:
Thank you for figuring this out @jorgemarpa. LGTM! 🎉

PR #955, opened 2021-02-18
Adjustment to 1.1 Tutorials

rebekah9969:
Several of the tutorials have functions bit with no link to the API. I have added in these links where appropriate.


barentsen:
Thanks @rebekah9969!! 👍 

Would you be willing to click "clear outputs" in the notebooks that are changed, and commit them again? We can only store non-executed notebooks in git, otherwise the size of the repository would run quickly out of control.

rebekah9969:
Cleaned output from the files.

barentsen:
Oops, when running `make html` in the `docs` directory, I find that some of the links you added do not render correctly in the HTML version of the docs.  For example, some of the links you added to the "What are LightCurve objects?" tutorial look like this:

Screen Shot 2021-02-18 at 8 25 53 PM

I bet these links rendered correctly when you added them in the Jupyter notebook.  However, when the notebooks are compiled into the html webpages, a complicated conversion takes place (markdown => rst => html).  This conversion does not allow the text of a link to contain backticks.

For example, to avoid the problem shown in the screenshot above, you'd have to make sure the notebook uses ```[flatten()]``` instead of ```[`flatten()`]``` to create the link for flatten.

Sorry about that!  You couldn't have known about this.  It's one of those software things...

Would you be willing to fix?

rebekah9969:
Adjusted hyperlinks

barentsen:
Thanks, this looks good now! :+1:

PR #966, opened 2021-02-22
Tutorial2.1

rebekah9969:
Added and fixed hyperlinks


barentsen (approved):
This is great stuff @rebekah9969!

Thank you for catching the `table['observation'] ` => `table['mission']` change, and for making sure the output has been removed. You rock! :+1:

PR #969, opened 2021-02-22
:sparkles: Add support for KEPSEISMIC

barentsen:


barentsen:
This PR doesn't pass the tests yet, in part because the KEPSEISMIC data products at MAST appear to have `sequence_number=17` (i.e. `quarter=17`) set in the Portal API data, which seems inaccurate because these light curves cover all quarters.  I reached out to Scott to understand if this is expected.

Issue #967, opened 2021-02-22
No support for KEPSEISMIC lightcurves


warrickball:
#### Problem description

Possibly more of a feature request than a bug but my searches through Lightkurve don't return any lightcurves from the KEPSEISMIC HLSP, although the corresponding MAST searches do.

#### Example

```python
import lightkurve

print(lk.search_lightcurve('KIC 1292147', mission='Kepler')) # only contains author 'Kepler'
```
produces
```
SearchResult containing 17 data products.

 #       mission      year author exptime  target_name  distance
                                     s                   arcsec 
--- ----------------- ---- ------ ------- ------------- --------
  0 Kepler Quarter 01 2009 Kepler    1800 kplr001292147      0.0
  1 Kepler Quarter 02 2009 Kepler    1800 kplr001292147      0.0
  2 Kepler Quarter 03 2009 Kepler    1800 kplr001292147      0.0
  3 Kepler Quarter 04 2010 Kepler    1800 kplr001292147      0.0
  4 Kepler Quarter 05 2010 Kepler    1800 kplr001292147      0.0
  5 Kepler Quarter 06 2010 Kepler    1800 kplr001292147      0.0
  6 Kepler Quarter 07 2010 Kepler    1800 kplr001292147      0.0
  7 Kepler Quarter 08 2011 Kepler    1800 kplr001292147      0.0
  8 Kepler Quarter 09 2011 Kepler    1800 kplr001292147      0.0
  9 Kepler Quarter 10 2011 Kepler    1800 kplr001292147      0.0
 10 Kepler Quarter 11 2012 Kepler    1800 kplr001292147      0.0
 11 Kepler Quarter 12 2012 Kepler    1800 kplr001292147      0.0
 12 Kepler Quarter 13 2012 Kepler    1800 kplr001292147      0.0
 13 Kepler Quarter 14 2012 Kepler    1800 kplr001292147      0.0
 14 Kepler Quarter 15 2013 Kepler    1800 kplr001292147      0.0
 15 Kepler Quarter 16 2013 Kepler    1800 kplr001292147      0.0
 16 Kepler Quarter 17 2013 Kepler    1800 kplr001292147      0.0
```
i.e. nothing with author KEPSEISMIC.

#### Expected behavior
[...]

PR #972, opened 2021-02-23
Add support for reading KEPSEISMIC data products

warrickball:
Here goes! This PR adds functions to read KEPSEISMIC data products. This partly resolves issue #967 by allowing the files to be read. PR #969 covers the question of returning the data products in the search.

### DIFF
--- a/src/lightkurve/io/__init__.py
+++ b/src/lightkurve/io/__init__.py
@@ -2,7 +2,7 @@
 from .detect import *
 from .read import *
 
-from . import kepler, tess, qlp, k2sff, everest, pathos, tasoc
+from . import kepler, tess, qlp, k2sff, everest, pathos, tasoc, kepseismic
 from .. import LightCurve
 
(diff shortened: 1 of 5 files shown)


barentsen (approved):
Thanks so much @warrickball.  This looks perfect!! :+1:

I didn't appreciate the kepseismic files only have two columns, that makes life easy =)

Sanity check, reading in a kepseismic file using this PR:

Screen Shot 2021-02-23 at 7 45 00 AM

barentsen:
Note: the LightCurve constructor enforces the presence of a `flux_err` column (populated with NaNs by default), which is why the Kepseismic LightCurve object shows one, despite the absence of such a column in the FITS file.  This enforcement enables users to assume that `time`, `flux`, and `flux_err` columns always exist, which simplifies a lot of code.

I will go ahead and merge and add an entry to `CHANGES.rst` in a separate commit.  Thanks again @warrickball!! :pray:

PR #945, opened 2021-02-11
[WIP] CDIPS reader

stephtdouglas:
Includes functions to read in a CDIPS light curve, in response to #918 

One particular issue that came up is the fact that the CDIPS lightcurves do not include the standard bitflags for quality. Based on [Bouma et al. 2019](https://ui.adsabs.harvard.edu/abs/2019ApJS..245...13B/abstract), the authors removed CoarsePoint (4) and momentum dump/desat (32) flags by default, as well as other cadences identified in the data release notes for each sector. Those deleted cadences are 0s or nans in the resulting lightcurve. Therefore, I removed the ability for the user to pass a quality_bitmask to the cdips reader, and simply deleted any datapoints that the CDIPS authors flagged as bad. 

This is still a WIP until I make sure the CI tests pass.

PR #980, opened 2021-02-24
[WIP] CDIPS reader, take 2

stephtdouglas:
Includes functions to read in a CDIPS light curve, in response to #918. Follows #945, but redone to follow the new directory structure in the latest version. 

This is a WIP until the tests run. I did get pytest running on my machine (finally), but it won't run on the new directory structure for some reason (Might be be related to [this](https://stackoverflow.com/questions/41748464/pytest-cannot-import-module-while-python-can))

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,11 @@
 2.0.7dev (unreleased)
 =====================
 
+- Added support for reading CDIPS light curves. [#980]
+
+- Fixed a bug which caused ``LightCurve`` object instantiation to fail if
+  a string-typed column had a unit set. [#980]
+
 - Fixed a bug in ``CBVCorrector`` causing it to not work on K2 data. [#1012]
 
 - Modified ``LightCurveCollection`` to have a more simple and generic ``__repr__()``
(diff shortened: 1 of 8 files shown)


stephtdouglas:
@barentsen I'm not entirely sure why this particular error is only happening for one of the tests. It seems to be an issue with the quality array, which is an array of strings (not the usual bitflags), and astropy doesn't like that it's not an array of numbers. I can't figure out where that default is coming from within the rest of lightkurve. 

I can just not pass the quality array to `read_generic_lightcurve`, and do the filtering within the CDIPS reader itself, unless you think there's a better option than that.

barentsen:
@stephtdouglas I apologize for letting two weeks pass by without looking into your question.  I appreciate your PR and will look into the issue ASAP!

barentsen:
@stephtdouglas Thank you again for opening this PR, you did a great job!

This was tricky to debug!!  The failing tests in this PR were caused by the fact that AstroPy `QTable` objects cannot be instantiated when a string-type column has a unit set.  For example, this snippet of code fails:
```
from astropy.table import QTable, Column
col = Column(data=["a", "b", "c"], unit='unitless')
QTable(col)
```

This affected `LightCurve` because it is built on top of `QTable`, and the CDIPS FITS files happened to be the first data product we encountered that contain string-typed columns for which a unit is set in the FITS files.

I added a bugfix and test for this issue to this PR.  Everything looks good to be merged now!

stephtdouglas:
@barentsen thanks, and thanks for solving the bug!

PR #991, opened 2021-03-02
Add a `TargetPixelFile.animate()` method powered by `matplotlib.FuncAnimation`

barentsen:
This PR proposes to add a matplotlib-based animation method to the TargetPixelFile object.

Screenshot:

Screen Shot 2021-03-01 at 8 25 13 PM

### DIFF
--- a/src/lightkurve/targetpixelfile.py
+++ b/src/lightkurve/targetpixelfile.py
@@ -1147,7 +1147,9 @@ class TargetPixelFile(object):
                         ax.add_patch(rect)
         return ax
 
-    def _to_matplotlib_animation(self, step: int = None, interval: int = 200, **plot_args) -> "matplotlib.animation.FuncAnimation":
+    def _to_matplotlib_animation(
+        self, step: int = None, interval: int = 200, **plot_args
+    ) -> "matplotlib.animation.FuncAnimation":
         """Returns a `matplotlib.animation.FuncAnimation` object.
 
         The animation shows the flux values over time by calling `tpf.plot()` for multiple frames.
@@ -1174,7 +1176,7 @@ class TargetPixelFile(object):
             return ax.images
 
         def animate(i):
-            frame = i*step
+            frame = i * step
             ax.images[0].set_data(self.flux[frame])
             ax.set_title(f"Frame {frame}")
             return ax.images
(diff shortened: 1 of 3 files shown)


barentsen:
Open question: what is the right name for this feature?
* `tpf.animate()`
* `tpf.plot_animation()`
* `tpf.interact_matplotlib()`
* ...something else... ?

barentsen:
The suggestion to use matplotlib's FuncAnimation here comes from @christinahedges.

@christinahedges Would you be willing to review?

christinahedges:
This seems great! Normally I would be hesitant, because we have a lot of this functionality in `tpf.interact`, however, I think there is use for this stand alone, and right now `interact` is quite broken (a bokeh update has made the pixels look funny).

Re name, `tpf.to_animation` might also be good? But `animate` is nice too...i like it.

I need a minute to test out the functionality then I'll review!

src/lightkurve/targetpixelfile.py
```diff
@@ -1146,6 +1147,62 @@ def plot(
                         ax.add_patch(rect)
         return ax
 
+    def _to_matplotlib_animation(self, step=None, interval=200, **plot_args):
```
christinahedges, line 1150: If you are hard coding a default later (`step = 50` perhaps the default should be set up here rather than `None`, `None` steps never makes sense.

src/lightkurve/targetpixelfile.py
```diff
@@ -1146,6 +1147,62 @@ def plot(
...
+        step : int
+            Spacing between frames.  Default: show 50 frames.
+        interval : int
+            Delay between frames in milliseconds.
+        **plot_args : dict
+            Optional parameters passed to tpf.plot().
+        """
+        if step is None:
+            step = len(self) // 50
+            if step < 1:
```
christinahedges, line 1166: I believe this is not indented correctly

src/lightkurve/targetpixelfile.py
```diff
@@ -1146,6 +1147,62 @@ def plot(
...
+        ----------
+        step : int
+            Spacing between frames.  Default: show 50 frames.
+        interval : int
+            Delay between frames in milliseconds.
+        **plot_args : dict
+            Optional parameters passed to tpf.plot().
+        """
+        from IPython.display import HTML
+        return HTML(self._to_matplotlib_animation(**plot_args).to_jshtml())
```
christinahedges, line 1204: We need a nice error message if we call this from outside an ipython notebook

tests/test_targetpixelfile.py
```diff
@@ -864,3 +864,8 @@ def test_fluxmode():
     assert lc_sum.flux.value[0] == np.nansum(tpf.flux.value[0])
     assert lc_med.flux.value[0] == np.nanmedian(tpf.flux.value[0])
     assert lc_mean.flux.value[0] == np.nanmean(tpf.flux.value[0])
+
+
+def test_animate():
+    tpf = read(filename_tpf_one_center)
+    tpf.animate()
```
christinahedges, line 871: Won't this break when we run tests because they aren't being run in ipython?

src/lightkurve/targetpixelfile.py
christinahedges, line 1150: This function would enable a `to_mp4` method, which might be quite nice.

christinahedges (changes requested):
This is a great feature, I think even though it does duplicate some of the features of interact, this is more robust and has a place in lightkurve. I will personally use this a lot! 

I've left some minor comments, but I have no major comments! I'm saying changes because I think there is a typo.

Issue #663, opened 2020-01-16
Restore `test_bin` and `test_bin_quality`


barentsen:
I temporarily skipped the tests `test_bin` and `test_bin_quality` over in `test_lightcurve.py`.  These tests pass fine (locally) using SciPy v1.3.x and earlier, but the newly-released SciPy v1.4.0 has a change which causes binning to fail (cf. https://github.com/scipy/scipy/issues/11365).

I'm opening this issue to make sure we re-enable these two tests as soon as the underlying SciPy issue is resolved.

barentsen:
Over in #707 I have updated `requirements.txt` to exclude SciPy v1.4.0 and v1.4.1 for now.  Fingers crossed the issue will be fixed in v1.4.2...

texadactyl:
The scipy folks said version "1.5".

I did this: ```pip3 install git+"https://github.com/scipy/scipy"```

```
Collecting git+https://github.com/scipy/scipy
  Cloning https://github.com/scipy/scipy to /tmp/pip-req-build-xphlmq9t
  Running command git clone -q https://github.com/scipy/scipy /tmp/pip-req-build-xphlmq9t
  Running command git submodule update --init --recursive -q
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Building wheels for collected packages: scipy
  Building wheel for scipy (PEP 517) ... done
  Created wheel for scipy: filename=scipy-1.6.0.dev0+a49e960-cp38-cp38-linux_x86_64.whl size=62739928 sha256=6d70207fee0517b89c5f5c9153b1ba9e8e2508df8c59709742d207371fb7db3a
  Stored in directory: /tmp/pip-ephem-wheel-cache-fqj7cyhd/wheels/d1/2c/5a/72e8bc20052944163183e7301df53307818e86bd2b709619b0
Successfully built scipy
Installing collected packages: scipy
  Attempting uninstall: scipy
    Found existing installation: scipy 1.3.3
    Uninstalling scipy-1.3.3:
      Successfully uninstalled scipy-1.3.3
Successfully installed scipy-1.6.0.dev0+a49e960

```
Now, your test program from the issue report i.e.
```
import scipy.stats, numpy as np
x = [0.5, 0.5, 1.5, np.nan]
values = [10, 20, 30, 40]
scipy.stats.binned_statistic(x, values, bins=3)
```
gets this result:
```
Traceback (most recent call last):

[...]

PR #995, opened 2021-03-11
MNT: Restored (similar functionality) to `bins` parameter and binsize tests

dhomeier:
Found a leftover branch here where I was trying to address some of the interface changes in 2.x with `timeseries`; namely emulating the old functionality of setting the number of bins with `bins=N`, taking this parameter to calculate the appropriate `time_bin_size` that can then be passed to `aggregate_downsample`.
I was also looking into some of the test failures in #663 and found that the results are not actually wrong; the problem just seems to be that `aggregate_downsample` takes _closed_ intervals of length `time_bin_size`, which results in the first bin having one time point more, and the last one less. Accordingly their errors are scaled down/up by `sqrt(N + 1)`, `sqrt(N - 1)`.
I have re-enabled the tests with the expected results adjusted; if we want all the bins to be of equal size, some further investigation into the `aggregate_downsample` interface would be required.
The failure/NaNs with `time_bin_edges = [0, 10, 20, 30, 40, 50, 60, 70, 80]` ~~also seems to be a problem with how `aggregate_downsample` handles them, as it apparently takes the inverse of the values somewhere. Haven't tracked this down and left that part of the test marked `skip`~~ remains as long as only a single bin size can be handled.

Note that there is also PR https://github.com/astropy/astropy/pull/11266 for extending `aggregate_downsample` that may finally allow to restore more of the former `LightCurve.bin()` functionality.

### DIFF
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -1224,6 +1224,7 @@ class LightCurve(QTimeSeries):
         time_bin_start=None,
         n_bins=None,
         aggregate_func=None,
+        bins=None,
         binsize=None,
     ):
         """Bins a lightcurve in equally-spaced bins in time.
@@ -1236,16 +1237,21 @@ class LightCurve(QTimeSeries):
         Parameters
         ----------
         time_bin_size : `~astropy.units.Quantity`, float
-            The time interval for the binned time series. (Default unit: days.)
+            The time interval for the binned time series.
+            (Default: 0.5 days; default unit: days.)
         time_bin_start : `~astropy.time.Time`, optional
             The start time for the binned time series. Defaults to the first
             time in the sampled time series.
         n_bins : int, optional
             The number of bins to use. Defaults to the number needed to fit all
-            the original points.
+            the original points. Note that this will create this number of bins
+            of length ``time_bin_size`` independent of the lightkurve length.
         aggregate_func : callable, optional
             The function to use for combining points in the same bin. Defaults
             to np.nanmean.
+        bins : int or iterable of ints
+            The number of bins to divide the lightkurve into. In contrast to
+            ``n_bins`` this sets the length of ``time_bin_size`` accordingly.
         binsize : int
             In Lightkurve v1.x, the default behavior of `bin()` was to create
             bins which contained an equal number data points in each bin.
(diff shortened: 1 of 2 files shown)


barentsen:
Thank you @dhomeier!!  At first glance this looks fabulous!

I expect to review & merge later today!

dhomeier:
OK, took a bit of re-orientation – passing sequences as `bins` now raises an error, which is checked in the respective tests.

Once https://github.com/astropy/astropy/pull/11266 is merged, which supports array-typed `time_bin_start`, it should be possible to restore the alternative format with minimal wrapping.

barentsen:
Thanks so much for restoring `bins` and re-instating the tests @dhomeier, that is super useful! 

It is of course confusing to have both `n_bins` and `bins` parameters which do different things, but we are mainly doing this for backwards compatibility, i.e. to ease the transition to Lightkurve v2.0, so that is a good reason.

@dhomeier I'm inclined to say we ought to mark the `bins` parameter as `@deprecated`, given that custom binning will be supported in a different way if the AstroPy PR is merged. Does that make sense to you?

dhomeier:
> @dhomeier I'm inclined to say we ought to mark the `bins` parameter as `@deprecated`, given that custom binning will be supported in a different way if the AstroPy PR is merged. Does that make sense to you?

I've been thinking a bit about this and had also been inclined to suggest at least deprecating the other (currently not implemented) uses of `bins`, since the meanings seemed quite different. I think in general overloading a parameter like this is not recommended in Astropy space, but then again, it maps precisely onto the way `bins` is used in `astropy.stats.calculate_bin_edges`, so from that point it seems sensible to keep it.

My main issue with `n_bins` has been the way `time_bin_size` takes precedent, even if only present as default value. This can lead to rather surprising results, e.g. binning a lightcurve of 1 day length to `n_bins=80` with no other inputs set, returning a 40-day lightcurve with 78 empty bins. Of course one might still change this, so for `n_bins` given and `time_bin_size=None` set `time_bin_size` to TimeDelta / `n_bins` just as it is done for `bins`, instead of ½ day; but that would mean a change in functionality.

Had to check the v1.11 version to find that the option of setting `bins` as iterable is in fact still different from the direct functionality of a `time_bin_start` array, as it is specifying the indices rather than the times. This would also provide a rather convenient replacement of `binsize` in its original meaning of a fixed number of samples rather than a length in time (as in `time_bin_size`), by setting `bins=np.arange(:len(lc):binsize)`.

Since this should all be rather easy to implement by passing the bins from `calculate_bin_edges` to `time_bin_start` in Astropy 4.3, I'd vote for keeping this option.

barentsen (approved):
Thanks @dhomeier!  I will go ahead and merge because this is definitely an improvement for now, and we can revisit the bin API when the AstroPy PR is merged!

dhomeier:
Thanks for the review! Yes, I also wanted to discuss if there is interest in integrating some of that functionality (bins of given number of samples) directly on the `timeseries.downsample` side.

PR #997, opened 2021-03-12
Remove `numba` as a dependency

barentsen:
This PR proposes to make `numba` an optional dependency, due to recent installation issues encountered with `numba` and its dependency `llvmlite`.

We only use Numba to decorate the `_spline_basis_vector` with `@numba.jit`.  This would only happen now if numba can be imported.


barentsen:
After discussing with @christinahedges, we struggled to find strong evidence that `numba.jit` is speeding up sparse spline matrices, so we decided to go ahead and remove it as a dependency altogether.  I updated this PR to reflect this.

Below is the benchmark that led us to this conclusion:

## With numba

Screen Shot 2021-03-12 at 4 20 39 PM


## Without numba

Screen Shot 2021-03-12 at 4 21 24 PM

Issue #139, opened 2018-06-06
from_archive() returns two target pixel files if you ask for quarter 11 K2 data


christinahedges:
This is actually expected behavior because there are two campaigns. Similarly if you use

`KeplerTargetPixelFile.from_archive('K2TargetName', quarter=9)`

You will get back two TPFs, one from 9a and one from 9b. If you want just one of them you have to ask for 

`KeplerTargetPixelFile.from_archive('K2TargetName', quarter=91)`

I think this is good behavior but it's not what people expect when they do from_archive, they're expecting just one TPF? 

I think this will be alleviated by #138 which will throw an error if people try to e.g. use to_lightcurve() on a collection.

christinahedges:
@barentsen how do you feel about this behavior?

barentsen:
I'm getting ever more convinced that the Right Way forward is to have `from_archive` always return either 1 file or a helpful error message, and require the use of `conesearch` for multiple files. It's going to be less confusing!

christinahedges:
Yep I think that sounds good to me.

PR #1010, opened 2021-03-22
query split campaigns separately

astrobatty:
Hi, 
I opened this PR to solve the issue of querying split campaigns, which was originally reported in [105](https://github.com/lightkurve/lightkurve/issues/105) and [139](https://github.com/lightkurve/lightkurve/issues/139),  and has been resolved in `Lightkurve 1.x`. Now, in version 2.x, it has reappeared. However, due to certain technical problems ([see K2 release notes](https://keplergo.github.io/KeplerScienceWebsite/k2-data-release-notes.html)) it is useful to let the user query the split campaigns separately as well.

### DIFF
--- a/src/lightkurve/search.py
+++ b/src/lightkurve/search.py
@@ -927,13 +927,14 @@ def _search_products(
     # passed a radius value), because strict target name search does not apply.
     if filetype.lower() == "ffi" and radius is None:
         radius = 0.0001 * u.arcsec
+    # No campaign should be given otherwise split K2 campaigns will be filtered out
     observations = _query_mast(
         target,
         radius=radius,
         project=mission,
         provenance_name=provenance_name,
         exptime=exptime,
-        sequence_number=campaign or sector,
+        sequence_number=sector,
         **extra_query_criteria,
     )
     log.debug(
@@ -1212,6 +1213,13 @@ def _filter_products(
     if "kepler" in provenance_lower and campaign is None and sector is None:
         mask |= _mask_kepler_products(products, quarter=quarter, month=month)
 
+    # K2 data needs a special filter for split campaigns
+    mask &= ~np.array(
+        [prov.lower() == "k2" for prov in products["provenance_name"]]
+    )
+    if "k2" in provenance_lower and quarter is None and sector is None:
+        mask |= _mask_k2_products(products, campaign=campaign)
+
     # HLSP products need to be filtered by extension
     if filetype.lower() == "lightcurve":
         mask &= np.array(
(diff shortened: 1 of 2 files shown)


barentsen:
Thank you for opening this issue @astrobatty!!

This isn't documented anywhere (my fault), but removing the support for split campaigns was an intentional simplification of the code in Lightkurve v2.x.

The main reason is that search queries run much faster if we can filter by quarter/campaign/sector on the server side, i.e., if we can pass a "sequence_number" filter in our underlying call to `astroquery.mast.Observations.query_criteria()`.  This makes the code less complicated, less prone to errors, and much faster because it means we no longer have to retrieve the full list of all data products across campaigns/sectors to filter them on the client side.

Unfortunately, it means limiting the queries to the value of `sequence_number` in the MAST API, which is "9" for both C91 and C92 data products.

Users can continue to filter C91 vs C92 products as follows:
```
search = lk.search_targetpixelfile("EPIC 228162462")
mask = ["c91" in obsid for obsid in search.table['obs_id']]
search[mask]
```

Do you think the syntax above is sufficient, or do you think there is a strong reason to continue supporting "campaign=91"-style syntax?  I'd love your feedback!

astrobatty:
I think supporting split campaign syntax should be continued, but based on your reply not in the previous way. A filtering by `obsid` may be applied after MAST query under the hood.

Why do I vote to keep the previous syntax? The main problem is as follows.
If you query and download e.g. campaign 9 data, then you will be warned that there are more than one file and the user should narrow the search result:
image

However, if you look at the current search result you will see that there are files with same metadata:
image

Based on this, how would you expect anyone to narrow their result? Unfortunately, this output is not intuitive and not helpful at all.

My personal opinion is that Lightkurve is a really great tool to work on space-based data, but if someone has to dig into to the details to get to know how to perform basic steps, it loses a bit of its value.

barentsen:
Thank you @astrobatty!

I hear you!  At minimum, it is an issue that the SearchResult table does not indicate any difference between the two results.  I wish we could have it show the exact date range for each data set, but that meta data is not consistently available from MAST.  Let me ponder this for a bit!

Thanks++ for raising this issue.

astrobatty:
I appreciate your feedback @barentsen!

Hopefully, we can somehow overcome this issue.

barentsen:
Thanks again for raising this issue @astrobatty!  Do you think the modifications made in #1018 are adequate, or is there a very strong reason to be able to also specify the split campaigns vie the `campaign` keyword?  Please explain the use case if so, so I can better understand!

astrobatty:
The modifications are adequate only if someone is searching the database manually. My problem is that we developed a pipeline to perform photometry on K2 TPFs, called [autoEAP](https://github.com/zabop/autoeap). Now, I'm trying to make it compatible with Lightkurve 2.x, which is not easy due to the several new features, but it seems like everything is working. The only problem is the querying. Unfortunately, we have to separate the split campaigns due some technical issues otherwise we cannot perform photometry, nor run K2 systematics correction properly. If Lightkurve won't allow to specify the split campaigns, then we have to force the users to use Lightkurve 1.19 or I have to handle this issue myself.

PR #1012, opened 2021-03-24
Issue 1008: Fixed issue with CBVCorrector not working with K2 data

jcsmithhere:
This was an honest bug when using CBVCorrector with K2 data. The fix was really simple.

However, while running the unit tests I also noticed another issue when not extrapolating. It looks like numpy.linalg.matrix_rank (which is used to validate a DesignMatrix) no longer likes NaNs in the matrix (it throws and error). I had to change the functionality so that if extrapolate=False then the outside values to be extrapolated are filled with zeros (instead on NaNs).

All unit tests now pass.


barentsen:
Thank you for the quick fix @jcsmithhere! Looks good to me! :+1:

PR #1018, opened 2021-04-01
Delineate sections of split K2 campaigns in SearchResult

nksaunders:
In response to #1010, I've added code to list the different segments of split campaigns separately with clear labels in the search result. (PR opened separately because this is a much simpler proposed change.) 

This approach requires no additional queries, and uses the `obs_id` value to label campaigns. It doesn't allow you to specifically search for half of a campaign, but it does disambiguate the search result. I think this strikes a balance between maintaining the fast query speed (important for large samples of data) and keeping it intuitive to users.

Here it is in action for campaigns 9 and 11:

Screen Shot 2021-03-31 at 2 52 49 PM

Because they're always listed in order, you can index the `SearchResult` to get the desired section in an automated way, e.g. you could use

```python
lk.search_targetpixelfile('EPIC 228162462', cadence='long', campaign=9)[0]
``` 
if you just want `9a` (FWIW that's true of the current behavior as well, it's just not as clear). 

The major issue with this change is that the labels for split campaign 10 on MAST are not consistent with the [K2 data release notes](https://keplergo.github.io/KeplerScienceWebsite/k2-data-release-notes.html) or campaigns 9 and 11. The DR notes say for campaign 10
> The two segments are identified in the archive products as c101 and c102, respectively.

However, for the c10 targets I checked, both segments are listed as `{id}-c10_lc` instead of `{id}-c101_lc` and `{id}-c102_lc` as they are for 9 and 11. 

I'll keep investigating if there's a simple way to check this for c10, but for now feel free to provide any feedback @barentsen and @astrobatty.

### DIFF
--- a/src/lightkurve/search.py
+++ b/src/lightkurve/search.py
@@ -980,7 +980,7 @@ def _search_products(
             # listed separately in the table with suffixes "-1" and "-2"
             if obs_project == "K2" and result["sequence_number"][idx] in [9, 10, 11]:
                 for half,letter in zip([1,2],['a','b']):
-                    if f"c{tmp_seqno}{half}" in result["obs_id"][idx]:
+                    if f"c{tmp_seqno}{half}" in result["productFilename"][idx]:
                         obs_seqno = f"{int(tmp_seqno):02d}{letter}"
             result["mission"][idx] = "{} {} {}".format(
                 obs_project, obs_prefix.get(obs_project, ""), obs_seqno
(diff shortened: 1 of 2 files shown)


barentsen:
Thank you @nksaunders!  I think this is a good solution!

We should be able to support C10 as well if we use the `productFilename` field instead of `obs_id`, which does encode "c101" and "c102", for example:

Screen Shot 2021-04-01 at 10 48 35 AM

Do you want to modify the PR to try this?

nksaunders:
Nice! Thanks @barentsen, that works for 9, 10, and 11. Are there strong opinions about `10a` and `10b` versus `10-1` and `10-2`? Or even just `101` and `102`? I'm happy with any of them, but prefer the first or second.

barentsen:
I'd vote for "10a" and "10b" as well.

astrobatty:
Thanks for the help @nksaunders! That's a step forward. I'd vote for "10a" and "10b" too.

barentsen (approved):
LGTM! Thanks @nksaunders!! 👍

PR #1035, opened 2021-04-22
remove unused doc build dependency file

orionlee:
Remove the now unused `doc/requirments.txt`
- the doc build dependency is now in `extras_require` of setup.py


barentsen:
Thank you @orionlee!

I will note that the `pyproject.toml` file in the root is now the main place to record dependencies (which gets converted to an equivalent `setup.py` file for backwards compatibility with some tools, e.g. the GitHub dependency tracker doesn't appear to understand a `pyproject.toml` file yet).

PR #1037, opened 2021-04-23
tpf.plot_pixels - improve legibility when show flux = True

orionlee:
This is a continuation of PR #906, which addresses issue #894

To handle legibility of the plot in yellow pixels, case `show_flux=True` (the [discussion of options](https://github.com/lightkurve/lightkurve/pull/906#issuecomment-728345949)), I make `markersize` customizable so that 
1. there is no change in typical cases
2. when an user encounters a particular plot that is not too legible, the user can increase the marker size to improve it. 

The usage is added as examples in the api doc.

### DIFF
--- a/src/lightkurve/targetpixelfile.py
+++ b/src/lightkurve/targetpixelfile.py
@@ -1801,6 +1801,7 @@ class TargetPixelFile(object):
         corrector_func=None,
         style="lightkurve",
         title=None,
+        markersize=0.5,
         **kwargs,
     ):
         """Show the light curve of each pixel in a single plot.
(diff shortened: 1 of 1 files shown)


barentsen:
This is an excellent improvement.

Also: thank you for adding an `Examples` section to the docstring, that is brilliant!

orionlee:
Regarding the examples, I believe it's needed because the options added to handle the use cases all end up quite indirect (e.g., changing size by creating `ax` aprior). So the examples serve as mini cookbook.

For the legibility: I personally prefer to have black plots on yellow pixel option. But based on prior discussion, it might also mislead some people. Given there is no clear winning solution, I end up with this PR  as a compromise.

Another approach I considered is to let user supply an optional customization function, in the form `(pixel_face_color) -> dict` that returns a dictionary of parameters to customize `plot()` call for pixels.

With such an approach, I can still choose to have black plots on yellow pixels for my own use,  without having it baked it the code. 😄 

I have not done it this way, because I feel it might make the API overly complicated. If you think it is acceptable, I think open another PR to work on it.

PR #1039, opened 2021-04-24
MAST query persistent cache

### LINKED ISSUE #934: Investigate using `diskcache` for persistent MAST query caching
We could look into using the `diskcache` package to cache MAST search operations:

https://pypi.org/project/diskcache/

The package is a disk-based key-value store which makes it easy to specify an expiration time for each entry.  This would allow us to e.g. cache search queries for 24 hours, which would significantly speed up searches across different Python processes, whilst largely avoiding the problem of stale search data.

orionlee:
Closes #934

Currently a working prototype.


barentsen:
This is awesome!  Thank you for exploring this!! :+1: 

Here are four quick items of feedback:

---
**Item 1**: I see that this PR proposes to use a custom `SearchResultCache._memoize()` decorator to enable the cache dir to be set after the decoration took place.  The implementation for this looks great, and I'd be happy to merge it, but I do wonder if being able to change the cache directory at runtime is a sufficiently important requirement to warrant the extra lines of code it needs.  If we decided that changing the cache directory at runtime is not necessary, we may be able to get away with just using two lines of code to define `sr_cache`:

```python
CACHEDIR = Path.home() / ".lightkurve-cache"
sr_cache = diskcache.Cache(directory=CACHEDIR)
```
... and then we'd use `sr_cache.memoize()` to decorate the search functions.  It would not be possible to change `CACHEDIR` at runtime in this scenario, but we could decide to read it from a config file so that at least it can be changed before runtime.  A config file could also specify the default expiration time.

Do you have a specific use case in mind that justifies adding our own `SearchResultCache` class instead of going with the decorator already provided out of the box by `diskcache`?

--- 

**Item 2**: If the expiration time is configurable, we can likely support disabling the cache via that setting.  A user could also call `sr_cache.clear()` after every search operation, which may be sufficient for most (if not all?) use cases where a user wants to disable caching.

--- 

**Item 3**: I'm surprised that sqlite is giving you trouble in VSCode/Jupyter because the `sqlite3` module is part of the Python standard library, i.e., it's not an extra dependency.  Feel free to copy/paste the error message if encountered one?

---

**Item 4**: Just for fun: note that Python 3.4 added the neat `pathlib` module to the standard library, which we can use now that Lightkurve v2 requires Python >= 3.6.  For example, it enables this snippet
```python
import os
os.path.join(os.path.expanduser("~"), ".lightkurve-cache")
```
to be rewritten as follows:
```python
from pathlib import Path
Path.home() / ".lightkurve-cache"
```
There's nothing wrong with using `os.path` of course, `pathlib` just makes the code a bit more readable.

orionlee:
I divide the response between 
1. requirements / API changes needed for first release 
2. implementation details

### Requirements / API changes
1. Changing cache directory: I don't think changing the directory at runtime is a requirement per se, but specifying it once (in an app) is needed.
Having the cache dir (and expiration time) specified in a config file, e.g., `~/.lightkurve.cfg` is a possibility. But it has its own issues:
  a. hard to specify in some environment, e.g., hosted jupyter such as google collab.
  b. if a user want to specify different values for different projects, the user can't do so, unless we introduce say, a magic environment variable specifying the location of the config file.

Having said that, configuring it as `cache.sr_cache` does not allow users to have their own global settings shared across all their work.

2. disabling cache: I'd like to gauge if it's a hard requirement for the first release.
Note: setting expiration to 0 is **not** disabling the cache, at least with `disckcache`. It just means no *new* entries will be added to the cache. Existing cache will still be used.

3. sqlite3 import error with VS Code + Jupyter:

The specific error is `ImportError: DLL load failed: The specified module could not be found.` The general issue (sqlite3 import failure in some environment) has been discussed in various cases (e.g., a [stackoverflow question](https://stackoverflow.com/questions/54876404/unable-to-import-sqlite3-using-anaconda-python)).

The main point I'd like to check is from a requirement perspective, is it something we should deal with (to reduce friction / pain points), to ensure lightkurve can still be used if the import failure happens?


### Implementation details

1. the custom `memoize` decorator codes: it's true only needed because cache dir is specified at runtime. However, 
  a. if we want to support disabling cache, it probably needs to be implemented in the custom decorator as well. The default `diskcache` `memoize` has no option to disable it. (And setting expiration to 0 will not be the case.
  b. if we want to support graceful handling of import diskcache failure, we probably need to do it in the decorator as well.

2. regarding `SearchResultCache` class: in addition to the custom decorator, it's meant to be an entry point exposing cache related config / features without exposing underlying `diskcache` implementation to users. So depending on the discussion of all others, it may stay or go.

PR #1041, opened 2021-04-25
Fix lc.bin() missing original meta

### LINKED ISSUE #1040: LightCurve.bin() - binned LightCurve missing meta data
#### Problem description

1. The binned lightcurve (from`lc.bin()`) does not carry the original meta data over.
2. Additionally, I think the binned lightcurve should also contain some basic bin information, e.g., bin size or possibly some bin statistics.

Motivational use case: I binned a lightcurve and use it to do `query_solar_system_objects()` to make  it run faster. It failed because binned lightcurve does not have RA / Dec in  its meta (the meta is empty).

#### Example

```python
import lightkurve

# lc is any regular LC downloaded from MAST
> lc_b = lc.bin(time_bin_size = 20 * u.minute)
> lc_b.meta
OrderedDict()

```

#### Environment

orionlee:
Close #1040

- The issue also has an open ended suggestion that some bin parameters / stats be stored in binned lc. 
If there is a consensus and easy to fix, I can add it to this PR
- The PR also removed outdated comments regarding temporarily skipped bin tests (no longer the case)

### DIFF
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -1382,9 +1382,7 @@ class LightCurve(QTimeSeries):
             ts.remove_column(colname)
             ts.add_column(tmpcol, name=colname, index=idx)
 
-        binned_lc = self.__class__(ts)
-        binned_lc.meta.update(self.meta)
-        return binned_lc
+        return self.__class__(ts, meta=self.meta)
 
     def estimate_cdpp(
         self, transit_duration=13, savgol_window=101, savgol_polyorder=2, sigma=5.0
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -1382,7 +1382,9 @@ class LightCurve(QTimeSeries):
             ts.remove_column(colname)
             ts.add_column(tmpcol, name=colname, index=idx)
 
-        return self.__class__(ts)
+        binned_lc = self.__class__(ts)
+        binned_lc.meta.update(self.meta)
+        return binned_lc
 
     def estimate_cdpp(
         self, transit_duration=13, savgol_window=101, savgol_polyorder=2, sigma=5.0
--- a/tests/test_lightcurve.py
+++ b/tests/test_lightcurve.py
@@ -482,7 +482,6 @@ def test_cdpp_tabby():
     assert np.abs(lc2.estimate_cdpp().value - lc.cdpp6_0) < 30
 
 
-# TEMPORARILY SKIPPED, cf. https://github.com/lightkurve/lightkurve/issues/663
 def test_bin():
     """Does binning work?"""
     with warnings.catch_warnings():  # binsize is deprecated
@@ -605,7 +604,6 @@ def test_bins_kwarg():
     #   - Bins = 310.0
 
 
-# TEMPORARILY SKIPPED, cf. https://github.com/lightkurve/lightkurve/issues/663
 def test_bin_quality():
     """Binning must also revise the quality and centroid columns."""
     lc = KeplerLightCurve(
@@ -1129,7 +1127,6 @@ def test_river():
         plt.close()
 
 
-# TEMPORARILY SKIPPED, cf. https://github.com/lightkurve/lightkurve/issues/663
 def test_bin_issue705():
     """Regression test for #705: binning failed."""
     lc = TessLightCurve(time=np.arange(50), flux=np.ones(50), quality=np.zeros(50))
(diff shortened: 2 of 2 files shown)


src/lightkurve/lightcurve.py
```diff
@@ -1382,7 +1382,9 @@ def bin(
             ts.remove_column(colname)
             ts.add_column(tmpcol, name=colname, index=idx)
 
-        return self.__class__(ts)
+        binned_lc = self.__class__(ts)
+        binned_lc.meta.update(self.meta)
+        return binned_lc
```
dhomeier, line 1387:
```suggestion
        return self.__class__(ts, meta=self.meta)
```
Might do it even as concisely as the original, though that is perhaps a matter of taste. But when adding it afterwards I see a
```
UserWarning: Lightkurve doesn't allow columns or meta values to be created via a new attribute name.A new attribute is created. It will not be carried over when the object is copied. - see https://docs.lightkurve.org/api/lightkurve.lightcurve.LightCurve.html
```

orionlee: I prefer the shortened version you suggested as well.

Regarding the `UserWarning`, I did not see it when I ran the test. 

I did notice an existing unrelated `test_attr_access_others()` does emit the warning. Could it be possible the one you saw came from `test_attr_access_others()`?

barentsen:
Thanks so much for opening this PR!   I will review this and related issues asap!

src/lightkurve/lightcurve.py
dhomeier: I did not run the tests for this branch, just got the warning when adding a `meta` in normal code. Could be that they are marked ignore in the tests.

orionlee: I am not sure what you mean by "when adding a `meta` in normal code"...

The change you suggested, `self.__class__(ts, meta=self.meta)`, does not cause warnings.  

I guess you noticed the (spurious) warning emitted by another test `test_attr_access_others()`. The issue has just been fixed in PR #1045.

dhomeier: No, again, I did not run any tests whatsoever, just made a simple check to verify the meta is lost in binning
```python
>>> lc = lk.LightCurve(time=[1, 2, 3, 4], flux=[1, 1, 1, 1])
>>> lc.meta = {'TELESCOP': 'Kepler'}
<stdin>:1: UserWarning: Lightkurve doesn't allow columns or meta values to be created via a new attribute name.A new attribute is created. It will not be carried over when the object is copied. - see https://docs.lightkurve.org/api/lightkurve.lightcurve.LightCurve.html
>>> lb = lc.bin(2)
>>> lb.meta
OrderedDict()
```
Note also that it explicitly warns that it will not be preserved on copying, so that might have been designed deliberately at some point; however the docs link is broken, and I did not see any other information on that so far.
But I don't see a strong reason not to copy it over to the binned lightcurve (apart perhaps from some information that would only apply to the original, like number of points...).
Regarding other stats, `time_bin_start` and `time_bin_size` are already stored; others like no. of points per bin may not be so easy to get out of `aggregate_downsample`.

orionlee: 1. On the UserWarning upon `lc.meta = {'TELESCOP`, 'Kepler'}, it is a bug that is just filed as #1046 . The warning is for intended for users creating their own  custom attribute, e.g., `lc.some_new_attr = 123`, and should not appear when dealing with built-in `lc.meta`

2. Reasons for copying metadata over binned lightcurve. Some LightCurve functions require metadata, e.g., `query_solar_system_object()`, `plot()`, etc.

3. Regarding other stats, I didn't realize `time_bin_size` is already there. I mistakenly looked for it in `lc.meta['time_bin_size']`. So I guess we're okay.

barentsen:
@orionlee @dhomeier Apologies I did not get to this issue yet.  I am currently on vacation and will be off the grid until Monday May 3rd.  I can't wait to carefully look at this then!!

orionlee:
@barentsen I didn't realize you just made a new release till now.
From my perspective, this PR is ready to be merged.

orionlee:
@barentsen Any chance this can get merged soon? It's a simple fix, and shouldn't be controversial.

barentsen (approved):
Apologies for the slow response!  This is a sweet, simple, and very useful bug fix indeed.

Thanks so much for this PR @orionlee!  I will add an entry to CHANGES.rst after merging.

barentsen:
Just to confirm: I added a change log entry via 6698e3bdeab666dce88ac818c7186b9e4ac590c9.

Issue #1028, opened 2021-04-15
Bin Function won't bin by a number of points


christinahedges:
I'm finding it very hard to bin by a number of points in lightkurve right now. This used to be the default argument, but now the default argument is a time, which means I have to think about it and I don't love thinking. I think there is definitely a bug because of what I'm seeing below:

![image](https://user-images.githubusercontent.com/14965634/114910676-086db580-9dd3-11eb-96a2-7573e84bd2cf.png)

I think one answer to how we should bin -could- be that the first argument is a "time resolution" e.g. 

`lc.bin(2)`

would half the time resolution or something

barentsen:
One potential solution that comes to mind is that the `.bin()` method should perhaps have continued to be reserved for "binning by equal number of data points", and a new method called `.resample()` or `.downsample()` should have been added for "binning in bins of equal time duration".

It is unclear, however, whether undoing the change made to `bin()` in Lightkurve v2.0 would cause more or less pain to users, who have been gradually adapting to the changes.

On a related note: the `aggregate_downsample` function in AstroPy TimeSeries is about to make binning with custom bins much easier, thanks to this PR: https://github.com/astropy/astropy/pull/11266#discussion_r597806758

christinahedges:
![image](https://user-images.githubusercontent.com/14965634/114913810-79fb3300-9dd6-11eb-8ca7-e0c894ecf4fe.png)

dhomeier:
> One potential solution that comes to mind is that the `.bin()` method should perhaps have continued to be reserved for "binning by equal number of data points", and a new method called `.resample()` or `.downsample()` should have been added for "binning in bins of equal time duration".

I am looking into how to implement this with the improved `aggregate_downsample` options, but was surprised to find that the "old" behaviour was actually binning into constant lengths in time as well (at least in v1.11 where I was looking).
`bin_edges` were determined by the `astropy.stats` `calculate_bin_edges` function, which for an unevenly sampled series returns this:
```python
>>> lc.time.mjd
array([57469.52119213, 57469.52119572, 57469.52119907, 57469.52121528, 57469.52122685, 57469.52123843])
>>> calculate_bin_edges(lc.time.mjd, 3)[:-1]
array([57469.52119213, 57469.52120756, 57469.52122299])
>>> lc.time.mjd[::2]
array([57469.52119213, 57469.52119907, 57469.52122685])
```
[...]

PR #1042, opened 2021-04-25
Restore old `lc.bin()` functionality with new `aggregate_downsample`

dhomeier:
Follow-up of #995 to recreate the original `binsize` functionality of specifying a fixed number of points per bin (+ the sequence and string options for the special `astropy.stats.histogram` methods).
This **requires** the enhanced `aggregate_downsample` from astropy/astropy#11266. Could also address #1028; the behaviour of a scalar `bins` in the current version is unchanged from the existing one (or afaict from that in v1.11), i.e. a number `bins` of _equal-length_ bins; but it could be discussed if that should be changed to _equal-sample-size_, see comments in the function.
How to support all combinations of the `time_bin_start`, `time_bin_end`, `time_bin_size` with the existing parameters, and in what priority, probably also needs further discussion.
The tests have been updated to reflect results with and without the new astropy branch; I have disabled the check for `binned_lc.quality` entirely since there does not seem to be any mechanism implemented to bin this correctly (with bitwise `or`), and previously the test apparently just averaged by chance to the correct values.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,9 @@
 - Deprecated ``download_kepler_cbvs`` and ``download_tess_cbvs`` in favor of
   ``load_kepler_cbvs`` and ``load_tess_cbvs``. [#1122]
 
+- Restored original ``bins`` functionality from v1.x in ``LightCurve.bin()`` and
+  enabled iterable inputs to ``time_bin_start``, ``time_bin_end``, ``time_bin_size``
+  for custom bin sizes (requires Astropy 5.0 or higher). [#1042]
 
 
 2.0.11 (2021-09-22)
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -1355,7 +1355,7 @@ class LightCurve(QTimeSeries):
             bin. If ``time_bin_end`` is an array, the time bins do not need to be
             contiguous. If this argument is provided, ``time_bin_size`` should not
             be provided. This option, like the iterable form of ``time_bin_start``,
-            requires Astropy 4.3.
+            requires Astropy 5.0.
         n_bins : int, optional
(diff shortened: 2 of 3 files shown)


dhomeier:
1 failure now with 
```
>       assert underfit_metric_neighbors(lc_sap, min_targets=3, max_targets=3) < 0.9
E       assert 0.9259854773984688 < 0.9
E        +  where 0.9259854773984688 = underfit_metric_neighbors(<TessLightCurve length=18187 LABEL="TIC 388857263" SECTOR=11 AUTHOR=SPOC FLUX_ORIGIN=pdcsap_flux>\n       time         ...35  1.3812323e+05 ... -4.5290943e-03  1.6911790e-02\n1623.8955752075456  1.3825589e+05 ... -1.5050994e-03  2.1134768e-02, min_targets=3, max_targets=3)
```
not performing quite as poorly as it's supposed to; can't see an obvious connection to the changes in `bin()`.

barentsen:
Good news: I believe the `underfit_metric_neighbors` failure is not related to this PR.  Instead, I believe it is caused by changes at the data archive (specifically, TESS data for certain sectors have been replaced with re-processed data).

barentsen:
I fixed the failing test on the `main` branch via 81a370cabaa58534267e94a0bbfbf9f37feb7cd5, merged the changes into this PR branch, and started re-running the tests.  Fingers crossed!

dhomeier:
Great, thanks!
I noted I had left some comments on further To-Do items like prioritising and testing different/conflicting bin kwargs; I had marked this ready for review since the basic functionality is there, but if you feel better test coverage for those cases is needed, leave a comment.

barentsen (approved):
Thanks again @dhomeier!  This looks like an excellent PR and a valuable fix. :+1:

 I intend to merge this in the next few hours, unless anybody objects?

PR #1050, opened 2021-05-03
tpf.interact_sky() misc enhancement

### LINKED ISSUE #1049: TargetPixelFile.interact_sky() misc. enhancement
#### Problem description
I'd like to propose a few minor enhancement on `interact_sky()` that could help to reduce an user's mental work:
1. In hover pop-in, format the x/z (the pixel row/column) as simple integers
2. Indicate the target's position
3. let users copy the GAIA ID of the hovered star. (One cannot move the mouse over to the hover tooltip to copy it, as the tooltipwill be gone).
4. (TESS-specific) provide TESS-related information when available, specifically, the TIC and TESS magnitude.

#### Example

The current `interact_sky()`:

image

With the enhancement:

image

image

For the copying of GAIA number: one solution is change the code so that when an user taps a star, some of the star's meta data is displayed in an adjacent text area, so that after the tooltip disappears when the user moves the mouse away, the meta data is still in the text area.

#### Environment

orionlee:
Closes #1049 

Not yet ready for detailed review.
- The specific changes desired is pending the discussion on the issue.
- Some of the codes are not yet formatted for merge. It's formatted so that changes for individual features can be seen more easily.
- The codes for adding TIC data could use some cleanup, and some work to make it more robust (e.g., ensuring `interact_sky()` will still work in case downloading data about TICs from ExoFOP fails)

---
TODOs

- [x] Ensure it does not break non TESS SPOC tpfs
  - [x] Kepler
  - [x] TESSCut
- [x] Refine star details output
  - [X] Pop-in output: rename x / y to row / column
  - [X] pixel coordinate with 1 decimal point precision
  - [X] link to SIMBAD
  - [x] Polish text area output
  - [x] For stars with no TIC, avoid showing it as `TIC NaN`
- [x] Bug fix that for some TICs, an error occurred: `WARNING: OverflowError converting to IntType in column TIC ID, reverting to String. [astropy.io.ascii.core]`
- [x] Ensure that in case ExoFOP is not available (for getting nearby TIC data), the plot would still work (maybe with a warning)
- [x] Add tests (if practical)

Bonus
- Generalize SIMBAD link to let user specify the links they want to include.


src/lightkurve/interact.py
```diff
@@ -257,6 +274,32 @@ def make_lightcurve_figure_elements(lc, lc_source, ylim_func=None):
     return fig, vertical_line
 
 
+def _add_nearby_tics_if_tess(tpf, source, tooltips):
+    tic_id = tpf.meta['TICID']
```
barentsen, line 278: I'm sure you are aware, but this line of course raises a `KeyError` for a Kepler TPF right now, e.g. this fails:
```python
lk.search_targetpixelfile("Kepler-10", mission="Kepler").download().interact_sky()
```

barentsen:
These are brilliant improvements!!  Some thoughts:

* To avoid ambiguity, should we use the words `column` and `row` instead of `x` and `y` as the label for the pixel coordinates?
* I would consider showing one decimal for the pixel coordinates, e.g. `column: 1450.6` instead of `column: 1450`.
* I love that the details of a star now show up in the adjacent text area after clicking.  This is a brilliant improvement.  However: is there a reason the information shown in the text area is different from the information shown in the tooltip.  Could we make them the same for simplicity? (Apart from links, they will only work in the text area of course.)

src/lightkurve/interact.py
```diff
@@ -346,23 +404,53 @@ def add_gaia_figure_elements(tpf, fig, magnitude_limit=18):
...
+    # a widget that displays some of the selected star's metadata
+    # so that they can be copied (e.g., GAIA ID).
+    # It is a workaround, because bokeh's hover tooltip disappears as soon as the mouse is away from the star.
+    message_selected_target = Div(text="")
+
+    def show_target_info(attr, old, new):
+        if len(new) > 0:
+            msg = "Selected:<br><br>"
+            for idx in new:
+                tic_id = source.data['tic'].iat[idx]
```
barentsen, line 427: This line raised a KeyError when clicking on a target in a TPF obtained from TessCut, e.g. using
```python
lk.search_tesscut("Pi Men", sector=34).download(cutout_size=20).interact_sky()
```

orionlee:
@barentsen The issues and suggestion you made all made sense. 

Regarding the discrepancy between the output in pop-in and the adjacent text area upon tap. It's because I was lazy. The pop-in is backed by bokeh-based template that cannot be ported directly to text area output. 
On what's desirable: the text area output should be the same as the pop-in with links added.

I've attached a list of action items in the PR's top comment:   https://github.com/lightkurve/lightkurve/pull/1050#issue-629297098

barentsen:
Thanks @orionlee.  Once the action items you listed are addressed, I'd be super excited to merge this PR! 👍

orionlee:
On linking SIMBAD: I thought about it too!

We could possibly just add a link to SIMBAD without verifying whether it is there or not.

Given a Gaia identifier, we could produce a reasonable SIMBAD link in the following pattern:

http://simbad.u-strasbg.fr/simbad/sim-id?Ident=Gaia+DR2+6385548541499112448&NbIdent=1&Radius=2&Radius.unit=arcmin&submit=submit+id

ExoFOP TESS does something similar, with a link to (cruder) match by coordinates. 

image

Most of the times it gives back the right object. Occasionally the match is false (in a crowded field) or no match is found.

orionlee:
@barentsen All UI-related changes are complete, and can be reviewed.

image

Case 2 stars are selected:
image

orionlee:
The PR is ready for review, both the UI and the actual codes.

One final UI change is for the SIMBAD links, now there are 2 links, one by Gaia ID and one by coordinate. 

While matching by Gaia ID is more accurate, anecdotally, it seems that there are noticeable amount of cases that a star exists in SIMBAD, but its entry does not have Gaia ID. So a search by coordinate is added as a backup per-se.

I don't know if we necessarily want to include both links in the release, but I thought it'd be helpful to add it now. It can easily be removed if we choose to do it.

barentsen (changes requested):
Thanks so much @orionlee!  The code changes are in a great shape.  I particularly liked your improvements to the unit tests.

My only major comment is that there appears to be an issue with the proper motion correction.  To test the correction, I checked two objects with a very large proper motion: Wolf 359 in K2 and Proxima Cen in TESS.

For Wolf 359 in K2, the circle and crosshair markers are located far away from the source position, and they are mutually inconsistent:

Screen Shot 2021-05-14 at 5 55 44 PM

For Proxima Cen in TESS, the circle marker appears correct, but the crosshair marker is slightly offset:

Screen Shot 2021-05-14 at 5 57 43 PM

I did not yet manage to understand the origin of the bug.  We may be mis-interpreting the equinox of either the FITS headers or the catalog, but we should also not discard the possibility of FITS header values being wrong.  We'll want to check the values by hand.  I did not have time to do this yet, but I'm happy to help if you get stuck with this.  Proper motions are somehow always complicated! =)

tests/test_interact.py
```diff
@@ -147,24 +150,52 @@ def ylim_func_unitless(lc):
...
+    from lightkurve.interact import (
+        prepare_tpf_datasource,
+        make_tpf_figure_elements,
+        add_gaia_figure_elements,
+    )
+    import lightkurve.interact as lk_interact
+    def mock_raise(*args):
+        raise IOError("simulated service unavailable")
+
+    monkeypatch.setattr(lk_interact, "_search_nearby_of_tess_target", mock_raise)
```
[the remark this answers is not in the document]
orionlee: The test function name, `test_interact_sky_functions_case_nearby_tics_failed`, is not self explanatory enough then. :-(

The test is to cover the case that if ExoFOP is not available, `interact_sky()` will still work.

orionlee:
On the proper motion issue. I have not completely solved it yet. But so far:
- I don't believe there is any regression for plotting stars from Gaia (the circle)
- there are issues in plotting the target star (the cross, the new feature)

Details:
1. For Wolf 359 in K2, the circle: it is not an issue because the circle does not represent Wolf 359. It is just a [nearby star](http://simbad.u-strasbg.fr/simbad/sim-id?Ident=Gaia%20DR2%203864973140467689216). Apparently Wolf 359 is not in Gaia DR2.
2. For Wolf 359 in K2, the cross: one issue identified is that in the K2 tpf, the proper motion is in the unit of arcsec / year while the code assumes milliarcsec / year. I suspect it might not be the only issue, however (see below).
3. For Proxima Cen in TESS, the cross: I don't know if it's the issue in my interpretation of equinox, or the header value in fits file might not be accurate. I don't think the general logic in computing proper motion is the issue (as it is used by plotting the circles).

For equinox (at least in TESS), the only description I found in the [data product description document](https://archive.stsci.edu/missions/tess/doc/EXP-TESS-ARC-ICD-TM-0014-Rev-F.pdf) is
```
EQUINOX = 2000.0 / equinox of celestial coordinate system
```
I interpret it as J2000, i.e., January 1, 2000, noon TT, but I am not 100% sure.

I suspect deviation in equinox definition is not sufficient to explain the discrepancy seen though. The reason is that in my local build, I tried to put in some arbitrary equinox to see how the cross gets displayed. An equinox of around year 1986 will put the cross near, but not on the circle.

orionlee:
On target proper motion correction: I have committed a fix. Please review.

I have tested the following targets, which have large or moderately large proper motion.
They all work, with the exception of Proxima Cen.

| Mission | Target | PM RA (mas/yr) | PM Dec (mas/yr) |
| --- | --- | --- | --- |
| TESS | Proxima Cen | -4020.83 | 764.86 |
| K2 | Wolf 359 | -3842 | -2725 |
| Kepler | Kepler-444 | 97.69999999999999 | -631.6 |
| TESS | pi Men | 310.712 | 1049.66 |
| TESS | LTT 1445 A | -362.46 | -277.001 |

Note: for Wolf 359: there is no circle because Wolf 359 is not in Gaia DR2.

I am inclined to believe that it means the current logic is correct. The problem with Proxima Cen lies with the data in the FITS file. 

An indirect evidence: if I LTT 1445 A, which has a small PM among the targets, if I do a plot without the proper motion corrected, the difference is still noticeable.


The current logic with proper motion corrected:
image


If I temporarily remove proper motion correction, the difference is noticeable:

image

orionlee:
Proper motion further fix: I have replaced the existing proper motion correction logic with astropy's [`SkyCoord.apply_space_motion()`](https://docs.astropy.org/en/stable/api/astropy.coordinates.SkyCoord.html#astropy.coordinates.SkyCoord.apply_space_motion).

Now they all lined up. For Proxima Cen, it is almost perfect

image

I tried improving it further by hardcoding Proxima Cen's parallax in the calculation (TIC data has no parallax). It yields no noticeable difference. So I conclude that the visible difference is a result of the difference of the underlying data (Gaia DR2 on the one hand, and TIC in fits header on the other hand).

---

One final edge case is for TPF file from `search_tesscut`: the target is somewhat likely to be off, as the TESSCut tpf has no proper motion supplied.

Currently, it will simply display the target without proper motion correction, with a warning so that users could guess what is going on:
```python
 LightkurveWarning: Proper motion correction cannot be applied, as none is supplied. pm_ra: 0.0 arcsec / yr, pm_dec: 0.0 arcsec / yr. Use RA Dec as is.
  category=LightkurveWarning)
```

barentsen:
Thank you @orionlee!  I started reviewing the fixes.

barentsen (approved):
I played around with the current version of this PR and it all looks good to me now!! 👍 

> I have replaced the existing proper motion correction logic with astropy's SkyCoord.apply_space_motion().

Ohhh, thank you, I didn't know SkyCoord provided this method.  This is very helpful!

> LightkurveWarning: Proper motion correction cannot be applied

I'm on the fence about this warning because there is not much the user can do to fix this problem, and it doesn't significantly harm the tool, so this warning will likely be ignored all the time.  I like to avoid emitting warnings that will be ignored because it will make users less sensitive to warnings that shouldn't be ignored.

Having said that, having the warning there might trigger us to fix the underlying issue, so I'm fine with keeping it for now!

barentsen:
I resolved the merge conflict in CHANGES.rst and intend to merge this PR as soon as the tests pass.

Thank you again for this awesome improvement @orionlee!

orionlee:
>> LightkurveWarning: Proper motion correction cannot be applied

>I'm on the fence about this warning because there is not much the user can do to fix this problem, and it doesn't significantly harm the tool, so this warning will likely be ignored all the time. I like to avoid emitting warnings that will be ignored because it will make users less sensitive to warnings that shouldn't be ignored.

The warning is primarily intended for scenario that when a user sees the target (cross) off target, they could understand that it could be due to the absence of proper motion correction.

One improvement could be really just catching that specific scenario and change the warning message, explicitly saying that the target might be noticeably off.

PR #1051, opened 2021-05-04
Bugfix lc.meta assignment bogus Warning and outdated API doc links

### LINKED ISSUE #1046: Incorrect UserWarning in assigning LightCurve.meta
#### Problem description

1. When user assigns value to `lc.meta`, an UserWarning` is issued. It should not be there.
2. The URL in the UserWarning is outdated. It is now https://docs.lightkurve.org/reference/api/lightkurve.LightCurve.html

#### Example

The warning should not appear in the following code.
```python
import lightkurve as lk
>>> lc = lk.LightCurve(time=[1, 2, 3, 4], flux=[1, 1, 1, 1])
>>> lc.meta = {'TELESCOP': 'Kepler'}
<stdin>:1: UserWarning: Lightkurve doesn't allow columns or meta values to be created via a new attribute name.A new attribute is created. It will not be carried over when the object is copied. - see https://docs.lightkurve.org/api/lightkurve.lightcurve.LightCurve.html
```

The `UserWarning` appeared **whenever** `lc.meta` is assigned, not just the initial assignment in the above code.

#### Environment

-  lightkurve version (e.g. 1.0b6): 2.0.9 (should affect all older 2.0.x)

orionlee:
Closes #1046

Additional, outdated API links in various tutorials have been fixed too.

(API doc URLs were changed from 
`https://docs.lightkurve.org/api/...` to `https://docs.lightkurve.org/reference/api/...`

### DIFF
--- a/docs/source/tutorials/1-getting-started/what-are-targetpixelfile-objects.ipynb
+++ b/docs/source/tutorials/1-getting-started/what-are-targetpixelfile-objects.ipynb
@@ -97,7 +97,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "You can find the full list of properties in the [API documentation](https://docs.lightkurve.org/api/lightkurve.targetpixelfile.KeplerTargetPixelFile.html#lightkurve.targetpixelfile.KeplerTargetPixelFile) on this object."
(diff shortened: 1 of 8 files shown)


barentsen:
Thank you for fixing the API doc urls in addition to fixing the .meta warning!

The `--remote-data` unit tests appear to be failing due to a transient HTTPError, likely due to maintenance at the data archive, so we can safely ignore that failure

Thank you @orionlee!!

PR #1052, opened 2021-05-04
Include source in apidoc output

orionlee:
api doc output is tweaked to that it includes sources, similar to what astropy and various python library do.

Changes:
image

Click source:

image


barentsen:
Ohhhh, I like it.  This is very cool.  Thank you @orionlee!! :+1:

Issue #268, opened 2018-09-26
Tutorial overhaul


gully:
As lightkurve has grown, our docs have become outdated.  I propose a bit of an overhaul of the tutorials to bring them up to date with the most recent version.  

lightkurve contributor @ojhall94 has identified some specific bugs, below:

### How to recover a known planet in Kepler data?
 - Plots are no longer scatterplots as depicted in the tutorial
 - The package transit_periodogram, used in this tutorial, has been depricated according to its Github page.
 - In the plot function, need to define ax = ax as a kwarg, the tutorial just lists it as an arg. (This one crops up a lot)

### How to use lightkurve for asteroseismology?
- Do not get identical results to the tutorial, but the differences are negligible.
- Some unexplained terms: CDPP?
- Direct code copied from the tutorial doesn't work anymore: aperture_mask = aper required as a keyword when creating a light curve with a slightly larger aperture.

### How to make a supernova lightcurve?
- A second observation has been made of KSN2015k since the tutorial was made, campaign should be specified to 6.
- tpf interact is commented out in the tutorial example
- tpf interact doesn't seem to work for me, but could be my machine (we've discussed this already, it was the localhost issue)
- tpf interact doesn't work with lc_clean because lc_clean has had frames removed, and so has fewer cadences than the tpf
- None of the tqdm progress bars shown in the tutorial output are in the code anymore

### How to combine lightcurves from different Kepler quarters
- Window length of the flatten() method risks imposing systematics if its not set manually. Might be useful to clarify the meaning of this kwarg in the tutorial so people (like me!) don't learn to use it blindly

My two cents-- I'd like to see the figures rendered high dpi for retina displays, and we should decide whether it's desirable to show `%matplotlib inline` and other configuration lines on every single tutorial.  I propose that we do not, and instead make a new "Troubleshooting" tutorial that introduces how to make those sorts of configuration changes, and addresses questions about how to run demos outside the Jupyter notebook, etc.  We may also want to revisit the notebook naming convention moving forward.

gully:
Idea: should we add a conda `environment.yml` file for lightkurve?  The virtue of such a file resides primarily as a common quick-start environment for tutorials and developers.  

For example, if we want to aid students in running the documentation locally, they would need to clone this repo, then have Jupyter installed, the right bokeh version, etc.  It would be helpful to instruct students to simply run:

```
conda env create -f environment.yml
conda activate lightkurve
```

barentsen:
[...]

Issue #1032, opened 2021-04-22
Problems in creating development environment during poetry install


orionlee:
#### Problem description

I tried to create a development environment following [http://docs.lightkurve.org/about/contributing.html#contributing](the doc). `poetry install` resulted in an error saying `lightkurve is not a package`.

When I examined the installation,
- `lightkurve` is not on `pip list`, but the required modules are there (have been installed)
- when I tried to import `lightkurve` in a python shell, it is somewhat there, but not completely

```python
(lkv2_dev) c:\dev\lightkurve>python
Python 3.7.10 (default, Feb 26 2021, 13:06:18) [MSC v.1916 64 bit (AMD64)] :: Anaconda, Inc. on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import lightkurve
>>> lightkurve
<module 'lightkurve' (namespace)>

                       ^^^ for a normal module, its path would be shown, rather than (namespace)
```

Any idea on what's wrong?

environment: Windows 10, Anaconda, python 3.7.10

#### Example

```python

# start from a minimal python 3.7.10 environment
(lkv2_dev) c:\dev\lightkurve>python -m pip install --upgrade pip poetry

(lkv2_dev) c:\dev\lightkurve>poetry install
...
  • Installing sphinxcontrib-rawfiles (0.1.1)
  • Installing tqdm (4.60.0)
  • Installing uncertainties (3.1.5)

  ValueError

  lightkurve is not a package.

  at c:\pkg\_winnonportables\anaconda3\envs\lkv2_dev\lib\site-packages\poetry\core\masonry\utils\package_include.py:80 in check_elements
      76│                 self._package = root.name
      77│                 self._elements = sorted(list(root.glob("**/*")))  # type: List[Path]
      78│
      79│                 if not self.is_stub_only() and not self.has_modules():
    → 80│                     raise ValueError("{} is not a package.".format(root.name))
      81│
      82│                 self._is_package = True
      83│             else:
      84│                 self._package = root.stem


```

orionlee:
The problem is resolved by cloning a brand new git repository lightkurve, rather than reusing existing one.

Specifically, in my existing repository, it has a legacy `lightkurve` directory (the sources were moved from `lightkurve` to `src/lightkurve` a while back) that contained a bunch of local `.pyc` files. The legacy `lightkurve` directory seemed to confuse python and poetry, causing the import failure.

PR #1056, opened 2021-05-11
Add a conda environment.yml file for standardizing a quickstart dev env

gully:
In [this comment](https://github.com/lightkurve/lightkurve/issues/268#issuecomment-443364489) from 2018, I thought it'd be convenient to provide a quickstart developer environment to get students or developers started on a common, standardized, sandboxed python environment.  I still think that's a good idea!  So here is the start of a pull request for such an `environment.yml` file.  

This approach has some drawbacks.  First there are new cool ways to install Python dependencies, including poetry, etc.  I am not as familiar with these!  Second, we'd have to support this environment file and constantly keep it up-to-date with whatever new versions of dependencies we hinge upon.  That sounds like a nightmare.  So definitely some drawbacks.  But the plus side is that it serves as a quick go-to for many who want to try out a project without disturbing their existing python installs.

The location in the root of the repo need not endure.  We could put it as a gist or as a documentation resource for example.  The root of the repo is nice because someone new to the project who is familiar with conda envs can start that-much-more-easily.  But the root of the repo is already cluttered, so either way is fine.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,40 @@
+name: lightkurve_env
+channels:
+  - defaults
+  - conda-forge
+dependencies:
+  - python=3.7
+  - numpy
+  - scipy
+  - matplotlib
+  - astropy>=4.1
+  - astroquery>=0.3.10
+  - beautifulsoup4
+  - tqdm
+  - pandas
+  - bokeh
+  - jupyter
+  - oktopus
+  - ipython
+  - seaborn
+  - scikit-learn
+  - requests
+  - pytest
+  - sphinx # for documentation and testing
+  - nbsphinx # for documentation and testing
+  - numpydoc # for documentation and testing
+  - sphinx-automodapi # for documentation and testing
+  - black
+  - isort
+  - tox
+  - uncertainties
+  - mypy
+  - pip
+  - pip:
+      #- sphinxcontrib-rawfiles
+      #- twine
+      #- pydata-sphinx-theme
+      #- pytest-remotedata
+      #- pytest-doctestplus
+      #- pytest-cov
+      - fbpca


gully:
I should emphasize that the contents of `environment.yml` probably need some review--- is the list of developer dependencies complete?  Should some be pinned?  And if we include this, we should probably add some info to the documentation explaining how to use it.

barentsen:
Thanks for opening the PR @gully!

For this PR to be merged, I would like to see an entry included in Lightkurve's `Makefile` which would let us automatically update the `environment.yml` file in the future without extra manual work.  Can you help me figure out how to do this?

It looks like the `dephell` tool (`pip install dephell`) allows for conversion from `pyproject.toml` to `environment.yml` using the following syntax:
```bash
$ dephell deps convert --from-path pyproject.toml --from-format poetry --to-path environment.yml --to-format conda
```
...however I haven't tested the resulting file at all.  I notice that `conda-forge` does not appear as a channel in the resulting yml file, so a bit more research is needed to understand if and how this really works.  Would you be willing to test this route?


Note: we offer two options right now to install Lightkurve in an isolated environment.  I detail them below for reference:

### Option 1: using `venv` and `pip`

Python's builtin `pip` and `venv` features work very well these days because all popular packages now provide binary wheels (similar to conda packages).  Long gone are the days where we had to use `conda` because using `pip` meant you'd be compiling packages from source.
 
Example use:

```bash
$ python -m venv lkenv
$ source lkenv/bin/activate
$ python -m pip install lightkurve
```


### Option 2: using `poetry` to setup a dev environment

Poetry is useful for setting up a development environment in which you can actively edit and test Lightkurve source code.  I recently documented the steps needed to set up a Lightkurve dev environment with `poetry` here: 

https://docs.lightkurve.org/about/contributing.html#install-the-development-environment

orionlee:
As a data point, I have recently followed the `poetry` based instruction in conda environment (*) and got it work.
 https://docs.lightkurve.org/about/contributing.html#install-the-development-environment

The only caveat is that I ran into an issue because my local git repo still had legacy pre v2.0 release directories (issue #1032), which should not be a problem for new students / developers.

(*) My environment is Windows 10 + Anaconda .

gully:
Maybe the poetry solution is adequate after all, I come from a conda perspective but there are many other options now.  You're right that manually maintaining the environment file would be unsustainable, this dephell solution seems interesting.  I could test the resulting environment next chance I get.

barentsen:
Closing this issue for now, but happy to add an entry to the `Makefile` to make it easy to create a conda environment file, if anyone ends up figuring out how to make this work with `dephell`!

Issue #990, opened 2021-03-01
Couldn't select pixels of TPFs in interact()


NaveenTime:
**Problem description**
interact() is not working properly. I could not select the pixels in TPFs. When I try to select a pixel, the bottom row pixels automatically get selected. I could not select any pixel other than the bottom row pixels.

**Example**
```python
import lightkurve
tpf = lk.search_targetpixelfile("HL Tau", author='K2', campaign=13, cadence='long').download()
tpf.interact()
```

![Screenshot (354)](https://user-images.githubusercontent.com/65549947/109490657-8a697f80-7aae-11eb-9273-e8d9965ab510.jpg)
![Screenshot (356)](https://user-images.githubusercontent.com/65549947/109491164-3e6b0a80-7aaf-11eb-89b2-498b73482737.png)

**Expected behavior**
When I click a pixel in the tpf stamp,  the pixel actually should get selected. But it doesn't. Instead, the pixels of the lowest row get selected. I couldn't select any pixel other than the pixels of the lowest row.

**Environment**
-  Windows
-  lightkurve version 2.0.3
-  installation method - pip
-  Executed the code in Python 3 in JupyterNotebook (version 6.2.0)

barentsen:
Oops, that sounds like a bug! :bug:

Thank you for taking the time to report this, @NaveenTime! :pray: 

@gully Would you be interested diagnosing the issue?  It may or may not be related to Lightkurve v2 changes, so a good first step may be to compare whether the bug can be reproduced in v1.11 vs v2.0.

gully:
Hi @NaveenTime thanks for your interest in lightkurve especially my favorite tool `.interact()`! 🚀 

My first instinct was that it could be a Windows issue, since Windows may not have as-extensive user testing, so I did a fresh install of lightkurve on Windows 10 and was not able to reproduce this erroneous interact pixel-selection bug--- the pixels selected as intended.  

I seem to recall having some strange selection problems with dual monitors a while ago.  Are you using dual monitors?

gully:
What browser are you using for Jupyter?  As another workaround and debugging step, I would recommend seeing if the bug reproduces in other browsers. I found no bug in Firefox with `bokeh` version `2.2.3`

NaveenTime:
Hi @gully . I am not using dual monitors. I am using Jupyter Notebook in Google Chrome browser in Windows 10 with bokeh version 2.3.0.  I can reproduce the bug in Firefox also.

[...]

PR #1057, opened 2021-05-11
[WIP] An attempted fix at Issue #990, currently not working

gully:
:construction: Under construction!  
This PR is aimed at fixing #990 by flattening the `x`, `y` pixels matrix into vectors to be compliant with a `ColumnDataSource`.

I emphasize this PR is not yet working.  It does not resolve the problem, but it's a start in the right direction.  Anyone who wants to help feel free to take over the PR or message me for a pair coding session.  :pray:

### DIFF
--- a/src/lightkurve/interact.py
+++ b/src/lightkurve/interact.py
@@ -119,12 +119,15 @@ def prepare_tpf_datasource(tpf, aperture_mask):
         Bokeh object to be shown.
     """
     npix = tpf.flux[0, :, :].size
-    pixel_index_array = np.arange(0, npix, 1).reshape(tpf.flux[0].shape)
-    xx = tpf.column + np.arange(tpf.shape[2])
-    yy = tpf.row + np.arange(tpf.shape[1])
+    ncadences, ny, nx = tpf.shape
+    pixel_index_array = np.arange(0, npix, 1)
+    xx = tpf.column + np.arange(nx)
+    yy = tpf.row + np.arange(ny)
     xa, ya = np.meshgrid(xx, yy)
+    xa = xa.flatten()
+    ya = ya.flatten()
     tpf_source = ColumnDataSource(data=dict(xx=xa.astype(float), yy=ya.astype(float)))
-    tpf_source.selected.indices = pixel_index_array[aperture_mask].reshape(-1).tolist()
+    tpf_source.selected.indices = pixel_index_array[aperture_mask.reshape(-1)].tolist()
     return tpf_source
 
 
@@ -750,10 +753,11 @@ def show_interact_widget(
             tpf, selected_pixel_indices, transform_func=transform_func
         ):
             """Create the lightcurve from the selected pixel index list"""
+            tpf.flux[0].shape
             selected_indices = np.array(selected_pixel_indices)
-            selected_mask = np.isin(pixel_index_array, selected_indices)
-            lc_new = tpf.to_lightcurve(aperture_mask=selected_mask)
-            lc_new.meta["APERTURE_MASK"] = selected_mask
+            selected_mask = np.isin(pixel_index_array.reshape(-1), selected_indices.reshape(-1))
+            lc_new = tpf.to_lightcurve(aperture_mask=selected_mask.reshape(tpf.flux[0].shape))
+            lc_new.meta["APERTURE_MASK"] = selected_mask.reshape(tpf.flux[0].shape)
             if transform_func is not None:
                 lc_transformed = transform_func(lc_new)
                 if len(lc_transformed) != len(lc_new):
@@ -765,7 +769,7 @@ def show_interact_widget(
                     )
                 else:
                     lc_new = lc_transformed
-                    lc_new.meta["APERTURE_MASK"] = selected_mask
+                    lc_new.meta["APERTURE_MASK"] = selected_mask.reshape(tpf.flux[0].shape)
             return lc_new
 
         def update_upon_pixel_selection(attr, old, new):


gully:
The key idea for `.interact()` is that we show the tpf image, and overplotted on that are invisible rectangles---one for each pixel.  The rectangles become visible when either someone clicks on them, or by default if they are in the `aperture_mask`.  When the user taps each rectangle, they go from invisible to visible.  In the code that means getting added to the "selected_indices" list.  The hiccup comes because we used to be able to hand in a 2D array to this "selected_indices" list (technically an abuse of bokeh), making it straightforward to map a 2D image to a list of 2D pixels.  The bokeh updates are a bit more scrupulous about accepting only columns of data (1D) and not matrices of data (2D)---a completely reasonable expectation for a data source object with "Column" in its name.  That's no problem:  we can simply reshape the 2D into 1D.  In practice this requires some careful look at all the places that used to be 2D, and make them 1D, up until the last moment when lightkurve applies the new aperture_mask, when we need to reshape it to 2D again.  

That's what this PR is doing--- going through the interact code to spot check the places where 2D arrays need to become 1D.

orionlee:
The PR actually works in my environment with bokeh 2.3.2.

I have some setup problems in switching bokeh version such that `interact()` is not rendered at all.
In browser's developer tools console, it says there is a version mismatch that JS is still in 2.2.3 and Python is 2.3.2 .

In this specific case, I got it work by creating a new notebook and run the codes from there. (*)

(*) In the past, I often got by the JS/Python version mismatch problem by a combination of clearing the book's widget state, reloading the notebook, etc.  But somehow it didn't work this time.

I have a few other minor comments - it might just be that this is intended to be a draft so the codes are not yet cleaned up.

src/lightkurve/interact.py
```diff
@@ -119,12 +119,15 @@ def prepare_tpf_datasource(tpf, aperture_mask):
...
+    ncadences, ny, nx = tpf.shape
+    pixel_index_array = np.arange(0, npix, 1)
+    xx = tpf.column + np.arange(nx)
+    yy = tpf.row + np.arange(ny)
     xa, ya = np.meshgrid(xx, yy)
+    xa = xa.flatten()
+    ya = ya.flatten()
     tpf_source = ColumnDataSource(data=dict(xx=xa.astype(float), yy=ya.astype(float)))
-    tpf_source.selected.indices = pixel_index_array[aperture_mask].reshape(-1).tolist()
+    tpf_source.selected.indices = pixel_index_array[aperture_mask.reshape(-1)].tolist()
```
orionlee, line 130: in some lines, flatten is done by `.flatten()` while some others, it is done by `.reshape(-1)`. 
It'd be better if one form is used consistently.

src/lightkurve/interact.py
```diff
@@ -750,10 +753,11 @@ def _create_lightcurve_from_pixels(
             tpf, selected_pixel_indices, transform_func=transform_func
         ):
             """Create the lightcurve from the selected pixel index list"""
+            tpf.flux[0].shape
```
orionlee, line 756: this new line has no effect. Maybe leftover from some debug codes?

src/lightkurve/interact.py
```diff
@@ -750,10 +753,11 @@ def _create_lightcurve_from_pixels(
...
             selected_indices = np.array(selected_pixel_indices)
-            selected_mask = np.isin(pixel_index_array, selected_indices)
-            lc_new = tpf.to_lightcurve(aperture_mask=selected_mask)
-            lc_new.meta["APERTURE_MASK"] = selected_mask
+            selected_mask = np.isin(pixel_index_array.reshape(-1), selected_indices.reshape(-1))
```
orionlee, line 758: 1. `pixel_index_array` and `selected_indices` are already in 1d. The additional reshape is unncessary.
2. We might want to rename `selected_mask` to `selected_mask_1d`, so make it clear this is a 1d version. Or we could convert it to 2D right here.

src/lightkurve/interact.py
```diff
@@ -119,12 +119,15 @@ def prepare_tpf_datasource(tpf, aperture_mask):
...
     npix = tpf.flux[0, :, :].size
-    pixel_index_array = np.arange(0, npix, 1).reshape(tpf.flux[0].shape)
-    xx = tpf.column + np.arange(tpf.shape[2])
-    yy = tpf.row + np.arange(tpf.shape[1])
+    ncadences, ny, nx = tpf.shape
+    pixel_index_array = np.arange(0, npix, 1)
+    xx = tpf.column + np.arange(nx)
+    yy = tpf.row + np.arange(ny)
     xa, ya = np.meshgrid(xx, yy)
+    xa = xa.flatten()
```
orionlee, line 127: Maybe a 1 line comment would be helpful, explaining why `xa`, `ya` and `tpf_source.selected.indices` are flattened here, when someone read the codes say a year from now.

orionlee:
@gully from what I reviewed and traced, the PR as-is does correctly convert 2D apertures to 1D selection and back. So I don't quite understand why it does not work for you. Have you confirmed it is not due to unrelated bokeh JS / Python version mismatch (if you changed bokeh version in your environment)?

barentsen:
Good news: it looks like this issue has been resolved via #1177!

gully:
Huzzah!!! Thank you so much @orionlee !! This bug has bitten many and I feel remorse that I did not further this stalled PR, and delighted that you did!  
Thank you x10! 🙏

orionlee:
I came across a few new lightkurve users who got bitten by it, wasted some time to provide workaround, and became motivated to close the issue.

PR #1058, opened 2021-05-13
Fix test_search_tesscut error

orionlee:
It should fix the persistent remote-data test failure in CI.

Somehow the expected TESSCut product "TIC 206669860", sector=2 is no longer there. There is one from sector 28. Changed the test code to use the one that exists.

BTW, is that something that should happen (a product seems to be removed from TESSCut service) at all on TESSCut side?

### DIFF
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -147,7 +147,7 @@ def test_search_tesscut():
     assert len(search_tesscut("pi Mensae", sector=1).table) == 1
     assert len(search_tesscut("pi Mensae").table) > 1
     # Cutout by TIC ID
-    assert len(search_tesscut("TIC 206669860", sector=2).table) == 1
+    assert len(search_tesscut("TIC 206669860", sector=28).table) == 1
     # Cutout by RA, dec string
     search_string = search_tesscut("30.578761, -83.210593")
     # Cutout by SkyCoord


barentsen:
Thanks for spotting this @orionlee!!

What's happening here is that MAST is currently replacing the TESS Sector 2 data with a newly re-processed version.  We can tell this based on the [TESS Holdings Status Page](https://outerspace.stsci.edu/display/TESS/TESS+Holdings+Available+by+MAST+Service).

As part of this process, Sector 2 data temporarily disappears from the various services.  This may be the case for Sector 3 data next week etc.

I think I'll go ahead and merge this PR because it's nice to have the unit tests pass, but of course we might see temporary failures elsewhere next week, so this type of issue will likely re-appear.

I see three more permanent solutions:
1. We could decide not to run the unit tests that utilize MAST for each PR, but then again, the data search features are very important in Lightkurve and we want to make sure they don't break. 
2. We could implement the MAST-based unit tests in a way that utilizes a permanent cache or mock function, and only occasionally run the tests against the real services.  The [casette](https://cassette.readthedocs.io/en/latest/) or `diskcache` packages may help us do this, but it risks making the tests much more complicated and we risk overlooking real bugs.
3. We could accept that the "remote data" tests frequently fail.  Perhaps we could configure GitHub Actions to show the test status as "green" even if the remote data tests fail.

orionlee:
On remote data testing failure: I haven't thought through various solutions and their trade off yet, other than that I feel we do need to do some test search / download against MAST. It is a critical integration point that we want to ensure it works.

However, but for the narrower class of the problem that led to this PR: test failures when data products are removed from MAST temporarily (could be for days though),
- it is an issue as long as we agree that some testing on search / download against MAST is needed.
- can we assume the issue applies primarily to TESS, i.e., Kepler / K2 related products are quite stable?
- if so, are there a certain set of TESS products that are stable that we can count on? E.g., are there some test (or maybe simulation) products that are likely to be stable? 
- if the answer is yes, we could possibly tweak the tests (in `test_search.py`) to use those products.

PR #1064, opened 2021-05-17
Fix tesscut dowload phantom error with downlooad_dir 

### LINKED ISSUE #1063: search_tesscut() download partially failed when download_dir is specified
#### Problem description

When trying to download TESSCut with "download_dir" specified, the codes 
1. will throw `FileNotFoundError`,
2. the file is actually there. If you rerun the same code, it will return the TESSCut tpf (correctly reading the TESSCut from the location specified the `download_dir`)

#### Example
```python
import lightkurve as lk

tpf = lk.search_tesscut("LTT 1445 A", sector=4).download(download_dir="data", cutout_size=11) # TessCut
```

It will result in
```python
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-200-9e807cdff825> in <module>
---> 16 tpf = lk.search_tesscut("LTT 1445 A", sector=4).download(download_dir="data", cutout_size=11) # TessCut
...
C:\dev\lightkurve\src\lightkurve\utils.py in wrapper(*args, **kwargs)
    553             sys.stdout = devnull
    554             try:
--> 555                 return f(*args, **kwargs)
    556             # restore to default
    557             finally:

C:\dev\lightkurve\src\lightkurve\search.py in download(self, quality_bitmask, download_dir, cutout_size, **kwargs)
    440             download_dir=download_dir,
    441             cutout_size=cutout_size,
--> 442             **kwargs,
    443         )
    444 

C:\dev\lightkurve\src\lightkurve\search.py in _download_one(self, table, quality_bitmask, download_dir, cutout_size, **kwargs)
    335 
    336             return read(
--> 337                 path, quality_bitmask=quality_bitmask, targetid=table[0]["targetid"]
    338             )
    339 
...
FileNotFoundError: [Errno 2] No such file or directory: 'data\\data\\tesscut\\tess-s0004-2-4_45.464137_-16.593340_11x11_astrocut.fits'
```

Rerun the same line and it will work.

In the initial download, the path of tesscut file is incorrect, as if `download_dir` is appended twice in the path.

#### Environment

-  lightkurve version (e.g. 1.0b6): 2.0.9 (*)

(*) I ran into the same issue back in lightkurve v 1.x, and just never got around to report it.

orionlee:
Closes #1063

The old code for generating the path
```python
 os.path.join(download_dir, cutout_path[0][0])
```
would append an addtional download_dir.

It works in typical case, because in typical case the download dir is an absolute path, e.g. `/usr/lk/.lightkurve-cache`. The os path join's behavior is such that it will detect that the path is already part of the argument, and won't join it again. 

In other words, the `join()` function is smart enough to hide the bug in the case where the path is absolute.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,9 @@ htmlcov/
 nosetests.xml
 coverage.xml
 *,cover
+# temp directory generated in test
+#  (that should have been cleaned up automatically in nomral test run)
+temp_lk_cache_4test_*
 
 # Translations
 *.mo
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,9 @@
 - Fixed a bug in ``TargetPixelFile.interact()`` when ``ylim_func`` is specified and
   users tap to select custom apertures. [#1033]
 
+- Fixed a bug in ``search_tesscut().download()`` when ``download_dir`` is specified.
+  [#1063]
+
 
 2.0.9 (2021-03-31)
 ==================
--- a/src/lightkurve/search.py
+++ b/src/lightkurve/search.py
@@ -531,7 +531,7 @@ class SearchResult(object):
             cutout_path = TesscutClass().download_cutouts(
                 coords, size=cutout_size, sector=sector, path=tesscut_dir
             )
-            path = os.path.join(download_dir, cutout_path[0][0])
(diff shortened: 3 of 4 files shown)


barentsen (approved):
This looks perfect.  Thank you @orionlee! :+1:

PR #1065, opened 2021-05-18
Mark remote data for applicable tests

### LINKED ISSUE #1062: Mark all tests that require external data with remote_data
#### Problem description

Should we mark all the tests that require external data on internet with `@pytest.mark.remote_data`?

There are tests that do not appear to have external data dependency on surface, but actually do.
E.g., testing of PRF Model is not marked as remote_data, but it does rely as the implementation rely on reading calibration file hosted on stsci.edu.

As an example, when the following test run (non remotedata) failed, it is not immediately apparent that the error is a transient error due to stsci.edu not available:
https://github.com/lightkurve/lightkurve/pull/1050/checks?check_run_id=2603894239

It is apparent once one digs through the log:
```python
tests\seismology\test_stellar_estimators.py .......                      [100%]

================================== FAILURES ===================================
_______________ test_keplerprf_gradient_against_simplekeplerprf _______________

self = <urllib.request.HTTPSHandler object at 0x000001F275311790>
http_class = <class 'http.client.HTTPSConnection'>
req = <urllib.request.Request object at 0x000001F275311070>
...

tests\prf\test_prfmodel.py:79: 
src\lightkurve\prf\prfmodel.py:71: in __init__
    ) = self._prepare_prf()
src\lightkurve\prf\prfmodel.py:260: in _prepare_prf
    ) = self._read_prf_calibration_file(prffile, i + 1)
src\lightkurve\prf\prfmodel.py:216: in _read_prf_calibration_file
    prf_cal_file = pyfits.open(path)    
```

orionlee:
Closes #1062 

Ready, pending discussion on exactly what we should do:
1. The PR currently will make all PRF / TPF Model tests run only on remote-data, currently linux 3.8 in the CI. I suspect we still want the logic underlying PRF / TPF models be tested in a variety of OS/Python version. Thoughts?

2. Outside of tests involving PRF / TPF models, there are 4 tests marked as remote-data. I think most, if not all, can be converted to using local data.
  i. 3 tests in `test_interact.py`: they all rely on the remote Kepler `TABBY_TPF` file. We could replace it with a local test FIS file with 1st cadence only. If there is concern that 1st cadence only is not realistic enough, we could use a local FITS file , say, 1 day worth of cadence.
  ii. `test_lightcurve.py::test_regression_346()`: it relies on a remote `K2_C08` lc file. The test seems to need test files with specific characteristics. So a simple trim to first cadence would not do.

3. I tried to tweak the CI so that we have a test config that simulates environments without network, so that any future tests missing `remote-data` mark would be exposed right away during CI tests. The tweak, which involves `pytest-socket` plugin, unfortunately doesn't fully  work so I backed it out. It is blocked by https://github.com/miketheman/pytest-socket/issues/65.

### DIFF
--- a/tests/prf/test_prfmodel.py
+++ b/tests/prf/test_prfmodel.py
@@ -12,6 +12,7 @@ from lightkurve.prf import KeplerPRF, SimpleKeplerPRF
 from lightkurve.targetpixelfile import KeplerTargetPixelFile
 
 
+@pytest.mark.remote_data  # PRF relies on calibration files on stsci.edu
 def test_prf_normalization():
     """Does the PRF model integrate to the requested flux across the focal plane?"""
     for channel in [1, 20, 40, 60, 84]:
@@ -26,6 +27,7 @@ def test_prf_normalization():
                 assert np.isclose(prf_sum, flux, rtol=0.1)
 
 
+@pytest.mark.remote_data  # PRF relies on calibration files on stsci.edu
 def test_simple_kepler_prf():
     """Ensures that concentric PRFs have the same values."""
     prf_1 = SimpleKeplerPRF(channel=16, shape=[20, 20], column=0, row=0)
@@ -54,6 +56,7 @@ def test_simple_kepler_prf_interpolation_consistency():
     np.isclose(np.sum(np.abs(sprf_data - cal_prf_subsampled_normalized)), 0)
 
 
+@pytest.mark.remote_data  # PRF relies on calibration files on stsci.edu
 def test_get_model_prf():
     tpf_fn = get_pkg_data_filename("../../tests/data/test-tpf-star.fits")
     tpf = KeplerTargetPixelFile(tpf_fn)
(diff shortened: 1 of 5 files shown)


barentsen:
@orionlee Thank you for fixing the missing `@pytest.mark.remote_data`!

> I suspect we still want the logic underlying PRF / TPF models be tested in a variety of OS/Python version. Thoughts?

One solution would be to introduce a separate "pre-release CI workflow" which uses the `--remote-data` flag on many more platforms/versions but is only run once in a while (e.g. as part of the [Lightkurve release procedure](https://docs.lightkurve.org/about/release-procedure.html), rather than for *every* commit in a PR).

> I think most, if not all, can be converted to using local data.

It's definitely true that we could convert more tests to use local files.  This would be a good improvement, even if we don't convert 100%.

>  I tried to tweak the CI so that we have a test config that simulates environments without network

I like this idea!!  I will keep an eye on the issue you opened over in `pytest-socket`.

barentsen:
Re-running the tests because the remote-data checks encountered a (transient?) TessCut error.

I intend to merge this PR if the checks pass and perhaps open a new issue to document the ongoing discussion.

PR #1066, opened 2021-05-18
interact_sky minor tweak: improve proper motion correction warning

orionlee:
Follow up on the comment for the warning in #1050
https://github.com/lightkurve/lightkurve/pull/1050#pullrequestreview-661600676

The PR changes the warning so that it is specific to the use case. 

E.g., when one displays a TESSCut tpf or (many if not all) Kepler tpf files, a warning would be issued:

```python
interact.py:503: LightkurveWarning: Proper motion correction cannot be applied to the target, as none is available. Thus the target (the cross) might be noticeably away from its actual position, if it has large proper motion.
```


barentsen:
Thank you for improving the warning @orionlee.  I like it!!

PR #1072, opened 2021-05-20
add flux column option to LombScarglePeriodogram.from_lightcurve

keatonb:
This very minor change allows a user to specify in a call to `to_periodogram` which column to compute the periodogram of. Default behavior is the same, but now that user columns can be added to LightCurve objects, the periodograms of these quantities may be of interest too.  Pretty please accept this so I can utilize it in my [Pyriod](https://github.com/keatonb/Pyriod) project :)

### DIFF
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -2086,7 +2086,7 @@ class LightCurve(QTimeSeries):
         ``minimum_frequency``, ``maximum_frequency``, ``mininum_period``,
         ``maximum_period``, ``frequency``, ``period``, ``nterms``,
         ``nyquist_factor``, ``oversample_factor``, ``freq_unit``,
-        ``normalization``, ``ls_method``.
+        ``normalization``, ``ls_method``, ``flux_column``.
 
         Optional keywords accepted if ``method='bls'`` are
         ``minimum_period``, ``maximum_period``, ``period``,
--- a/src/lightkurve/periodogram.py
+++ b/src/lightkurve/periodogram.py
@@ -653,6 +653,7 @@ class LombScarglePeriodogram(Periodogram):
         freq_unit=None,
         normalization="amplitude",
         ls_method="fast",
+        flux_column="flux",
         **kwargs
     ):
         """Creates a `Periodogram` from a LightCurve using the Lomb-Scargle method.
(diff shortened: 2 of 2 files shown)


barentsen:
Thanks for opening this PR @keatonb!

Idea: instead of adding a `flux_column` parameter to `to_periodogram()`, do you think it would be acceptable if we added a `LightCurve.set_flux()` method instead, so you could do the following:
```python
lc.set_flux("your_flux_column").to_periodogram(...)
```

The reason I ask is because my intention for Lightkurve v2 has been to make the `time`, `flux`, and `flux_err` columns special in the sense that:
* "*they shall be the columns upon which all light curve operations shall operate*"
* "*they shall always be present*" (populated with NaN values if need be)

The benefit of adopting these commandments is that it simplifies a lot of code, i.e., you never have to check if the flux column exists, and there is no need to add a `flux_column` parameter to every method.

My worry is that if we add a `flux_column` to the `to_periodogram` method, we're going to end up adding it to every LightCurve method, which adds a lot of extra code to maintain and extra surfaces for bugs to grow.

For completeness, I'll mention that the current syntax for creating a periodogram from a custom flux column is to do the following:
```python
lc.flux = lc.your_flux_column
lc.flux_err = lc.your_flux_column_error
pg = lc.to_periodogram()
```
...but this is definitely not very fluent or user-friendly, so I do think adding `set_flux()` is overdue at minimum!

keatonb:
Hey Geert,

Yeah, your proposed solution to add a set_flux() function would work fine, and I agree that would be useful in other applications.  I suppose the additional overhead will hardly be noticeable.

I will note that this PR really adds a `flux_column` parameter to the `from_lightcurve` method of the `LombScarglePeriodogram` class, so not so clear a violation of the commandments, and it probably wouldn't creep into other LightCurve methods, because it isn't one, really.  But `set_flux()` would be enough for me.

Let me know how I can help to get this into the next release expediently as I'm itching to release the next version of Pyriod that will take full advantage of this feature (now instead of have three separate `LightCurve` objects holding the original light curve, the sampled model, and the residuals, they are just three columns of a single `LightCurve` object---I love it!).

barentsen:
I hope to get this done later today!   In meanwhile, you can of course use the workaround in your code, i.e.:

```
lc.flux = lc.your_flux_column
lc.flux_err = lc.your_flux_column_error
```

keatonb:
Closing because #1076 is provides an alternative.

PR #1075, opened 2021-05-21
API doc fixes - collections, MPLSTYLE, windows build

### LINKED ISSUE #898: Lightcurve plot methods: linecolor not according to style when ax is supplied
#### Problem description

For `Lightcurve.scatter()` and other plot methods, when `ax` is supplied as an argument, the line color is not `Lightkurve`'s default black. Instead it is blue-ish color from matplotlib.

#### Example

```python
ax = plt.figure(figsize=(6, 3)).gca()
# scatter color is blue-ish (matplotlib default), rather than lightkurve's default black
lc.scatter(ax=ax) 

lc.scatter();
```

image

#### Environment

-  lightkurve version (e.g. 1.0b6): 1.11.3

### LINKED ISSUE #1068: TargetPixelFileCollection and LightCurveCollection missing in apidoc
#### Problem description
`TargetPixelFileCollection` and `LightCurveCollection` are not included in apidoc.
They are mentioned in `download_all()`, but there is no link to the actual classes.

I suggest we can include them in "Downloading data" section of the api doc, if we don't want to create a section for them.

#### Environment

orionlee:
Closes #1068 (for collections)
Closes #898 (`MPLSTYLE`)

Couple other minor api doc bug fixes
- case of meta keyword access
- windows build

### DIFF
--- a/src/lightkurve/search.py
+++ b/src/lightkurve/search.py
@@ -366,7 +366,7 @@ class SearchResult(object):
     ):
         """Download and open all data products in the search result.
 
-        This method will returns a `~lightkurve.TargetPixelFileCollection` or
+        This method will return a `~lightkurve.TargetPixelFileCollection` or
         `~lightkurve.LightCurveCollection`.
 
         Parameters
--- a/src/lightkurve/search.py
+++ b/src/lightkurve/search.py
@@ -366,8 +366,8 @@ class SearchResult(object):
     ):
         """Download and open all data products in the search result.
 
-        This method will returns a `~lightkurve.collections.TargetPixelFileCollection` or
-        `~lightkurve.collections.LightCurveCollection`.
+        This method will returns a `~lightkurve.TargetPixelFileCollection` or
+        `~lightkurve.LightCurveCollection`.
 
         Parameters
         ----------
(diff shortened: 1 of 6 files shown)


orionlee:
Collections: added to Downloading data section.

image

---

`MPLSTYLE` documentation: added to the LightCurve > plot subsection.

image

image

orionlee:
`MPLSTYLE` doc current contains the full local path of the stylesheet, which is probably not we want.

There seems to be a way to suppress the value in [autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autodata). But MPLSTYLE's rst doc is generated by autosummary. I don't know how to override it.

src/lightkurve/search.py
```diff
@@ -366,8 +366,8 @@ def download_all(
     ):
         """Download and open all data products in the search result.
 
-        This method will returns a `~lightkurve.collections.TargetPixelFileCollection` or
-        `~lightkurve.collections.LightCurveCollection`.
+        This method will returns a `~lightkurve.TargetPixelFileCollection` or
```
barentsen, line 369: Tiny typo:
```suggestion
        This method will return a `~lightkurve.TargetPixelFileCollection` or
```

barentsen:
Thank you for these useful changes!

It's annoying that MPLSTYLE contains the full local path for whoever compiled the docs, but I'm happy to live with this for now, because the alternative is to not document this feature!

PR #1076, opened 2021-05-21
Add a `select_flux()` method to the LightCurve class

barentsen:
This PR intends to address #1072 by adding a `set_flux` method to the `LightCurve` class.

This method enables users to assign a different column to act as the main `flux` (and `flux_err`) column, which is what most `LightCurve` methods operate on by default.

For example, the syntax for assigning the values of the `sap_flux` column to the main `flux` column will now be:

```python
lc = lc.set_flux(flux_column="sap_flux")
```

Optionally, the error column can be specified, though it already defaults to `f"{flux_column}_err"` (if available) for convenience, so specifying it is not actually necessary in this example:

```python
lc = lc.set_flux(flux_column="sap_flux", flux_err_column="sap_flux_err")
```

The main reason for adding `set_flux` is that it remove the need for every LightCurve method to accept and support a `flux_column` parameter. For example, a periodogram can now be created from `sap_flux` data as follows:

```python
import lightkurve as lk
lc = lk.search_lightcurve("Pi Men", sector=1, author="SPOC").download()
lc.set_flux("sap_flux").to_periodogram("lombscargle").plot()
```

Does anyone have questions/thoughts/concerns?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,9 @@
 - Fixed a bug in ``LightCurveCollection`` and ``TargetPixelFileCollection``, where
   indexing by slice, e.g., ``lc_collection[0:2]``, incorrectly returns a ``list`` [#1077]
 
+- Added the ``LightCurve.select_flux()`` method to make it easier to use a different
+  column to populate the ``flux`` and ``flux_err`` columns. [#1076]
+
 
 2.0.9 (2021-03-31)
 ==================
--- a/docs/source/reference/lightcurve.rst
+++ b/docs/source/reference/lightcurve.rst
@@ -83,7 +83,7 @@ The following methods all return a new `LightCurve` object.
   LightCurve.normalize
   LightCurve.remove_nans
   LightCurve.remove_outliers
-  LightCurve.set_flux
+  LightCurve.select_flux
   LightCurve.tail
   LightCurve.truncate
 
--- a/docs/source/reference/lightcurve.rst
+++ b/docs/source/reference/lightcurve.rst
@@ -83,6 +83,7 @@ The following methods all return a new `LightCurve` object.
   LightCurve.normalize
   LightCurve.remove_nans
   LightCurve.remove_outliers
+  LightCurve.set_flux
   LightCurve.tail
   LightCurve.truncate
 
--- a/src/lightkurve/io/generic.py
+++ b/src/lightkurve/io/generic.py
@@ -115,5 +115,6 @@ def read_generic_lightcurve(
     tab.meta["RA"] = hdulist[0].header.get("RA_OBJ")
     tab.meta["DEC"] = hdulist[0].header.get("DEC_OBJ")
     tab.meta["FILENAME"] = filename
+    tab.meta["FLUX_ORIGIN"] = flux_column
 
     return LightCurve(time=time, data=tab)
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -63,8 +63,8 @@ class QTimeSeries(TimeSeries):
         """
         # string-typed columns should not have a unit, or it will make convert_col_for_table crash!
         # see https://github.com/lightkurve/lightkurve/pull/980#issuecomment-806178939
-        if hasattr(col, "dtype"):
-            if hasattr(col, "unit") and col.dtype.kind in {"U", "S"}:
+        if hasattr(col, 'dtype'):
+            if hasattr(col, 'unit') and col.dtype.kind in {'U', 'S'}:
                 del col.unit
 
         col = super()._convert_col_for_table(col)
@@ -366,7 +366,7 @@ class LightCurve(QTimeSeries):
                 name not in self.__dict__
                 and not name.startswith("_")
                 and not self._new_attributes_relax
-                and name != "meta"
+                and name != 'meta'
             ):
                 warnings.warn(
                     (
@@ -387,7 +387,7 @@ class LightCurve(QTimeSeries):
         result = f"<{self.__class__.__name__}"
         if "LABEL" in self.meta:
             result += f" LABEL=\"{self.meta.get('LABEL')}\""
-        for kw in ["QUARTER", "CAMPAIGN", "SECTOR", "AUTHOR"]:
+        for kw in ["QUARTER", "CAMPAIGN", "SECTOR", "AUTHOR", "FLUX_ORIGIN"]:
             if kw in self.meta:
                 result += f" {kw}={self.meta.get(kw)}"
         result += ">"
@@ -402,7 +402,7 @@ class LightCurve(QTimeSeries):
             descr_vals.append("length={}".format(len(self)))
             if "LABEL" in self.meta:
                 descr_vals.append(f"LABEL=\"{self.meta.get('LABEL')}\"")
-            for kw in ["QUARTER", "CAMPAIGN", "SECTOR", "AUTHOR"]:
+            for kw in ["QUARTER", "CAMPAIGN", "SECTOR", "AUTHOR", "FLUX_ORIGIN"]:
                 if kw in self.meta:
                     descr_vals.append(f"{kw}={self.meta.get(kw)}")
         return super()._base_repr_(html=html, descr_vals=descr_vals, **kwargs)
@@ -437,8 +437,12 @@ class LightCurve(QTimeSeries):
     def flux_err(self, flux_err):
         self["flux_err"] = flux_err
 
-    def set_flux(self, flux_column, flux_err_column=None):
+    def select_flux(self, flux_column, flux_err_column=None):
         """Assign a different column to be the flux column.
+
+        This method returns a copy of the LightCurve in which the ``flux``
+        and ``flux_err`` columns have been replaced by the values contained
+        in a different column.
 
         Parameters
         ----------
@@ -454,6 +458,14 @@ class LightCurve(QTimeSeries):
         -------
         lc : LightCurve
             Copy of the ``LightCurve`` object with the new flux values assigned.
+
+        Examples
+        --------
+        You can use this function to change the flux data on which most Lightkurve
+        features operate.  For example, to view a periodogram based on the "sap_flux"
+        column in a TESS light curve, use::
+
+            >>> lc.select_flux("sap_flux").to_periodogram("lombscargle").plot()  # doctest: +SKIP
         """
         # Input validation
         if flux_column not in self.columns:
@@ -466,11 +478,15 @@ class LightCurve(QTimeSeries):
         if flux_err_column:  # not None
             lc["flux_err"] = lc[flux_err_column]
         else:
-            flux_err_column = flux_column + "_err"
+            # if `flux_err_column` is unspecified, we attempt to use
+            # f"{flux_column}_err" if it exists
+            flux_err_column = f"{flux_column}_err"
             if flux_err_column in lc.columns:
                 lc["flux_err"] = lc[flux_err_column]
             else:
                 lc["flux_err"][:] = np.nan
+
+        lc.meta['FLUX_ORIGIN'] = flux_column
         return lc
 
     # Define deprecated attributes for compatibility with Lightkurve v1.x:
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -63,8 +63,8 @@ class QTimeSeries(TimeSeries):
         """
         # string-typed columns should not have a unit, or it will make convert_col_for_table crash!
         # see https://github.com/lightkurve/lightkurve/pull/980#issuecomment-806178939
-        if hasattr(col, 'dtype'):
-            if hasattr(col, 'unit') and col.dtype.kind in {'U', 'S'}:
+        if hasattr(col, "dtype"):
+            if hasattr(col, "unit") and col.dtype.kind in {"U", "S"}:
                 del col.unit
 
         col = super()._convert_col_for_table(col)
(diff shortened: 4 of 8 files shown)


orionlee:
The direction sounds good to me. I have my own helpers doing similar things - so it looks like I am not alone.

Thoughts:
- should we deprecate the `column` parameter in `plot()`, etc.? Or are we providing 2 ways of doing the same thing? The `column` is more limited though.
- I propose we make `LightCurve` object keeping track of the origin column of flux, so that users or codes have ways to know the type of flux in a given object. E.g.
```pyhton
>>> lc1 = lc.set_flux("sap_flux")
>>> print(lc1.meta["FLUX_ORIGIN_COLUMN"])  # set_flux keeps track of the column in metadata
'sap_flux'
```
One immediate usage could be that in plotting methods: the plot can remind users the type of flux it is showing.

E.g., the existing `lc.scatter(column=...)` functions indicate the type of flux in y-axis label
image

- The name, `set_flux()`, is a bit misleading: I thought it is modifying the `LightCurve` object itself, and didn't realize it's creating a new object until I browse the actual code changes. Some suggestion:
  - `lc.to_lc(flux_column="sap_flux")`
  - `lc.with_flux("sap_flux")`

orionlee:
A loosely related subject: I noticed the commits have some lines tweaked, probably by black formatter.

If we are committed to making the code style consistent, I suggest we encourage and/or coerce PRs be formatted in black.
1. Encourage developers to have their codes formatted by commit time. It looks like we can ask them to setup some [pre commit hooks](https://github.com/adrn/delicatessen/blob/master/.pre-commit-config.yaml) somewhat easily.
2. A github action that formats the PRs. It seems like such actions exist, e.g., [run black formatter action](https://github.com/marketplace/actions/run-black-formatter). 
It's a last resort for people who don't have the codes pre-formatted. It might also nudge users to have their codes formatted.

keatonb:
Hey thanks for this PR addressing my issue.  I have one minor concern about this approach, and the comments from @orionlee give me an idea of how to address this.

My concern is, for generic LightCurve objects (not TessLightCurve or whatever), we expect maybe time, flux, fluxerr to be present, but these are generally not backed up as, e.g., "PDCSAP_FLUX" is in a TESS data product.  If I have another user-defined column I want to operate on (e.g., "residuals") and I do set_flux("residuals"), "flux" gets replaced and possibly lost (if lc is overwritten as `lc = lc.set_flux(flux_column="residuals")`).  There is no separate `FLUX_ORIGIN_COLUMN` in the case of the original LightCurve object, it was just "flux".

What about instead of replacing `flux`, `flux_err` and `FLUX_ORIGIN_COLUMN`, we just introduce some parameter similar to `FLUX_ORIGIN_COLUMN` that holds the name of the column that the lightkurve operations should be run on, maybe `ACTIVE_COLUMN` which is `flux` by default.  This seems like an especially simple solution that enables every lightcurve operation to be run on any column without having to add a `column=` argument to every function, and it avoids copying/overwriting data that adds minor computational cost.  I realize this violates the guiding idea of `flux`, etc. "they shall be the columns upon which all light curve operations shall operate," but not without the user taking a specific step to specify a non-default active column, so maybe worth considering.

barentsen:
@orionlee Fully agreed we ought to have a GitHub action that applies `black`!  (It wasn't smart of me to include unrelated changes from black in this PR, it can confuse code review!)

@keatonb Thanks, that's an interesting suggestion!  In your vision, would we still have an `lc.flux` attribute at all, or would we make users use e.g. `lc.pdcsap_flux` or `lc[ACTIVE_COLUMN]`?

barentsen:
> should we deprecate the column parameter in plot(), etc.? Or are we providing 2 ways of doing the same thing?

I'm inclined to keep it because it is useful for plotting non-flux data, e.g. `lc.plot(column="centroid_col")` makes sense but `lc.set_flux("centroid_col").plot()` would be weird.
 
> I propose we make LightCurve object keeping track of the origin column of flux

Agreed 100%!

> The name, set_flux(), is a bit misleading: I thought it is modifying the LightCurve object itself, and didn't realize it's creating a new object

I believe that all other methods which return a `LightCurve` object are returning a copy right now, which led me to return a copy here as well, but you are quite right that the word "set" suggests an in-place change.  I suppose we could `return self` without making a copy as long as it's well-documented.  Would the name `select_flux()` be better?

keatonb:
> @keatonb Thanks, that's an interesting suggestion! In your vision, would we still have an lc.flux attribute at all, or would we make users use e.g. lc.pdcsap_flux or lc[ACTIVE_COLUMN]?

I imagined there still being a flux column; I like the idea of this being a special, required column that all operations act on _by default_.  

> I suppose we could return self without making a copy as long as it's well-documented.

Please return a copy, whichever the approach.  I agree that `lc.set_flux("centroid_col").plot()` is a bit awkward, but `lc.select_col("centroid_col").plot()` is less so.  If I do plot that column though, it would be confusing if `"centroid_col"` replaced `flux` inplace.

orionlee:
> Please return a copy, whichever the approach. 

I also support returning a copy. My comment is about that the name `set_flux()` is somewhat misleading.

@barentsen raised a good point of retaining `column` argument in plot. So for cases such as `centroid_col`, users can still use `lc.plot(column="centroid")` (I myself use it in my notebook!).

---

> If I have another user-defined column I want to operate on (e.g., "residuals") and I do set_flux("residuals"), "flux" gets replaced and possibly lost (if lc is overwritten as lc = lc.set_flux(flux_column="residuals")).

By `operate`, do you mean you want to do some math using lc's notation, e.g., `lc = lc * 1.5`, `lc.normalize()`, etc.? 

> for generic LightCurve objects (not TessLightCurve or whatever), we expect maybe time, flux, fluxerr to be present, but these are generally not backed up as, e.g., "PDCSAP_FLUX" is in a TESS data product.

In the `FLUX_ORIGIN_COLUMN` approach, if the lightcurve is original, then `FLUX_ORIGIN_COLUMN` would be `None`. 

I admit I haven't fully figured out what `FLUX_ORIGIN_COLUMN` exactly would mean. E.g., 
```python
lc = LightCurve.read("some-TESS-lc.fits")
lc.meta["FLUX_ORIGIN_COLUMN"]  # should be pdcsap_flux
lc = lc.normalize() # should it sill be pdcsap_flux? I think so but not 100%.
lc = lc.flatten() # should it sill be pdcsap_flux? I think so but not 100%.
```

---

>>@keatonb Thanks, that's an interesting suggestion! In your vision, would we still have an lc.flux attribute at all, or would we make users use e.g. lc.pdcsap_flux or lc[ACTIVE_COLUMN]?
>I imagined there still being a flux column; I like the idea of this being a special, required column that all operations act on by default.

So to make sure we understand what the `ACTIVE_COLUMN` proposal mean, could we run through some typical case here? The following is my understanding based on what is discussed so far.

```python
lc = LightCurve.read("some-TESS-lc.fits")
print(lc.meta["ACTIVE_COLUMN"])  # "flux"
lc.flux  # still exists, as a copy of lc.pdscsap_flux

# Now I want to operate  about sap_flux
lc.meta["ACTIVE_COLUMN"] = "sap_flux"]
lc = lc.normalize()  # lc["sap_flux"] gets normalized, lc.flux remains unchanged, a copy of lc.pdcsap_flux
lc.plot()  # plot lc["sap_flux"]
lc.to_periodogram("lombscargle").plot() # again, based on sap_flux
```

The above codes would work, but 
- it might break some expectation that `lc.flux` is special. Specifically, `lc.flux` will be just another column once user sets a new `ACTIVE_COLUMN` (It is not an alias to the active column.)
- the original `sap_flux` is gone (some users might care while others won't)

barentsen:
@orionlee Thank you for posting an example use case -- this is a great way to explore API changes!

Based on the example, I am concerned that the `ACTIVE_COLUMN` mechanism does add quite a lot of complexity.  It doesn't feel very intuitive to me, but perhaps neither is the current behavior of `lc.flux`?

## Proposed solution

At this time, my preferred solution is to make two changes:

1. Introduce the `lc.select_flux(column)` method which returns a copy of `lc` in which `lc['flux']` (and `lc['flux_err']`) have been populated with the values of `lc['column']` (and `lc['column_err']`).
2. Introduce a `lc.meta["FLUX_ORIGIN"]` keyword which is a free-form string which documents the original name of the column from which the flux data was populated.  We may want to display its value as part of `LightCurve.__repr__`, but its value is allowed to be empty and there won't be a requirement for the `FLUX_ORIGIN` column to exist in the LightCurve object, i.e. we'd populate it on a best-effort basis for display purposes only.


## Is it ok for the `flux` values to be lost?

Earlier in this thread @keatonb correctly pointed out that it is pretty easy right now for the values in `lc['flux']` to be replaced and lost.  For some data products the original values will be backed up in a different column (e.g. `lc['pdcsap_flux']` for TESS products), but that is not necessarily true for a home-made light curve object.

Over the years, we have often wondered whether `LightCurve` objects should be able to track their history, and perhaps even be able to revert operations applied on them.  My opinion is that this would be really expensive and complicated to implement correctly, so my preference is to not offer this capability, and rely on the user to preserve any intermediate values if desired.

For example,
```python
lc["old_flux"] = lc["flux"]
```
```python
lc_old_flux = lc
lc_new_flux = lc.
```



 I'm in favor of adding more meta data to track the operations that have been applied (e.g. it would be nice to add a string to `lc.meta["HISTORY"]` every time a )

keatonb:
@barentsen Your proposed solution will work great, and will be smoothest for new users to get used to.  `FLUX_ORIGIN` helps keep things straight, even if this isn't "flux" data.  I suppose one possible issue is that `FLUX_ORIGIN` won't be updated if users set the flux column manually the way they might currently, with `lc.flux = lc["sap_flux"]` or whatever.  

@orionlee, that is a good example of how I envisioned the `ACTIVE_COLUMN` proposal working.  This admittedly may be less intuitive and too complex, and basically `select_flux` will enable all the same functionality.  Are we ok with whatever column is selected for analysis being called "flux" even if it isn't a flux quantity?

barentsen:
> Are we ok with whatever column is selected for analysis being called "flux" even if it isn't a flux quantity?

Do you have a few specific examples in mind where non-flux values would benefit from having LightCurve operations applied to them?

I can imagine someone wanting to e.g. remove outliers from centroid measurements.  If that's sufficiently common, I guess it would be fine to have the `remove_outliers` method accept a `column` argument, but in this scenario we ought to start thinking about migrating the `remove_outliers()` method to the `TimeSeries` parent class to enable the `LightCurve` class to keep its focus on dealing with flux measurements.  (Of course `LightCurve` is a great place to trial generic time series methods before moving them to `TimeSeries`.)

Do you have a few specific uses cases in mind?  I'd love to understand them better!

keatonb:
> Do you have a few specific examples in mind where non-flux values would benefit from having LightCurve operations applied to them?

One example off the top of my head, I have worked with ground-based telescope data where the tracking varied periodically with a gear frequency. If the flatfield correction was less than perfect, a systematic signal could appear in the light curve at the mechanical gear frequency.  To identify these spurious signals, we'd record the star locations alongside the extracted light curves and look for signals in the periodogram of the star location that correspond to any photometric frequencies.  

Yes, probably I am thinking of more general time series methods that are (just barely) outside the scope of lightkurve.  I am using lightkurve under the hood for Pyriod, a time series analysis program that is also primarily developed for light curve analyses, but I keep thinking, why wouldn't this be a fine framework for analyzing, e.g., radial velocity data? The only problem I can think of is that the label "flux" is hardcoded in as the analysis column. Sure, that isn't really enough to stop anybody, but names can cause confusion (i.e., "power" for amplitudes in LombScarglePeriodogram objects confuse my students fairly often).

orionlee:
On `lc.meta["FLUX_ORIGIN"]`:
1. It is updated when users call `select_flux()`
2. It should be property initialized in initial `read()` of existing TESS/Kepler FITS files
3. We can't do anything if user call `lc.flux = lc["sap_flux"]`. Users could  set their own value, e.g. `lc.flux = some_user_custom_flux_func()`, and we can't really tell.
4. keatonb, how would the proposal work with the residual example you raised earlier? I also wonder for analyzing other types of data, e.g., radial velocity, would the base astropy `TimeSeries` be more appropriate?  What does `Lightkurve` add on top of `TimeSeries` in those cases?

keatonb:
> 4. keatonb, how would the proposal work with the residual example you raised earlier? I also wonder for analyzing other types of data, e.g., radial velocity, would the base astropy TimeSeries be more appropriate? What does Lightkurve add on top of TimeSeries in those cases? Maybe the visualization functions (plot / scatter) for convenience?
This proposed solution will work just fine.  Yes, the base astropy TimeSeries objects are probably more appropriate for other analyses, but lightkurve does have very nice conveniences besides plotting, like getting a convenient periodogram, smoothing data, etc.

barentsen:
Update:
* I renamed `set_flux` into `select_flux` everywhere.
* I made sure we are setting `lc.meta["FLUX_ORIGIN"]` in both `LightCurve.select_flux()` and the `read_generic_lightcurve()` function in `io/generic.py`.
* I modified the `LightCurve` repr functions to show the FLUX_ORIGIN keyword if it is available.

Example repr (note the "FLUX_ORIGIN" information at the top right):

Screen Shot 2021-06-03 at 8 56 53 AM

(Note: the information at the top of the repr header is getting quite chaotic...  I wonder if it can be presented in a better way!)

barentsen:
I think I'm happy to merge this, but will wait a day to see if anyone has comments!

orionlee:
Looks good to me. 

It'd be a bonus if there is a test for the change in reading FITS file, if it's not too much of troubles.

barentsen:
> It'd be a bonus if there is a test for the change in reading FITS file

Good point!  Added a few tests just now!

barentsen:
Thanks for the discussion everyone!  I plan to release a new version of Lightkurve in the next day.

keatonb:
> Thanks for the discussion everyone! I plan to release a new version of Lightkurve in the next day.

Thanks @barentsen !!  I look forward to utilizing this functionality in the next version of Pyriod!

PR #1078, opened 2021-05-25
Remove the use of remote_data in some tests due to remote TABBY_TPF

orionlee:
A follow-up from PR #1065

It replaced the use of remote `TABBY_TPF` with a local  light version (~ 2 days of data, first 100 cadences) in the tests I can replace it confidently.

The use of remote `TABBY_TPF` is untouched in :
- tests that have other remote dependencies: remote `TABBY_Q8` lightcurve file
- tests that I am not as sure whether the use of trimmed down local version is appropriate: in PLD corrector test


barentsen (approved):
Thanks @orionlee, this is a great improvement!

I agree that altering the PLDCorrector test may need a bit more experimentation.

PR #1082, opened 2021-06-01
More Interact_sky tweak: arrow at selection, etc.

### LINKED ISSUE #1081: More interract_sky usability enhancement
#### Problem description

This is a follow up on #1049. Some more suggestion on tweaking `interact_sky`
1. upon selecting a star by tapping it, show at arrow pointing at the star, rather than hiding all other stars in the view.
I don't know what the use cases are for hiding other stars. If the purpose is to make it clear the star being selected, an arrow would do. 
When I use screenshots from `interact_sky`, I repeatedly add an arrow to the screenshots, and wish `interact_sky()` would do it for me.

2. Replace the box multi-select control with box zoom.  
I am not sure how useful box multi-selecting the stars in practice. I find that at times I want to zoom in, and mistakenly thought the multi-select control as box zoom, and wish box zoom is there.

3. (Optionally), highlight aperture pixels.
If implemented, it will make it easier to cross-reference the skyview plot here with the plots from `interact` and `plot_pixels`.

---

Illustrating (1) and (2):

image

#### Environment

orionlee:
Closes #1081 

The PR currently implements arrow at selection, and is subjected to further changes, depending on the decision in #1081.


orionlee:
Implemented the three changes in #1081 
- selected star is pointed by an arrow (instead of hiding the others)
- optionally, display aperture pixels specified in `aperture_mask` parameter
- box zoom is added, replacing box select 
  - box select make sense in `interact` for selecting apertures, but does not make much sense in `interact_sky`

```python
tpf_trunc.interact_sky(aperture_mask="pipeline")
```

image

orionlee:
Rebased to resolve test conflicts.
Ready for review.

orionlee:
Rebased to resolve conflicts.
Ready for review / merge.

src/lightkurve/interact.py
```diff
@@ -507,15 +509,78 @@ def add_gaia_figure_elements(tpf, fig, magnitude_limit=18):
...
+
+            # workaround: the arrow_head color should have been specified once
+            # in its creation, but it seems to hit a bokeh bug, resulting in an error
+            # of the form  ValueError("expected ..., got {'value': 'red'}")
+            # in actual websocket call, it seems that the color value is
+            # sent as "{'value': 'red'}", but they are expdecting "red" instead.
+            # somehow the error is bypassed if I specify it later in here.
+            #
+            # The issue is present in bokeh 2.2.3 / 2.1.1, but  not in bokeh 2.3.1
+            # I cannot identify a specific issue /PR on github about it though.
```
barentsen, line 542: I'd be open to update the Bokeh dependency to be e.g. `bokeh>=2.3.1`, because Bokeh v1.1 is already three years old at this stage, or do we think supporting older Bokeh versions makes dependency issues much easier for our users?

barentsen:
I've played around with this PR and I think this is brilliant!

Specifically, I agree with all three changes, i.e.:
1. I love the arrow.  I don't see any reason to preserve the previous behavior (i.e. deselecting other stars rather than showing an arrow).
2. I love the new `aperture_mask` parameter.
3. I agree with the change to use box zoom rather than box select.  Selecting multiple objects served no purpose.

I would be happy to merge this in its current state.  I would change the version number of the main branch  to `v2.2.0dev` (rather than `2.1.2dev`) because this really represents a feature change rather than a bug fix, but that's easy to do.

I only have two minor comments/questions:
1. I wonder if it would be easier to upgrade the minimum bokeh dependency to a more recent version to avoid bug workarounds (both here and in future PRs).
2. I encountered a benign `UnitWarning` when using the following snippet:
```python
import lightkurve as lk
tpf = lk.search_targetpixelfile("Proxima Cen")[8].download()
tpf.interact_sky("http://localhost:8892", aperture_mask="threshold")
```
which yielded
```
WARNING: UnitsWarning: Unit 'e' not supported by the VOUnit standard.  [astropy.units.format.vounit]
WARNING:astroquery:UnitsWarning: Unit 'e' not supported by the VOUnit standard. 
```

src/lightkurve/interact.py
orionlee: I don't think we can require `bokeh>=2.3.1` yet, until #990 is resolved (pixel selection in `tpf.interact()` does not work for bokeh>=2.3`

orionlee:
minimum bokeh dependency: 
I suggest we table it until #990 is resolved (hopefully soon. See PR #1177)

---
The warning `UnitsWarning: Unit 'e' not supported by the VOUnit standard. [astropy.units.format.vounit] ...`:
I suspect it is not directly triggered by changes in this PR. I have seen the warning sporadically since using Lightkurve v2.1 in different places, and have not tracked it down yet.

orionlee:
I confirmed the warning ```UnitsWarning: Unit 'e' not supported by the VOUnit standard. [astropy.units.format.vounit] ...``` comes from doing Vizier search on Gaia DR2 with `astroquery`.

So some recent version of `astroquery` (and/or `astropy`) caused the warning. It's a bit perplexing but harmless.

barentsen:
PR merged and change log updated.  Thanks for the awesome improvement!

barentsen:
Update: these changes have been included in the Lightkurve v2.2.0 release.

PR #1096, opened 2021-07-01
fix lc.bin() excessive memory usage in some cases

### LINKED ISSUE #1092: Memory leak when using lc.bin()
#### Problem description

When I use the bin() method with the keyword `bins`, the memory usage increases over time until it reaches 100%. The program then starts using swap memory and everything is running very slowly. The python process has to be stopped manually. This seems to happen whatever number is passed to `bins`, including negative number or a number that is > to len(lc).

lk_memory_usage
(Bottom plot is memory usage. RAM in purple red, Swap memory in green)

#### Example

```python
import lightkurve as lk

d = lk.search_lightcurve('TIC382160069',cadence='short').download_all()
full_lc = d.stitch().remove_outliers()
binned_lc = full_lc.bin(bins=10000)
```

#### Possible cause

The issue seems to be caused by np.searchsorted() used here in the source code (in lightcurve.py): 

```python
   1386         if time_bin_size is None:
   1387             if bins is not None:
-> 1388                 i = len(self.time) - np.searchsorted(
   1389                     self.time, time_bin_start - 1 * u.ns
   1390                 )
```
as running the following line will result in the same problem as described above:
```python
np.searchsorted(full_lc.time, full_lc.time[0] - 1*u.ns)
```

Apparently NumPy does not like astropy's Time objects. For instance, the following code works fine and returns 0:
```python
np.searchsorted(full_lc.time.value, full_lc.time[0].value - 1e-10)
``` 

#### Expected behavior

I would expect lc.bin() to return the binned lightcurve (or raise an error if the number passed to `bins` is invalid).

#### Environment

-  platform (e.g. Linux, OSX, Windows): Linux (Ubuntu 20.04.2)
- numpy version : 1.21.0

orionlee:
Closes #1092 

- Fixed by avoiding the astropy bug 
  - The alternative, raising minimum astropy to 4.2, would require us to drop Python 3.6 support
- A test is added. It is run in isolation in a new CI check `3.8-memtest` (on Linux)

### DIFF
--- a/.github/workflows/lightkurve-tests.yml
+++ b/.github/workflows/lightkurve-tests.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        name: [3.6, 3.7, 3.8-remote-data, 3.9]
+        name: [3.6, 3.7, 3.8-remote-data, 3.8-memtest, 3.9]
         include:
           - name: 3.6
             python-version: 3.6
--- a/.github/workflows/lightkurve-tests.yml
+++ b/.github/workflows/lightkurve-tests.yml
@@ -31,6 +31,9 @@ jobs:
           - name: 3.8-remote-data
             python-version: 3.8
             pytest-command: poetry run pytest --remote-data --doctest-modules
+          - name: 3.8-memtest
+            python-version: 3.8
+            pytest-command: poetry run pytest -m memtest
           - name: 3.9
             python-version: 3.9
             pytest-command: poetry run pytest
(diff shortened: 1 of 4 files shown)


orionlee:
Issue: astropy 4.2 requires Python >=3.7 . So if we upgrade to 4.2, we need to drop Python 3.6 support. Thoughts?

barentsen (approved):
Thank you for investigating this issue so quickly @orionlee!  The fix looks excellent, and the tests you added are thorough!

Python 3.7 was first released ~3 years ago, so I think it would be acceptable for future Lightkurve versions to drop Python 3.6 support in favor of relying on AstroPy 4.2.  Of course it looks like this bug has been resolved without requiring 4.2 for now, so I guess we can keep Python 3.6 support for now...

PR #1103, opened 2021-07-09
Centroid estimates are off by 0.5 pixels in column and row position

barentsen:
Until ~one year ago, Lightkurve followed the convention that pixel coordinates refer to the corner of a pixel.  This was incorrect, because the Kepler/TESS pipeline products adopt the convention that pixel coordinates refer to the middle of a pixel (cf. #755).  For example, (column, row) = (10.0, 20.0) refers to the middle of a pixel rather than its lower left corner.

Over in #755, we previously updated Lightkurve's plotting features to follow the correct convention, but we forgot to update Lightkurve's centroid estimation feature, which was also affected by the incorrect convention.

As a result, the centroid positions estimated by `TargetPixelFile.estimate_centroids()` have been off by exactly 0.5 pixels in both the row and column positions (see example screenshots below). 

The PR fixes this bug and introduces a regression test.

## Old behavior (incorrect)
Screen Shot 2021-07-09 at 1 41 05 PM

## New behavior (correct)
Screen Shot 2021-07-09 at 1 33 51 PM

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@
   does not have the target's proper motion / coordinate; or it is from
   `tpf.cutout()` call. [#1088]
 
+- Fixed a bug in ``TargetPixelFile.estimate_centroids`` which caused the column
+  and row coordinates reported to be off by 0.5.
+
 
 2.0.10 (2021-06-04)
 ===================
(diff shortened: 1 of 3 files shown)


barentsen:
I should mention that both `TargetPixelFile.estimate_centroids(method="moments")` and `TargetPixelFile.estimate_centroids(method="quadratic")` were affected.  This PR fixes and tests both methods.

PR #1105, opened 2021-07-12
Fix FoldedLightCurve.odd_mask / even_mask

### LINKED ISSUE #1104: FoldedLightCurve.odd_mask / even_mask is incorrect
#### Problem description
`FoldedLightCurve.odd_mask` / `even_mask` is incorrect when the phase is literal time (the default) rather than normalized phase (lightkurve v1 behavior).

It is a bug stemming in lightkurve v2, as the existing logic assumes the phase is normalized. Thus when given the phase is literal time, i.e, the folded `lc.time` is actual time rather than normalized `[-1, 1]`, the resulting mask is bogus.

See `FoldedLightCurve.odd_mask` implementation. The cycle calculation assumes `lc.time` is normalized:
https://github.com/lightkurve/lightkurve/blob/81bc8697480afd371e01a9c526487f25e414ba7d/src/lightkurve/lightcurve.py#L2807-L2811

Update: the odd/even mask can still be incorrect even when the phase is normalized. See the next comment.

#### Example

For the following lightcurve with a period of 4 days, the `even_mask` produced is incorrect:
```python
import lightkurve as lk

lc = lk.LightCurve(time=Time(np.linspace(0, 24, num=2400, endpoint=False), format="btjd"))
epoch_time, period = 3, 4
lc.flux = np.sin((period * 0.75 + lc.time.value - epoch_time) * 2 * np.pi / period)
ax = lc.plot()

lc_f = lc.fold(epoch_time=epoch_time, period=period)  # the default: normalize_phase=False

lc_e = lc_f[lc_f.even_mask]
ax.scatter(lc_e.time_original.value, lc_e.flux, c="red", label="even_mask (incorrect)")
ax.legend();
```

image

#### Expected behavior

What the even_mask should look like, assuming the first cycle is cycle 0 (even): 
image

#### Environment

orionlee:
Close #1104 

The PR also provided a new attribute, `FoldedLightCurve.cycle`.


barentsen:
Awesome! 👍   This looks good to be merged!

I plan to give it a careful manual test myself tomorrow.  Stay tuned.

gully:
I applaud the addition of this `cycle` attribute, I was recently seeking the need for something like this.

orionlee:
It'd be great if the PR can be reviewed and be moved forward.

To aid the review, one can consider to run new test method, `test_lightcurve_fold_odd_even_masks()` standalone, say in a Jupyter notebook. It'd plot the result. I find it easier to verify the result.

https://github.com/lightkurve/lightkurve/blob/e79a592018973aef6b1d215d92e0aba589e307ac/tests/test_lightcurve.py#L270-L272

barentsen:
Sanity check: this PR appears to resolve the bug reported in #1104 correctly. Screenshot:

Screen Shot 2022-04-29 at 9 06 23 AM

barentsen:
Thanks so much @orionlee.  This looks good to merge!

PR #1106, opened 2021-07-15
Add a sort_by argument for stitch

ckm3:
When the order of a LightCurveCollection is not sorted by time, the stitched light curve would require additional sorting. I have added an argument for sorting a stitched light curve by time.

### DIFF
--- a/src/lightkurve/collections.py
+++ b/src/lightkurve/collections.py
@@ -171,7 +171,7 @@ class LightCurveCollection(Collection):
         will be removed soon."""
         return LightCurveCollection([lc.SAP_FLUX for lc in self])
 
-    def stitch(self, corrector_func=lambda x: x.normalize()):
+    def stitch(self, corrector_func=lambda x: x.normalize(), sort_by='time'):
         """Stitch all light curves in the collection into a single `LightCurve`.
 
         Any function passed to `corrector_func` will be applied to each light curve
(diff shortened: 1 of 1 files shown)


barentsen:
Thank you for this contribution @ckm3!  This is an interesting idea, and I agree we should probably have this feature in some form.

I do have a question I'd love feedback on from anyone who reads this: **should `sort_by='time'` be the default behavior of `stitch()`?**  I imagine this is what most people want (?), but I wonder whether or not it is intuitive for the verb `stitch()` to be a synonym for "concatenate and sort". Perhaps this is fine, I just want to think carefully about method names because they're hard to change after the facts.  The alternative is to give this "concatenate and sort" operation a different method name.

(Note: this PR is currently making some of the unit tests fail, e.g. `test_lightcurve_append_multiple()`.  @ckm3 would you be willing to look into editing the failing tests to match the new behavior?  You can find more information about running the unit tests [here](https://docs.lightkurve.org/about/testing.html).  Thanks again for your contribution!!)

orionlee:
From API perspective, I'd suggest
- add a `sort()` function to `lightkurve`'s `Collection`.
https://github.com/lightkurve/lightkurve/blob/1187df1d7d731e4b7d8402a19f0973098584d952/src/lightkurve/collections.py#L19-L21
- We might want to tweak `stitch()`, so that it would always (or by default) sort the collection object before doing the stitching. As barentsen said, stitching out-of-order collection makes little if any sense in practice.

The new `lk.Collection.sort()` function could possibly sort in-place or return a new sorted collection. Returning a new sorted collection is more inline with the general style used in `lightkurve`, but both general python and numpy's `sort()` do it in place (and return `None`) so we might not want to break the more general convetion.

ckm3:
Sorry for my late. Add a `sort()` function to `Collection` would require extra data reading. I think applying sort method in the stitch function is the most conservative way. The `LightCurve.append` method should behave like list and ndarray, the `sort_by` argument should be None by default.

PR #1108, opened 2021-07-16
fix outdated doc on testing / dev install

### LINKED ISSUE #1071: running tests causes "Fatal Python Error"
Following the instructions at http://docs.lightkurve.org/about/testing.html?highlight=tests#how-do-i-run-tests-locally

When I try to run tests (or run any Python) from within the `tests` directory, I get:
> Fatal Python error: init_sys_streams: can't initialize sys standard streams
> AttributeError: module 'io' has no attribute 'OpenWrapper'

This is caused by the new addition of a subdirectory called `io` that Python finds deeply troubling.

Renaming `tests/io` to anything else solves this problem.

orionlee:
Close #1071

Update the doc to use poetry-based install / pytest

### DIFF
--- a/docs/source/about/contributing.rst
+++ b/docs/source/about/contributing.rst
@@ -44,6 +44,8 @@ Head to a local directory of your choice and download your fork:
     $ git clone https://github.com/YOUR-GITHUB-USERNAME/lightkurve.git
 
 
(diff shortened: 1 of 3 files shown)


barentsen:
Awesome PR.  Thank you!!

Issue #1024, opened 2021-04-06
Use `format="tjd"` or `format="tess"` instead of `format="btjd"` in astropy Time objects?


barentsen:
Lightkurve v2 added support for Kepler and TESS times (i.e. BTJD and BKJD) by registering "btjd" and "bkjd" as AstroPy time formats in `src/lightkurve/time.py`.  This allows for the following convenient manipulation of time values:

```python
>>> import lightkurve as lk
>>> from astropy.time import Time
>>> t = Time(2000, format="btjd", scale="tdb")
>>> t
<Time object: scale='tdb' format='btjd' value=2000.0>
>>> t.iso
'2020-05-30 12:00:00.000'
>>> t.bkjd
4167.0
```

In hindsight, the choice of `format='btjd'` in this context is likely misleading because the acronym encodes both the time format (i.e. JD - 2457000) and the time scale (the "b" stands for "tdb", i.e. "Barycentric Dynamical Time").  This could be confusing because these are separate parameters in AstroPy.

I wonder if we ought to correct the syntax by adopting
```python
t = Time(2000, format="tjd", scale="tdb")
```
or
```python
t = Time(2000, format="tess", scale="tdb")
```
instead of `format="btjd"`.  We'd make a similar change for Kepler.

Thoughts, anyone?

barentsen:
(Side note: we also ought to deprecate `btjd_to_astropy_time` and `bkjd_to_astropy_time` in `lightkurve.utils` now that Kepler and Tess times are supported directly in Time objects.)

orionlee:
While I don't know the intricacy of time systems, what @barnesten raised makes sense.
If we want to make changes, I suggest we should do it soon, as the `format` is visible to lightkurve users. Changing it could potentially break user codes, albeit unlikely.
```python
# users might have relied in checking the format value in their codes.
> lc.flux.format
'btjd'
```

1. The suggestion of `tjd` in place of `btjd` looks good to me.
2. Alternatively, we could possibly retain `btjd` as a format, but have the scaled defaulted to `tdb`. A few astropy built-in formats have such behavior (see [doc](https://docs.astropy.org/en/stable/time/index.html?highlight=Time#time-from-epoch-formats)), e.g., `unix`.

Option (1) would be more generic. While option (2) is more user friendly in typical cases. 
Is there any use case of having TESS time in a scale that is not `tdb`? If the answer is no, option (2) could be a viable alternative.

barentsen:
[...]

PR #1112, opened 2021-07-21
Improve the implementation of the BKJD and BTJD time formats

barentsen:
This PR addresses #1024 by revising the implementation of the `TimeBKJD` and `TimeBTJD` AstroPy time formats to utilize the `TimeFromEpoch` base time format available in AstroPy.

The benefits of this change are two-fold:
1. It simplifies the implementation.
2. It ensures that times use the Barycentric Dynamical Time (TDB) scale by default, because the `TimeFromEpoch` base class includes logic to ensure values are in the time scale of the epoch.

This PR also adds improved unit tests for verification.

We may eventually wish to add these time formats to AstroPy, but for the time being, Lightkurve provides a good testing ground.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,10 @@
 - Fixed a bug in ``TargetPixelFile.estimate_centroids`` which caused the column
   and row coordinates reported to be off by 0.5. [#1103]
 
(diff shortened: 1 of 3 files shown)


orionlee:
The PR looks good to me.

Issue #1031, opened 2021-04-21
CBVs are not found, despite they are available


ninahernitschek:
#### Problem description

I see quite often CBVs are not found, despite they are available.

#### Example

Start the Jupyter notebook 2-3-how-to-use-cbvcorrector.ipynb, and load this light curve

```
lc = search_lightcurve('TIC 231838912', author='TESS-SPOC', sector=1).download(flux_column='sap_flux')
```

and then try running the notebook until

```cbvCorrector.cbvs[0].plot();
```
Nothing is plotted, and the the cbvs will be empty (NaN).
However, the appropriate CBVs are there:
https://archive.stsci.edu/hlsps/tess-spoc/s0001/cbv/hlsp_tess-spoc_tess_phot_1-3-s0001_tess_v1_cbv.fits.gz

#### Expected behavior

I would expect the CBVs as in the link are found and applied.

#### Environment

-  platform Linux
-  lightkurve version 2.0
-  installation method pip

jcsmithhere:
Hello @ninahernitschek, 

So very sorry for taking so long to address this issue. When CBVCorrector loads the CBVs is "aligns" the CBVs to the passed light curve using the cadence numbers `cadenceno`. it appears for this Sector and CCD the CBV cadence numbers do not align with the light curve cadence numbers. I should explore why. It could be an error in the data products. To get around this problem you can tell CBVCorrector to interpolate the CBVs to the light curve. This uses the cadence times, not the cadence numbers, and those appear to line up. See attached example.

![Screenshot from 2021-07-21 23-26-42](https://user-images.githubusercontent.com/6098349/126598175-68e29737-63f6-4e8c-8c31-2d1645a559fe.png)

jcsmithhere:
Given that the error is in the data products and not in CBVCorrector, I should pass a warning message to the user when it appears the CBVs do not line up with the light curve and then recommend the user uses `interpolate_cbvs=True`.

jcsmithhere:
I'm adding in a warning message so people do not get tripped up by this in the future:
![Screenshot from 2021-07-22 00-13-10](https://user-images.githubusercontent.com/6098349/126602115-777c64cd-7ec6-4b1d-8b0a-0696c65c2ddf.png)

jcsmithhere:
Created PR #1113 to address this issue.

barentsen:
The warning message has been added via #1113.

@ninahernitschek  Many thanks for reporting this issue! 👍

PR #1113, opened 2021-07-22
Added warning message if the CBVs are poorly aligned to the light curve.

jcsmithhere:
In Issue #1031 it became apparent that if the loaded CBVs are not at all aligned to the light curve the user will get no indication. A warning is now issued recommending the user set `interpolate_cbvs=True`.


barentsen (approved):
Thanks @jcsmithhere!

Issue #1111, opened 2021-07-20
underfit_metric_neighbors() cannot handle different shapes for neighborhood fluxes


akar2014:
#### Problem description

Underfitting metric cannot broadcast different numpy array shapes, possibly due to neighboring fluxes have different sized lightcurves.
#### Example

```
import lightkurve as lk
from lightkurve.correctors import CBVCorrector
search_result = lk.search_lightcurve('TIC 77897915', cadence=120, author = 'SPOC',mission='TESS').download(flux_column='sap_flux')
cbvCorrector = CBVCorrector(search_result)
cbvCorrector.correct(cbv_type=['SingleScale','Spike'],cbv_indices=[np.arange(1,9),'ALL'])
```

The above code snippet runs into the following error:

```
  File "/opt/anaconda3/lib/python3.8/site-packages/lightkurve/correctors/metrics.py", line 201, in underfit_metric_neighbors
    fluxMatrix[:, idx] = lc_neighborhood_flux[idx]

ValueError: could not broadcast input array from shape (17601,) into shape (17696,)
```

#### Expected behavior

The above snippet works for other TIC IDs or other sectors for the same TIC, possibly because neighboring fluxes have the same shape/cadences as the target.

#### Environment

-  platform: OSX
-  lightkurve version: 2.0.10
-  installation method: conda

jcsmithhere:
Hello @akar2014, Yes, this is a bug but also a philosophical debate for the CBVCorrector class. By default the CBVs and any neighboring targets are "aligned," meaning the cadence numbers for the CBVs and neighboring targets are broadcast to the cadence numbers in the light curve. However, here we have a case were doing so results in an error, which I need to fix. 

The workaround is to set `inteprolate_cbvs=True` as shown below. This is slightly dangerous because the CBVs and neighboring targets are interpolated using PCHIP which can result in introduced signals and possible corruption of the CBVs and neighboring targets. This concern is why "aligning" is the default and one must force `interpolate_cbvs=True`. 

Nonetheless, the code should still run when aligning and simply broadcast the neighboring targets to the shape of the light curve under study. I will fix this bug.


![Screenshot from 2021-07-22 21-38-46](https://user-images.githubusercontent.com/6098349/126738002-dd50301f-f997-4940-9921-f53618316182.png)

jcsmithhere:
Created PR #1120 to resolve this Issue.

barentsen:
[...]

PR #1120, opened 2021-07-26
Fixed bug in underfit_metric_neighbors

jcsmithhere:
In Issue #1111 a bug in `underfit_metric_neighbors` was discovered. This PR fixes the bug and better aligns neighboring targets to the light curve under study.


barentsen (approved):
Thanks @jcsmithhere!

Issue #1107, opened 2021-07-16
option to load in CBVs from data directory


benjaminpope:
#### Problem description
CBVCorrector currently loads MAST CBVs by default, or else you have to set up design matrices yourself. But I want to run CBVCorrector on an HPC system without access to the internet.

#### Expected behavior
I have already downloaded all TESS CBVs to a directory on the cluster. It would be nice to just pass it a "cbv_dir='my_cbv_dir'" option and it will look for them in there, but otherwise display the same behaviour.

orionlee:
In the interim, there appears to have an (ugly) workaround. You'd need to write some logic to load the local CBVs though.

```python
    corrector = CBVCorrector(lc, do_not_load_cbvs=True)
    corrector.cbvs = custom_cbv_loading_from_local_dir(lc)
```

See: 
http://docs.lightkurve.org/reference/api/lightkurve.correctors.CBVCorrector.html#lightkurve.correctors.CBVCorrector.__init__

barbierimauro:
I have a similar issue of BenjaminPope, but related to lightcurves.
I have all the lightcurves downloaded in a local archive, and I have also the list of the filenames for each TICID.
There is a way to tell to lightkurve to open all the local fits files of one TICID ? (I mean creating a unique object)

jcsmithhere:
Hi @benjaminpope and @orionlee, 

Thank you for the feedback. This is a great recommendation! I see a way I can add an option to pass a path to load the CBVs without much work at all. But in the mean time, yes, you can pass a custom design matrix and still use the fitting algorithms in the CBVCorrector class. However, I think I can add the local path option in the near future.

benjaminpope:
Thanks! Yeah I think the main issue is that there is currently a very nice API for automatically selecting which CBVs to use. It would be nice to be able to use the *same code* inside an HPC core, pointing to a pre-saved directory of CBVs, as you would in a notebook.

orionlee:
@jcsmithhere An option to pass a path to load the CBVs would also be helpful in software maintenance. 

Right now, in the continuous integration tests, the CBV codes are only tested on python 3.8+Linux environment (as all tests rely on external data do). With a local path option, some CBV tests can be added to the regular test suite that would in turn be tested in a variety of platform/Python.

orionlee:
[...]

PR #1122, opened 2021-07-27
Load CBVs from Data Directory

benjaminpope:
This would resolve Issue #1107, and also a side issue I encountered along the way where `to_fits() `didn't save cadenceno. 

- Creates new function in `cbvcorrector.py`, `load_tess_cbvs`. This has exactly the same behaviour as `download_tess_cbvs`, except takes an additional argument `cbvdir` which must contain the required CBVs.
- Also updates the syntax in `download_tess_cbvs` to just use string formatting to generate the filename correctly, rather than logic over the size of `sector`.
- When testing, also encountered a problem where, for whatever reason, `dir(lk.LightCurve)` didn't include all column names. I don't understand this behaviour but it is fixed if you use `hasattr(self,'cadenceno')` instead, and appears to save fits files correctly now.

Included a `cbv_test.ipynb` demonstrating new functionality but this should be dropped on merge.


benjaminpope:
I should probably add, it is worth doing an identical thing for Kepler/K2 CBVs but this isn't necessary for my use case yet and I didn't want to get ahead of myself until I had feedback on this pull request.

barentsen:
Thanks @benjaminpope!

@jcsmithhere Would you agree with this change?

jcsmithhere:
Thanks for working on this, @benjaminpope! I was going to begin this work soon but you got to it before me!

I'll take a look at your code in the very near future. My first reaction is that we probably do not need two functions `load_tess_cbvs` and `download_tess_cbvs`. Perhaps we should merge them into a single function, called `retrieve_tess_cbvs`. If a `cbvdir` file path is passed then it searches the directory, otherwise it downloads from MAST. What do you think?

benjaminpope:
Great. Yes I actually agree - I just wanted to provide a minimal working example. Should be easy to write. Want me to do this?

benjaminpope:
Added a `load_tess_cbvs` function that either loads or queries MAST. Seems to work ok

src/lightkurve/correctors/cbvcorrector.py
```diff
@@ -41,7 +40,7 @@ class CBVCorrector(RegressionCorrector):
     """Class for removing systematics using Cotrending Basis Vectors (CBVs)
     from Kepler/K2/TESS.
 
-    On construction of this object, the relevent CBVs will be downloaded from
+    On construction of this object, the relevant CBVs will be downloaded from
```
jcsmithhere, line 43: Thanks for fixing my spelling errors!

jcsmithhere (changes requested):
This looks good overall but I haven't tested the code yet. Before we merge we would need to create a corresponding `load_kepler_cbvs`. Also, both the loading CBVs and using CBVCorrector tutorials will need to be updated.

benjaminpope:
Made requested changes and pushed - including a `load_kepler_cbvs` method. Haven't tested this - is there a preferred Kepler example to run on?

orionlee:
Regarding `lc.to_fits()` fix, of replacing column check by `dir(lk.LightCurve)` with `hasattr(self,'cadenceno')`: I think it is an issue in Lightkurve v1 to v2 changes.

In lightkurve v2, currently `dir(lc)` does not return all the columns because
a. the columns are natively accessed in the form of lc['cadenceno'], the same as the underlying astropy `TimeSeries` / `Table`.
b. ligtkurve provides a syntactic sugar of allowing a column to be accessed as an attribute, e.g., `lc.cadenceno`
c. but such columns won't show up as the list of attributes by `dir()`, (with the exception of `time`, `flux`, and `flux_err`, as they have been predefined).

The current behavior (not all columns would show up in `dir()`) is probably acceptable. @barentsen thoughts?

orionlee:
Regarding the CBV changes:
- For API doc, it'd be great if we can specify where the Cbv fits files are expected to be under `cbv_dir`, e.g., must they be directly under `cbv_dir`, or can they be in some sub directory under `cbv_dir`, e.g., `<cbv_dir>/ffi/s0017/2019/279/4-4/tess2019279210107-s0017-4-4-0161-s_cbv.fits`?

benjaminpope:
> The current behavior (not all columns would show up in `dir()`) is probably acceptable.

@orionlee Re this - probably fine for most purposes, but it needs to be able to save cadenceno for custom lightcurves, which the current `dir()` call prevents. Hence checking `hasattr instead.

> * For API doc, it'd be great if we can specify where the Cbv fits files are expected to be under `cbv_dir`, e.g., must they be directly under `cbv_dir`, or can they be in some sub directory under `cbv_dir`, e.g., `<cbv_dir>/ffi/s0017/2019/279/4-4/tess2019279210107-s0017-4-4-0161-s_cbv.fits`?

re this - hm, good question. Personally I think they should be in that exact directory, in case people mess up the path. An overly specific directory structure (eg latter example) matches MAST but would just add overhead for users who want to use their local CBVs. (eg on our HPC implementation we just dump them all in a single directory)

orionlee:
> > The current behavior (not all columns would show up in `dir()`) is probably acceptable.
> 
> but it needs to be able to save cadenceno for custom lightcurves, which the current `dir()` call prevents. Hence checking `hasattr instead.

Yes, I agree that `to_fits()` not saving cadenceno by default is a regression that should be fixed. Sorry I wasn't clearer. 
It'd be great if the test suite is updated to verify it too:
https://github.com/lightkurve/lightkurve/blob/e44b1d9b959831e2d08e522ad5ad96eabca22638/tests/test_lightcurve.py#L771


> 
> > * For API doc, it'd be great if we can specify where the Cbv fits files are expected to be under `cbv_dir`, e.g., must they be directly under `cbv_dir`, or can they be in some sub directory under `cbv_dir`, e.g., `<cbv_dir>/ffi/s0017/2019/279/4-4/tess2019279210107-s0017-4-4-0161-s_cbv.fits`?
> 
> re this - hm, good question. Personally I think they should be in that exact directory, in case people mess up the path. 

That's fine. I'm just wondering if the docstring needs to make this clear. From `lightkurve` users' perspective, it might not be clear.

E.g., the existing `download()` function accepts a `download_dir` parameter for where the fits file for lc/tpfs to be saved. The `download_dir` specified is the base where fits files are saved in a specific directory structure, e.g., `<download_dir>/mastDownload/TESS/tess2020324010417-s0032-0000000737578805-0200-s/tess2020324010417-s0032-0000000737578805-0200-s_lc.fits` So users might expect `cbv_dir` to allow or require some directory structure.

benjaminpope:
Ok, makes sense - it can simply parse the entire directory structure under `cbv_dir` and the existing solution with searching a glob list should allow it to retrieve the correct cbv. I don't think it should _require_ and directory structure though.

orionlee:
I am not advocating that it should require any directory structure either. My comment is about making the behavior clear and stated in docstring.

benjaminpope:
Ok - I think the simplest case is pointing to a specific directory. Have updated the docstrings accordingly. What do you think?

orionlee:
Sounds good!

jcsmithhere:
@benjaminpope, your code looks good to me but the unit tests fail. Looks like they need to be updated to reflect the new function names. The tutorials will also need to be updated.

benjaminpope:
Should be ok now I think

benjaminpope:
Updated docs at `lightkurve/docs/source/tutorials/2-creating-light-curves/2-3-how-to-use-cbvcorrector.ipynb`. This includes changes to the ylims of a couple of plots, where the new functionality from the latest lk updates changes normalization by a very small factor. I think everything looks good now.

benjaminpope:
Want to review / CI / merge?

jcsmithhere:
Sorry, plenty of other things to work on... I hope to do one more look at this soon and get feedback to you. Provided the unit tests run and the tutorials have been updated appropriately then I think it looks good!

jcsmithhere:
@benjaminpope, the unit tests fail for me. It looks like `load_kepler_cbvs` line 1789 does not define `soup`. Do you know how to run the unit tests? Here is some guidance:

https://docs.lightkurve.org/about/testing.html

Also, you need to have the package `pytest-remotedata` installed. Once installed run:

`pytest ./test_cbvcorrector.py --remote-data`

to run the CBVCorrector unit tests.

jcsmithhere:
Also, now that we can load the CBVs locally, we really should update the unit tests to test this functionality. This would entail adding a very simple CBV FITS test file in `lightkurve/tests` and then writing a unit test to read in the CBVs and confirm proper functionality. 

Are you familiar with writing unit tests?

benjaminpope:
New unit test created for loading local files in tests/data for each of TESS, Kepler and K2, and all unit tests now passing.

benjaminpope:
Bumping this - can we review and merge?

jcsmithhere (changes requested):
Code looks good! Unit tests all pass. The CBVCorrector tutorial successfully runs. However the `how-to-use-cbvs` tutorial still needs to be updated. With that change I think we will be all set!

The only other comment is I noticed the CBV FITS files added to the test data directory are full length CBV FITS files totaling about 50MB. This significantly increases the size of the total git repository. @barentsen might want the data files added to be shrunk down to minimize the increase in size of the full repo. That is his call. Problem is, the larger data files are already in the commit so they will be in the repo if that commit is merged. We might need to reverse out the commit with the larger data files before we merge to the main branch. This is all @barentsen's call, not mine.

benjaminpope:
Ok I have cut the CBV files to 100 cadences and updated the tutorial.

benjaminpope:
Merge in?

jcsmithhere:
I can't do the merge. Someone like @barentsen has to do it.

benjaminpope:
Just checking in - @barentsen ?

barentsen:
@benjaminpope Regarding the change of `**extra_data` into `extra_data={}` in `LightCurve.to_fits()`: would you be willing to open a separate issue/PR for this change with an example?  I ask because this change may potentially break existing user code, plus it is somewhat dangerous to accept an empty dictionary as a default argument (e.g. [see explanation here](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument)).

I've removed the change from this PR, so we can get it merged without having to discuss this separate issue here, if that sounds good?

barentsen:
Unfortunately the `test_CBVCorrector_retrieval` test is failing right now because MAST is in the process of ingesting re-processed Sector 10 data.  I am going to wait a few hours to give this test a chance to run and pass, because it directly covers the code being changed by this PR.

Otherwise I believe this is OK to merge.

barentsen:
@scfleming Thanks for the update!  The test happens to use Sector 10 data (by chance).  We can probably just wait until the Sector 10 data is back up again -- please post below if someone needs this PR merged more urgently.

benjaminpope:
@barentsen fine with the extra data thing - will create a separate PR and issue. Long story short, functionality on my machine is that it was saving no user-specified extra data at all, as it was getting overriden with the kepler/tess specific keywords. The CBV corrector won't depend on that but I was just fixing other issues as I went...

barentsen:
It looks like Sector 10 data continues to be unavailable for now, but I'll keep checking and get this merged asap.

benjaminpope:
Hi @barentsen - looks like the docs at https://docs.lightkurve.org/tutorials/2-creating-light-curves/2-2-how-to-use-cbvs.html and https://docs.lightkurve.org/tutorials/2-creating-light-curves/2-3-how-to-use-cbvcorrector.html still use the deprecated `download_tess_cbvs` functionality. Did we merge everything into the docs? Should I do another pull request?

barentsen:
Thanks for noticing this @benjaminpope!

The changes were merged in fine, but unfortunately the notebook `2-2-how-to-use-cbvs.ipynb` does not build successfully right now, which explains why the online docs haven't updated yet.

I am seeing the following error:

```python
$ make docs
Notebook error:
CellExecutionError in tutorials/2-creating-light-curves/2-2-how-to-use-cbvs.ipynb:
------------------
from lightkurve import search_lightcurve
from lightkurve.correctors import load_tess_cbvs, load_kepler_cbvs
import numpy as np
lc = search_lightcurve('TIC 99180739', author='SPOC', sector=10).download(flux_column='sap_flux')
cbvs = load_tess_cbvs(sector=lc.sector, camera=lc.camera, ccd=lc.ccd, cbv_type='MultiScale', band=2)

Exception: CBVS were not found
```

Haven't had time yet to investigate why the CBVs are not found for this example.  I was hoping it might be a transient error (e.g. due to a new data set being loaded into MAST), but it appears to be persistent.

scfleming:
I did a quick check on MAST side and confirmed the CBV files for Sector 10 are present on disk where they should be.

barentsen:
Update: the failing notebook (`2-2-how-to-use-cbvs.ipynb`) works again.  There must have been a transient issue with retrieving TESS Sector 10 CBVs that has gone away.

I have now refreshed the online docs to make sure they consistently refer to `load_tess_cbvs` instead of `download_tess_cbvs` (same for Kepler).

PR #1126, opened 2021-07-30
Show the dependencies used during test

orionlee:
At times I wanted to know the versions of some dependent packages used during CI tests.

The "install dependencies" in CI output often does not have the information (possibly due to some form of caching), e.g., I cannot find astropy in https://github.com/lightkurve/lightkurve/runs/3176571199?check_suite_focus=true#step:4:1

As a workaround, this PR adds a dummy test that prints out the version of the main dependent packages

image

### DIFF
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -1,0 +1,15 @@
+import lightkurve as lk
+import numpy as np
+import astropy
+import scipy
+import astroquery
+
+def test_show_environment(capsys):
+    """Show the environment the test suite is running. """
+    with capsys.disabled():
+        print("Dependencies used: ")
+        print("lightkurve", lk.__version__)
+        print("astropy", astropy.__version__)
+        print("numpy", np.__version__)
+        print("scipy", scipy.__version__)
+        print("astroquery", astroquery.__version__)


barentsen:
Thanks @orionlee.  I agree it is very useful to be able to see the package versions!

When I go to the [CI output link you shared](https://github.com/lightkurve/lightkurve/runs/3176571199?check_suite_focus=true#step:4:1), I can see the AstroPy version installed under the "Install dependencies" tab.  Screenshot:

Screen Shot 2021-07-31 at 11 35 05 AM

The "Install dependencies" tab shows the output of running `poetry install`, which already appears to print the version numbers that are being installed.

If we want more information, we could print the version numbers using an extra command (e.g. `poetry show` or `poetry run pip list`).  We could add such a command after line 70 of our [GitHub CI workflow config file](https://github.com/lightkurve/lightkurve/blob/main/.github/workflows/lightkurve-tests.yml).  This seems redundant however, because `poetry install` already appears to print the version numbers.

I'd be hesitant to use a unit test to print this information, because it adds more noise to the tests (potentially causing e.g. important warnings to be overlooked).  I think a golden rule is that tests should show ~no output when everything is fine.

Does that make sense?

orionlee:
@barentsen You're correct that the astropy version is already in "Install dependencies" tab. 

My bad: I used browser's Find in page to look for it and couldn't find it, and incorrectly presume there was some kind of smart caching happening such that it is pre-installed.

Lesson learned: use the CI's search log feature instead.

PR #1127, opened 2021-07-30
lc.truncate() : support optional column

### LINKED ISSUE #1116: LightCurve.truncate() - let users optionally specify the column (of time) as the basis
#### Problem description
Use case:  I have a FoldedLightCurve, `lc_f`. Occasionally, I want to plot only part of it, e.g., there are variation over time that pollutes the folded lightcurve. I want to see if the pattern holds better in a short duration.

#### Proposal

Add an optional `column` argument to `LightCurve.truncate()` function. An user can easily select part of the `FoldedLightCurve` to plot

```python
lc_f.truncate(column="time_original", 1400, 1430).scatter();
```

If PR #1105 is merged, users can also specify the truncate duration in cycle too (with the new `FoldedLightCurve.cycle`)

```python
# Want to see if the pattern holds better in the first 5 cycles 
lc_f.truncate(column="cycle", 0, 5).scatter();
```

#### Environment

orionlee:
Close #1116

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,7 @@
 
 - Various improvements to the online documentation. [#1102]
 
-- Fixed a bug in ``TargetPixelFile.estimate_centroids`` which caused the column
+- Fixed a bug in ``TargetPixelFile.estimate_centroids()`` which caused the column
   and row coordinates reported to be off by 0.5. [#1103]
 
 - Fixed a bug which caused a light curve's meta data to be lost after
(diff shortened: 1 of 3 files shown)


orionlee:
I want to make note that the changes could let users truncate based on any numeric column-like data. E.g., conceivably, someone can use it to clip outliers:
```python
lc.turncate(95, 105, column="flux").plot();
```

barentsen:
@orionlee Thank you for this excellent change!

I rebased the PR and added a change log entry.  I intend to merge as soon as the tests pass!

PR #1128, opened 2021-07-30
lc.normalize() - remove warning on relative_unit

### LINKED ISSUE #1115: LightCurve.normalize() - remove flux in relative units warning (or make it optional)
#### Problem description

Use case:

I have some helpers to plot a LightCurve in some consistent style, including normalized to percentage.

```python
def plot_helper(lc):
    ...
    lc = lc.normalize(unit="percent")
    ax = lc.scatter()
    ...
```

Depending on situation, sometimes the incoming `lc` object is already normalized, but in some other units, e.g., unscaled.

In such case, `lightkurve` will emit a warning
```
 LightkurveWarning: The light curve already appears to be in relative units; `normalize()` will convert the light curve into relative units for a second time, which is probably not what you want.
```

To avoid the such warning, user would need to do additional `warnings.filterwarnings` before calling normalize.

``` python
        warnings.filterwarnings("ignore", category=lk.LightkurveWarning, message=".*in relative units.*")
        lc = lc.normalize(unit="percent")
```

### Proposal

Add an optional argument `emit_relative_units_warning=True`, to `normalize()`. Users can then conveniently do normalize without the warning in 1 line:
```python
    lc = lc.normalize(unit="percent", emit_relative_units_warning=False)
```

#### Environment

orionlee:
Close #1115

### DIFF
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -1053,15 +1053,6 @@ class LightCurve(QTimeSeries):
                 "not what you want".format(median_flux),
                 LightkurveWarning,
             )
-        # Warn if the light curve was already normalized before
-        if self.meta.get("NORMALIZED"):
-            warnings.warn(
-                "The light curve already appears to be in relative "
-                "units; `normalize()` will convert the light curve "
-                "into relative units for a second time, which is "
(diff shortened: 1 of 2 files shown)


barentsen (approved):
Thanks @orionlee! :+1:

PR #1133, opened 2021-08-05
new bias background corrector

rebekah9969:
During TESS's primary mission relatively dim and/or crowded target stars were often subject to overestimated background levels, which would result in underestimated flux values and therefore overestimated transit depths.

In Sector 27 a procedure was added to improve the accuracy of the background correction. 

This update has been applied to all 2-minute and  20-second targets. Previous Sector data however, has not been corrected. While the change in the background estimates and relative transit depths is generally small (< 2% for transit depths and < 1% for planet radii),  for some dim and/or crowded targets the effect is larger.

This corrector class calculates the scaler offset (pdcCorrection) to apply to the PDCSAP_FLUX in Years 1 and 2, to correct for the measured background bias . It also calculates the predicted bias (transitBias) in the transit depth given the bias in the background estimate. This value is diagnostic only, but useful in determining if the background bias is significant for a given target object.

### DIFF
--- a/tests/correctors/test_biasbackgroundcorrector.py
+++ b/tests/correctors/test_biasbackgroundcorrector.py
@@ -1,0 +1,100 @@
+"""Unit tests for the `BiasBackgroundCorrector` class."""
+import pytest
+import warnings
+
+import numpy as np
+from numpy.testing import assert_almost_equal
+from scipy import ndimage
+
+import lightkurve as lk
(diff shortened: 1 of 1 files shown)


rebekah9969:
Fixed pervious issue which failed the check. Also adjusted saturation bound.

rebekah9969:
Adjusted for units and meta data

PR #1139, opened 2021-09-12
Add line "from matplotlib import animation"

isaac-yong0804:


barentsen:
@isaac-yong0804 Thank you for opening a PR! :+1:

Question: would you be willing to clarify why you think `from matplotlib import animation` needs to be added at the top of `targetpixelfile.py`?  I don't see any line in the file that accesses a function from `animation.*`.

Did you encounter a specific error that was resolved by this addition?

Thanks++!

isaac-yong0804:
@barentsen Hi. I was using this library in colab. When I do `pixelfile.animate()`, I got `AttributeError: module 'matplotlib' has no attribute 'animation'` in lightkurve/targetpixelfile.py line 1208.
I guess the reason is stated [here]( https://stackoverflow.com/a/14812396).

barentsen:
@isaac-yong0804 Thank you for the additional details! :+1:

I am able to reproduce this issue on Google Colab, which uses matplotlib v3.2.2 by default.  Upgrading matplotlib to v3.4.3 on Colab appears to resolve the issue, which explains why I wasn't initially able to reproduce this problem locally.

Thank you for this contribution!  I intend to merge this PR after adding a change log entry.

PR #1144, opened 2021-09-23
Duplicate

rebekah9969:
This code helps the user to identify the duplicate and split stars which are a result of the TIC 8.2 version update. This issue was caused by updating the catalog from 2MASS to Gaia. 

searchduplicate creates a table that indicates if the object of interest is part of the <1% of stars affected by the update and also lists the kind of issue present.

Then TIC82Correct adjusts the SAP flux for the affected object - correcting it and making it more comparable to that in later sectors. 
The user will have to correct the SAP flux for background and noise still and replace the PLDSAP flux.


rebekah9969:
Need to re-run now changes have been made to lightkurve to fix dependancies.

PR #1155, opened 2021-11-04
CBVCorrector working with 20-second CBVs.

jcsmithhere:
Here is a branch to get the CBVCorrector working with 20-second "fast" CBVs.

CBVCorrector will now automatically identify if the TESS light curve under study is 20-second or 2-minute cadence data and download the appropriate CBVs. The tutorials have been updated to explain the new functionality. 

For Kepler, only 30-minute CBVs are available so not much changed there.

New meta fields have been created in the light curve objects to help identify the cadence type:
```
lc.meta["EXPTIME"] = lc.meta.get("FRAMETIM") * lc.meta.get("NUM_FRM")
lc.meta["CADENCE_TYPE"] = (TESS: 'long', 'short', 'fast) or (KEPLER/K2: 'ffi', 'long', 'short')
```
Helper functions created are `io/tess.get_tess_cadence_type` and `io/kepler.get_kepler_cadence_type` to determine the cadence type from the exposure time. A third helper function is `cbvcorrector._determine_exptime_from_lc` but that one is not intended for public use.

It would be appreciated if someone could go through the tutorials and see if everything works for them and makes sense.

### DIFF
--- a/docs/source/tutorials/2-creating-light-curves/2-3-how-to-use-cbvcorrector.ipynb
+++ b/docs/source/tutorials/2-creating-light-curves/2-3-how-to-use-cbvcorrector.ipynb
@@ -28,16 +28,29 @@
    "metadata": {},
    "source": [
     "There are three basic types of CBVs: \n",
-    "\n",
     "- **Single-Scale** contains all systematic trends combined in a single set of basis vectors. \n",
-    "\n",
     "- **Multi-Scale** contains systematic trends in specific wavelet-based band passes. There are usually three sets of multi-scale basis vectors in three bands.\n",
-    "\n",
     "- **Spike** contains only short impulsive spike systematics.\n",
     "\n",
-    "There are two different correction methods in PDC: Single-Scale and Multi-Scale. Single-Scale performs the correction in a single bandpass. Multi-Scale performs the correction in three separate wavelet-based bandpasses. Both corrections are performed in PDC but we can only export a single PDC light curve for each target. So, PDC must choose which of the two to export on a per-target basis. Generally speaking, single-scale performs better at preserving longer period signals. But at periods close to transiting planet durations multi-scale performs better at preserving signals. PDC therefore mostly chooses multi-scale for use within the planet finding pipeline and for the archive. You can find in the light curve FITS header which PDC method was chosen (keyword “PDCMETHD”). Additionally, a seperate correction is alway performed to remove short impulsive systematic spikes.\n",
+    "There are two different correction methods in PDC: Single-Scale and Multi-Scale. Single-Scale performs the correction in a single bandpass. Multi-Scale performs the correction in three separate wavelet-based bandpasses. Both corrections are performed in PDC but the mission can only export a single PDC light curve for each target. So, PDC must choose between the two on a per-target basis. Generally speaking, single-scale performs better at preserving longer period signals. But at periods close to transiting planet durations multi-scale performs better at preserving signals. PDC therefore mostly chooses multi-scale for use within the planet finding pipeline and for the archive. You can find in the light curve FITS header which PDC method was chosen (keyword “PDCMETHD”). Additionally, a seperate correction is alway performed to remove short impulsive systematic spikes.\n",
     "\n",
-    "For an individual's research needs, the mission supplied PDC lightcurves might not be ideal and so the CBVs are provided to the user to perform their own correction. All three CBV types are provided at MAST for TESS, however only Single-Scale is provided at MAST for Kepler and K2. Also for Kepler and K2, Cotrending Basis Vectors are supplied for only the 30-minute target cadence."
+    "For an individual's research needs, the mission supplied PDC lightcurves might not be ideal and so the CBVs are provided to the user to perform their own correction. All three CBV types are provided at MAST for TESS, however only Single-Scale is provided at MAST for Kepler and K2."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Cadence Type CBV Availability"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "For Kepler and K2 CBVs are available for only the 30-minute 'long' cadence (technically 29.4 minute cadence).\n",
+    "\n",
+    "For TESS, CBVs are available for 2-minute 'short' and 20-second 'fast' cadence."
    ]
   },
   {
@@ -51,7 +64,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "One can directly obtain the CBVs with `load_tess_cbvs` and `load_kepler_cbvs`, either from MAST by default or from a local directory `cbv_dir`. However when generating a [CBVCorrector](https://docs.lightkurve.org/reference/api/lightkurve.correctors.CBVCorrector.html?highlight=cbvcorrector) object the appropriate CBVs are automatically downloaded from MAST and aligned to the lightcurve. Let's generate this object for a particularily interesting TESS variable target. We first download the SAP lightcurve."
+    "One can directly obtain the CBVs with `load_tess_cbvs` and `load_kepler_cbvs`, either from MAST by default or from a local directory `cbv_dir`. However when generating a [CBVCorrector](https://docs.lightkurve.org/reference/api/lightkurve.correctors.CBVCorrector.html?highlight=cbvcorrector) object, the appropriate CBVs are automatically downloaded from MAST and aligned to the lightcurve. Let's generate this object for a particularily interesting TESS variable target. We first download the SAP lightcurve."
    ]
   },
   {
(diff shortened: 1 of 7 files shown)


barentsen:
Thanks Jeff -- I expect to test/review this first thing tomorrow!

jcsmithhere:
Thanks, @barentsen! I also asked a couple other people to look at it.

barentsen:
> I also asked a couple other people to look at it.

Perfect.  The more feedback the better!!

src/lightkurve/io/tess.py
```diff
@@ -46,4 +46,48 @@ def read_tess_lightcurve(
...
+def get_tess_cadence_type(exptime):
+    """ Returns the TESS cadence type as a string based on the exposure time.
+
+    Alternatively, if a string is pass then this function checks for a valid 
+    cadence type and then returns it. It raises an exception if not valid.
+
+    The options are:
+    if   exptime < 60   : "fast"  (i.e. 20-second)
+    elif exptime < 300  : "short" (i.e. 2-minute)
+    else                : "long"  (i.e. 30 or 10 minute FFI)
```
barentsen, line 62: I wonder if we want to use the string `"ffi"` instead of `"long"`, because I often hear people say "TESS FFI light curves" and I've ~never heard "TESS long cadence light curves".  (But I might be wrong -- is "long" is already adopted somewhere?)

barentsen (changes requested):

src/lightkurve/io/tess.py
jcsmithhere: I'm debating this too. I've heard both. The mission uses FFI but I've seen others say "long". I would prefer "ffi'. If you are good with that then I'll change it to 'ffi'

barentsen: I think "ffi" is the way to go!

jcsmithhere: Terminology changed to 'ffi' and merged to branch.

dacmess:
@jcsmithhere I'm am but an egg and am not sure of the correct way to comment on your pull request, but I think for general comments, I can just put them here(?). 
In Tutorial 2-2-how-to-use-cbvs:  This tutorial worked fine for me and was clear.
-In the intro, you might change "Single-Scale" and "Multi-Scale" to "SingleScale" and "MultiScale" to match the actual python keyword values for cbv_type.
-You mention setting extrapolate=False sets the CBV values to 0 and will likely lead to a bad fit, but you might caution about extrapolate=True also likely being problematic and maybe it's better to crop the LC (I realize this was in the original code and not just introduced).

docs/source/tutorials/2-creating-light-curves/2-3-how-to-use-cbvcorrector.ipynb
```diff
@@ -146,9 +166,9 @@
...
+    "- **Under-fitting metric**: Measures the mean residual target to target Pearson correlation between the target under study and a selection of neighboring targets. This metric will find and download a selection of neighboring SPOC SAP targets in RA and Decl. until a minimum number is found. It attempts to find lightcurves with the same cadence type as the target under study. The metric is callibrated such that a value of 0.95 means the residual correlations in the target is equivalent to chance correlations of White Gaussian Noise.\n",
     "\n",
-    "_The Goodness Metrics are not perfect!_ They are an estimate of over- and under-fitting and are to be used as a guideline along other other metrics to assess the quality of your light curve."
+    "_The Goodness Metrics are not perfect!_ They are an estimate of over- and under-fitting and are to be used as a guideline along other metrics to assess the quality of your light curve."
```
dacmess, line 171: "...used as a guideline along with other metrics...."

dacmess:
On tutorial 2-3-how-to-use-cbvcorrector:
typo in description after correct_gaussian_prior() cell: should be "The very strong dips right at the beginning..." rather than "lips"
More generally, my correction for TIC 99180739 doesn't look like a good correction, am I missing something?
Screen Shot 2021-11-10 at 12 53 35 PM

And here's my optimized-Alpha fit result. Doesn't seem consistent with the tutorial discussion...?
Screen Shot 2021-11-10 at 1 11 11 PM

dacmess:
On tutorial 2-3-how-to-use-cbvcorrector:
The 20-sec fast cadence example worked great. I think that is good.
However, I'm getting an error in the Goodness Metrics and CBVCorrector with other Design Matrices section. I suspect it might be related to my python/package configuration, but I'm not sure. It is on the line:
> cbvcorrector.goodness_metric_scan_plot(cbv_type=cbv_type, cbv_indices=cbv_indices, ext_dm=dm);
This is probably not something you need to diagnose for this pull request. 
The bottom of the error list is:
Screen Shot 2021-11-10 at 2 17 37 PM

docs/source/tutorials/2-creating-light-curves/2-3-how-to-use-cbvcorrector.ipynb
jcsmithhere: Fixed

jcsmithhere:
> -In the intro, you might change "Single-Scale" and "Multi-Scale" to "SingleScale" and "MultiScale" to match the actual python keyword values for cbv_type. 

@dacmess, I added a statement saying that we reference the CBV types using CamelCase, that being the convention used by the missions. 

> -You mention setting extrapolate=False sets the CBV values to 0 and will likely lead to a bad fit, but you might caution about extrapolate=True also likely being problematic and maybe it's better to crop the LC (I realize this was in the original code and not just introduced).

I added a sentence warning about the dangers of extrapolation.

jcsmithhere:
> More generally, my correction for TIC 99180739 doesn't look like a good correction although the description says that it should look ok. Am I missing something?

@dacmess, I think there might be something wrong with your configuration. You seem to get different results. Your light curve also has additional ungapped cadences. Compare your results with those from the older, published tutorial :

https://docs.lightkurve.org/tutorials/2-creating-light-curves/2-3-how-to-use-cbvcorrector.html 

I wonder if the way your LightKurve configuration is handling quality flags is different than the default. Did you try restarting your kernel and running all cells in the tutorial with no changes?

jcsmithhere:
> I reran both tutorials using a poetry install of this branch, rather than adjusting my system path, and I am getting the same behavior and same error as above. Which is on this line in the tutorial:
> % cbvcorrector.goodness_metric_scan_plot(cbv_type=cbv_type, cbv_indices=cbv_indices, ext_dm=dm);
> Cells past this one worked fine, so it appears to just be the scan_plot that is having an issue. I'm still not sure if this is a local setup issue for me, but is poetry supposed to take care of that?

@dacmess, I wonder if it is an incompatibility with your astropy installation. Here is my Lightkurve and astropy install versions:

```
$ conda list | grep lightkurve
lightkurve                2.0.11.dev0               dev_0    <develop>
$ conda list | grep astropy
astropy                   4.2              py38h497a2fe_2    conda-forge
```

dacmess:
@jcsmithhere I restarted everything and got similar errors in the 2.3 tutorial. The first difference I noticed from the saved version of the tutorial is on cell #4: cbvCorrector.cbvs[0].plot(); 
My version appear to have only CBVs with any content, whereas the saved version has 16 (see plots below). 
I am running the jupyter notebook through poetry after doing "poetry install" in the lightcurve directory. My various version are:
(base) ➜  lightkurve git:(cbvcorrector_feature/load_20-sec_cbvs) ✗ conda list |grep lightkurve
lightkurve                2.0.6                    pypi_0    pypi
(base) ➜  lightkurve git:(cbvcorrector_feature/load_20-sec_cbvs) ✗ poetry version
lightkurve 2.0.12dev
(base) ➜  lightkurve git:(cbvcorrector_feature/load_20-sec_cbvs) ✗ conda list |grep astropy
astropy                   4.2                      pypi_0    pypi
My conda version of lightkurve is 2.06, by the version running in poetry is 2.0.12dev. Are the two versions stepping on each other, or have I messed up when using poetry?
caldwell_cbv_plot
saved_cbv_plot

jcsmithhere:
@dacmess, it does look like you indeed do have two versions of LighKurve stepping on each other. However, it is not clear to me how that would result in you seeing different CBVs in the plot tool than the tutorial. I don't think I changed anything in my code that would result in half the CBVs being zero. It's possible the code you're using is loading only the first 8 CBVs, but I don't think I did anything to change that. 

I really have very little experience with poetry so I can't help there.

dacmess:
@jcsmithhere I'll stop spamming this ticket after this, since I think you've got what you need for the 20 sec CBV changes. 

I started from scratch with a new python environment (python 3.8.12) and only installed lightkurve (2.0.12.dev0) and what was needed to run jupyter notebook. Astropy is at version 4.3.1. I am getting the exact same behavior as before; i.e. starting with only 8 valid CBVs. That got me to thinking that the problem might be reprocessed S10 data vs the original? Didn't we switch from delivering all 16 CBVs to not delivering CBVs beyond 8 (or vice versa) with the reprocessing? Could this be the source of the difference between what I'm seeing and the saved tutorial? Of course that doesn't explain the error I'm still getting on the scan_plot when combining CBV and DM corrections.

Issue #1161, opened 2021-11-11
More `MaskedNDArray` failures with Astropy 5.0


dhomeier:
#### Problem description
Found a couple more test failures with astropy 5.0rc2 that I missed previously (likely were present with 5.0rc1 as well, but I did not test with remote-data). All passing with 4.3.1 in the same environment; I have not looked into details, but they all seem in some way related to handling `MaskedNDArray` (rather than unmasked NDData or `np.ma.array`) now.

The first test e.g. could be fixed by substituting
`assert np.isnan(getattr(lc.flux, 'data', lc.flux)).any()  # ensure the test data has nan in flux`

#### Example
```python
poetry run pytest --remote-data

______________________________________________________________________ test_preprocess_lc ______________________________________________________________________

    @pytest.mark.remote_data
    def test_preprocess_lc():
        """Test to ensure the lightcurve is pre-processed before applying BLS for correctness and consistent output"""
        from lightkurve.interact_bls import _preprocess_lc_for_bls
    
        lc = KeplerLightCurve.read(KEPLER10)
>       assert np.isnan(lc.flux).any()  # ensure the test data has nan in flux
E       assert MaskedNDArray(False)
E        +  where MaskedNDArray(False) = <bound method MaskedNDArray.any of MaskedNDArray([  ———, False, False, ..., False, False, False])>()
E        +    where <bound method MaskedNDArray.any of MaskedNDArray([  ———, False, False, ..., False, False, False])> = MaskedNDArray([  ———, False, False, ..., False, False, False]).any
E        +      where MaskedNDArray([  ———, False, False, ..., False, False, False]) = <ufunc 'isnan'>(<MaskedQuantity [      ———, 575100.6 , 575187.25, ..., 575059.7 ,\n                 575059.3 , 574996.6 ] electron / s>)
E        +        where <ufunc 'isnan'> = np.isnan
E        +        and   <MaskedQuantity [      ———, 575100.6 , 575187.25, ..., 575059.7 ,\n                 575059.3 , 574996.6 ] electron / s> = <KeplerLightCurve length=1011 LABEL="KIC 11904151" QUARTER=4 AUTHOR=Kepler FLUX_ORIGIN=pdcsap_flux>\n       time       ...58  5.7505931e+05 ...  6.4287381e-03  4.3517463e-02\n373.21904601071583  5.7499662e+05 ...  6.4082127e-03  4.3601818e-02.flux

tests/test_interact_bls.py:85: AssertionError
[...]

PR #1162, opened 2021-11-15
Make Lightkurve compatible with Astropy v5.0 and Python 3.10

barentsen:
As reported by @dhomeier in https://github.com/lightkurve/lightkurve/issues/1161, a number of Lightkurve unit tests are currently failing with Astropy v5.0rc2.

I intend to use this PR to investigate and resolve those test failures.

### DIFF
--- a/.github/workflows/lightkurve-tests.yml
+++ b/.github/workflows/lightkurve-tests.yml
@@ -1,10 +1,10 @@
 # GitHub Actions workflow for Lightkurve's continuous integration.
 #
 # This file configures seven different jobs:
-# 1) pytest on Linux, Python 3.7.
-# 2) pytest on Linux, Python 3.7.
-# 3) pytest on Linux, Python 3.8, with --remote-data enabled (i.e. get data from MAST).
-# 4) pytest on Linux, Python 3.9.
+# 1) pytest on Linux, Python 3.8, with -m memtest (i.e. stress memory)
+# 2) pytest on Linux, Python 3.8, with --remote-data enabled (i.e. get data from MAST).
+# 3) pytest on Linux, Python 3.9.
+# 4) pytest on Linux, Python 3.10.
 # 5) pytest on Windows, Python 3.8.
 # 6) pytest on OSX, Python 3.8.
 # 7) flake8 syntax check.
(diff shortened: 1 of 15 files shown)


barentsen:
Note: a few outstanding issues affecting `LightCurve.bin()` are caused by edge case bugs in AstroPy's `aggregate_downsample` function.  I opened an AstroPy PR to resolve some of them: https://github.com/astropy/astropy/pull/12527

The implication is that we may want to support AstroPy v5.0.1 rather than v5.0.0.

dhomeier:
I found another more serious bug with the new `MaskedQuantity`: this breaks `LombScarglePeriodogram` since `np.isnan(lc.flux)` is never triggered on flux with masked `NaN`s, returning `MaskedNDArray(False)` instead.
`lc.remove_nans()` itself works, but is not called because of this. `timeseries.LombScargle` however still returns all-NaN with the masked NaN inputs.
This is relatively easily caught, but I think `np.isnan` is such an ubiquitous test that it's probably worth another issue in Astropy; ideally I guess `timeseries.LombScargle` would just work with those inputs – I think that was the idea for the new `Masked` classes, but no idea how easily that might be realised.

barentsen:
@dhomeier Interesting!  The issue you describe wasn't even on my radar yet.  Do you happen to have a minimum example that reproduces it?

dhomeier:
Basically any lightcurve with NaNs in flux, I'd think – I ran into it trying to recreate the example from the [Periodogram Tutorial](https://docs.lightkurve.org/tutorials/3-science-examples/periodograms-creating-periodograms.html#2.-Creating-a-Periodogram-from-Kepler-Data)
```python
lc = lk.search_lightcurve('KIC 10264202', author="Kepler", quarter=10).download()
pg = lc.normalize(unit='ppm').to_periodogram()
np.isfinite(pg.power).any()
np.isnan(pg.power).all()
```

barentsen:
Update: this PR is starting to near completion.  I made various small bugfixes to ensure the unit tests now pass with AstroPy v5.0.0 on Python 3.8, 3.9, and 3.10.  This ensures that Lightkurve can continue to be used with the latest versions of AstroPy and Python.

Note: this PR updates the minimum dependencies of Lightkurve to become python>=3.8 and astropy>=5.0.  These new requirements would take effect from the next release (v2.1).  Users of older versions of AstroPy and Python can of course continue to use Lightkurve v2.0.x, which we can maintain as needed.

One outstanding issue is the possible bug reported by @dhomeier above (https://github.com/lightkurve/lightkurve/pull/1162#issuecomment-983847177).  I expect to investigate this issue tomorrow.

barentsen:
I was able to reproduce and fix the issue reported by @dhomeier above in https://github.com/lightkurve/lightkurve/pull/1162#issuecomment-983847177.  I also added a regression test.

I intend to merge this PR shortly and use it as the base for a Lightkurve v2.1 release soon.

I intend to make further modifications to the CI system by introducing the use of `tox`, but I will complete this work in a separate PR.

Issue #873, opened 2020-10-11
One Sector 6 cadence has all-zero flux in TESSCut data


barentsen:
I find that TESS Sector 6 data of TIC 123835353 near BTJD=1477.015 contains all-zero flux in TESSCut data, even though the Target Pixel File appears fine.

**Target Pixel File flux values**

<img width="769" alt="Screen Shot 2020-10-10 at 17 15 22" src="https://user-images.githubusercontent.com/817669/95667538-5cde6280-0b1c-11eb-8b2a-cfb8bee0e094.png">

**TESSCut flux values**

<img width="671" alt="Screen Shot 2020-10-10 at 17 15 56" src="https://user-images.githubusercontent.com/817669/95667540-6c5dab80-0b1c-11eb-8423-7f5811b2d653.png">

I haven't been able figure out yet if the original FFI data are indeed all-zero, or if this is an issue in TessCut instead. /cc @ceb8

ceb8:
@barentsen Oh weird, I will take a look. Thanks for noticing. 😝

barentsen:
Closing this issue because the Lightkurve side of the issue has been fixed, but I encourage anyone to post updates below as needed!

jaymedina:
Hi Geert! I wanted to confirm your latest update that this was no longer a problem for `Lightkurve` by running the command in your second screenshot: 

`fromLK = lk.search_tesscut("TIC 123835353", sector=6).download(cutout_size=5)`

and got the following error: 

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/miniconda3/envs/astroquery395/lib/python3.9/site-packages/lightkurve/search.py in _download_one(self, table, quality_bitmask, download_dir, cutout_size, **kwargs)
    230                 )
--> 231                 path = self._fetch_tesscut_path(
    232                     table[0]["target_name"],

~/miniconda3/envs/astroquery395/lib/python3.9/site-packages/lightkurve/search.py in _fetch_tesscut_path(self, target, sector, download_dir, cutout_size)
    505         # this is necessary to ensure cutouts are not downloaded multiple times
--> 506         sec = TesscutClass().get_sectors(coords)
    507         sector_name = sec[sec["sector"] == sector]["sectorName"][0]

TypeError: get_sectors() takes 1 positional argument but 2 were given

During handling of the above exception, another exception occurred:

SearchError                               Traceback (most recent call last)
[censoring for security] in <module>
[...]

PR #1176, opened 2022-03-24
Fix TESSCut downloads with Astroquery v0.4.6

barentsen:
This PR fixes the issue that TESSCut downloads via Lightkurve no longer work when users upgrade to Astroquery v0.4.6 (released two days ago).  

The issue was first discovered and reported by @jaymedina over in https://github.com/lightkurve/lightkurve/issues/873#issuecomment-1077052232.  Thank you!

### Example

The following snippet:
```python
lk.search_tesscut("TIC 123835353", sector=6).download(cutout_size=5)
```
yields a
```
TypeError: get_sectors() takes 1 positional argument but 2 were given
```

### Solution

The issue is caused by the fact that `astroquery.mast` changed a few arguments into keyword-only arguments (cf. https://github.com/astropy/astroquery/pull/2317).  Lightkurve was using these arguments as positional arguments.

The fix is trivial.  I plan to release a new version of Lightkurve later today (v2.1.1) to make this bugfix easily available.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -53,6 +53,7 @@ The following GitHub members directly contributed code and bugfixes:
 - `Warrick Ball <https://github.com/warrickball>`_
 - `Isaac Yong <https://github.com/isaac-yong0804>`_
 - `Gutsycat <https://github.com/Sniperq2>`_
+- `Jennifer V Medina <https://github.com/jaymedina>`_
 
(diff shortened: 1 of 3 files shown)


barentsen:
Looks like GitHub Actions is having issues right now.  I verified the tests pass locally.  Merging!

PR #1188, opened 2022-04-05
retire QColumn

### LINKED ISSUE #1180: Retire QColumn
`QColumn`, and associated `QMaskedColumn`, `QTimeSeries`, can all be retired, given `Astropy.Column` now provides `.value` alias natively.

https://github.com/lightkurve/lightkurve/blob/ad8eb62ae0f6b9d7ecf4183c391fe9891a62a969/src/lightkurve/lightcurve.py#L69

Version: Lightkurve 2.1.1

orionlee:
replaced by native astropy support.
Close #1180

### DIFF
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -48,78 +48,8 @@ def _to_unitless_day(data):
     else:
         return data
 
-class QColumn(Column):
-    """(Temporary) workaround to provide ``.value`` alias to raw data, so as to match ``Quantity``.
-    """
 
-    @property
-    def value(self):
-        return self.data
-
-
-class QMaskedColumn(MaskedColumn):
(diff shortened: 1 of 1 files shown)


barentsen:
This is brilliant.  All the tests pass!

barentsen:
I think we can safely remove these classes.  They were only ever introduced to make Lightkurve v2.0.x compatible with AstroPy v4.2 and earlier.  We no longer need them because Lightkurve v2.1.0 raised the minimum dependency of AstroPy to v5.0 to take advantage of the new `MaskedQuantity` feature and various bugfixes.

I double-checked the code base to verify that `QColumn`, `QMaskedColumn`, and `QTimeSeries` are not used elsewhere in the code. (They are not.)  They have also never been advertised publicly.

I will merge this PR and add a change log entry.

Thanks so much @orionlee!!

PR #1189, opened 2022-04-06
Fix pandas deprecation warning

keyuxing:
This is a small PR to replace `DataFrame.append` (deprecated in [pandas v1.4.0](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.4.0.html#whatsnew-140-deprecations-frame-series-append)) with `pandas.concat` in `_query_solar_system_objects`.

Without this fix, executing `lc.query_solar_system_objects` currently yields the following deprecation warning:
```python
/opt/homebrew/Caskroom/mambaforge/base/lib/python3.10/site-packages/lightkurve/utils.py:742: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.
  df = df.append(res)
```

### DIFF
--- a/src/lightkurve/utils.py
+++ b/src/lightkurve/utils.py
@@ -739,7 +739,7 @@ def _query_solar_system_objects(
             if df is None:
                 df = res
             else:
-                df = df.append(res)
+                df = pd.concat([df, res])
     if df is not None:
         df.reset_index(drop=True)
     return df


orionlee:
@barentsen The change looks good to me. 

I've also verified that using `pandas.concat` is safe in the sense that it is supported by the minimal pandas (1.1.4) needed for lightkurve. https://github.com/lightkurve/lightkurve/blob/b709aeb6066f8f87f7737713006b92c0b68c420f/pyproject.toml#L37

https://pandas.pydata.org/pandas-docs/version/1.1/reference/api/pandas.concat.html

barentsen:
Thank you @keyuxing and @orionlee! :+1: :+1:

PR #1177, opened 2022-03-28
Fix TPF interact select pixels problems with bokeh 2.3.x or later

orionlee:
Close #990

The starting point is the stalled PR #1057.
If there are interests to get the issue resolved, I'd work on polishing it. Let me know.

Anecdotally, I came across a few people new to Lightkurve, could not get `tpf.interact()` to work because of this issue and gave it up. Resolving it would improve new user experience.

PR #1193, opened 2022-04-09
tutorail tweak: note the shift-clicking limiation on Jupyter Lab.

orionlee:
As a follow-up on #1177, I suggest we state the shift-clicking limitation on the tutorial, to limit possible confusion by newbies.

I also changed the "deselect pixel" from shift-clicking to shit-ctrl-clicking, as this is how it works on Windows. I am not sure how it works on Mac, and probably needs to be furhter reworded.

### DIFF
--- a/docs/source/tutorials/1-getting-started/interactively-inspecting-data.ipynb
+++ b/docs/source/tutorials/1-getting-started/interactively-inspecting-data.ipynb
@@ -176,13 +176,14 @@
     "Interaction modes:\n",
     "\n",
     "- Clicking on a single pixel shows the time series light curve of that pixel alone.  \n",
-    "- `shift`-clicking on multiple pixels shows the light curve using that pixel mask.\n",
-    "- `shift`-clicking on an already selected pixel will *de*select that pixel.\n",
+    "- `shift`-clicking on multiple pixels shows the light curve using that pixel mask. (*)\n",
+    "- `shift`-`ctrl`-clicking on an already selected pixel will *de*select that pixel.\n",
     "- Clicking and dragging a box will make a rectangular aperture mask — individual pixels can be deselected from this mask by shift-clicking (box deselecting does not work).\n",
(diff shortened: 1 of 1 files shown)


barentsen:
Thank you for this useful update to the docs!

Using Jupyter Notebook on Mac OSX, I was not able to use `shift`-`ctrl`-click to deselect a pixel.  This keyboard combination always triggers a browser user menu to pop up (i.e. `shift`-`ctrl`-click appears to behave identical to right-click).

I tried other combinations of `shift` / `ctrl` / `option` / `command` and was not able to find a combination of keys that would let me deselect a pixel.

Adding a pixel using `shift`-click did work fine.

docs/source/tutorials/1-getting-started/interactively-inspecting-data.ipynb
```diff
@@ -176,13 +176,14 @@
     "Interaction modes:\n",
     "\n",
     "- Clicking on a single pixel shows the time series light curve of that pixel alone.  \n",
-    "- `shift`-clicking on multiple pixels shows the light curve using that pixel mask.\n",
-    "- `shift`-clicking on an already selected pixel will *de*select that pixel.\n",
+    "- `shift`-clicking on multiple pixels shows the light curve using that pixel mask. (*)\n",
+    "- `shift`-`ctrl`-clicking on an already selected pixel will *de*select that pixel.\n",
```
barentsen, line 180: Suggested edit (see previous comment):
```suggestion
    "- `shift`-`ctrl`-clicking on an already selected pixel will *de*select that pixel. (May only work on Windows systems.)\n",
```

orionlee:
I did some web search. It looks like deselecting on Mac OS seems to be an unresolved bokeh issue:
 https://github.com/bokeh/bokeh/issues/10317

gully:
At some point I was able to deselect on Mac with some combination of `shift` and maybe `ctrl` or `command` and click.  The one hiccup was that I could not deselect the *most-recently-selected-pixel*, because of the logic of how we tracked pixel changes.  Interesting to see that there may be a platform dependent friction point here.

PR #1203, opened 2022-04-28
Fix CDIPS tests due to changes in CDIPS DR5 release

orionlee:
They've updated naming convention so the old URL won't work anymore.
I've picked another TIC with up-to-date filename and is available.


barentsen:
Perfect.  Thank you!!

PR #1204, opened 2022-04-28
interact_sky(): include nearby TICs not in Gaia 

orionlee:
Close #1186
Also close #1183

Suggested changelog:
```
- Improved ``TargetPixelFile.interact_sky()``  to include TIC information even when
  the TICs are not found in Gaia. [#1204]
- Removed extraneous warnings from ``TargetPixelFile.interact_sky()``. [#1204]
```

Example of the effects:

The nearby TIC 233087857 below, without a Gaia DR2ID, are not shown in the current `interact_sky()`. 
With this PR, such TICs are also plotted:

image

PR #1209, opened 2022-05-10
fix interact_sky() add nearby tic, case type checking fails.

orionlee:
A bug fix for `interact_sky()`, the not yet released PR# #1204 (include TICs not found in Gaia)

In some cases (e.g., TIC 135100529), some  dtype of the Gaia result pandas dataframe is weird, e.g., `df['Source'].dtype` is an instance of `pd.Int64Dtype`, rather than the type class itself.

This PR includes the fix and a unit test that covers such case.

Changelog: I don't think it's needed, as PR# #1204 is not yet released.


barentsen:
Thanks @orionlee!  This looks like a good fix and test.

I agree a change log entry is not needed here!

PR #1197, opened 2022-04-16
Correction of typos in docs & changed contribution link

DeepekshGupta:
There were some minor typos in the docs like "which which"
in the install.rst file.
Also changed the link in the contributing.rst file from
https://docs.lightkurve.org/about to
https://docs.lightkurve.org/about/contributing.html
so that people can access the contribution guidelines page directly.
this is my first time creating a pull request, hope this is helpful

Issue #1101, opened 2021-07-04
Feature request: include the calculated errors of the exoplanet's period & transit depth


vvilhelmus:
#### Problem description

The .show_properties() of a periodogram does not include errors on attributes like period_at_max_power & depth_at_max_power

#### Expected behavior

When a value for a period or transit depth is obtained; some standardised calculated error or (even better) an upper & lower error should be given as well please.

#### Environment

-  lightkurve version (e.g. 1.0b6): 2.0.10

orionlee:
If you're referring to `BoxLeastSquaresPeriodogram`, would `compute_stats()` address your need?


https://github.com/lightkurve/lightkurve/blob/81bc8697480afd371e01a9c526487f25e414ba7d/src/lightkurve/periodogram.py#L1186-L1189

It is basically a wrapper over the underlying astropy method:

https://docs.astropy.org/en/v4.1/api/astropy.timeseries.BoxLeastSquares.html#astropy.timeseries.BoxLeastSquares.compute_stats

vvilhelmus:
Oh wow, thank you so much for this information, Sam @orionlee ! :) Now I feel silly for not knowing this... 
Upon closer inspection the .compute_stats() function is nowhere to be found in the API:
http://docs.lightkurve.org/reference/api/lightkurve.periodogram.Periodogram.show_properties.html
Other functions, which I have encountered in the tutorials, seem to be missing in the API as well, like the .get_transit_model() function for example. Could these missing functions perhaps be added to the API and/or tutorials, so newbie users like me can learn about them please?
But this .compute_stats() function does indeed give the error of the transit depth, which solves half of my request. :)
Is there also some hidden function, which can give the error in the exoplanet's period?

orionlee:
1. `compute_stats()` missing in apidoc is certainly an issue. Maybe you can open a PR to fix it? 

2. For error in exoplanet's period from BLS: I am not an expert on BLS so I can't speak with surety, but I think the way BLS works is not equipped with providing standard errors for identified periods. The power you get back in the Periodogram does give you a measure on how well a epoch / particular period describes the data. You have control on the fit criteria with `objective`. See:

https://docs.astropy.org/en/latest/api/astropy.timeseries.BoxLeastSquares.html#astropy.timeseries.BoxLeastSquares.power

vvilhelmus:
Assuming PR is Pull Request I am going to create one, sorry, I am new to this sort of thing.

[...]

PR #1210, opened 2022-05-24
compute_stats and get_transit_model added to API list

vvilhelmus:
Did this work?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,12 +15,14 @@
 
 - Fixed a bug in ``LightCurve['time', '<some-column>']``, which triggered an
   ``AttributeError`` when selecting a subset of columns including ``time``. [#1199]
-  
-- Fixed a bug in ``LightCurve()`` constructor, which triggered an 
+
+- Fixed a bug in ``LightCurve()`` constructor, which triggered an
   ``AttributeError`` when `data` parameter is a ``list`` or ``ndarray`` [#1199]
 
 - Fixed a bug in ``LightCurve.query_solar_system_objects`` which caused an
   excessive search radius to be used for TESS light curves. [#1208]
+
+- Various improvements to the online documentation. [#1197, #1210]
 
 
 
--- a/docs/source/reference/periodogram.rst
+++ b/docs/source/reference/periodogram.rst
@@ -45,3 +45,5 @@ Methods
   Periodogram.smooth
   Periodogram.to_seismology
   Periodogram.to_table
+  BoxLeastSquaresPeriodogram.compute_stats
+  BoxLeastSquaresPeriodogram.get_transit_model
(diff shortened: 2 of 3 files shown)


orionlee:
This looks good!

I just take a look the actual doc of the 2 methods. It'd be a bonus if we update the outdated reference to `astropy.stats.bls` (Astropy 3.x) with one referencing the updated class/method at:

https://docs.astropy.org/en/stable/api/astropy.timeseries.BoxLeastSquares.html

Syntax for referencing astropy classes/methods can be found in various other docstrings, e.g., the docstring of `from_lightcurve()` function.

orionlee:
@vvilhelmus will this PR satisfy issue #1101 from your perspective?

barentsen:
> At section 5 "pull upstream main" did not work for me

Oops, you are right, it should say `git pull upstream main` instead of `pull upstream main`.

orionlee:
@vvilhelmus 
1. Would you be interested in fixing the outdated `astropy.stats.bls` references in the doc?
2. For issues in https://docs.lightkurve.org/about/contributing.html , you're welcome to open a PR to fix them, or file them in an issue so that we can keep track of it.

Thanks!

orionlee (approved):

vvilhelmus:
Sam @orionlee , I have fixed the outdated BLS references, tested the new updated BLS references on my system, and managed to push the relevant periodogram.py file to this thread, yay! :raised_hands: 

When I was going to fix the contributing.rst file, I discovered that the missing 'git'-part had already been added by someone else, however, I cannot find any documentation about this recent fix yet, so I am not sure if I should push the change I did not fix myself to my fork on my Github account or do something else about it? :shrug:

orionlee (approved):

orionlee:
@vvilhelmus Thanks! 

The updated BLS reference looks good.

For `contributing.rst` file, it looks like the missing `git` was indeed fixed in PR #1197 . So it does not need to be done twice:

https://github.com/lightkurve/lightkurve/blame/main/docs/source/about/contributing.rst#L145

The change was not pushed to docs.lightkurve.org yet, which generally gets updated when a new release is made. 
It should be mentioned in [CHANGES.rst](https://github.com/lightkurve/lightkurve/blob/main/CHANGES.rst), but it seems to have slipped through the crack. @barentsen or I would fix the `CHANGES.rst` when this PR is merged.

vvilhelmus:
@orionlee & @barentsen  thank you for taking the time to teach me this. These fixes could more easily and much faster be done, when doing it yourselves. 
I didn't know about the CHANGES.rst file, I was actually browsing this forum looking for something that matches and after a while of finding nothing I just gave up.  :grin: 
Every time I post something here, I am learning something new, so again, thank you. :pray:

orionlee:
@vvilhelmus 
1. In case there is confusion: `CHANGES.rst` is a convention used as a summary for user-facing changes (including documentation), e.g., test-only changes are usually not covered. For definitive actual changes, one should consult commit history / `git blame`, etc.
2. We'd like to include you in [AUTHORS.rst](https://github.com/lightkurve/lightkurve/blob/main/AUTHORS.rst). What'd be the name you prefer to be credited as?

Issue #691, opened 2020-02-13
Should we standardize lightkurve warnings?


christinahedges:
Inside lightkurve we have two different ways to warn users about problems:

`log.warning('Cannot find Campaign or Quarter number.')`
`warnings.warn('`transit_midpoint` is deprecated, please use `t0` instead.',
                          LightkurveWarning)`

It's not clear to me why we have two different ways to warn users, and it feels like we should use the second way everywhere. Should we standardize this so that people contributing code know how to raise warnings and errors? @barentsen can you tell us the right™️ answer?

barentsen:
## Geert's Right Answer:tm:

We should (almost) never use `log.warning` or `log.error`.  Here's my take:

1. Whenever possible, we should design APIs that prevent errors.  For example, when an unobserved target is requested, `search` functions should return an empty `SearchResult` rather than an ugly error.
2. For emitting messages that help diagnose what Lightkurve is doing, we should use `log.debug()` and `log.info()`.  This may e.g. help a user understand why the `SearchResult` was empty.
3. When a user makes a benign mistake that should trigger them to fix their code, `warnings.warn` should be used to notify them.  The classic example is a `DeprecationWarning`.
4. When a fatal mistake is encountered, an `Exception` should be raised.

## Background

The Python docs (https://docs.python.org/3/howto/logging.html#when-to-use-logging) recommend the following:
* use warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning
* use logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted.

...however I feel that we should design the API to avoid the latter as much as possible.  A rare exception is when a feature is designed to be called exclusively by hand (e.g. `tpf.interact()`), in which case using `log.error("you need to set the notebook url")` leads to a more user-friendly experience than raising an exception.

## Footnote 

The occurence of `log.warning` you refer to shouldn't exist:

https://github.com/KeplerGO/lightkurve/blob/b93605f0b753e725d546527fa7b1657c2fd7bc7a/lightkurve/lightcurve.py#L1899

...when a light curve does not have a quarter or campaign keyword, the API should just gracefully set it to `None` and not be noisy about it.

Does that make sense?

PR #1214, opened 2022-06-06
make default download_dir configurable 

### LINKED ISSUE #1212: Make default `download_dir` configurable
#### Problem description

Right now, when users download lightcurve / TPF files, 
1. they'd by default be saved at `~/.lightkurve-cache`, unless users override it with `download_dir` parameter in `download()` / `download_all()` call.
2. When lightcurve files are downloaded as part of `CBVCorrector`, users cannot change the location at all (there is no `download_dir`). See https://github.com/lightkurve/lightkurve/blob/2f1dda2a9efaa884c66edc2e07bed6b77e8443fc/src/lightkurve/correctors/metrics.py#L330

#### Feature Request

1. let users optionally specify the default `download_dir`
2. Bonus: For `CBVCorrector` case, also expose `download_dir` so that users can override it at per-object basis.

orionlee:
Close #1212

Tasks:
- [x] make default `download_dir` configurable: a new config item `search_result_download_dir`
- [x] move the default `download_dir` from `~/.lightkurve-cache` to `~/.lightkurve/cache`
- [x] Support users in migration from `~/.lightkurve-cache` to `~/.lightkurve/cache`: issue an warning, and/or provide helper function for migration
- [x] Support override the default `download_dir` with `XDG_CACHE_HOME` environment variable, analogous to [Astropy config](https://docs.astropy.org/en/stable/config/index.html)
- [x] Update documentation:
  - [`download()`](https://docs.lightkurve.org/reference/api/lightkurve.SearchResult.download.html#lightkurve.SearchResult.download) and [`download_all()`](https://docs.lightkurve.org/reference/api/lightkurve.SearchResult.download_all.html#lightkurve.SearchResult.download_all) functions.
  - dev doc: use of `XDG_CACHE_HOME` to speed up remote-data test in local dev environment

Maybe:
- Let users specify `download_dir` for `CBVCorrector` (used when CBV corrector searches and downloads neighbor lightcurves).

### DIFF
--- a/docs/source/reference/config.rst
+++ b/docs/source/reference/config.rst
@@ -24,3 +24,36 @@ Access configuration values
   conf
   config.get_cache_dir
   config.get_config_dir
+
+
+Default Cache Directory Migration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Starting from ``Lightkurve`` version ``2.3.0``, the default cache directory is
+at ``~/.lightkurve/cache`` . The data files cached at the legacy location,
+``~/.lightkurve-cache``, will not be used.
+
+A warning is issued if the legacy ``~/.lightkurve-cache`` directory still exists.
+
+Migration suggestions for handling various scenarios:
+
+* To use the existing data files cached, move all the contents under ``~/.lightkurve-cache``
+  to ``~/.lightkurve/cache``, and remove ``~/.lightkurve-cache`` directory itself.
+
+* If you need to use older version of ``Lightkurve``, e.g., because of the requirements
+  of other packages / applications, you can:
+
+  #. Keep the cache at the legacy location ``~/.lightkurve-cache``
+  #. Instruct current ``Lightkurve`` to use the legacy location. In the user's ``lightkurve.cfg``, add:
+
+      [config]
+
+      cache_dir = /<your-home-directory>/.lightkurve-cache
+
+  #. The warning will no long appear once a custom ```cache_dir``` is specified.
+
+* To suppress the warning for any reason, you can set ``warn_legacy_cache_dir`` in the user's ``lightkurve.cfg``.
+
+    [config]
+
+    warn_legacy_cache_dir = False
--- a/docs/source/reference/config.rst
+++ b/docs/source/reference/config.rst
@@ -22,4 +22,5 @@ Access configuration values
   :toctree: api/
 
   conf
+  config.get_cache_dir
   config.get_config_dir
--- a/docs/source/reference/config.rst
+++ b/docs/source/reference/config.rst
@@ -29,7 +29,7 @@ Access configuration values
 Default Cache Directory Migration
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Starting from ``Lightkurve`` version ``2.3.0``, the default cache directory is
+Starting from ``Lightkurve`` version ``2.4.0``, the default cache directory is
 at ``~/.lightkurve/cache`` . The data files cached at the legacy location,
 ``~/.lightkurve-cache``, will not be used.
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,12 @@
+2.4.0 (unreleased)
+==================
+
+- Added the ability to configure the default cache directory used by
+  ``SearchResult.download()`` / ``SearchResult.download_all()``. [#1214]
+- Moved the default cache directory from ``$HOME/.lightkurve-cache``
+  to ``$HOME/.lightkurve/cache``. [#1214]
+
+
 2.3.0 (2022-07-07)
 ==================
 
--- a/docs/source/about/testing.rst
+++ b/docs/source/about/testing.rst
@@ -46,6 +46,17 @@ Running some of our tests requires external data, e.g. some require data to be d
 
     poetry run pytest test_targetpixelfile.py --remote-data
 
+.. note::
+    To speed up the tests, you can instruct the tests use a specific directory for data file cache, so that when the tests are re-run,
+    data files in the local directory will be used, instead of being downloaded over the network.
+
+    Specify the data file cache directory via `XDG_CACHE_HOME` environment variable. For example,
+
+    .. code-block:: bash
+
+        $ XDG_CACHE_HOME=/my_dev/lk-test-cache poetry run pytest test_targetpixelfile.py --remote-data
+
+    The data files will be cached at `/my_dev/lk-test-cache/lightkurve` in the above example.
 
 
 My tests passed, but I got warning messages
--- a/src/lightkurve/__init__.py
+++ b/src/lightkurve/__init__.py
@@ -68,6 +68,8 @@ class Conf(_config.ConfigNamespace):
     cache_dir
         Default cache directory for data files downloaded, etc. Defaults to ``~/.lightkurve/cache`` if not specified.
 
+    warn_legacy_cache_dir
+        If set to True, issue warning if the legacy default cache directory exists. Default is True.
     """
     # Note: when using list or string_list datatype,
     # the behavior of astropy's parsing of the config file value:
@@ -89,6 +91,12 @@ class Conf(_config.ConfigNamespace):
         module="lightkurve.config"
     )
 
+    warn_legacy_cache_dir = _config.ConfigItem(
+        True,
+        "If set to True, issue warning if the legacy default cache directory exists.",
+        cfgtype="boolean",
+        module="lightkurve.config"
+    )
 
 conf = Conf()
 
@@ -105,3 +113,6 @@ from .convenience import *
 from .collections import *
 from .io import *
 from .search import *
+
+from . import config
+config.warn_if_default_cache_dir_migration_needed()
--- a/src/lightkurve/__init__.py
+++ b/src/lightkurve/__init__.py
@@ -71,6 +71,13 @@ class Conf(_config.ConfigNamespace):
         module="lightkurve.search"
     )
 
+    search_result_download_dir = _config.ConfigItem(
+        None,
+        "Default download directory for data files",
+        cfgtype="string",
+        module="lightkurve.search"
+    )
+
 
 conf = Conf()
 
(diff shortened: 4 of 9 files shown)


orionlee:
User Documentation changes:

`download()` (`download_all()` has similar changes):
image

The new `get_cache_dir()`:
image

---

Developer Documentation changes (for running unit test locally)

image

orionlee:
@barentsen Feel free to provide any feedback / comments.

There is still an outstanding task of supporting users in migration from `~/.lightkurve-cache` to `~/.lightkurve/cache` (issue an warning, and/or provide helper function for migration, etc.), but is largely isolated use cases and code changes that shouldn't affect the main changes in codes and docs.

orionlee:
@barentsen Feedback appreciated on `~/.lightkurve-cache` migration

Originally I intended to provide some migration help: check whether `~/.lightkurve-cache` exists, and issue a warning if it does.

However, such warnings could become nuisances if users also use codes that rely on old versions of lightkurve (in particular lightkurve v1.x): `~/.lightkurve-cache` will still need to be there for such cases.

A migration by making `~/.lightkurve-cache` a symlink to the new `~/.lightkurve/cache` does not work either: the existing lightkurve codes require `~/.lightkurve-cache` to be a directory, and reject symlink (treated as a boundary condition and will use current directory instead).

To handle the migration, I think we have the following alternatives:

1. We warn users if `~/.lightkurve-cache` exists. For users who still require old versions of `lightkurve`, they have to suppress the warning themselves, or we provide a config parameter to disable the warning.
2. We only document that the migration is needed. For users who still require old versions of `lightkurve`, we suggest that they set the new `search_result_download_dir` config parameter to `~/.lightkurve-cache` (the old location) so that a single cache directory can be used.
3. Back off from moving `~/.lightkurve-cache` to `~/.lightkurve/cache`

---

Some (probably a subset of) packages that depend on `lightkurve`:
https://libraries.io/pypi/lightkurve/dependents

barentsen:
@orionlee Would it make sense to name the parameter `download_dir` instead of `search_result_download_dir`?

Apart from being shorter, it will make it easier to re-use the parameter when downloading files not related to the SearchResult API (e.g. when downloading CBV files?)

I believe Astropy config files can have top-level parameters, so perhaps we could have a top-level `download_dir` parameter instead of the `search_result_download_dir` parameter inside the `[search]` section?

orionlee:
@barentsen I was just thinking about renaming `search_result_download_dir` to `cache_dir` before receiving your comment. I think `cache_dir` is a bit more appropriate. In addition to the possibility of hosting CBV files, conceivably it'd serve as the basedir of persistent search result cache. 

**Planned action**: rename `search_result_download_dir` to a top level `cache_dir`.

Additionally, feedback would be appreciated for migration from `~/.lightkurve-cache` to `~/.lightkurve/cache`.
- I'm sightly leaning towards warning users if `~/.lightkurve-cache` exists, with a config option to turn off the warning (for cases that users are stuck with `~/.lightkurve-cache` because they still use , say, Lightkurve v1.x ).
- I also think we should give a heads-up to the users of a upcoming migration, e.g., an announcement on the Discussions board. I'm unclear on the reach of the Discussions board though.

orionlee:
- I've renamed `search_result_download_dir` to `cache_dir` under `[config]` section. See the end for the reasoning for moving it to `[config]` section rather than keeping it at top level.
- all default cache dir logic is now centralized in `config.get_cache_dir()` function.

---
Reason for keeping `cache_dir` in `[config]` section: 
- syntactically it can be kept as a top level config parameter. I feel it can be a bit error prone.

Scenario. If an user has an existing config:
```cfg
[search]
search_result_display_extra_columns = proposal_id,
```
The user wants to config `cache_dir` and append the config to the end.
```cfg
[search]
search_result_display_extra_columns = proposal_id,

cache_dir = /my_research/cache
```
The `cache_dir` *would not* be used, as it is seen as under `[search]` section.

While the mistake could be avoided by putting `cache_dir` ahead of the existing parameters, I think it's better to keep `cahce_dir` under `[config]` section, so that mistakes similar to the above scenario could be avoided. It also maps nicely to the API semantics, as it does affect the value of `config.get_cache_dir()` after the refactoring in the latest commit.

orionlee:
The documentation is now centralized in `config.get_cache_dir()`. Search's `download()` / `download_all()` now reference `config.get_cache_dir()`

image

barentsen:
> I think `cache_dir` is a bit more appropriate.

Yes, `cache_dir` sounds good to me!

> We warn users if `~/.lightkurve-cache` exists. 

I think this is a good solution, in particular if we ensure that (i) the warning is only raised once (upon `import lightkurve`!?) and (ii) we include a helpful suggestion on the action the user should take (deleting the directory?)

> I also think we should give a heads-up to the users of a upcoming migration, e.g., an announcement on the Discussions board.

While I don't expect an announcement in the Discussion board to reach a lot of users, it does seem like the perfect place to ensure that users who Google the warning message will easily find a webpage with an explanation. TL;DR I think it's a good idea if we copy/paste the full warning message in the thread!

> I've renamed `search_result_download_dir` to `cache_dir` under `[config]` section.

I agree your reasoning!  Perhaps the `cache_dir` parameter could be defined as a parameter of the `[io]` module, but I do think it's better to have it in the same place as the `config_dir`, i.e. I think `[config]` is the way to go.

:+1::+1::+1:

orionlee:
A warning is now issued when importing `lightkurve`:

image

I considered to issue the warning in `download()` / `download_all()` call, but feel the warning might become excessive, so I have the check in the package init instead.


The documentation that considers migration suggestion referenced by the warning:

image

barentsen:
I like the `UserWarning`!  Would it make sense to raise it only when the `SearchResult.download()` or `SearchResult.download_all()` method is called, to avoid confusing users who may not be downloading data products?

orionlee:
Yeah, raising the `UserWarning` in `SearchResult.download()` / `.download_all()` would be a more natural place, and work better for users who don't use `lightkurve` to download files as you suggested.

I ended up doing it in `import lightkurve` because if some users choose to ignore the warning for whatever reasons, repeated warnings when calling `download()` could be a bit annoying. 

I understand in theory, a warning should only show up once only by default, but somehow it often is not the case in practice in my experience (I never understand the reasons). On the flip side, one could argue that users really should not ignore the warning anyway, especially they are given a way to suppress it.

I was on the fence between raising the warning at `download()` vs at `import lightkurve` to begin with. If you feel also feel `download()` is better, then let's go with it.

barentsen:
> I ended up doing it in import lightkurve because if some users choose to ignore the warning for whatever reasons, repeated warnings when calling download() could be a bit annoying.

That would be super annoying.  Fortunately, I believe warnings issued via `warnings.warn()` only show up once by default (unless configured otherwise, e.g. via `python -W all`), whereas warnings issued via `log.warning()` more commonly display multiple times.  We'd want to use `warnings.warn()` in this case for reasons summarized in https://github.com/lightkurve/lightkurve/issues/691#issuecomment-586016365

orionlee:
GRA™️ - :rofl:

Anyway, so let's move the warning raising from `import lightkurve` to `download() / donwload_all()` then.

orionlee:
@barentsen I've moved the raising of default cache dir migration warning so that it is triggered only when `download()` / `download_all()` is called. 

However, when testing in my own environments (command line and Jupyter notebooks), the warning is repeatedly emitted to the display (see the sample script at the end for an example).

However, in a simpler isolated script, the `UserWarning` is indeed emitted to the display once only. 

I don't know what's missing. I did verify it is not due to warning filters configuration changes after importing `lightkurve`.


The lightkurve script that shows repeated `UserWarning` with this PR.
```python
import warnings
import pprint
import lightkurve as lk

sr = lk.search_lightcurve('tic77031414', sector=2, exptime=120, author='spoc')
print(sr)

lc = sr[0].download()
print("Download #1 ", len(lc))

lc = sr[0].download()
print("Download #2 ", len(lc))

lc = sr[0].download()
print("Download #3 ", len(lc))

lcc = sr.download_all()
print("Download_all #1 ", len(lcc))


print("Warning filters config")
pp = pprint.PrettyPrinter(indent=4)
pp.pprint(warnings.filters)
```


Isolated script that shows `UserWarning` only once, even with `import lightkurve`:
```
import warnings
import pprint

# see if the act of importing lightkurve will lead to UserWarning being emitted repeatedly
# (the answer is no)
import lightkurve as lk

def try_warning():
    warnings.warn("Some warning", category=UserWarning)


try_warning()
try_warning()
try_warning()

print("Warning filters config")
pp = pprint.PrettyPrinter(indent=4)
pp.pprint(warnings.filters)
```

orionlee:
I can't make `UserWarning` show up once only.

I also tried explicitly add a warning filter with `default` action in `config/__init__.py`. It did not have any effect.

```python
warnings.filterwarnings(
    "default",
    message=".*default Lightkurve cache directory has been moved.*",
    category=UserWarning,
    module=__name__)
```

- I tried action `once`. It did not work either.
- When I specified action `ignore`, the warning is properly suppressed.
- So the filter is catching the warning, but somehow `default` or `once` did not work.

orionlee:
Let's go with emitting warnings once upon `import lightkurve` then.

We could amend the warning message to state that it is related to `donwload()`, so that the users would know what the warning is related to download even when it shows up after `import lightkurve`:

Amendment in **_bold italic_**:
> UserWarning: The default Lightkurve cache directory, **_used by download(), etc.,_** has been moved to [new-dir]. Please move all the files in the legacy directory [legacy-dir] to the new location and remove the legacy directory. Refer to https://docs.lightkurve.org/reference/config.html#default-cache-directory-migration for more information.

docs/source/reference/config.rst
```diff
@@ -22,4 +22,38 @@ Access configuration values
...
 
   conf
+  config.get_cache_dir
   config.get_config_dir
+
+
+Default Cache Directory Migration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Starting from ``Lightkurve`` version ``2.3.0``, the default cache directory is
```
orionlee, line 32: @barentsen I intend to merge this PR in the next few days and bump the version to v2.4.0 . 
The version needs to be synced up with the one specified in the migration doc above.

Let me know if there is any concern.

PR #1217, opened 2022-06-13
Adding ability to query `eleanor-lite` light curves to `search`

vbkostov:
This code lets the user fetch 30-min candence GSFC-ELEANOR-LITE lightcurves from MAST. More information can be found at https://archive.stsci.edu/hlsp/gsfc-eleanor-lite. At the time of writing, only Sector 1 is available. 

New function has been added to /src/ilghtkurve/io, “eleanorlite.py”, following the established structure, along with necessary updates in detect.py, read.py, __init__.py, and search.py:

```python
lc = lk.search_lightcurve('TIC 140068425', author = 'GSFC-ELEANOR-LITE').download_all(quality_bitmask="none")
```

A corresponding test function has also been added as well, /tests/io/test_eleanorlite.py


christinahedges:
@vbkostov thanks for your pull request! I've worked with @vbkostov to make sure this PR has most of the things we need.  The `eleanor-lite` pipeline will have a lot of FFI light curves and will be released over the coming months, and so I think this is a valuable contribution for the community. I've run the tests so we can check this is passing, and I have added some comments for you that are mostly minor working changes. Once you've addressed these comments, I recommend we ask @barentsen for his input on this PR.

src/lightkurve/io/detect.py
```diff
@@ -47,6 +48,10 @@ def detect_filetype(hdulist: HDUList) -> str:
     if "mit/qlp" in hdulist[0].header.get("origin", "").lower():
         return "QLP"
 
+    # Is it a GSFC-ELEANOR-LITE light curve?
+    if hdulist[0].header.get("LITE") == True:
```
christinahedges, line 52: Will this work if this keyword doesn't exist? Maybe @barentsen can weigh in on whether this is the right way to do things, or if this is not good practice?

src/lightkurve/io/detect.py
```diff
@@ -47,6 +48,10 @@ def detect_filetype(hdulist: HDUList) -> str:
     if "mit/qlp" in hdulist[0].header.get("origin", "").lower():
         return "QLP"
 
+    # Is it a GSFC-ELEANOR-LITE light curve?
+    if (hdulist[0].header.get("AUTHOR") == 'Adina D. Feinstein') and (hdulist[0].header.get("LITE") == True):
```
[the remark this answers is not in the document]
christinahedges: Ah I see that you used both, maybe both is ok. I am worried that `hdulist[0].header.get("LITE")` will return something weird and this will break if the keyword doesn't exist?

@vbkostov can you show that if you run this on a fits file that doesn't have that keyword in the header that this statement works?

christinahedges (changes requested):

christinahedges:
This PR has very similar functionality to #1220, we should touch base and make sure that we get the best out of both PRs, and work with @orionlee and @barentsen!

src/lightkurve/io/detect.py
orionlee: Is the reader intended to work with GSFC-ELEANOR-LITE exclusively, or any FITS file generated from vanilla `eleanor`?

Based on what I've scanned the codes in the vanilla `eleanor`, it appears that the following headers are generated by the custom version of GSFC-ELEANOR-LITE:
```
AUTHOR = 'Adina D. Feinstein'
TVERSION= '0.0.1.dev0'
GITHUB = 'https://github.com/afeinstein20/eleanor'
```

orionlee:
Misc. feedback:
- `FLUX_ERR` Column in Eleanor FITS file is to the error of `RAW_FLUX` (cf. [Eleanor doc](https://adina.feinste.in/eleanor/api.html#eleanor.TargetData.flux_err)). Having it associated with `CORR_FLUX` (the default flux chosen) is probably misleading.
- the units of various columns are missing. (Eleanor FITS file has no units annotated)
- The reader tries to map `FFIINDEX` to cadence number `cadenceno_column="FFIINDEX"`. But it does not appear to work.
- Similarly, the  parameters `centroid_col_column="X_CENTROID"` and `centroid_row_column="Y_CENTROID",` do not appear to have any effect. 
  - Note: Eleanor's X_CENTROID / Y_CENTROID use a different reference point, compared with the centroids in SPOC lightcurves (cf. [eleanor doc](https://adina.feinste.in/eleanor/api.html#eleanor.TargetData.centroid_xs)). Using them as `centroid_col` / `centroid_row` might cause some confusion.
- Eleanor FITS file does not have quality. The parameter `quality_column="QUALITY"` in reader result in a quality column with all zeros.
- Some fields in `meta` typically present in SPOC lightcurves are missing. I suggest we add them for compatibility reasons:
```python
    tic = lc.meta.get("TIC_ID")
    if tic is not None:
        # compatibility with SPOC, QLP, etc.
        lc.meta["TARGETID"] = tic
        lc.meta["TICID"] = tic
        lc.meta["OBJECT"] = f"TIC {tic}"
        # for Lightkurve's plotting methods
        lc.meta["LABEL"] = f"TIC {tic}"
```

tests/io/test_eleanorlite.py
```diff
@@ -0,0 +1,33 @@
...
+from lightkurve.io.eleanorlite import read_eleanorlite_lightcurve
+from lightkurve.io.detect import detect_filetype
+
+
+@pytest.mark.remote_data
+def test_eleanor_lite():
+    """Can we read in GSFC-ELEANOR-LITE light curves?"""
+    url = "https://mast.stsci.edu/api/v0.1/Download/file?uri=mast:HLSP/gsfc-eleanor-lite/s0001/0000/0001/4006/8425/hlsp_gsfc-eleanor-lite_tess_ffi_s0001-0000000140068425_tess_v1.0_lc.fits"    
+    with fits.open(url, mode="readonly") as hdulist:
+        # Can we auto-detect a QLP file?
```
orionlee, line 17: Nitpick: the comment should be updated from QLP to GSFC-ELEANOR-LITE

src/lightkurve/io/detect.py
vbkostov: Yes, it is exclusively intended to work with GSFC-ELEANOR-LITE.

tests/io/test_eleanorlite.py
vbkostov: Thank you, it's fixed now.

src/lightkurve/io/detect.py
vbkostov: > Will this work if this keyword doesn't exist? Maybe @barentsen can weigh in on whether this is the right way to do things, or if this is not good practice?

If there is no keyword, detect_filetype returns “None”

vbkostov:
> * `FLUX_ERR` column in Eleanor FITS file is to the error of `RAW_FLUX`

Thanks, I added a brief description to eleanorlite.py

vbkostov:
>

vbkostov:
> Misc. feedback:
> 
>     * `FLUX_ERR` column in Eleanor FITS file is to the error of `RAW_FLUX` (cf. [Eleanor doc](https://adina.feinste.in/eleanor/api.html#eleanor.TargetData.flux_err)). Having it associated with `CORR_FLUX` (the default flux chosen) is probably misleading.
> 
>     * the units of various columns are missing. (Eleanor FITS file has no units annotated)
> 
>     * The reader tries to map `FFIINDEX` to cadence number `cadenceno_column="FFIINDEX"`. But it does not appear to work.
> 
>     * Similarly, the  parameters `centroid_col_column="X_CENTROID"` and `centroid_row_column="Y_CENTROID",` do not appear to have any effect.
>       
>       * Note: Eleanor's X_CENTROID / Y_CENTROID use a different reference point, compared with the centroids in SPOC lightcurves (cf. [eleanor doc](https://adina.feinste.in/eleanor/api.html#eleanor.TargetData.centroid_xs)). Using them as `centroid_col` / `centroid_row` might cause some confusion.
> 
>     * Eleanor FITS file does not have quality. The parameter `quality_column="QUALITY"` in reader result in a quality column with all zeros.
> 
>     * Some fields in `meta` typically present in SPOC lightcurves are missing. I suggest we add them for compatibility reasons:
> 
> 
> ```python
>     tic = lc.meta.get("TIC_ID")
>     if tic is not None:
>         # compatibility with SPOC, QLP, etc.
>         lc.meta["TARGETID"] = tic
>         lc.meta["TICID"] = tic
>         lc.meta["OBJECT"] = f"TIC {tic}"
>         # for Lightkurve's plotting methods
>         lc.meta["LABEL"] = f"TIC {tic}"
> ```
Screen Shot 2022-06-15 at 5 31 11 PM

Can you please let me know the TIC and sector you are looking into? The quality flags, x- and y-centroids, and ffiindex work for me:
Screen Shot 2022-06-15 at 5 31 11 PM

src/lightkurve/io/detect.py
vbkostov: > Is the reader intended to work with GSFC-ELEANOR-LITE exclusively, or any FITS file generated from vanilla `eleanor`?
> 
> Based on what I've scanned the codes in the vanilla `eleanor`, it appears that the following headers are generated by the custom version of GSFC-ELEANOR-LITE:
> 
> ```
> AUTHOR = 'Adina D. Feinstein'
> TVERSION= '0.0.1.dev0'
> GITHUB = 'https://github.com/afeinstein20/eleanor'
> ```

Yes, it is exclusively intended to work with GSFC-ELEANOR-LITE.

vbkostov:
> Misc. feedback:
> 
>     * `FLUX_ERR` column in Eleanor FITS file is to the error of `RAW_FLUX` (cf. [Eleanor doc](https://adina.feinste.in/eleanor/api.html#eleanor.TargetData.flux_err)). Having it associated with `CORR_FLUX` (the default flux chosen) is probably misleading.
> 
>     * the units of various columns are missing. (Eleanor FITS file has no units annotated)
> 
>     * The reader tries to map `FFIINDEX` to cadence number `cadenceno_column="FFIINDEX"`. But it does not appear to work.
> 
>     * Similarly, the  parameters `centroid_col_column="X_CENTROID"` and `centroid_row_column="Y_CENTROID",` do not appear to have any effect.
>       
>       * Note: Eleanor's X_CENTROID / Y_CENTROID use a different reference point, compared with the centroids in SPOC lightcurves (cf. [eleanor doc](https://adina.feinste.in/eleanor/api.html#eleanor.TargetData.centroid_xs)). Using them as `centroid_col` / `centroid_row` might cause some confusion.
> 
>     * Eleanor FITS file does not have quality. The parameter `quality_column="QUALITY"` in reader result in a quality column with all zeros.
> 
>     * Some fields in `meta` typically present in SPOC lightcurves are missing. I suggest we add them for compatibility reasons:
> 
> 
> ```python
>     tic = lc.meta.get("TIC_ID")
>     if tic is not None:
>         # compatibility with SPOC, QLP, etc.
>         lc.meta["TARGETID"] = tic
>         lc.meta["TICID"] = tic
>         lc.meta["OBJECT"] = f"TIC {tic}"
>         # for Lightkurve's plotting methods
>         lc.meta["LABEL"] = f"TIC {tic}"
> ```

Thanks for the feedback!

orionlee:
vbkostov:

> I added a comment to eleanorlite.py to explicitly state that that FLUX_ERR is for RAW_FLUX.

I feel this could still be misleading, given it doesn't exactly describe the error of CORR_FLUX. But I think we should hear the feedback from @christinahedges , @barentsen, etc.

Reason: `Lightkurve` follows the convention (and used in various parts of the codes) that the error of a given column is in `<column-name>_err` . As it is , when codes process `lc.flux` (which is CORR_FLUX), it will treat `lc.flux_err` as the associated error.


> Can you please let me know the TIC and sector you are looking into? The quality flags, x- and y-centroids, and ffiindex work for me:

1. quality: you're correct. I'm mistaken to believe quality column is missing. 
However, it raises another issue. For GSFC-ELEANOR-LITE, is the semantics of the quality flag the same as the standard SPOC? Because if it is not, then the filtering with `quality_bitmask` during `download()` would produce erroneous result.

2. There is a misunderstanding on the issue regarding x- and y-centroids, and ffiindex. By convention, the the column names for them are `centroid_col`, `centroid_row`, `cadenceno` respectively. In the PR, they only exist in their eleanor name. 
To have them appear as `centroid_col`, `centroid_row`, `cadenceno` columns as well, you'd need to pass the actual column names to `read_generic_lightcurve()` call.

src/lightkurve/io/detect.py
orionlee: If the reader is to work with GSFC-ELEANOR-LITE exclusively, detection via `LITE = T` would not work, because vanilla `eleanor` also produces `LITE=T` header if the users chooses to do so in  [`TargetData.save()`](https://adina.feinste.in/eleanor/api.html#eleanor.TargetData.save)

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,74 @@
+"""Reader for GSFC-ELEANOR-LITE light curve files. 
+Details can be found at https://archive.stsci.edu/hlsp/eleanor
```
orionlee, line 2: The URL of GSFC-ELEANOR-LITE is actually:
https://archive.stsci.edu/hlsp/gsfc-eleanor-lite

src/lightkurve/io/detect.py
vbkostov: Good point. @christinahedges, @barentsen can you weight it on this? Should we use "AUTHOR" and/or "TVERSION" to distinguish between eleanor and elenor-lite?

src/lightkurve/io/eleanorlite.py
vbkostov: Thanks, I listed both links just in case

vbkostov:
> vbkostov:
> 
> > I added a comment to eleanorlite.py to explicitly state that that FLUX_ERR is for RAW_FLUX.
> 
> I feel this could still be misleading, given it doesn't exactly describe the error of CORR_FLUX. But I think we should hear the feedback from @christinahedges , @barentsen, etc.
> 
> Reason: `Lightkurve` follows the convention (and used in various parts of the codes) that the error of a given column is in `<column-name>_err` . As it is , when codes process `lc.flux` (which is CORR_FLUX), it will treat `lc.flux_err` as the associated error.
> 
> > Can you please let me know the TIC and sector you are looking into? The quality flags, x- and y-centroids, and ffiindex work for me:
> 
>     1. quality: you're correct. I'm mistaken to believe quality column is missing.
>        However, it raises another issue. For GSFC-ELEANOR-LITE, is the semantics of the quality flag the same as the standard SPOC? Because if it is not, then the filtering with `quality_bitmask` during `download()` would produce erroneous result.

Agreed, I set the quality bitmask to None 

> 
>     2. There is a misunderstanding on the issue regarding x- and y-centroids, and ffiindex. By convention, the the column names for them are `centroid_col`, `centroid_row`, `cadenceno` respectively. In the PR, they only exist in their eleanor name.
>        To have them appear as `centroid_col`, `centroid_row`, `cadenceno` columns as well, you'd need to pass the actual column names to `read_generic_lightcurve()` call.

Ah, yes, thank you. I updated the reader

christinahedges:
> the units of various columns are missing. (Eleanor FITS file has no units annotated)

@vbkostov I don't see an update to the code that includes the units. Maybe you have implemented this and just need to push. You can use @orionlee's excellent solution on #1220 as a template to see how to propagate these errors.

I also see there are other comments that aren't addressed but you've discussed above (e.g. metadata for `TICID` etc) so maybe you just need to push your updates.

orionlee:
Regarding @christinahedges 's suggestion
1. keywords to use for detection: that seems to work
2. flux_err: agreed, though I think it needs to be documented stating it is a derived value and the derivation.
3. quality: The suggestion of mapping `eleanor` specific 4096 to 128 sounds sensible. One thing I wonder is whether 4096 appears in practice for GSFC-ELEANOR-LITE. E.g., does GSFC-ELEANOR-LITE exclude all such cadences in the final output?


@ethankruse from GSFC-ELEANOR-LITE did open an issue regarding GSFC-ELEANOR-LITE. Maybe he can provide some feedback too?

ethankruse:
Ok so chiming in. In the headers, there is a 'lite' flag. hdr['lite] == True when you saved as a GSFC-lite file format and hdr['lite'] == False when it's in the standard full eleanor file. You can use that to easily distinguish.

Yes, I believe the flux errors are as described though I'll go check.

As for quality flags, there are actually 2 eleanor ones. Ben added a second eleanor quality flag. And it's no longer 4096. SPOC added several quality flags (sometime around S19-22 I think?) so eleanor had to shift its own quality flags up. 

I believe they're bits 18/19 (or maybe 17/18) depending on 0-index or not. I can go dig in and double check.

However, I strongly advise against putting them as manual exclude. @christinahedges you've been in those Zoom calls. eleanor is rather aggressive in setting those flags and it's often best to ignore at least one and sometimes both of them. And definitely important to know what are the actual SPOC manual excludes. What breaks if you keep them as is? How can I help?

vbkostov:
@christinahedges, @orionlee , @ethankruse thank you for the fantastic discussion and suggestions!

I updated the PR to:

1) Use hdr['lite] == True to unique specify GSFC-ELEANOR-LITE
2) Use @orionlee implementation to include the units
3) Get flux_err from corr_flux*flux_err/raw_flux

What would be a good solution for the quality flags?

src/lightkurve/io/detect.py
ethankruse: Hang on, I'm just catching up on all this, but is there any reason not to let this generically read in any eleanor light curves? Right now lk can't handle any eleanor light curves. Why not allow any, whether user-generated or from GSFC? As far as I know, none of the differences between lite and regular will have any effect on the rest of the changes.

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,77 @@
...
+    for colname in ["barycorr"]:
+        if colname in lc.colnames:
+            lc[colname].unit = u.day
+
+    # In Eleanor fits file, raw_flux's error is in flux_err, which breaks Lightkurve convention.
+    # To account for this, the corr_flux error is calculated from corr_flux_err = corr_flux*raw_flux_err/raw_flux. For completeness, 
+    # the original raw_flux's error is added as a "raw_flux_err" column
+    lc["raw_flux_err"] = lc["flux_err"]
+    if flux_column.lower() != 'raw_flux':
+        lc["flux_err"] = lc["corr_flux"]*lc["flux_err"]/lc["raw_flux"]
```
ethankruse, line 69: This isn't generic enough. I believe you can just use
`lc["flux_err"] = lc[flux_column.lower()]*lc["flux_err"]/lc["raw_flux"]`

vbkostov: Thanks, done!

ethankruse (approved):

orionlee:
1. quality flag: given what @ethankruse stated that eleanor shifted the bits, so that eleanor-specific ones no longer clash with SPOC quality flag. I suppose we can leave it as is. Users can apply  additional eleanor-specific additional exclusion.

2. On detection using hdr['lite] == True : FITS file generated from vanilla `eleanor` [`(TargetData.save(lite=True)`](https://github.com/afeinstein20/eleanor/blob/cf19a998a99d0cea09c8426f8db2248fdc55f12c/eleanor/targetdata.py#L1314) will also be accepted. It probably doesn't hurt accepting vanilla eleanor FITS file.

3. Unit handling. I noticed the majority has been added. I list the remaining issues at the end.

4. header / meta:  the following keywords should be added for compatibility reasons
```python
        lc.meta["TICID"] = tic
        lc.meta["OBJECT"] = f"TIC {tic}"
        # for Lightkurve's plotting methods
        lc.meta["LABEL"] = f"TIC {tic}"
```

---

Units handlings:
- flux_bkg: should be percent `u.percent`
- centroid_col, centroid_row  (and the original x_centroid, y_centroid): should be pixel `u.pix` . I think the similar eleanor-specific `x_com` and `y_com` is also pixel too.

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,77 @@
...
+    for colname in ["barycorr"]:
+        if colname in lc.colnames:
+            lc[colname].unit = u.day
+
+    # In Eleanor fits file, raw_flux's error is in flux_err, which breaks Lightkurve convention.
+    # To account for this, the corr_flux error is calculated from corr_flux_err = corr_flux*raw_flux_err/raw_flux. For completeness, 
+    # the original raw_flux's error is added as a "raw_flux_err" column
+    lc["raw_flux_err"] = lc["flux_err"]
+    if flux_column.lower() != 'raw_flux':
+        lc["flux_err"] = lc[flux_column.lower()]*lc["flux_err"]/lc["raw_flux"]
```
christinahedges, line 69: Do you mean to overwrite this in this way?
```suggestion
        lc["flux_err"] = lc[flux_column.lower()]*lc["raw_flux_err"]/lc["raw_flux"]
```

These few lines seem a bit confusing.

christinahedges:
@ethankruse thanks so much for weighing in, I think I'm happy with adding more quality flags then since you point out they are very aggressive. @barentsen discusses [here](https://github.com/lightkurve/lightkurve/pull/1218#issuecomment-1159165912) that we might make each pipeline's reader be able to deal with each pipeline's quality flags. This is definitely an interesting idea! It would be a future update though. For now, let's go ahead and use the quality flags as they come, and add bits 17/18.

### Can we make this a generic `ELEANOR` reader?
At the moment this is one team's light curves using `ELEANOR`, but other teams may put together light curves in the future. I can envisage a future where there are multiple light curve products, and we want them to be separated by "author" during search, to make sure 1) not to confuse users 2) to give light curve product authors credit. So for now, I believe we're better off making this a reader for specifically this product.

We can definitely keep an eye out for this as more products come using the `ELEANOR` pipeline.

orionlee:
Quality flag effect. I use pi Men  sector 1 as an example.

From GSFC-ELEANOR-LITE:
image

7 sigma clipped:
image


Cadences with Eleanor-specific bits marked:
image

---
Counterpart from SPOC 30-minute cadence (PDCSAP_FLUX and SAP_FLUX):

image

src/lightkurve/io/eleanorlite.py
vbkostov: > Do you mean to overwrite this in this way?
> 
> These few lines seem a bit confusing.

Yes, thank you, I updated the line accordingly

vbkostov:
> 1. quality flag: given what @ethankruse stated that eleanor shifted the bits, so that eleanor-specific ones no longer clash with SPOC quality flag. I suppose we can leave it as is. Users can apply  additional eleanor-specific additional exclusion.
> 
>     2. On detection using hdr['lite] == True : FITS file generated from vanilla `eleanor` [`(TargetData.save(lite=True)`](https://github.com/afeinstein20/eleanor/blob/cf19a998a99d0cea09c8426f8db2248fdc55f12c/eleanor/targetdata.py#L1314) will also be accepted. It probably doesn't hurt accepting vanilla eleanor FITS file.
> 
>     3. Units handling: I noticed the majority has been added. I list the remaining issues at the end.
> 
>     4. header / meta:  the following keywords should be added for compatibility reasons
> 
> 
> ```python
>         lc.meta["TICID"] = tic
>         lc.meta["OBJECT"] = f"TIC {tic}"
>         # for Lightkurve's plotting methods
>         lc.meta["LABEL"] = f"TIC {tic}"
> ```
> 
> Units handling:
> 
>     * flux_bkg: should be percent `u.percent`
> 
>     * centroid_col, centroid_row  (and the original x_centroid, y_centroid): should be pixel `u.pix` . I think the similar eleanor-specific `x_com` and `y_com` columns are pixel too.

Thanks again. I added the units and the header/meta.

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,94 @@
...
+            else:
+                lc[colname].unit = "electron/s"
+
+    for colname in ["flux_bkg"]:
+        if colname in lc.colnames:
+            lc[colname].unit = u.percent
+
+    for colname in ["centroid_col", "centroid_row", "x_centroid", "y_centroid", "x_com", "y_com"]:
+        if colname in lc.colnames:
+            lc[colname].unit = "pixels"
```
orionlee, line 66: "pixels" is not a recognized unit unfortunately. Use `u.pix` instead.

orionlee (changes requested):

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,94 @@
...
+    centroid_col_column="X_CENTROID",
+    centroid_row_column="Y_CENTROID",
+    quality_column="QUALITY",
+    quality_bitmask=None, 
+    **kwargs):
+    """Returns a `TessLightCurve` object given a light curve file from the GSFC Eleanor-lite Pipeline.
+
+    By default, eleanor's `CORR_FLUX` column is used to populate the `flux` values,
+    and 'FLUX_ERR' is used to populate `flux_err`. Note that the "FLUX_ERR" column in the 
+    Eleanor FITS file is referred to the uncertainty of "RAW_FLUX", not "CORR_FLUX"
```
orionlee, line 24: The comment about `flux_err` is no longer accurate after the changes.

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,94 @@
...
+    Parameters
+    ----------
+    filename : str
+        Local path or remote url of a QLP light curve FITS file.
+    flux_column : 'RAW_FLUX', 'CORR_FLUX', 'PCA_FLUX', or 'FLUX_BKG'
+        Which column in the FITS file contains the preferred flux data?
+        By default the "Corrected Flux" flux (CORR_FLUX) is used.
+    flux_err_column: 'FLUX_ERR'
+      Which column in the FITS file contains the preferred flux_err data?
+      Note that the "FLUX_ERR" column in the Eleanor FITS file is referred to the uncertainty of "RAW_FLUX", not "CORR_FLUX"
```
orionlee, line 35: similarly, it is no longer accruate.

orionlee (changes requested):

src/lightkurve/io/eleanorlite.py
vbkostov: Thank you, it's updated now

src/lightkurve/io/eleanorlite.py
vbkostov: Good catch, thanks! It's fixed

src/lightkurve/io/eleanorlite.py
vbkostov: Fixed

orionlee (approved):

orionlee:
I think the PR is largely ready to be merged. The 2 remaining issues are:
1. public documentation
2. whether to support vanilla `eleanor` (related to detection logic)

--- 

#### public documentation
At the minimal, we should point out that the default `err_flux` does not come from the product, but is a derived approximation.

One issue is there is no product-specific information in public documentation currently. Some options are:
- add a product subsection in https://docs.lightkurve.org/reference/lightcurve.html 
  - In this case, we need to add and maintain the text in the correspond `lightcurve.rst`
- expose various reader functions in the API doc. 
  - While the readers are not intended to be used directly, product-specific information is naturally there.

#### whether to support vanilla `eleanor` 

I suggest the following:
1. we support reading vanilla `eleanor` (the reader is already doing that). `AUTHOR` header is set to `ELEANOR` .
2. For GSFC-ELEANOR-LITE lightcurve,  `AUTHOR` header is set to `GSFC-ELEANOR-LITE`.

Effectively, users can read arbitrary `eleanor` files, but they can still easily identify GSFC-ELEANOR-LITE lightcurves via `AUTHOR` header.

I did a quick check locally and confirmed the reader codes can handle vanilla `eleanor` files, both the lite version and the full version (which has extra information). The only change is the detection logic, e.g., `hdulist[0].header.get("LITE") is not None`  (`eleanor` outputs `LITE` keyword in both full and lite version).

---

<details>
<summary>Sample vanilla eleanor header (the full version)</summary>
<pre>
SIMPLE  =                    T / conforms to FITS standard                      
BITPIX  =                    8 / array data type                                
NAXIS   =                    0 / number of array dimensions                     
EXTEND  =                    T                                                  
WCAX4   =                    2 / number of WCS axes                             
1CTYP4  = 'RA---TAN'           / right ascension coordinate type                
2CTYP4  = 'DEC--TAN'           / declination coordinate type                    
1CRPX4  =      15.961971696515 / [pixel] reference pixel along image axis 1     
2CRPX4  =       15.75568746722 / [pixel] reference pixel along image axis 2     
1CRVL4  =      84.258970696725 / [deg] right ascension at reference pixel       
2CRVL4  =   -80.47178663280999 / [deg] declination at reference pixel           
1CUNI4  = 'deg     '           / physical unit in column dimension              
2CUNI4  = 'deg     '           / physical unit in row dimension                 
1CDLT4  =                  1.0 / [deg] pixel scale in RA dimension              
2CDLT4  =                  1.0 / [deg] pixel scale in DEC dimension             
11PC4   =     0.00355196986142 / Coordinate transformation matrix element       
12PC4   =   0.0043416588633886 / Coordinate transformation matrix element       
21PC4   =  -0.0041451948610946 / Coordinate transformation matrix element       
22PC4   =   0.0032446836544245 / Coordinate transformation matrix element       
WCSN4P  = 'PHYSICAL'           / table column WCS name                          
WCAX4P  =                    2 / table column physical WCS dimensions           
1CTY4P  = 'RAWX    '           / table column physical WCS axis 1 type, CCD col 
2CTY4P  = 'RAWY    '           / table column physical WCS axis 2 type, CCD row 
1CUN4P  = 'PIXEL   '           / table column physical WCS axis 1 unit          
2CUN4P  = 'PIXEL   '           / table column physical WCS axis 2 unit          
1CRV4P  =                 1616 / table column physical WCS ax 1 ref value       
2CRV4P  =                  244 / table column physical WCS ax 2 ref value       
1CDL4P  =                  1.0 / table column physical WCS a1 step              
2CDL4P  =                  1.0 / table column physical WCS a2 step              
1CRP4P  =                    1 / table column physical WCS a1 reference         
2CRP4P  =                    1 / table column physical WCS a2 reference         
WCAX5   =                    2 / number of WCS axes                             
1CTYP5  = 'RA---TAN'           / right ascension coordinate type                
2CTYP5  = 'DEC--TAN'           / declination coordinate type                    
1CRPX5  =      15.961971696515 / [pixel] reference pixel along image axis 1     
2CRPX5  =       15.75568746722 / [pixel] reference pixel along image axis 2     
1CRVL5  =      84.258970696725 / [deg] right ascension at reference pixel       
2CRVL5  =   -80.47178663280999 / [deg] declination at reference pixel           
1CUNI5  = 'deg     '           / physical unit in column dimension              
2CUNI5  = 'deg     '           / physical unit in row dimension                 
1CDLT5  =                  1.0 / [deg] pixel scale in RA dimension              
2CDLT5  =                  1.0 / [deg] pixel scale in DEC dimension             
11PC5   =     0.00355196986142 / Coordinate transformation matrix element       
12PC5   =   0.0043416588633886 / Coordinate transformation matrix element       
21PC5   =  -0.0041451948610946 / Coordinate transformation matrix element       
22PC5   =   0.0032446836544245 / Coordinate transformation matrix element       
WCSN5P  = 'PHYSICAL'           / table column WCS name                          
WCAX5P  =                    2 / table column physical WCS dimensions           
1CTY5P  = 'RAWX    '           / table column physical WCS axis 1 type, CCD col 
2CTY5P  = 'RAWY    '           / table column physical WCS axis 2 type, CCD row 
1CUN5P  = 'PIXEL   '           / table column physical WCS axis 1 unit          
2CUN5P  = 'PIXEL   '           / table column physical WCS axis 2 unit          
1CRV5P  =                 1616 / table column physical WCS ax 1 ref value       
2CRV5P  =                  244 / table column physical WCS ax 2 ref value       
1CDL5P  =                  1.0 / table column physical WCS a1 step              
2CDL5P  =                  1.0 / table column physical WCS a2 step              
1CRP5P  =                    1 / table column physical WCS a1 reference         
2CRP5P  =                    1 / table column physical WCS a2 reference         
WCAX6   =                    2 / number of WCS axes                             
1CTYP6  = 'RA---TAN'           / right ascension coordinate type                
2CTYP6  = 'DEC--TAN'           / declination coordinate type                    
1CRPX6  =      15.961971696515 / [pixel] reference pixel along image axis 1     
2CRPX6  =       15.75568746722 / [pixel] reference pixel along image axis 2     
1CRVL6  =      84.258970696725 / [deg] right ascension at reference pixel       
2CRVL6  =   -80.47178663280999 / [deg] declination at reference pixel           
1CUNI6  = 'deg     '           / physical unit in column dimension              
2CUNI6  = 'deg     '           / physical unit in row dimension                 
1CDLT6  =                  1.0 / [deg] pixel scale in RA dimension              
2CDLT6  =                  1.0 / [deg] pixel scale in DEC dimension             
11PC6   =     0.00355196986142 / Coordinate transformation matrix element       
12PC6   =   0.0043416588633886 / Coordinate transformation matrix element       
21PC6   =  -0.0041451948610946 / Coordinate transformation matrix element       
22PC6   =   0.0032446836544245 / Coordinate transformation matrix element       
WCSN6P  = 'PHYSICAL'           / table column WCS name                          
WCAX6P  =                    2 / table column physical WCS dimensions           
1CTY6P  = 'RAWX    '           / table column physical WCS axis 1 type, CCD col 
2CTY6P  = 'RAWY    '           / table column physical WCS axis 2 type, CCD row 
1CUN6P  = 'PIXEL   '           / table column physical WCS axis 1 unit          
2CUN6P  = 'PIXEL   '           / table column physical WCS axis 2 unit          
1CRV6P  =                 1616 / table column physical WCS ax 1 ref value       
2CRV6P  =                  244 / table column physical WCS ax 2 ref value       
1CDL6P  =                  1.0 / table column physical WCS a1 step              
2CDL6P  =                  1.0 / table column physical WCS a2 step              
1CRP6P  =                    1 / table column physical WCS a1 reference         
2CRP6P  =                    1 / table column physical WCS a2 reference         
WCAX7   =                    2 / number of WCS axes                             
1CTYP7  = 'RA---TAN'           / right ascension coordinate type                
2CTYP7  = 'DEC--TAN'           / declination coordinate type                    
1CRPX7  =      15.961971696515 / [pixel] reference pixel along image axis 1     
2CRPX7  =       15.75568746722 / [pixel] reference pixel along image axis 2     
1CRVL7  =      84.258970696725 / [deg] right ascension at reference pixel       
2CRVL7  =   -80.47178663280999 / [deg] declination at reference pixel           
1CUNI7  = 'deg     '           / physical unit in column dimension              
2CUNI7  = 'deg     '           / physical unit in row dimension                 
1CDLT7  =                  1.0 / [deg] pixel scale in RA dimension              
2CDLT7  =                  1.0 / [deg] pixel scale in DEC dimension             
11PC7   =     0.00355196986142 / Coordinate transformation matrix element       
12PC7   =   0.0043416588633886 / Coordinate transformation matrix element       
21PC7   =  -0.0041451948610946 / Coordinate transformation matrix element       
22PC7   =   0.0032446836544245 / Coordinate transformation matrix element       
WCSN7P  = 'PHYSICAL'           / table column WCS name                          
WCAX7P  =                    2 / table column physical WCS dimensions           
1CTY7P  = 'RAWX    '           / table column physical WCS axis 1 type, CCD col 
2CTY7P  = 'RAWY    '           / table column physical WCS axis 2 type, CCD row 
1CUN7P  = 'PIXEL   '           / table column physical WCS axis 1 unit          
2CUN7P  = 'PIXEL   '           / table column physical WCS axis 2 unit          
1CRV7P  =                 1616 / table column physical WCS ax 1 ref value       
2CRV7P  =                  244 / table column physical WCS ax 2 ref value       
1CDL7P  =                  1.0 / table column physical WCS a1 step              
2CDL7P  =                  1.0 / table column physical WCS a2 step              
1CRP7P  =                    1 / table column physical WCS a1 reference         
2CRP7P  =                    1 / table column physical WCS a2 reference         
WCAX8   =                    2 / number of WCS axes                             
1CTYP8  = 'RA---TAN'           / right ascension coordinate type                
2CTYP8  = 'DEC--TAN'           / declination coordinate type                    
1CRPX8  =      15.961971696515 / [pixel] reference pixel along image axis 1     
2CRPX8  =       15.75568746722 / [pixel] reference pixel along image axis 2     
1CRVL8  =      84.258970696725 / [deg] right ascension at reference pixel       
2CRVL8  =   -80.47178663280999 / [deg] declination at reference pixel           
1CUNI8  = 'deg     '           / physical unit in column dimension              
2CUNI8  = 'deg     '           / physical unit in row dimension                 
1CDLT8  =                  1.0 / [deg] pixel scale in RA dimension              
2CDLT8  =                  1.0 / [deg] pixel scale in DEC dimension             
11PC8   =     0.00355196986142 / Coordinate transformation matrix element       
12PC8   =   0.0043416588633886 / Coordinate transformation matrix element       
21PC8   =  -0.0041451948610946 / Coordinate transformation matrix element       
22PC8   =   0.0032446836544245 / Coordinate transformation matrix element       
WCSN8P  = 'PHYSICAL'           / table column WCS name                          
WCAX8P  =                    2 / table column physical WCS dimensions           
1CTY8P  = 'RAWX    '           / table column physical WCS axis 1 type, CCD col 
2CTY8P  = 'RAWY    '           / table column physical WCS axis 2 type, CCD row 
1CUN8P  = 'PIXEL   '           / table column physical WCS axis 1 unit          
2CUN8P  = 'PIXEL   '           / table column physical WCS axis 2 unit          
1CRV8P  =                 1616 / table column physical WCS ax 1 ref value       
2CRV8P  =                  244 / table column physical WCS ax 2 ref value       
1CDL8P  =                  1.0 / table column physical WCS a1 step              
2CDL8P  =                  1.0 / table column physical WCS a2 step              
1CRP8P  =                    1 / table column physical WCS a1 reference         
2CRP8P  =                    1 / table column physical WCS a2 reference         
EXTNAME = 'PIXELS  '                                                            
INHERIT =                    T                                                  
BACKAPP =                  0.0 / background is subtracted                       
CDPP0_5 =  / RMS CDPP on 0.5-hr time scales                                     
CDPP1_0 =  / RMS CDPP on 1.0-hr time scales                                     
CDPP2_0 =  / RMS CDPP on 2.0-hr time scales                                     
CROWDSAP=  / Ratio of target flux to total flux in op. ap.                      
DEADAPP =                  1.0 / deadtime applied                               
DEADC   =                0.792 / deadtime correction                            
EXPOSURE=       0.016499993395 / [d] time on source                             
FLFRCSAP=  / Frac. of target flux w/in the op. aperture                         
FRAMETIM=                  2.0 / [s] frame time [INT_TIME + READTIME]           
FXDOFF  =              3355400 / compression fixed offset                       
GAINA   =                 5.25 / [electrons/count] CCD output A gain            
GAINB   =    5.170000076293945 / [electrons/count] CCD output B gain            
GAINC   =    5.119999885559082 / [electrons/count] CCD output C gain            
GAIND   =    5.199999809265137 / [electrons/count] CCD output D gain            
INT_TIME=                 1.98 / [s] photon accumulation time per frame         
LIVETIME=   0.0164999933953277 / [d] TELAPSE multiplied by DEADC                
MEANBLCA=                 6574 / [count] FSW mean black level CCD output A      
MEANBLCB=                 6574 / [count] FSW mean black level CCD output B      
MEANBLCC=                 6573 / [count] FSW mean black level CCD output C      
MEANBLCD=                 6278 / [count] FSW mean black level CCD output D      
NREADOUT=                  720 / number of read per cadence                     
NUM_FRM =                  900 / number of frames per time stamp                
READNOIA=    10.23750019073486 / [electrons] read noise CCD output A            
READNOIB=     7.23799991607666 / [electrons] read noise CCD output B            
READNOIC=    7.270399570465088 / [electrons] read noise CCD output C            
READNOID=    9.515999794006348 / [electrons] read noise CCD output D            
READTIME=                 0.02 / [s] readout time per frame                     
TIERRELA=             1.16E-05 / [d] relative time error                        
TIMEDEL =  0.02083333333333333 / [d] time resolution of data                    
TIMEPIXR=                  0.5 / bin time beginning=0 middle=0.5 end=1          
TMOFST11=  / (s) readout delay for camera 1 and ccd 1                           
VIGNAPP =                  1.0 / vignetting or collimator correction applied    
WCS_FFI = 'tess2018221072942-s0001-4-2-0120-s_ffic.fits' / FFI used for cutout W
EXTVER  =                    1 / extension version number (not format version)  
SIMDATA =                    F / file is based on simulated data                
ORIGIN  = 'STScI/MAST'         / institution responsible for creating this file 
DATE    = '2021-05-11'         / file creation date.                            
TSTART  =     1325.31384463669 / observation start time in BTJD                 
TSTOP   =     1353.16770459837 / observation stop time in BTJD                  
DATE-OBS= '2018-07-25T19:30:46.993' / TSTART as UTC calendar date               
DATE-END= '2018-08-22T16:00:20.493' / TSTOP as UTC calendar date                
CREATOR = 'astrocut'           / software used to produce this file             
PROCVER = '0.8     '           / software version                               
FILEVER = '1.0     '           / file format version                            
TIMVERSN= 'OGIP/93-003'        / OGIP memo number for file format               
TELESCOP= 'TESS    '           / telescope                                      
INSTRUME= 'TESS Photometer'    / detector type                                  
DATA_REL=                   42 / data release version number                    
ASTATE  =                    T / archive state F indicates single orbit processi
SCCONFIG=                  120 / spacecraft configuration ID                    
RADESYS = 'ICRS    '           / reference frame of celestial coordinates       
FFIINDEX=                 4697 / number of FFI cadence interval                 
EQUINOX =               2000.0 / equinox of celestial coordinate system         
CRMITEN =                    T / spacecraft cosmic ray mitigation enabled       
CRBLKSZ =                   10 / [exposures] s/c cosmic ray mitigation block siz
CRSPOC  =                    F / SPOC cosmic ray cleaning enabled               
CAMERA  =                    4 / Camera number                                  
CCD     =                    2 / CCD chip number                                
SECTOR  =                    1 / Observing sector                               
RA_OBJ  =     84.2911879979852 / [deg] right ascension                          
DEC_OBJ =   -80.46911979699409 / [deg] declination                              
TIMEREF = 'SOLARSYSTEM'        / barycentric correction applied to times        
TASSIGN = 'SPACECRAFT'         / where time is assigned                         
TIMESYS = 'TDB     '           / time system is Barycentric Dynamical Time (TDB)
BJDREFI =              2457000 / integer part of BTJD reference date            
BJDREFF =                  0.0 / fraction of the day in BTJD reference date     
TIMEUNIT= 'd       '           / time unit for TIME, TSTART and TSTOP           
TELAPSE =    27.85385996167997 / [d] TSTOP - TSTART                             
OBJECT  = '' / string version of target id                                      
TCID    =                    0 / unique tess target identifier                  
PXTABLE =                    0 / pixel table id                                 
PMRA    =                  0.0 / [mas/yr] RA proper motion                      
PMDEC   =                  0.0 / [mas/yr] Dec proper motion                     
PMTOTAL =                  0.0 / [mas/yr] total proper motion                   
TESSMAG =                  0.0 / [mag] TESS magnitude                           
TEFF    =                  0.0 / [K] Effective temperature                      
LOGG    =                  0.0 / [cm/s2] log10 surface gravity                  
MH      =                  0.0 / [log10([M/H])] metallicity                     
RADIUS  =                  0.0 / [solar radii] stellar radius                   
TICVER  =                    0 / TICVER                                         
TICID   =  / unique tess target identifier                                      
CHECKSUM= 'FULRHUKPFUKPFUKP'   / HDU checksum updated 2022-06-19T01:31:51       
DATASUM = '3252788793'         / data unit checksum updated 2022-06-19T01:31:51 
CREATED = '2022-06-18'                                                          
FILTER  = 'TESS    '           / Filter keyword                                 
VERSION = '2.0.4   '           / eleanor version used for light curve production
LITE    =                    F / eleanor-lite keyword for saving light curves   
TIC_ID  =            261136679 / TESS Input Catalog ID                          
TMAG    =               5.1054 / TESS mag                                       
TIC_V   =             20190415 / TIC Version                                    
GAIA_ID =  4623036865373793408 / Associated Gaia ID                             
SECTOR  =                    1 / Sector                                         
CAMERA  =                    4 / Camera                                         
CCD     =                    2 / CCD                                            
CHIPPOS1=                 1616 / central x pixel of TPF in FFI chip             
CHIPPOS2=                  244 / central y pixel of TPF in FFI                  
POSTCARD= 'tesscut\tess-s0001-4-2_84.291188_-80.469120_31x31_astrocut.fits'     
POSTPOS1=                    0 / predicted x pixel of source on postcard        
POSTPOS2=                    0 / predicted y pixel of source on postcard        
CEN_RA  =     84.2911879979852 / RA of TPF source                               
CEN_DEC =   -80.46911979699409 / DEC of TPF source                              
TPF_H   =                   13 / Height of the TPF in pixels                    
TPF_W   =                   13 / Width of the TPF in pixels                     
TESSCUT =                    T / If TessCut was used to make this TPF           
BKG_LVL = 'PC_LEVEL'           / Stage at which background is subtracted        
PCORIGIN= 'MAST    '           / Provenance of eleanor Postcard                 
APERTURE= '4_circle_exact'     / Best aperture used for lightcurves in extension
</pre>
</details>

orionlee:
@christinahedges @vbkostov @barentsen I suggest we merge this PR as-is. For the [2 outstanding issues](https://github.com/lightkurve/lightkurve/pull/1217#issuecomment-1159607369) on documentation and  vanilla `eleanor` support, we could handle them separately afterwards.

orionlee:
@barentsen 

The `AUTHOR_LINKS` has been in the PR:


https://github.com/lightkurve/lightkurve/blob/7998c0c2697e87840b32754794f5821771efcc5d/src/lightkurve/search.py#L49-L51

---

For public documentation: let's review. The following is the current docstring as-is:

image

barentsen:
> The AUTHOR_LINKS has been in the PR:

Ooops, I was looking at the wrong branch. Please ignore that comment indeed!

vbkostov:
> @vbkostov On public documentation and `quality_bitmask` semantics.
> ### `quality_bitmask`
> 
> we should support `quality_bitmask`, rather than just ignore it, given that eleanor quality bits is basically standard TESS SPOC ones. The implementation in `read_tess_lightcurve()` would be a good starting point.
> 
> For the handling of the 2 additional eleanor-specific bits Ethan mentioned, I suggest we include them in the `"hardest"` case. Based on what Ethan described, that eleanor is rather aggressive in marking cadences to be excluded and probably should not be default, including them in the `"hardest"` would make sense (hardest means any flag set).
> ### public documentation and API tweaks
> 
>     * parameters `time_column='TIME', flux_err_column='FLUX_ERR', cadenceno_column='FFIINDEX', centroid_col_column='X_CENTROID', centroid_row_column='Y_CENTROID', quality_column='QUALITY'` should be hardcoded rather than exposed as parameters, because there are fixed in the underlying eleanor files.
> 
>     * Replace outdated references to QLP in docstring
> 
>     * Include a link to https://archive.stsci.edu/hlsp/gsfc-eleanor-lite in the docstring

Thanks for the suggestions @orionlee, I'll work on the quality flags and the public documentation

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,114 @@
...
+    flux_column="CORR_FLUX", 
+    flux_err_column="FLUX_ERR", 
+    cadenceno_column="FFIINDEX",
+    centroid_col_column="X_CENTROID",
+    centroid_row_column="Y_CENTROID",
+    quality_column="QUALITY",
+    quality_bitmask="default", 
+    **kwargs):
+    """Returns a `TessLightCurve` object given a light curve file from the GSFC Eleanor-lite Pipeline.
+
```
christinahedges, line 23: Add the link to the eleanor documentation here

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,114 @@
...
+    By default, eleanor's `CORR_FLUX` column is used to populate the `flux` values. Note that the "FLUX_ERR" column in the 
+    Eleanor FITS file is referred to the uncertainty of "RAW_FLUX", not "CORR_FLUX". 
+    Thus the uncertainty reported in the 'flux_err' column here is calculated as follows: corr_flux_err = corr_flux*raw_flux_err/raw_flux. 
+    For completeness, the original raw_flux's error is added as a "raw_flux_err" column.
+
```
christinahedges, line 28: Add explanation of quality bits, a la @ethankruse comments from #1217

vbkostov:
> @vbkostov On public documentation and `quality_bitmask` semantics.
> 
> ### `quality_bitmask`
> we should support `quality_bitmask`, rather than just ignore it, given that eleanor quality bits is basically standard TESS SPOC ones. The implementation in `read_tess_lightcurve()` would be a good starting point.
> 
> For the handling of the 2 additional eleanor-specific bits Ethan mentioned, I suggest we include them in the `"hardest"` case. Based on what Ethan described, that eleanor is rather aggressive in marking cadences to be excluded and probably should not be default, including them in the `"hardest"` would make sense (hardest means any flag set).
> 
> ### public documentation and API tweaks
> * parameters `time_column='TIME', flux_err_column='FLUX_ERR', cadenceno_column='FFIINDEX', centroid_col_column='X_CENTROID', centroid_row_column='Y_CENTROID', quality_column='QUALITY'` should be hardcoded rather than exposed as parameters, because there are fixed in the underlying eleanor files.
> * Replace outdated references to QLP in docstring
> * Include a link to https://archive.stsci.edu/hlsp/gsfc-eleanor-lite in the docstring

I updated the "hardest" quality_bitmask to handle the eleanor-specific bits 17 and 18 by setting them equal to ManualExclude(128). It's crude but it works...

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,114 @@
...
+from ..lightcurve import TessLightCurve
+from ..utils import TessQualityFlags
+from astropy import units as u
+
+from .generic import read_generic_lightcurve
+
+import numpy as np
+
+def read_eleanorlite_lightcurve(filename,
+    time_column="TIME",
```
christinahedges, line 13: These won't change, we can hard code them, don't need to make them parameters

christinahedges:
@vbkostov I translated @orionlee's comments to a review for you so you can see exactly where these comments are and address them. 

> I updated the "hardest" quality_bitmask to handle the eleanor-specific bits 17 and 18 by setting them equal to ManualExclude(128). It's crude but it works...

@vbkostov we had a discussion with @ethankruse and agreed not to put bits 17-18 as manual exclude because they are too aggressive, and users will want to use the "manual exclude" from the pipeline. 

> For the handling of the 2 additional eleanor-specific bits Ethan mentioned,
I suggest we include them in the "hardest" case. Based on what Ethan described, that eleanor is rather aggressive in marking cadences to be excluded and probably should not be default, including them in the "hardest" would make sense (hardest means any flag set).

@orionlee I don't know exactly how we would do that, the hardest bitmask is set for all light curves. Can you help me understand how we should achieve this?

src/lightkurve/io/eleanorlite.py
vbkostov: Thanks @christinahedges, I added the link

src/lightkurve/io/eleanorlite.py
vbkostov: Done

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,114 @@
...
+
+    By default, eleanor's `CORR_FLUX` column is used to populate the `flux` values. Note that the "FLUX_ERR" column in the 
+    Eleanor FITS file is referred to the uncertainty of "RAW_FLUX", not "CORR_FLUX". 
+    Thus the uncertainty reported in the 'flux_err' column here is calculated as follows: corr_flux_err = corr_flux*raw_flux_err/raw_flux. 
+    For completeness, the original raw_flux's error is added as a "raw_flux_err" column.
+
+    Parameters
+    ----------
+    filename : str
+        Local path or remote url of a QLP light curve FITS file.
```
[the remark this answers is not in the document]
vbkostov: Fixed

src/lightkurve/io/eleanorlite.py
vbkostov: Done

orionlee (changes requested):

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,117 @@
...
+        flux_column=flux_column.lower(),
+        flux_err_column = flux_err_column.lower(),
+        time_format="btjd",
+        quality_column= quality_column.lower(),
+        centroid_col_column = centroid_col_column.lower(),
+        centroid_row_column = centroid_row_column.lower(),
+        cadenceno_column = cadenceno_column.lower()
+    )
+
+    if quality_bitmask == "hardest":
```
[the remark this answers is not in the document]
vbkostov: Thanks @orionlee, very much appreciated! 

I updated the eleanor reader and also test_eleanorlite to assert that the default bitmask reads in the two extra eleanor flags but the hardest bitmask doesn't

src/lightkurve/io/eleanorlite.py
```diff
@@ -37,9 +37,9 @@ def read_eleanorlite_lightcurve(filename,
         be used to mask out bad cadences. If a string is passed, it has the
         following meaning:
             * "none": no cadences will be ignored (`quality_bitmask=0`).
-            * "default": cadences with flags indicating AttitudeTweak, SafeMode, CoarsePoint, EarthPoint, Desat, or ManualExclude will be ignored
-            * "hard": cadences with default flags, ApertureCosmic, CollateralCosmic, Straylight, or Straylight2 will be ignored
-            * "hardest": TBD: cadences with all the above flags will be ignored, in addition to cadences with GSFC-ELEANOR-LITE bit flags of 17 or 18. This is done by setting both of these flags to be equal to ManualExclude = 128
+            * "default": cadences with flags indicating AttitudeTweak, SafeMode, CoarsePoint, EarthPoint, Desat, or ManualExclude will be ignored.
+            * "hard": cadences with default flags, ApertureCosmic, CollateralCosmic, Straylight, or Straylight2 will be ignored.
+            * "hardest": cadences with all the above flags will be ignored, in addition to cadences with GSFC-ELEANOR-LITE bit flags of 16 or 17.
```
orionlee, line 42: Describing them as 16/17 (or 17/18) only is probably not very clear. The conversation so far proves that.

I suggest we describe them as bit 17 (decimal value 65536) and bit 18 (decimal value 131072) . 

I say bit 17/18 rather than 16/17 so that it is consistent with the convention used in [TESS documentation](https://outerspace.stsci.edu/display/TESS/2.0+-+Data+Product+Overview).

---
Overall, we probably want to state something to the effect that Eleanor uses the TESS SPOC quality flags. Additionally, it adds bit 17 and bit 18. 

For most users, who are familiar with standard TESS SPOC products, it'd be easier to understand.

orionlee (changes requested):

src/lightkurve/io/eleanorlite.py
orionlee: The docstring looks good to me. From my end, the only outstanding issue is the testing of quality bitmasks. ( I've commented on `test_eleanorlite.py`, in case they get lost among the messages. )

tests/io/test_eleanorlite.py
```diff
@@ -20,6 +20,11 @@ def test_eleanor_lite():
         lc = read_eleanorlite_lightcurve(url, quality_bitmask=0)
         assert lc.meta["FLUX_ORIGIN"] == "corr_flux"
         assert_array_equal(lc.flux.value, hdulist[1].data["CORR_FLUX"])
+        # Are the correct quality flags read in?
+        lc = read_eleanorlite_lightcurve(url, quality_bitmask='default')
+        assert 2**16 and 2**17 in lc["quality"]
+        lc = read_eleanorlite_lightcurve(url, quality_bitmask='hardest')
+        assert 2**16 and 2**17 not in lc["quality"]
```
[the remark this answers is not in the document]
vbkostov: Thanks again, I'll fix these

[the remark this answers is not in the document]
vbkostov: > similar issue.

The two bits eleanor adds are 17 and 18. I updated the reader and the test accordingly

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,119 @@
...
+            * "hard": cadences with default flags, ApertureCosmic, CollateralCosmic, Straylight, or Straylight2 will be ignored.
+            * "hardest": cadences with all the above flags will be ignored, in addition to cadences with GSFC-ELEANOR-LITE bit flags of 17 or 18.
+    """
+
+    time_column="TIME"    
+    flux_err_column="FLUX_ERR"
+    cadenceno_column="FFIINDEX"
+    centroid_col_column="X_CENTROID"
+    centroid_row_column="Y_CENTROID"
+    quality_column="QUALITY"
```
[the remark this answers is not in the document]
vbkostov: Thank you, I moved them read_generic_lightcurve

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,112 @@
...
+    In terms of quality flags, eleanor uses the TESS SPOC quality flags by identifying short-cadence targets that fall on each camera-CCD pairing for a given sector. However, eleanor, also adds two new quality flags -- bit 17 (decimal value 131072)) and bit 18 (decimal value 262144). 
+
+    Parameters
+    ----------
+    filename : str
+        Local path or remote url of a GSFC-ELEANOR-LITE light curve FITS file.
+    flux_column : 'RAW_FLUX', 'CORR_FLUX', 'PCA_FLUX', or 'FLUX_BKG'
+        Which column in the FITS file contains the preferred flux data?
+        By default the "Corrected Flux" flux (CORR_FLUX) is used.
+    flux_err_column: 'FLUX_ERR'
```
orionlee, line 31: `flux_err_column` is no longer a parameter. So it should be removed from Parameters section of the docstring.

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,112 @@
...
+from astropy import units as u
+
+from .generic import read_generic_lightcurve
+
+import numpy as np
+
+def read_eleanorlite_lightcurve(filename,
+    flux_column="CORR_FLUX",
+    quality_bitmask="default", 
+    **kwargs):
```
orionlee, line 15: What is the use case for the optional `**kwargs`? The other readers (TESS, QLP, etc.) don't have it.

If there is some use case, it needs to be documented.

If there is no use case, let's take it out of the function signature.

tests/io/test_eleanorlite.py
```diff
@@ -0,0 +1,37 @@
...
+from numpy.testing import assert_array_equal
+
+from lightkurve import search_lightcurve
+from lightkurve.io.eleanorlite import read_eleanorlite_lightcurve
+from lightkurve.io.detect import detect_filetype
+
+@pytest.mark.remote_data
+def test_eleanor_lite():
+    """Can we read in GSFC-ELEANOR-LITE light curves?"""
+    url = "https://mast.stsci.edu/api/v0.1/Download/file?uri=mast:HLSP/gsfc-eleanor-lite/s0001/0000/0003/3673/2616/hlsp_gsfc-eleanor-lite_tess_ffi_s0001-0000000336732616_tess_v1.0_lc.fits"
```
orionlee, line 14: - please remove trailing whitespace
- Style: For long URL, we  generally break it down so that each line is < 127 characters (usually < 100 characters if possible). See `test_lightcurve.py` for an example.
  - In general, we also try to limit each line to be less than 127 characters (preferrably less than 100 characters), for readability / ease of review, e.g., avoid the need of horizontal scroll when one reviews a PR. But this is not strictly enforced.

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,112 @@
...
+    """Returns a `TessLightCurve` object given a light curve file from the GSFC Eleanor-lite Pipeline (https://archive.stsci.edu/hlsp/gsfc-eleanor-lite).
+
+    By default, eleanor's `CORR_FLUX` column is used to populate the `flux` values. Note that the "FLUX_ERR" column in the 
+    Eleanor FITS file is referred to the uncertainty of "RAW_FLUX", not "CORR_FLUX". 
+    Thus the uncertainty reported in the 'flux_err' column here is calculated as follows: corr_flux_err = corr_flux*raw_flux_err/raw_flux. 
+    For completeness, the original raw_flux's error is added as a "raw_flux_err" column.
+    In terms of quality flags, eleanor uses the TESS SPOC quality flags by identifying short-cadence targets that fall on each camera-CCD pairing for a given sector. However, eleanor, also adds two new quality flags -- bit 17 (decimal value 131072)) and bit 18 (decimal value 262144).
```
orionlee, line 22: docstring style: I suggest we add a blank line before "In terms of quality flags, ", so that the quality flags text will have its own paragraph. As it is, the entire text, from `CORR_FLUX` to quality flags, are rendered as a single long paragraph that is a bit hard to read.

orionlee (changes requested):
@vbkostov  I think it looks pretty good! There are just a few minor things.
- some docstring editing
- the optional `**kwargs` should probably be taken out.
- code style: trailing whitespaces. Please remove them, as they tend to cause spurious diff down the road. To save the troubles, in most editors, you can configure it so that when a file is saved, trailing whitespaces are trimmed automatically.
- code style: long lines (> 127 characters). This is not strictly enforced, nor a showstopper, but it'd be a plus if the long lines are broken up (< 127 characters, preferrably <100 characters), primarily for readability , ease of review (to avoid horizontal scrolling)

orionlee:
The CI check failures on the Linux tests are unrelated to the PR. I ran the `--remote-data` variant locally (on Windows) with no issue found.

For the failing CI runs on Linux, somehow the `install poetry` failed with a `JSONDecodeError`:
https://github.com/lightkurve/lightkurve/runs/7078270024?check_suite_focus=true#step:4:108

- Comparing it with the [most recent run](https://github.com/lightkurve/lightkurve/actions/runs/2557187672) that worked, the failing CI run tried to install virtualenv v20.15.0 , instead of v20.14.1 (in the recent run that worked)
- however, windows / OSX tests also use virtualenv v20.15.0 without any issue. So it appears to be some combination of version of packages that cause the failure on Linux tests.

@barentsen Somehow I had to repeatedly approve the to let the checks run, for vbkostov is a first-time contributor. Is that the expected behavior (that checks need to be approved every single time, until the PR is merged and vbkostov is considered a contribtor)?

vbkostov:
> @vbkostov I think it looks pretty good! There are just a few minor things.
> 
>     * some docstring editing
> 
>     * the optional `**kwargs` should probably be taken out.
> 
>     * code style: trailing whitespaces. Please remove them, as they tend to cause spurious diffs down the road. To save the troubles, in most editors, you can configure it so that when a file is saved, trailing whitespaces are trimmed automatically.
> 
>     * code style: long lines (> 127 characters). This is not strictly enforced, nor a showstopper, but it'd be a plus if the long lines are broken up (< 127 characters, preferrably <100 characters), primarily for readability , ease of review (to avoid horizontal scrolling)

Thanks again! I removed the whitespaces and the optional **kwargs, and split the lines so that none is longer than 127 characters

tests/io/test_eleanorlite.py
```diff
@@ -0,0 +1,39 @@
...
+
+from lightkurve import search_lightcurve
+from lightkurve.io.eleanorlite import read_eleanorlite_lightcurve
+from lightkurve.io.detect import detect_filetype
+
+@pytest.mark.remote_data
+def test_eleanor_lite():
+    """Can we read in GSFC-ELEANOR-LITE light curves?"""
+    url =
+    "https://mast.stsci.edu/api/v0.1/Download/file?uri=mast:HLSP/gsfc-eleanor-lite/s0001/0000/"
```
orionlee, line 15: Unfortuantely, the syntax for breaking down the url to 2 line isn't correct, causing the tests to fail:
https://github.com/lightkurve/lightkurve/runs/7080678164?check_suite_focus=true#step:5:35

For it to work, brackets are needed to surround the lines:
```python
url = (
    "line1"
    "line2"
)
```

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,115 @@
...
+    tic = lc.meta.get("TIC_ID")
+    if tic is not None:
+        # compatibility with SPOC, QLP, etc.
+        lc.meta["TARGETID"] = tic
+        lc.meta["TICID"] = tic
+        lc.meta["OBJECT"] = f"TIC {tic}"
+        # for Lightkurve's plotting methods
+        lc.meta["LABEL"] = f"TIC {tic}"
+
+    return TessLightCurve(data=lc, **kwargs)
```
orionlee, line 115: With `**kwargs` removed, it needs to be taken out.

orionlee (changes requested):
Thanks for making the changes. 
There are a couple syntax errors introduced, however. (One of the reasons why it is a good idea to run unit tests locally before pushing to github).

src/lightkurve/io/eleanorlite.py
vbkostov: Good catch, parameter removed

tests/io/test_eleanorlite.py
vbkostov: Thanks, syntax updated

orionlee (changes requested):
A minor docstring formatting issue noted when I render the docstring.

orionlee:
The API doc (once the formatting issue is fixed) for review:

image

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,115 @@
...
+        Local path or remote url of a GSFC-ELEANOR-LITE light curve FITS file.
+    flux_column : 'RAW_FLUX', 'CORR_FLUX', 'PCA_FLUX', or 'FLUX_BKG'
+        Which column in the FITS file contains the preferred flux data?
+        By default the "Corrected Flux" flux (CORR_FLUX) is used.
+    quality_bitmask : str or int
+        Bitmask (integer) which identifies the quality flag bitmask that should
+        be used to mask out bad cadences. If a string is passed, it has the
+        following meaning:
+            * "none": no cadences will be ignored (`quality_bitmask=0`).
+            * "default": cadences with flags indicating AttitudeTweak, SafeMode, CoarsePoint, EarthPoint, Desat, or
```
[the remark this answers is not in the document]
vbkostov: Thanks again! Docstring fixed

src/lightkurve/io/eleanorlite.py
```diff
@@ -0,0 +1,115 @@
...
+               ManualExclude will be ignored.
```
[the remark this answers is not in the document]
vbkostov: Sorry about that, this should do it

orionlee:
@christinahedges @barentsen I think this PR is ready to be merged.

See the API doc:

image

barentsen (approved):
The changes look good to me.  Thanks everyone. :+1:

I intend to release Lightkurve v2.3 after this is merged.

orionlee:
@barentsen This PR does not contain the changes in `.rst` files so that the data product specific API doc (including Eleanor) is exposed.

A barebone fix is in: 
https://gist.github.com/orionlee/05e49c4ed15fe3986fda3bc51c8d6544

My intention is to have a small PR to fix them after this PR is merged.
From my perspective, the data product specific public API doc can be in v2.3 or the next release.

orionlee:
I intend to merge this PR by Wednesday July 6. Reply back if there is any concern.

barentsen:
:+1: Thanks all!

Issue #1163, opened 2021-11-22
Quality flags are missing from the TessQualityFlags class


mrtommyb:
#### Problem description

The TESS quality flags were updated. They are now 16 bit. They are described [here](https://archive.stsci.edu/files/live/sites/mast/files/home/missions-and-data/active-missions/tess/_documents/EXP-TESS-ARC-ICD-TM-0014-Rev-F.pdf), which was updated in Sep 2020.

We should add these flags to the TESSQualityFlags class. I think that it may be worth a discussion on whether any of the new flags get added to _default_ or _hard_, but they should be added to _hardest_.
![Screen Shot 2021-11-22 at 11 36 39 AM](https://user-images.githubusercontent.com/1900573/142899990-1c8f28eb-7635-4391-abf4-67e33186d152.png)

mrtommyb:
@rebekah9969 do you have a feel if bits 14-16 compromise the data at all, and should be excluded by default. I suspect the answer is no, but I don't know for sure.

barentsen:
@rebekah9969 Any update on whether bits 14-16 compromise the data?

I will go ahead and add these flags to the "hardest" mask only unless I hear otherwise.

rebekah9969:
@barentsen and @mrtommyb I contacted Michael about this and he said the following...

"Bit 14 is specific to planet science, and different users may find the data entirely acceptable.  You probably don't need to exclude this bit by default, although we haven't used it since year 2 and so it may never make a difference.

Bit 15 and 16 are pretty serious--- for 15, there shouldn't be any valid data except in the RAW_CNTS TPF column, I think.  Best to throw these data out by default.

Bit 16 means that PDC cannot apply a correction for that time stamp.  There is SAP_FLUX data at that time that may be useful for some researchers, but there is no PDC data at these times.  I don't know if what the default assumptions for lightkurve would be in this case.  I'd always look at the difference between SAP and PDCSAP flux and decide on a case by case basis, so I wouldn't want this bit removed by default.  But I could see how it may confuse someone if SAP_FLUX and PDCSAP_FLUX have different lengths."

I would therefore exclude 15 at minimum.

orionlee:
I'd suggest at the minimal we should add the flags to TESSQualityFlags class, and include them all in the `hardest` (unless we change its definition).

Anecdotally, there are noticeable amount of cadences that are flagged as such. 
The following are some examples from some 2-minute lightcurve files, loaded with default mask.

```
[...]

PR #1218, opened 2022-06-13
Update the `TessQualityFlags` class with new quality flags from the mission

christinahedges:
The TESS SPOC have updated their quality flags, adding bits 13-15. This PR adds those new keywords along with their descriptions.

### DIFF
--- a/src/lightkurve/utils.py
+++ b/src/lightkurve/utils.py
@@ -761,7 +761,6 @@ def show_citation_instructions():
     # because we can assume it is installed when notebook-specific features are called
     try:
         from IPython.display import HTML
-
         ipython_installed = True
     except ModuleNotFoundError:
         ipython_installed = False
--- a/src/lightkurve/utils.py
+++ b/src/lightkurve/utils.py
@@ -770,7 +770,7 @@ def show_citation_instructions():
         print(__citation__)
     else:
         from pathlib import Path  # local import to speed up `import lightkurve`
-        import astroquery  # local import to speed up `import lightkurve`
+        import astroquery         # local import to speed up `import lightkurve`
 
         templatefile = Path(PACKAGEDIR, "data", "show_citation_instructions.html")
         template = open(templatefile, "r").read()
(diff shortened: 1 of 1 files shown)


christinahedges:
@barentsen I requested your review here to check there's nothing missing from this PR, thank you for any help!

orionlee:
Close  #1163   

Comments: 

1. `HARDEST_BITMASK` should be updated accordingly.

https://github.com/lightkurve/lightkurve/blob/155708d5d52324eb190c72412c943d960f940a51/src/lightkurve/utils.py#L258-L259

2. #1163 also has some discussion on whether the default bitmask should be updated. There was no definite conclusion. It shouldn't block this PR anyway.
3.

christinahedges:
Thanks @orionlee, maybe the TESS GI office need to have a think about what we would make the recommended bitmask for most users and check that aligns with the default. I'll update the `HARDEST_BITMASK` and then I think we can get this merged.

orionlee:
@barentsen I think this can be merged too.

The issue about whether default bitmask needs to be changed is not a showstopper. From what I could tell, the new bits don't affect the result of PSCSAP_FLUX much anyway.

src/lightkurve/utils.py
```diff
@@ -250,7 +255,7 @@ class TessQualityFlags(QualityFlags):
         DEFAULT_BITMASK | ApertureCosmic | CollateralCosmic | Straylight | Straylight2
     )
     #: HARDEST bitmask identifies cadences with any flag set. Its use is not recommended.
-    HARDEST_BITMASK = 8191
+    HARDEST_BITMASK = 65535
```
barentsen, line 258: Sanity check: this value is correct because ``(32768 * 2) - 1 == 65535``.

barentsen:
> The issue about whether default bitmask needs to be changed is not a showstopper.

Agreed.  The concept of a default bitmask is getting tricky now that we have multiple pipelines which include quality flags that are often similar but not identical.  Default bitmask recommendations should probably be documented in the pipeline-specific `read_{pipeline}_lightcurve()` functions located in `io/{pipeline}.py`.  An unfortunate drawback is that these pipeline-specific reader functions are rather hidden in the call stack when opening light curves via `SearchResult.download()`.

barentsen (approved):
LGTM!

This fix will be part of the upcoming v2.3.0 release. I will add an entry to the change log after merging.

PR #1221, opened 2022-06-15
Perror

vvilhelmus:
Hello again,
because you added my name to the contributors and I felt like I didn't really make a contribution (I was being dragged over the finish line like a heavy ball with chain on someone else's leg doing all the actual work), I thought to better make a real contribution this time.
Last year I would have liked to see the error margins of the periodicity from the ice giants exoplanet's transits for my Master's degree Astrophysics science report, however, this feature seemed to be missing from Lightkurve, so I made my own one, which I am proposing here now.
This is the first time I created a definition in Python, so it is probably full of rookie mistakes and ugly as heck, hence why I am looking forward to your feedback. :crossed_fingers:

### DIFF
--- a/docs/source/reference/periodogram.rst
+++ b/docs/source/reference/periodogram.rst
@@ -45,3 +45,5 @@ Methods
   Periodogram.smooth
   Periodogram.to_seismology
   Periodogram.to_table
+  BoxLeastSquaresPeriodogram.compute_stats
+  BoxLeastSquaresPeriodogram.get_transit_model
--- a/src/lightkurve/periodogram.py
+++ b/src/lightkurve/periodogram.py
@@ -9,6 +9,7 @@ import warnings
 
 import numpy as np
 from matplotlib import pyplot as plt
+from scipy.signal import find_peaks
 
 import astropy
 from astropy.table import Table
@@ -1330,3 +1331,68 @@ class BoxLeastSquaresPeriodogram(Periodogram):
         raise NotImplementedError(
             "`smooth` is not implemented for `BoxLeastSquaresPeriodogram`. "
         )
+    
+    def perror(self, **kwargs):
+        """Calculates the lower & upper error of the periodicity of the first peak found,
+        that is higher than 90% the period's maximum power discovered in the periodogram,
+        using the Half Width at Half Maximum method, and plots the peak with the results
+        in a Matplotlib plot.
+        See scipy.signal.find_peaks documentation for more details.
+
+        Parameters
+        ----------
+        kwargs : dict
+            Dictionary of arguments to be passed to `Periodogram.plot`.
+
+        Returns
+        -------
+        plot : `~matplotlib.pyplot.Plot`
+            The matplotlib plot object.
+        """
+        peaks, stats = find_peaks(self.power,
+                                  prominence=float(self.max_power)*0.9,
+                                  width=1)
+        peak = self.period[peaks][0].to_value(unit=None)
+        lower_error_period_int_down = self.period[math.floor(stats["left_ips"][0])].to_value(unit=None)
+        lower_error_period_int_up = self.period[math.ceil(stats["left_ips"][0])].to_value(unit=None)
+        lower_error_int_remainder = stats["left_ips"][0] - math.floor(stats["left_ips"][0])
+        lower_error_period = lower_error_period_int_down + lower_error_int_remainder * (lower_error_period_int_up - lower_error_period_int_down)
+        lower_period_error = peak - lower_error_period
+        upper_error_period_int_down = self.period[math.floor(stats["right_ips"][0])].to_value(unit=None)
+        upper_error_period_int_up = self.period[math.ceil(stats["right_ips"][0])].to_value(unit=None)
+        upper_error_int_remainder = stats["right_ips"][0] - math.floor(stats["right_ips"][0])
+        upper_error_period = upper_error_period_int_down + upper_error_int_remainder * (upper_error_period_int_up - upper_error_period_int_down)
+        upper_period_error = upper_error_period - peak
+        plt.plot(self.period,
+                 self.power)
+        plt.hlines(y = stats["width_heights"][0],
+                   xmin = lower_error_period,
+                   xmax = upper_error_period,
+                   ls=":")
+        plt.vlines(x=self.period[peaks][0].to_value(unit=None),
+                   ymin=(self.power[peaks][0]-stats["prominences"][0]),
+                   ymax=self.power[peaks][0],
+                   ls=":")
+        plt.xlim(peak - 5*(peak - lower_error_period),
+                 peak + 5*(upper_error_period - peak))
+        plt.text(peak,
+                 self.power[peaks][0],
+                 'Period at maximum power: ',
+                 ha='right',
+                 va='center')
+        plt.text(peak,
+                 self.power[peaks][0],
+                 f'   {peak} days',
+                 ha='left',
+                 va='center')
+        plt.text(lower_error_period,
+                 stats["width_heights"][0],
+                 f'Lower error:\n- {lower_period_error:.5g}',
+                 ha='right',
+                 va='bottom')
+        plt.text(upper_error_period,
+                 stats["width_heights"][0],
+                 f'Upper error:\n+ {upper_period_error:.5g}',
+                 ha='left',
+                 va='bottom')
+        return plt.show()
--- a/src/lightkurve/periodogram.py
+++ b/src/lightkurve/periodogram.py
@@ -1186,7 +1186,7 @@ class BoxLeastSquaresPeriodogram(Periodogram):
     def compute_stats(self, period=None, duration=None, transit_time=None):
         """Computes commonly used vetting statistics for a transit model.
 
-        See astropy.stats.bls docs for further details.
+        See `~astropy.timeseries.BoxLeastSquares` docs for further details.
 
         Parameters
         ----------
@@ -1223,7 +1223,7 @@ class BoxLeastSquaresPeriodogram(Periodogram):
     def get_transit_model(self, period=None, duration=None, transit_time=None):
         """Computes the transit model using the BLS, returns a lightkurve.LightCurve
 
-        See astropy.stats.bls docs for further details.
+        See `~astropy.timeseries.BoxLeastSquares` docs for further details.
 
         Parameters
         ----------


orionlee:
It looks like there are 2 features:
1. calculate the half width at half maximum of the peak
2. plot to illustrate the peak

Calculate half width at half maximum of the peak
- I wonder if calling it error would be misleading. I'm not really well versed on this though. @barentsen thoughts?
- it should probably be 
  - available as a function, without being tied to the plotting features.
  - available for all `Periodogram`, not just BLS, and applicable to both period and frequency.

Plotting features
- I feel it's probably better exposed it as an optional feature of the existing `plot()` function. It'd avoid API bloat, and take advantage of the existing plotting styling, flexibility, etc.
- Stepping back, so far Lightkurve shies away from providing use case specific plotting features. I wonder if this might be a step too specific.


Example: existing `pg.plot()` result:
image

The new plot returned by `pg.perror()` in this PR:
image

vvilhelmus:
Hi Sam @orionlee 
thank you for your feedback.

> I wonder if calling it error would be misleading.

I have considered calling it lower & upper limit, but I found that even more mathematically confusing because of the limit function -> https://en.wikipedia.org/wiki/Limit_function, so I do not know what to call it else. Width? Cut-off?

> it should probably be
>   available as a function, without being tied to the plotting features.

The reason I added the plotting feature is because it shows the resolution of the peak. Perror does not generate a new periodogram, it only zooms in on the peak with max power in the periodogram that was previously created. In your example you can immediately see how the resolution of the peak is not smooth at all, you can see the straight lines. The Perror plot therefore shows that redoing the periodogram in higher resolution will give you error values, that might be more accurate and precise. This is something you cannot see by just looking at raw data numbers...

> it should probably be
>   available for all Periodogram, not just BLS, and applicable to both period and frequency.

The find_peaks function (which I learned about from the asteroseismology tutorial of @ojhall94 & @barentsen, thank you btw) can find multiple peaks, but it only takes those peaks from a 1-D array unfortunately, see https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.find_peaks.html webpage. The related periods/frequencies have to be calculated extra afterwards, which probably means your whole periodogram methodology needs to be rewritten entirely, and that's a rabbit hole I did not want to tumble into, being utterly new to Python coding myself, haha. :sweat_smile: 

> Git comments: the PR currents an unrelated commits https://github.com/lightkurve/lightkurve/commit/c92d966e9045b4a5f040d4d2f1cf80aa62668931 and https://github.com/lightkurve/lightkurve/commit/2eb1122b0e9eef916fc5cc9022789b36636c8458 (API doc fix) that were in your earlier PR. Please remove them to merge complication down the road. I'd suggest you git rebase your branch with the latest main, and ensure it contains only relevant commits in the PR. You probably need to edit the commits to really take it out.
> 
> On second thought, given you only have 1 relevant commit so far, it might be easier to create a new branch instead, if you're not familiar with git rebase and commits editing.

Yes, sorry, I saw that mistake after the git push command and didn't know how to fix it... I am not very proficient with Github yet. I'll see what I can do.

orionlee:
@vvilhelmus I also feel it's best to wait for feedback from @barentsen or others familiar with the domain before proceeding any further.

In the meantime, I think there is some misunderstanding on my comment about "... be available for all Periodogram, not just BLS, and applicable to both period and frequency."

What I have in mind in something like this:

```python

class Periodogram:
    def compute_hwhm_at_max_power(view=None):
        ... # the actual implementation
        return [witdh_lower, width_upper]
```


It can be used to get HWHM for any periodogram, be it in period view or frequency view.
```
    #  pg_bls is a BLS periodogram
  
    # return the lower/upper for peak period, because pg_bls.default_view == "period"
    width_lower, width_upper = pg_bls.compute_hwhm_at_max_power()

    # return the lower/upper for peak frequency
    width_lower, width_upper = pg_bls.compute_hwhm_at_max_power(view="frequency")


    #  pg_ls is a LombScarglePeriodogram

    # return the lower/upper for peak frequency, because pg_ls.default_view == "frequency"
    width_lower, width_upper = pg_bls.compute_hwhm_at_max_power()

```

vvilhelmus:
Hi again Sam @orionlee ,
thank you for your reply and explanation.
Yes, I think it is good to brainstorm this issue for a bit first before continuing. I have come up with some new ideas myself! :bulb: 

I have gone through the periodogram tutorials on the lightkurve.org website, but my experience and own usage is mostly limited to just the Box Least Squares periodograms only to be honest. Is there a need for the HWHM values of a _**single**_ peak with the plain periodogram and Lomb-Scargle periodogram methods? In the asteroseismology tutorial the HWHM value is used in a different way and on the whole **_collection_** of peaks, not on a single peak as far as I have understood it. -> https://docs.lightkurve.org/tutorials/3-science-examples/asteroseismology-estimating-mass-and-radius.html#3.-Estimating-the-Frequency-Spacing-\Delta\nu I might be wrong on this, because my asteroseismology experience is only limited to that tutorial, but if there is no usage for HWHM values of a single peak in plain & Lomb-Scargle periodograms, then I do not really see the point of it...

One new idea I have come up with is a 'list_peaks'-definition, that produces a table instead of a plot.
The 1st column will be the power values of all the peaks in descending order, so the max_power value will be on top.
The 2nd column will be the periodicity values of those respective peaks found in the periodogram.
The 3rd column will be the lower half width value at half maximum (LHWHM) value of those respective peaks.
The 4th column will be the upper half width value at half maximum (UHWHM) value of those respective peaks.
And perhaps a 5th column, which would give some indication, if the peak is an echo of the top max_power peak. That way the peaks of another transiting exoplanet periodicity can easily be spotted and detected. Also the Lightkurve users can take those values from the table and create their own plots (which is what the devs try to encourage with the Lightkurve program, if I got your philosophy correct?). I have no idea by the way how to create such a 5th column though, but that's the main idea what is playing on in my mind right now. :gear: 

My final question, I am still struggling with, is: if the lower & upper HWHM values cannot be used as lower & upper errors, how can the uncertainty of the exoplanet's periodicity then be calculated? :grey_question:
Because that's the whole point of this exercise... to show the statistics of the exoplanet's periodicity... :thinking:

orionlee:
I'm not well versed in the domain either so I can't really comment much about the appropriate uses of HWHM.

Regarding the errors of exoplanet's period: The values from periodogram is generally not used in any final published result. They are primarily used to look for potential candidates. Once candidates are identified, researchers generally perform modeling that comes up with parameters that can fit the observation while obeying various laws of physics. The planet parameters, including orbital period and the associated uncertainty, generally come from those modeling result. 

Having said that, the periodicity identified from periodogram has other uses, e.g., identifying rotational period for spotted stars. I am not as familiar with them enough to speak about the specifics.

vvilhelmus:
Good Monday morning Sam @orionlee and everyone else.

Uhm... I am hesitating to call the king (astrophysicist community) naked, because I might be missing something, but the uncertainty should be calculated based on observations, not on modelling results. This is bad practice and bad science, because every model makes assumptions and leaves a lot of complexities of reality out of the picture. 
> "All models are wrong, but some are useful" -- British statistician George E. P. Box

I didn't even have to look far for an example, I found one on my first attempt. -> https://exoplanetarchive.ipac.caltech.edu/overview/Kepler-25%20c#planet_Kepler-25-c_collapsible
For the ice giant Kepler-25c two periods were reported in 2019:
Mills et al. reported a period of 12.72070±0.00011 days
and Gajdos et al. reported a 12.7203749057±0.0000011577 days period.
The margins of those error bars aren't overlapping at all, so one period has to be wrong and the other period has to be right. However, if they are both based on modelling, then the calculated uncertainties are both wrong and way too underestimated, or am I missing out on something here?

I see many other examples as well, which I am not going to report here. Why are astrophysicists making the periodicity error bars so tiny and doing the statistics of the periodicity deliberately incorrect? Do they do it on purpose, so they can publish more significant figures to help others out with calculations of more precise future transits? I don't get it... the simplistic HWHM values would be more honest than those actual reported results.  :astonished: 

On a side note, I think I have figured something out for the 5th column of the previously proposed list_peaks table, which I am going to work on this week a little bit and I hope to report more on that later on... :hammer:

ojhall94:
Hey @vvilhelmus et al! I'm here to provide some degree of limited asteroseismology/rotation domain knowledge.

But to first answer your previous comment and since we're getting philosophical, I would generally take all uncertainties with a grain of salt, as they are often under-reported due to ie model fitting, as youve described. However a periodogram *is* already a model - since astronomical data are discrete, the Lomb Scargle periodogram approximates a "true" (mathematical) Fourier Transform as best as it can. The same is true for the BLS periodogram. So I would be hesitant to claim any form of uncertainty measurement is more fundamental than another, in this context. It is all interpreting periodograms and deciding how much we trust them.

Second, when you measure something like the tip of a peak and a HWHM and call that an 'error', the next person will immediately assume both those things are a Gaussian with that mean and standard deviation, and treat the error as such (because thats just what everyone does, because it makes things easier). So there's some information being lost there, too.

The latter point is the reason we never built in a specific uncertainty estimator for the tools included in the `seismology` subpackage. At the moment it is a first-look tool, and uncertainty measurement requires users to go away and seek some domain knowledge to interpret the results.  My opinion is that Lightkurve reporting uncertainties is fine, but we must be happy with the results and requested updates to the program that people may want (ie people citing the Lightkurve package to claim a particular mass for a star or planet, for example, and that coming under scrutiny). 

In terms of the two use cases I'm familiar with:
- Stellar rotation: I'm uncertain HWHM is an appropriate uncertainty estimate - if we are assuming the peak is Gaussian (as they often are for coherent oscillators, I believe) then a FWHM, converted to a standard deviation, would perhaps be more appropriate. 
- Asteroseismology: in both ACF measurements when approximating `numax` and `deltanu` Gaussians are only broadly approximating the shapes of the data, and so I'd be very reluctant to use a HWHM or any other tool to estimate uncertainties from these data - in particular because they will then immediately feed into measurements of mass, radius and logg. There are other packages, such as `PySYD`, that do this better and with the appropriate published literature.

Was fun thinking about this!

vvilhelmus:
Thank you for your feedback, Oliver @ojhall94 .
I am not sure what to think about "take all uncertainties with a grain of salt", since uncertainties should reflect the quality of the data. The entire data point becomes meaningless and perhaps even misleading when it comes to future transit predictions, if it isn't taken seriously at all...
Out of curiosity I calculated the standard deviation out of all the published periodicities of Kepler-25c reported on the https://exoplanetarchive.ipac.caltech.edu/overview/Kepler-25%20c#planet_Kepler-25-c_collapsible website with the following NumPy code:
```python
k25c = np.array([12.72070, 12.7203749057, 12.72037372, 12.720374354, 12.7203678, 12.7203724, 12.720, 12.7204, 12.7204])
np.mean(k25c), np.std(k25c)
```
The result is a periodicity of 12.72037±0.00017 days, but is this the level of uncertainty that should be aimed for?
Most of the published periodicities underestimate their uncertainties by 100-1000 fold compared to the above result.
In comparison the crude HWHM method overestimates the uncertainty only by 10 fold, however, Kepler-25c is an old exoplanet discovery and therefore a lot of periodicities have been reported, so is that crude HWHM method then an overestimation still? :shrug: 

Before I am going to create yet another new pull request, I wanted to share the following code, I created in Jupyter Notebook, first:
```python
import astropy
from astropy.table import Table
import lightkurve as lk
import numpy as np
from scipy.signal import find_peaks

def peaks_list(): #compute_hwhm_at_max_power(view=None):
    peaks, stats = find_peaks(pg.power,
                              prominence=(float(pg.max_power)/24),
                              width=1)
    lhwhm_int_down = pg.period[np.floor(stats["left_ips"]).astype(int)]
    lhwhm_int_up = pg.period[np.ceil(stats["left_ips"]).astype(int)]
    lhwhm_int_remainder = stats["left_ips"] - np.floor(stats["left_ips"])
    lhwhm_period = lhwhm_int_down + lhwhm_int_remainder * (lhwhm_int_up - lhwhm_int_down)
    lhwhm = lhwhm_period - pg.period[peaks]
    uhwhm_int_down = pg.period[np.floor(stats["right_ips"]).astype(int)]
    uhwhm_int_up = pg.period[np.ceil(stats["right_ips"]).astype(int)]
    uhwhm_int_remainder = stats["right_ips"] - np.floor(stats["right_ips"])
    uhwhm_period = uhwhm_int_down + uhwhm_int_remainder * (uhwhm_int_up - uhwhm_int_down)
    uhwhm = uhwhm_period - pg.period[peaks]
    result = Table(data=[stats['prominences'], pg.period[peaks], lhwhm, uhwhm],
                   names=('Power', 'Periodicity', 'Lower HWHM', 'Upper HWHM'))
    result.sort('Power',
                reverse=True)
    result['Periodicity Ratio'] = result['Periodicity'][0] / result['Periodicity']
    result['Power'].format = '.1f'
    result['Lower HWHM'].format = '.5g'
    result['Upper HWHM'].format = '+.5g'
    result['Periodicity Ratio'].format = '.3f'
    result['Periodicity Ratio'].unit = 'top peak:peak'
    return result#[0]['Lower HWHM'], result[0]['Upper HWHM']
```
The above definition won't work, if you don't name your periodogram as `pg` by the way.
Deleting the hashtags and the initial `peaks_list(): `-part in the above code will result into the suggested `compute_hwhm_at_max_power(view=None)` definition. The method used in the `perror()` definition had a small chance of obtaining the HWHMs of another peak, so the creation of a sortable table was required anyways. But I wanted to show off this table I envisioned first, which I have tested out on a periodogram of the Kepler-25 light curve and it produced the following table:
peaks_list
First column shows the BLS Power sorted in descending order, second column their respective periodicities, third & fourth column the HWHMs of their periodicity, and the last column shows the periodicity ratio with the max_power peak from the Kepler-25c exoplanet transit in this example. What sticks out in this last column is the 4th row, which is no longer a neat fraction, as in:
1/1
2/1
1/2
2.039
3/2
2/3
etc.
This 4th periodicity peak is from the second exoplanet transit: Kepler-25b.
I think this is a real easy detection method of secondary exoplanet transits.
So do you think this `peaks_list()` table would be a better addition to LightKurve than the suggested `compute_hwhm_at_max_power(view=None)` array?
I am looking forward to more of your feedback, thoughts, or suggestions. One change I want to create myself, is to make it possible for users to set the S/N threshold for the table themselves, but if the array is preferred over the table, then there is no point of course.

vvilhelmus:
This thread continues in #1255 where I hopefully have uploaded my tutorial Jupyter notebook webpage correctly, which will show why the errors in the periodicity are relevant. :smiley:

PR #1232, opened 2022-06-27
Reflect TESS GI proposal url changes

orionlee:
An amendment to the not-yet-released PR #1134.

At some point TESS office has changed the URL of the list of primary mission GI. 
This PR provides a fix to reflect th changes.

No need of a new entry in the changelog, but we could add the PR number to the existing one referencing #1134

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,7 @@
 
 - Allow users to include more columns in ``SearchResult`` display via a new
   ``SearchResult.display_extra_columns`` attribute, with defaults set by an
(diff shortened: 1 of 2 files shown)


orionlee:
@barentsen I suggest we merge this small fix for v2.3 release. 

The PR fixed the TESS GI proposal URLs for the primary mission proposals - at some point TESS office changed moved all primary proposals to a separate URL. 

It is not a showstopper per se, but it'd be better if we include this, so that the merged PR #1134 won't be released with a known bug.

barentsen (approved):
LGTM :+1:

PR #1233, opened 2022-06-28
Expose data product-specific doc as public API doc

orionlee:
It also:
- exposes the generic `lk.read()` function
- removes [LightCurve.read()](https://docs.lightkurve.org/reference/api/lightkurve.LightCurve.read.html) from public documentation. 
  - Users can't really directly use it to read lightcurve FITS file anyway (only concrete subclass `Kepler / TessLightCurve` would work).
  - It does have a auto-generated list of supported format. But I think the new `io` section serves similar purpose.

There are some editing that can be done (adding links to correspond HLSP doc, fixing broken links, etc.) on individual functions. If we agree on the general approach, the additional editing can be added to this PR or some subsequent ones.

The PR assumes GSFC-ELEANOR-LITE support PR #1217 will be merged, and contains references to it.
But even if #1217 is not merged, the PR won't break doc build (just causes some additional warning).

### DIFF
--- a/docs/source/reference/io.rst
+++ b/docs/source/reference/io.rst
@@ -1,0 +1,45 @@
+.. _api.io:
+
+=====================
+Reading Data Products
+=====================
+.. currentmodule:: lightkurve.io
+
+
+.. autosummary::
+  :toctree: api/
+
+    read
+
+In general, users only need to call `read` function, also available as ``lightkurve.read``,
+to read all supported Kepler / TESS data products.
+
+The data product-specific functions below **should not** be called directly.
+They do, however, contain information pertaining to specific products, as well as
+optional parameters that can be passed to the generic `read()`.
+
+
+Kepler Data Products
+~~~~~~~~~~~~~~~~~~~~~
+
+.. autosummary::
+  :toctree: api/
+
+    kepler.read_kepler_lightcurve
+    everest.read_everest_lightcurve
+    kepseismic.read_kepseismic_lightcurve
+    k2sff.read_k2sff_lightcurve
+
+
+TESS Data Products
+~~~~~~~~~~~~~~~~~~~~
+
+.. autosummary::
+  :toctree: api/
+
+    tess.read_tess_lightcurve
+    cdips.read_cdips_lightcurve
+    eleanorlite.read_eleanorlite_lightcurve
+    pathos.read_pathos_lightcurve
+    qlp.read_qlp_lightcurve
+    tasoc.read_tasoc_lightcurve
--- a/docs/source/reference/io.rst
+++ b/docs/source/reference/io.rst
@@ -11,12 +11,16 @@ Reading Data Products
 
     read
 
-In general, users only need to call `read` function, also available as ``lightkurve.read``,
-to read all supported Kepler / TESS data products.
+In general, users only need to call the `read` function, also available as
+``lightkurve.read``.  This function will auto-detect the type of data product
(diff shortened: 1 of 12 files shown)


orionlee:
A new top level section, Reading Data Products, is added (correspond to `io` module).

image

orionlee:
The now-exposed `lk.read()` API doc:

image

docs/source/reference/io.rst
```diff
@@ -0,0 +1,45 @@
...
+  :toctree: api/
+
+    read
+
+In general, users only need to call `read` function, also available as ``lightkurve.read``,
+to read all supported Kepler / TESS data products.
+
+The data product-specific functions below **should not** be called directly.
+They do, however, contain information pertaining to specific products, as well as
+optional parameters that can be passed to the generic `read()`.
```
barentsen, line 19: I recommend rewording this section slightly, because the phrase `The data product-specific functions below **should not** be called directly` makes it sound like the user's computer might break if they call this function 😄  

Suggestion:

```suggestion
In general, users only need to call the `read` function, also available as
``lightkurve.read``.  This function will auto-detect the type of data product
being opened and pass it on to a product-specific reader function.

Below we list the product-specific reader functions.  These functions
are not intended to be called directly because it is much easier to
simply call the `read` function.  We include the functions here because
their documentation lists the optional parameters that can be passed to
the generic `read()` function, and because they document information
pertaining to specific products.
```

barentsen:
Thanks @orionlee!  This looks very good!

I agree it looks like the docstrings of the various functions could be improved significantly (both content and rst syntax), but it doesn't necessarily have to happen as part of this PR.  I propose we merge this PR right before releasing v2.3 (likely in 1-2 days?), because it already provides a significant improvement as-is.  Does that sound right?

> removes [LightCurve.read()](https://docs.lightkurve.org/reference/api/lightkurve.LightCurve.read.html) from public documentation. -- Users can't really directly use it to read lightcurve FITS file anyway (only concrete subclass Kepler / TessLightCurve would work).

Good point!!  Perhaps we ought to override `LightCurve.read()` to remove any confusion between `lightkurve.read` and `LightCurve.read`, but removing it from the docs is certainly a good first step.

Side note: I think we eventually ought to deprecate and remove the `KeplerLightCurve` and `TessLightCurve` classes.  In Lightkurve v1.x these classes served as an interface to the official Kepler or TESS light curve files produced by the SOC/SPOC pipeline.  The Lightkurve v2.x philosophy is to have one pipeline-agnostic `LightCurve` object, surrounded by pipeline-specific reader and writer functions (though we haven't made any writer functions yet).

orionlee:
> I agree it looks like the docstrings of the various functions could be improved significantly (both content and rst syntax), but it doesn't necessarily have to happen as part of this PR. I propose we merge this PR right before releasing v2.3 (likely in 1-2 days?), because it already provides a significant improvement as-is. Does that sound right?

Sounds good. I might do a pass of fixing some of the docstrings tomorrow, but if we don't get to it. The PR as-is is a good enough improvement to be merged.

orionlee:
I have done a round of fixes of the now public docstring.
- add links to HLSPs
- fix broken links
- remove incorrect quality_bitmask values stated in various TESS readers docstring
- document `**kwargs` in `lk.read()`

From my end, this PR is ready to be merged.

orionlee:
I intend to merge this PR by Wednesday July 6. Reply back if there is any concern.

barentsen (approved):
LGTM :+1:

PR #1236, opened 2022-07-07
support vanilla eleanor in addition to GSFC-ELEANOR-LITE

orionlee:
This is a follow up of PR #1217 , regarding the support of vanilla eleanor


christinahedges:
This looks great, no notes from me! I think this is the right behavior. Thank you for opening!

barentsen:
My thinking is that different variations of data products from the same pipeline could go in one module, but I guess it's really not that important, i.e. `eleanorlite.py` will do!

orionlee:
On a second thought, I'd like to renaming it to `eleanor.py`. In the scenarios:
i. new pipelines also use `eleanor` to produce their lightcurves, or
ii. eleanor full is to supported.

The changes are likely to be done in the same module.

`eleanor.py` would be more appropriate. The naming is somewhat important as we're about to make it part of the public API doc.

orionlee:
`eleanortlite.py` is renamed to `eleanor.py`, along with various format names and functions (e.g., `read_eleanor_lightcurve()` .

barentsen:
Updated the change log via #1239

PR #1238, opened 2022-07-07
Make  github git blame view ignore formatting commits

orionlee:
It'd make git blame view on github ignore formatting commits, currently d9f9ba1145cad12ddc2b2bef0413827fc7e4a09c

See: 
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view


barentsen (approved):
I verified that `d9f9ba1145cad12ddc2b2bef0413827fc7e4a09c` is indeed a monster commit which applied `black` across the code base.  This is a very useful change.  Thank you!

PR #1262, opened 2022-11-17
Fix interact features for bokeh v3

### LINKED ISSUE #1261: interact features does not work with bokeh 3.x
#### Problem description

Various interact features (e.g., `tpf.interact()`, `lc.interact_bls()`, `Seismology.plot_echelle()` ) do not work with (the latest) bokeh 3.x. 

It is also seen in the regression test, with the following error:
```python
...
tests/test_interact.py:157: 
src/lightkurve/interact.py:288: in make_lightcurve_figure_elements
    fig = figure(
...

FAILED tests/test_interact.py::test_interact_functions - AttributeError: unexpected attribute 'plot_height' to figure, similar attributes are outer_height, height or min_height
```
#### Workaround

Downgrade bokeh from 3.x to 2.x (the latest being 2.4.3)

#### Root cause

- bokeh removed `plot_height` / `plot_width` in v3.x. They need to be replaced with `height` / `width`. 
- Even in bokeh 2.x, `plot_height` / `plot_width` are only meant to be compatibility alias for `height` / `width`.  (See [v2.4.3 doc](https://docs.bokeh.org/en/2.4.3/docs/reference/plotting/figure.html#bokeh.plotting.figure), in section Inherited from Plot.  The even older [v1.4 doc](https://docs.bokeh.org/en/1.4.0/docs/reference/plotting.html#bokeh-plotting) seems to suggest `plot_height` is the parameter. So it looks like `plot_height` is deprecated in v2.x, and removed in v3.x.)
  - see also: [Bokeh v3.0.0 migration guide](https://github.com/bokeh/bokeh/wiki/Migration-Guides#removals-) 

E.g., for `tpf.interact()`:

https://github.com/lightkurve/lightkurve/blob/9763e49b9ae7e794685fed7e0043b5cd67779564/src/lightkurve/interact.py#L288-L291

#### Actions needed

- replace `plot_height` / `plot_width` with `height` / `width` in various bokeh `figure()` calls. 
- raise minimum bokeh requirement from 1.1 to 2.x (otherwise we might need more codes to use plot_height or height depending bokeh version.)
  - 1.x is hard to be installed nowadays anyway. In a minimum python environment, after I install bokeh 1.4, there is some version incompatibility problem. In trying to use `bokeh` by calling ` from bokeh.io import show`, I get `ImportError: cannot import name 'Markup'` from 'jinja2'`. 

#### Environment

-  lightkurve version (e.g. 1.0b6): 2.3.0
-  installation method (e.g. pip, conda, source):

orionlee:
Fix #1261

- replace all `plot_height` / `plot_width` with `height` / `width` in bokeh calls.
- additional fix `tpf.interact()` to make pixel selection compatible with bokeh v3.

changelog to add:
```
- Fixed interactive features to work with bokeh v3. [#1262]
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,8 +3,11 @@
 
 - Added the ability to configure the default cache directory used by
   ``SearchResult.download()`` / ``SearchResult.download_all()``. [#1214]
+
 - Moved the default cache directory from ``$HOME/.lightkurve-cache``
   to ``$HOME/.lightkurve/cache``. [#1214]
+
+- Fixed interact features, e.g. ``tpf.interact()``, to work with bokeh v3.x [#1262]
 
 
 2.3.0 (2022-07-07)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,9 +38,8 @@ pandas = ">=1.1.4"
 uncertainties = ">=3.1.4"
 patsy = ">=0.5.0"
 fbpca = ">=1.0"
-# tpf.interact_sky() tap selection requires a bug fix due to
-#  https://github.com/bokeh/bokeh/issues/6508
-bokeh = ">=1.1"
+# bokeh v2+ requirement due to https://github.com/lightkurve/lightkurve/issues/1261
+bokeh = ">=2.0.0"
 memoization = { version = ">=0.3.1", python = ">=3.8,<4.0" }
 scikit-learn = ">=0.24.0"
 
--- a/src/lightkurve/interact.py
+++ b/src/lightkurve/interact.py
@@ -234,7 +234,10 @@ def prepare_tpf_datasource(tpf, aperture_mask):
     xa = xa.flatten()
     ya = ya.flatten()
     tpf_source = ColumnDataSource(data=dict(xx=xa.astype(float), yy=ya.astype(float)))
-    tpf_source.selected.indices = aperture_mask_to_selected_indices(aperture_mask)
+    # convert the ndarray from aperture_mask_to_selected_indices() to plain list
+    # because bokeh v3.0.2 does not accept ndarray (and causes js error)
+    # see https://github.com/bokeh/bokeh/issues/12624
+    tpf_source.selected.indices = list(aperture_mask_to_selected_indices(aperture_mask))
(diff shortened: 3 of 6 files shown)


orionlee:
The `plot_height` / `plot_width` change fixed most of the problems. 

However, `tpf.interact()` still does not work. It returned nothing in a jupyter notebook. In developer console, the following error message is shown
```
Error rendering Bokeh items: Error: Selection(p2753).indices given invalid value: {"0":37,"1":38,"2":39,"3":48,"4":49,"5":50,"6":51,"7":59,"8":60,"9":61,"10":62,"11":70,"12":71,"13":72,"dtype":"int32","shape":[14],"dimension":1}
    at q.validate (VM60 bokeh-3.0.2.min.js:182:2127)
    at q._update (VM60 bokeh-3.0.2.min.js:182:1842)
    at q.initialize (VM60 bokeh-3.0.2.min.js:182:1480)
    at _.initialize_props (VM60 bokeh-3.0.2.min.js:179:4432)
    at y._decode_object_ref (VM60 bokeh-3.0.2.min.js:212:4715)
    at y._decode (VM60 bokeh-3.0.2.min.js:212:1709)
    at y._decode_plain_object (VM60 bokeh-3.0.2.min.js:212:2270)
    at y._decode (VM60 bokeh-3.0.2.min.js:212:1035)
    at y._decode_object_ref (VM60 bokeh-3.0.2.min.js:212:4684)
    at y._decode (VM60 bokeh-3.0.2.min.js:212:1709)
```

I suspect the pixel selection feature, unique in `tpf.interact()` and not found in other interact features, is the issue.

orionlee:
@barentsen 
1. I'd like to get PR merged soon, which will also resolve CI problem.
2. We might want to release a new version of lightkurve after this PR is resolved.

The impact of the bokeh issue is that for new lightkurve users, they could easily run into the problem if they try interactive features, because for new users, their new install would pick up bokeh v3 as dependency.

orionlee:
@barentsen I plan to merge this PR in the next few days. It's relatively low risk and will fix the CI problem.

Issue #1263, opened 2022-11-17
`lc.remove_outliers()` raised "ValueError: could not broadcast input array from shape ..." in some environments


Nschanche:
lc.estimate_cdpp fails with the error "ValueError: could not broadcast input array from shape (16781,) into shape ()". The problem is in the call to SigmaClip and then remove_outliers. A quick fix is to modify the remove_outliers call to sigma_clip by changing the data to self.flux.value (rather than self.flux)

orionlee:
I cannot reproduce the problem. Please include specific examples, preferably a reproducible snippet that illustrates the problem.

E.g., in the following snippet, it produces `34.706248ppm` for pi Men in TESS sector 1.

```
import lightkurve as lk

lc = lk.search_lightcurve("pi Men").download()
lc.estimate_cdpp()
```

cdevaneprugh:
I have experienced the same problem, specifically with premade light curves. It looks like the issue pops up when you try to use `.remove_outliers()` after using `.flatten()`. For example, if I try: 

```
import lightkurve as lk

lc = lk.search_lightcurve('KIC 6922244', author="Kepler", 
                          cadence="long", quarter=4).download()
lc = lc.flatten()
lc = lc.remove_outliers()
```

I get a value error that says:  `could not broadcast input array from shape (4116,) into shape ()`

However, if I download a tpf file, then convert it to a light curve (like below), there is no issue.
```
tpf = lk.search_targetpixelfile('KIC 6922244', author="Kepler", 
                                cadence="long", quarter=4).download()
lc = tpf.to_lightcurve(aperture_mask=tpf.pipeline_mask)
lc = lc.flatten()
lc = lc.remove_outliers()
```

I also fixed the issue in a similar way, by removing the astropy units from the flux array in `.remove_outliers()`. It looks like the issue popped up when Astropy updated to version 5.0

orionlee:
Hmm, I cannot reproduce the problem indicted using the snippets for KIC 6922244 . 
The code snippets work on Google colab.

Could it some kind of incompatibility issue between modules? E.g., recent lightkurve only works with astropy 5.0+

You could help to narrow down the problem by reporting the environment you use.

Environment
```python
import platform; print(platform.platform())
import sys; print("Python", sys.version)
import lightkurve as lk; print("lightkurve", lk.__version__)
import astropy; print("astropy", astropy.__version__)
import numpy as np; print("numpy", np.__version__)
```

Example output (on google colab):
[...]

PR #1272, opened 2023-01-18
fixing np.int and np.float

christinahedges:
[`numpy` deprecated the `np.int` and `np.float` features in v1.20](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations) and so we need to replace them, because as of version 1.24 it causes exceptions. I'm replacing them with `np.int_` and `np.float_` and checking if the tests pass. We might be better off replacing them with `int` and `float`.

Note: K2-199 was used in a test for PLD, but now that name returns no results from our search. I've updated the target name to the EPIC number for that target [EPIC 212779596](https://simbad.u-strasbg.fr/simbad/sim-basic?Ident=EPIC212779596).

### DIFF
--- a/src/lightkurve/collections.py
+++ b/src/lightkurve/collections.py
@@ -51,7 +51,7 @@ class Collection(object):
         return len(self.data)
 
     def __getitem__(self, index_or_mask):
-        if isinstance(index_or_mask, (int, np.integer)):
+        if isinstance(index_or_mask, (int, np.int_)):
             return self.data[index_or_mask]
         elif isinstance(index_or_mask, slice):
             return type(self)(self.data[index_or_mask])
--- a/src/lightkurve/interact.py
+++ b/src/lightkurve/interact.py
@@ -1248,7 +1248,7 @@ def show_interact_widget(
                     exported_filename,
                     overwrite=True,
                     flux_column_name="SAP_FLUX",
-                    aperture_mask=lc_new.meta["APERTURE_MASK"].astype(np.int),
+                    aperture_mask=lc_new.meta["APERTURE_MASK"].astype(np.int_),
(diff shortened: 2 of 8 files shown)


orionlee:
The remote data CI erors of encountering `TypeError` in `test_cdpp_tabby()` might have the same root cause as #1263  (of which I looked into, but could not reproduce it).

orionlee:
I took a look at the four remaining failed tests in CI.

```
FAILED tests/test_targetpixelfile.py::test_missing_pipeline_mask - lightkurve.search.SearchError:
```

This is unrelated to numpy 1.24 . There seems to be a behavior change in testcut. At least for the target in test, `cutout_size=1` now raised Error. Change to`cutout_size=3` would make the test pass.
https://github.com/lightkurve/lightkurve/blob/843f0b5f20356d0ccc49d4b4f07005770d9d4e19/tests/test_targetpixelfile.py#L806

---

```
FAILED tests/test_lightcurve.py::test_cdpp_tabby - TypeError: cannot write to unmasked output
FAILED tests/test_lightcurve.py::test_regression_346 - TypeError: cannot write to unmasked output
```

Seems to be an astropy issue with numpy 1.24, filed at https://github.com/astropy/astropy/issues/14360

---

```
FAILED tests/test_lightcurve.py::test_to_fits - ValueError: operands could not be broadcast together with shapes (3113,) (3113,11,11)
```

Unclear on the specific reason but it is related to numpy 1.24 . I ran a [CI that uses numpy 1.23.5](https://github.com/orionlee/lightkurve/actions/runs/4092792989/jobs/7058724823). All tests passed, except `test_missing_pipeline_mask()` (due to changes in tesscut behavior).

PR #1279, opened 2023-02-09
Numpyfix v2

Nschanche:
Modification of the pull request #1272 

[numpy deprecated the np.int and np.float features in v1.20](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations) and so we need to replace them, because as of version 1.24 it causes exceptions. 

The change to numpy's masked_invalid in v24 also affected the call to sigma_clip in lightcurve.py (also noted in issue #1263), so changed to specify self.flux.data. 

Also modified the targetpixelfile test. There seems to be a behavior change in tesscut so that cutout_size=1 or cutout_size=2 raises an error. Changed to cutout_size=3.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@
 
 - Fixed ``SearchResult`` HTML display for links to TESS GI cycles 3/4 proposals. [#1260]
 
+- Fixed references to np.int and np.float due to changes in numpy v1.24. [#1279]
+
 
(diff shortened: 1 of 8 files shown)


christinahedges (approved):
These changes look great to me, revert the pyproject.toml file back and I'll merge. Closes #1272.

christinahedges (approved):

PR #1280, opened 2023-02-11
Workaround sigma_clip bug with MaskedQuantity

orionlee:
A follow up on PR #1279 on the workaround for Astropy `sigma_clip()` issue with `MaskedQuantity` (https://github.com/astropy/astropy/issues/14360).

PR #1279 uses a workaround by providing `self.flux.data` instead of `self.flux` to `sigma_clip()`.

Two potential concerns:
1. `self.flux.data` is an undocumented attribute `.data` for [`Quantity`](https://docs.astropy.org/en/stable/api/astropy.units.quantity.Quantity.html) / [`MaskedQuantity`](https://docs.astropy.org/en/stable/utils/masked/index.html) of type `memoryview` (rather than some array-like type). While it works as-is, it could be error prone to internal changes in Astropy.
2. in the case flux is `MaskedQuantity` (the default type for TESS / Kepler readers), it effectively passed unmasked values to sigma_clip().
In common usage, it is not an issue because the masked values are just NaNs. However, if users deliberately mask specific non-NaN flux values, it'd produce incorrect result (without any error).

The fix here is to convert `MaskedQuantity` to the unmasked version, substituting masked values with NaN explicitly.

### DIFF
--- a/src/lightkurve/correctors/regressioncorrector.py
+++ b/src/lightkurve/correctors/regressioncorrector.py
@@ -7,6 +7,7 @@ import warnings
 from astropy.stats import sigma_clip
 from astropy import units as u
 from astropy.utils.exceptions import AstropyUserWarning
+from astropy.utils.masked import Masked
 import matplotlib.pyplot as plt
 import numpy as np
 from scipy.sparse import issparse, csr_matrix
@@ -255,10 +256,14 @@ class RegressionCorrector(Corrector):
             )
             model = u.Quantity(model, unit=self.lc.flux.unit)
             residuals = self.lc.flux - model
-            # workaround for https://github.com/astropy/astropy/issues/14360
-            # in passing MaskedQuantity to sigma_clip
-            from astropy.utils.masked import Masked
             if isinstance(residuals, Masked):
+                # Workaround for https://github.com/astropy/astropy/issues/14360
+                # in passing MaskedQuantity to sigma_clip, by converting it to Quantity.
+                # We explicitly fill masked values with `np.nan` here to ensure they are masked during sigma clipping.
+                # To handle unlikely edge case, convert int to float to ensure filing `np.nan` work.
+                # The conversion is acceptable because only the mask of the sigma_clip() result is used.
+                if np.issubdtype(residuals.dtype, np.int_):
+                    residuals = residuals.astype(float)
                 residuals = residuals.filled(np.nan)
             with warnings.catch_warnings():  # Ignore warnings due to NaNs
                 warnings.simplefilter("ignore", AstropyUserWarning)
--- a/src/lightkurve/correctors/regressioncorrector.py
+++ b/src/lightkurve/correctors/regressioncorrector.py
@@ -6,6 +6,7 @@ import warnings
 
 from astropy.stats import sigma_clip
 from astropy import units as u
+from astropy.utils.exceptions import AstropyUserWarning
 import matplotlib.pyplot as plt
 import numpy as np
 from scipy.sparse import issparse, csr_matrix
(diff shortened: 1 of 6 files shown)


orionlee:
I also reviewed the usage of `sigma_clip()` in the rest of the codebase.

https://github.com/lightkurve/lightkurve/blob/2c4722e1f85effc1c244b80c4d0f8a70abbe40ce/src/lightkurve/correctors/regressioncorrector.py#L257-L260

- The code here seems to try wo workaround `sigma_clip()` issues with `MaskedQuantity` as well. 
I think it should be replaced with the fix in this PR. Passing `.unmasked` to sigma_clip() could lead to incorrect result.


https://github.com/lightkurve/lightkurve/blob/d863a8ad4d8ed022cd1d5cf9bb4007d5161d3fb2/src/lightkurve/interact.py#L259

- No immediate change is needed here because the source of the flux is from `TargetPixelFile`, which uses `Quantity` only (not `MaskedQuantity`).

src/lightkurve/correctors/regressioncorrector.py
```diff
@@ -254,10 +255,14 @@ def correct(
             )
             model = u.Quantity(model, unit=self.lc.flux.unit)
             residuals = self.lc.flux - model
-            # In Astropy>=5.0, residuals will be a MaskedQuantity
-            if hasattr(residuals, 'mask'):
-                residuals = residuals.unmasked
-            self.outlier_mask |= sigma_clip(residuals, sigma=sigma).mask
+            # workaround for https://github.com/astropy/astropy/issues/14360
+            # in passing MaskedQuantity to sigma_clip
+            from astropy.utils.masked import Masked
```
christinahedges, line 260: Can we have this import at the top of the file? Is there a reason it is in the middle?

christinahedges (commented):
Hi @orionlee thanks for opening this PR and pointing out this situation. It looks like you've raise:

1. The `flux.data` attribute is not an array like, which could become problematic down the line
2. The `sigma_clip` method is not masking values that are not `np.nan` values, even when they are masked

To address both of these issues, you have taken the `flux` attribute and used the `filled` function to make all values that are masked `np.nan`. This looks like a good fix to me, my one concern is checking that this does not overwrite values in flux. 

i.e. that this:
```python
flux = self.flux
if isinstance(flux, Masked):
                flux = flux.filled(np.nan)
```
is equivalent to
```python
flux = self.flux.copy()
if isinstance(flux, Masked):
                flux = flux.filled(np.nan)
```

If this is the case I think this is a good fix and I will merge. I've left three very easy comments below about the import statement and adding comments.

Nschanche:
I agree the .data call is not very robust for changes down the line, so I agree this explicit substitution is a better way to handle it. I think it looks like a good PR.

orionlee:
`Masked.filled()` does not change the underlying data. I just verified it.

```python
flux = astropy.utils.masked.Masked([1., 2., 3.], [False, True, False])
print(flux)
print(flux.filled(np.nan))
print(flux)  # .filled() does not change the underlying data
```
```python
[1. —— 3.]
[ 1. nan  3.]
[1. —— 3.]
```

---
Additionally,
- I reworded the comment to make it clearer on the workaround (convert `MaskedQuantity` to `Quantity`, with missing values substituted by `np.nan`)
- handle the (extremely unlikely) case that the flux is `int` rather than `float`
- An unrelated test tweak is added because somehow MAST just changed its behavior (hopefully only temporarily) that searching for "Kepler-10" no longer return anything.  I substituted "Kepler-10" with the correspond KIC.

christinahedges:
This looks good to me now, I'm merging it in, thanks @orionlee for the great fixes!

PR #1281, opened 2023-02-12
Revert column to be Quantity for Kepler/TESS LCs, pre lk v2.1 / Astropy5 behavior

orionlee:
This PR is to make Quantity-like columns for TESS / Kepler lightcurves, e.g., flux, sap_flux, etc. to be `Quantity`, rather than `MaskedQuantity`.

It essentially reverts to the behavior of Lightkurve 2.0.x with Astropy 4.x.
The reason is to avoid various complication with `MaskedQuantity`, most recently errors involving astropy `sigma_clip()` with numpy 1.24

### DIFF
--- a/src/lightkurve/io/generic.py
+++ b/src/lightkurve/io/generic.py
@@ -3,7 +3,7 @@ import logging
 import warnings
 
 from astropy.io import fits
-from astropy.table import Table
+from astropy.table import MaskedColumn, Table
 from astropy.time import Time
 from astropy.units import UnitsWarning
 import numpy as np
@@ -47,6 +47,17 @@ def read_generic_lightcurve(
         # out to be addressed at the archive level. (See issue #1216.)
         warnings.simplefilter("ignore", category=UnitsWarning)
         tab = Table.read(hdulist[ext], format="fits")
+
+    # Turn any MaskedColumn into regular Column. Later they will be turned
+    # to regular Quantity.
+    # This is done to avoid the use of MaskedQuantity.
+    #
+    # The logic mimics Astropy's built-in Kepler / TESS reader:
+    # https://github.com/astropy/astropy/blob/6fda4d5dc40e94e8944c2d23ba5a7004be330479/astropy/timeseries/io/kepler.py#L79
+    for colname in tab.colnames:
+        unit = tab[colname].unit
+        if unit and isinstance(tab[colname], MaskedColumn):
+            tab[colname] = tab[colname].filled(np.nan)
 
     # Make sure the meta data also includes header fields from extension #0
     tab.meta.update(hdulist[0].header)


christinahedges:
Hi @orionlee thanks for bringing this up! In this PR, if a user were to input mask data e.g. because of a flare or a cosmic ray, this data would be over-written with a `np.nan` value. Users may find that this isn't what they want, e.g. they may want a data value temporarily masked for some analysis. 

What do you think?

orionlee:
@christinahedges Could you elaborate on the temporary mask use case you mentioned? 

In my mind, the existing boolean indexing works just fine without using `MaskedQuantity`:
```python
mask = some_function_to_define_cosmic_ray_mask(lc)
some_analysis_function(lc[~mask])  # temporary mask out cadences with cosmic ray
```

Using `MaskedQuantity` is arguably more cumbersome, e.g.,
```python
# temporarily mask out flux at index 123
lc_tmp = lc.copy()
lc_tmp.flux.mask[123] = False
# if sap_flux is also used by the analysis, need to mask it too
lc_tmp.sap_flux.mask[123] = False
some_analysis_function(lc_tmp)
```

The change in this PR is only concerned with reading TESS / Kepler FITS file. 
- The current behavior is to replace nan values with mask.
- This PR exposes the underlying nan values (lightkurve 2.0.x behavior)
It has nothing direct bearing on user temporarily masking out specific cadence / values.

------

Irrespective on the merits of using `MaskedQuantity`
1. This PR does not stop users from using `MaskedQuantity`, but it is no longer the de-facto column type for TESS / Kepler LCs. If users have specific needs to use `MaskedQuantity`, they can still explicitly replace a column (of type `Quantity`) with the `MaskedQuantity` counterpart.
2. If it's decided that we do want to retain `MaskedQuantity` as the column type for TESS / Kepler LCs, I think we should:
  a. document it somewhere
  b. ensure we do always return `MaskedQuantity`. The current behavior is dependent on the contents of the specific FITS file: if a a column contains `nan`, the column will be `MaskedQuantity`, otherwise, it'd be regular `Quantity`. 
In other words, lightkurve does not guarantee a column would be `MaskedQuantity`. Users cannot exactly assume it to be the case.  `MaskedQuantity` becomes de-facto type for Kepler/TESS LCs because in most cases, columns have `nan` values.

christinahedges:
Hi @orionlee thanks for the explanation, when you say 

> The current behavior is to replace nan values with mask.

what do you mean here?

There is also a comment in the `astropy` Kepler file reader that [this is a workaround they want to get rid of when masked quantities are supported better](https://github.com/astropy/astropy/blob/6fda4d5dc40e94e8944c2d23ba5a7004be330479/astropy/timeseries/io/kepler.py#L78). Given that, do you think this is still a good fix to implement in lightkurve?

Issue #1277, opened 2023-02-01
Inconsistency identifying TESS data packages


garyhawk:
I see an inconsistency in identifying available data packages.  For example, if you identify EV Lac as 'EV Lac' I get a different number of data packages than if you identify it by HIP 112460. In LightKurve EV Lac only identifies one package, while HIP 11240 yields seven packages:

0TESS Sector 162019[SPOC](https://heasarc.gsfc.nasa.gov/docs/tess/pipeline.html)1201541016780.0
1TESS Sector 162019[TESS-SPOC](https://archive.stsci.edu/hlsp/tess-spoc)18001541016780.0
2TESS Sector 162019[QLP](https://archive.stsci.edu/hlsp/qlp)18001541016780.0
3TESS Sector 562022[SPOC](https://heasarc.gsfc.nasa.gov/docs/tess/pipeline.html)201541016780.0
4TESS Sector 562022[SPOC](https://heasarc.gsfc.nasa.gov/docs/tess/pipeline.html)1201541016780.0
5TESS Sector 572022[SPOC](https://heasarc.gsfc.nasa.gov/docs/tess/pipeline.html)201541016780.0
6TESS Sector 572022[SPOC](https://heasarc.gsfc.nasa.gov/docs/tess/pipeline.html)1201541016780.0  

I've seen a similar issue with other stars as well and other query packages, like Peranso. HIP and HD numbers seem to work better than constellation names.

Any thoughts?

garyhawk:
Pinging this thread.

christinahedges:
Hi @garyhawk thanks for getting in touch. I'm forwarding this to the TESS GI helpdesk as this is a question about how the archive is searched.

christinahedges:
The TESS GI team took a look at your question and we have some answers for you. @tylerapritchard was the one to investigate and find you a work around!

@garyhawk it looks like this object is a high proper motion target. You can see [here](http://simbad.u-strasbg.fr/simbad/sim-basic?Ident=EV+Lac&submit=SIMBAD+search) that there is significant proper motion. `lightkurve` relies on `astroquery`, and `astroquery` is searching the holdings at MAST. Because this a high proper motion target, the coordinates returned for the `astroquery` resolver (which searches MAST) seem to be different depending on the name used, see below. 

[...]

PR #1282, opened 2023-02-14
fix test for search lc by name

orionlee:
A further teak on the test fix in fc0cc0bcb81dd89d45f3631ecc45558cb2f46c66 for MAST's Kepler-10 issue.

One of the tests is to test search lc by name, so the fix of using KIC ID inadvertently rendered the test useless. The PR uses a different name.

P.S. Is the Kepler-10 issue someone in MAST (or the underlying name resolution service) should be contacted?  It seems to be specific to Kepler-10 only. I tried 
1. a few other Kepler names (e.g, Kepler-4). They still all work.
2. increase search radius to 2arcmin (`lk.search_lightcurve("Kepler-10", mission="Kepler", radius=120)`) in case somehow the cooridinate of the resolved Kepler-10 has shifted slightly. It still returns nothing.

### DIFF
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -103,8 +103,10 @@ def test_search_split_campaigns():
 @pytest.mark.remote_data
 def test_search_lightcurve(caplog):
     # We should also be able to resolve it by its name instead of KIC ID
+    # The name Kepler-10 somehow no longer works on MAST. So we use 2MASS instead:
+    #   https://simbad.cds.unistra.fr/simbad/sim-id?Ident=%405506010&Name=Kepler-10
     assert (
-        len(search_lightcurve("KIC 11904151", mission="Kepler", cadence="long").table)
+        len(search_lightcurve("2MASS J19024305+5014286", mission="Kepler", cadence="long").table)
         == 15
     )
     # An invalid KIC/EPIC ID or target name should be dealt with gracefully


christinahedges:
Hi @orionlee this looks great, and you're totally right about this defeating the purpose of the test. I do think we might need to get in touch with MAST and understand what's going on here with the name resolving. It's interesting because #1277 also seems to be having problems with the name resolver.

scfleming:
We were having problems with our name resolver cache using bad values from one of the NASA archives.  We think we've fixed the issues, but if any names are STILL not resolving to correct coordinates, please contact us through the MAST help desk email.

christinahedges:
Hi @scfleming thanks for commenting, I responded to #1277 with a temporary fix but it does seem that the astroquery position for EV Lac (a high proper motion target) isn't correct. Here's an example:
Screen Shot 2023-03-21 at 5 23 28 PM

It may be that this is what's occasionally tripping us up with Kepler-10:
Screen Shot 2023-03-21 at 5 21 18 PM

christinahedges:
I opened a helpdesk ticket at MAST about this. For now, this is a great PR, and fixes a test that is currently not testing the right functionality. I'm going to merge this, and will see what MAST say from the helpdesk ticket.

PR #1289, opened 2023-03-22
Fixing pysphinx docs version

christinahedges:
Fixing the `pyproject.toml` file to have the correct `pydata-sphinx-theme = "==0.8.1"` in the correct place.

### DIFF
--- a/.github/workflows/lightkurve-tests.yml
+++ b/.github/workflows/lightkurve-tests.yml
@@ -43,6 +43,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip poetry
+        poetry config installer.modern-installation false
         poetry install
     - name: Test with ${{ matrix.pytest-command }}
       run: |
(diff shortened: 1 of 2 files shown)


christinahedges:
It looks like poetry updated to 1.4.1 this week, and it's causing some issues with continuous integration (e.g. see [here](https://github.com/microsoft/debugpy/issues/1246)). I reverted the PRs I merged this morning (#1282 and #1284) and I'm going to put in some fix for poetry, then I will merge them back in. The tests for those PRs were run before the new poetry version, and so looked good.

christinahedges:
It looks like as of v1.4.0 `poetry` has added a new installation method, see [here](https://python-poetry.org/docs/configuration/#installermodern-installation). This seems to not play nicely with our `pydata-sphinx-theme` installation, and so turning it off in configuration ensures the tests pass. Setting `installer.modern-installation` to `false` should give the same behavior from pre `v1.4.0`. 

Because this configuration setting only exists in `v1.4.0` or higher, I'm explicitly installing poetry in our github actions as 

```
python -m pip install --upgrade pip 'poetry>=1.4.0'
poetry config installer.modern-installation false
```

christinahedges:
It seems like we're getting a ton of timeout messages from our remote tests.

Issue #1286, opened 2023-03-12
LightkurveError: Not recognized as a supported data product


diyiliaoya:
#### Example

```python
import lightkurve as lk 
lcs = lk.search_lightcurve('TIC140898436', exptime=1800,mission='TESS').download_all()
# insert code here ...
```

#### Expected behavior

LightkurveError: Not recognized as a supported data product:
/home/lishijie/.lightkurve/cache/mastDownload/HLSP/hlsp_tglc_tess_ffi_gaiaid-4624742688944261376-s0001-cam4-ccd2_tess_v1_llc/hlsp_tglc_tess_ffi_gaiaid-4624742688944261376-s0001-cam4-ccd2_tess_v1_llc.fits
This file may be corrupt due to an interrupted download. Please remove it from your disk and try again.

#### Environment

-  Linux
-  lightkurve 2.4.0
-  pip

gully:
Hi there, it's possible you may have attempted to pre-downloaded this file before and the query is retrieving a corrupted file.  What happens if you remove the file and attempt to re-run the query?

`home/lishijie/.lightkurve/cache/mastDownload/HLSP/hlsp_tglc_tess_ffi_gaiaid-4624742688944261376-s0001-cam4-ccd2_tess_v1_llc/hlsp_tglc_tess_ffi_gaiaid-4624742688944261376-s0001-cam4-ccd2_tess_v1_llc.fits`

diyiliaoya:
> Hi there, it's possible you may have attempted to pre-downloaded this file before and the query is retrieving a corrupted file. What happens if you remove the file and attempt to re-run the query?
> 
> `home/lishijie/.lightkurve/cache/mastDownload/HLSP/hlsp_tglc_tess_ffi_gaiaid-4624742688944261376-s0001-cam4-ccd2_tess_v1_llc/hlsp_tglc_tess_ffi_gaiaid-4624742688944261376-s0001-cam4-ccd2_tess_v1_llc.fits`

It's still the same error after I remove this file.

christinahedges:
Hi @diyiliaoya thanks for getting in touch. I can't reproduce this error unfortunately. I believe this could be to do with the migrating cache from #1214. On importing lightkurve I believe you will get a warning message stating:

[...]

PR #1291, opened 2023-03-22
added TGLC reader for new HLSP

christinahedges:
#1286 has pointed out that the new [TGLC HLSP](https://archive.stsci.edu/hlsp/tglc) can not be read in by `Lightkurve`, and will break the `download_all` function if there is a TGLC light curve in the search result. This PR adds the functionality to read in the `TGLC` light curves and checks they do not break searches. 

Note TGLC does not provide errors, but this is the case for other HLSPs we allow (e.g. K2SFF)

I've use the `cal_psf_flux` as the default flux here, as this seems to be their key data product. They also provide calibrated aperture flux as an option which can be obtained using
`lc.select_flux('cal_psf_flux')`

### DIFF
--- a/src/lightkurve/io/tglc.py
+++ b/src/lightkurve/io/tglc.py
@@ -19,9 +19,9 @@ def read_tglc_lightcurve(
     provided by this HLSP.
 
     Note this reader does not use the TGLC_FLAG extension to inform the bitmask for the
-    returned light curve, but can still be accessed.
+    returned light curve, but those flags can still be accessed.
 
-    More information on eleanor: https://archive.stsci.edu/hlsp/tglc
+    More information on TGLC: https://archive.stsci.edu/hlsp/tglc
 
     Parameters
     ----------
--- a/src/lightkurve/io/tglc.py
+++ b/src/lightkurve/io/tglc.py
@@ -1,0 +1,96 @@
+"""Reader for TGLC light curve files.
+Details can be found at https://archive.stsci.edu/hlsp/tglc
+"""
+import numpy as np
+from astropy import units as u
+
+from ..lightcurve import TessLightCurve
+from ..utils import TessQualityFlags
+from .generic import read_generic_lightcurve
+
+
+def read_tglc_lightcurve(
+    filename, flux_column="cal_psf_flux", quality_bitmask="default"
+):
+    """Returns a `~lightkurve.lightcurve.LightCurve` object given a light curve file from
+    TGLC HLSP
+
+    By default, TGLC's `cal_psf_flux` values are used for the `flux` column. No errors are
+    provided by this HLSP.
+
+    Note this reader does not use the TGLC_FLAG extension to inform the bitmask for the
+    returned light curve, but can still be accessed.
+
+    More information on eleanor: https://archive.stsci.edu/hlsp/tglc
+
+    Parameters
+    ----------
+    filename : str
+        Local path or remote url of a TGLC light curve FITS file.
+    flux_column : 'CAL_PSF_FLUX', 'CAL_APER_FLUX', 'PSF_FLUX', or 'APERTURE_FLUX'
+        Which column in the FITS file contains the preferred flux data?
+        By default the "Corrected PSF Flux" flux (CAL_PSF_FLUX) is used.
+    quality_bitmask : str or int
+        Bitmask (integer) which identifies the quality flag bitmask that should
+        be used to mask out bad cadences. If a string is passed, it has the
+        following meaning:
+            * "none": no cadences will be ignored (`quality_bitmask=0`).
+            * "default": cadences with flags indicating AttitudeTweak, SafeMode, CoarsePoint, EarthPoint, Desat, or
+              ManualExclude will be ignored.
+            * "hard": cadences with default flags, ApertureCosmic, CollateralCosmic, Straylight, or Straylight2 will be
+              ignored.
+            * "hardest": cadences with all the above flags will be ignored, in addition to cadences with GSFC-ELEANOR-LITE
+              bit flags of 17 (decimal value 131072) and 18 (decimal value 262144).
+    """
+
+    lc = read_generic_lightcurve(
+        filename,
+        time_column="time",
+        flux_column=flux_column.lower(),
+        quality_column="tess_flags",
+        cadenceno_column="cadence_num",
+        time_format="btjd",
+    )
+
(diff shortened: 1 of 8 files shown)


christinahedges:
Closes #1286

src/lightkurve/io/tglc.py
```diff
@@ -0,0 +1,96 @@
...
+    quality_mask = TessQualityFlags.create_quality_mask(
+        quality_array=lc["quality"], bitmask=quality_bitmask
+    )
+
+    # TGLC FITS file do not have units specified. re-add them.
+    for colname in ["psf_flux", "aperture_flux", "background"]:
+        if colname in lc.colnames:
+            if lc[colname].unit is not None:
+                # for case flux, flux_err, lightkurve has forced it to be u.dimensionless_unscaled
+                # can't reset a unit, so we create a new column
```
tylerapritchard, line 64: Is there any possibility that lc[colname].unit != u.dimensionless_unscaled ?  Should this be checked explicitly so a different unit value is not overwritten? 

The most common case I could imagine would just be re-writing a electron/s value which shouldn't matter if someone is reading in/saving the same file, but could imagine unintended consequences otherwise.

tests/io/test_tglc.py
```diff
@@ -0,0 +1,38 @@
...
+    with fits.open(url, mode="readonly") as hdulist:
+        # Can we auto-detect a GSFC-ELEANOR-LITE file?
+        assert detect_filetype(hdulist) == "TGLC"
+        # Are the correct fluxes read in?
+        lc = read_tglc_lightcurve(url, quality_bitmask=0)
+        assert lc.meta["AUTHOR"] == "TGLC"
+        assert lc.meta["FLUX_ORIGIN"] == "cal_psf_flux"
+        assert_array_equal(lc.flux.value, hdulist[1].data["cal_psf_flux"])
+        assert np.issubdtype(lc["cadenceno"].dtype, np.integer)
+
```
tylerapritchard, line 24: Should we test here: 
That flux units are correct? 
That TIC/TargetID/Object/Label are set as expected?
That any simple lightcurve features work with with HLSP as expected?

tylerapritchard (commented):
Looks good to me in general

tests/io/test_tglc.py
tylerapritchard: Good point - definitely agree, we should definitely stay consistent re: other HLSP testing.

tylerapritchard:
In general I copied over the branch and opened up a tglc file and things seemed to work.  

It looks like there is an issue saving a tglc hlsp lightcurve file using lightkurve.io.to_fits() that was read in from the tglc hlsp.  The TGLC hlsp lacks the centroid_col & centroid_row columns which leads to to_fits() breaking.  Adding the columns resolves the issue. 

image

going to approve the PR as this is probably more of a general hlsp/to_fits question than tglc specific

src/lightkurve/io/tglc.py
tylerapritchard: I'm happy with this -  my concern had been an edge case - if a user is reading in a TGLC lightcurve file -> modifying/saving it -> later re-reading the saved lightcurve that the units might get overwritten (e.g. if a user is doing a count-to-flux conversion).  However, after checking, it looks like to_fits strips the AUTHOR keyword so these values are preserved.  

This did lead to a concern regarding lightkurve.io.to_fits() and hlsp's without centroid coordinates, see end comment.

src/lightkurve/io/tglc.py
```diff
@@ -0,0 +1,96 @@
...
+    """Returns a `~lightkurve.lightcurve.LightCurve` object given a light curve file from
+    TGLC HLSP
+
+    By default, TGLC's `cal_psf_flux` values are used for the `flux` column. No errors are
+    provided by this HLSP.
+
+    Note this reader does not use the TGLC_FLAG extension to inform the bitmask for the
+    returned light curve, but can still be accessed.
+
+    More information on eleanor: https://archive.stsci.edu/hlsp/tglc
```
Nschanche, line 24: typo: replace eleanor with tglc

src/lightkurve/io/tglc.py
```diff
@@ -0,0 +1,96 @@
...
+    filename, flux_column="cal_psf_flux", quality_bitmask="default"
+):
+    """Returns a `~lightkurve.lightcurve.LightCurve` object given a light curve file from
+    TGLC HLSP
+
+    By default, TGLC's `cal_psf_flux` values are used for the `flux` column. No errors are
+    provided by this HLSP.
+
+    Note this reader does not use the TGLC_FLAG extension to inform the bitmask for the
+    returned light curve, but can still be accessed.
```
Nschanche, line 22: grammar: but <the extension> can still be accessed

src/lightkurve/io/tglc.py
```diff
@@ -0,0 +1,96 @@
...
+    quality_bitmask : str or int
+        Bitmask (integer) which identifies the quality flag bitmask that should
+        be used to mask out bad cadences. If a string is passed, it has the
+        following meaning:
+            * "none": no cadences will be ignored (`quality_bitmask=0`).
+            * "default": cadences with flags indicating AttitudeTweak, SafeMode, CoarsePoint, EarthPoint, Desat, or
+              ManualExclude will be ignored.
+            * "hard": cadences with default flags, ApertureCosmic, CollateralCosmic, Straylight, or Straylight2 will be
+              ignored.
+            * "hardest": cadences with all the above flags will be ignored, in addition to cadences with GSFC-ELEANOR-LITE
```
Nschanche, line 42: remove the cadences with gsvc-Eleanor-lite, I think

tests/io/test_tglc.py
```diff
@@ -0,0 +1,38 @@
...
+from astropy.io import fits
+from numpy.testing import assert_array_equal
+
+from lightkurve import search_lightcurve
+from lightkurve.io.detect import detect_filetype
+from lightkurve.io.tglc import read_tglc_lightcurve
+
+
+@pytest.mark.remote_data
+def test_gsfc_eleanor_lite():
```
Nschanche, line 12: change name of function to test_tglc? Also would change the comment after the first assertion.

christinahedges:
> It looks like there is an issue saving a tglc hlsp lightcurve file using lightkurve.io.to_fits() that was read in from the tglc hlsp. The TGLC hlsp lacks the centroid_col & centroid_row columns which leads to to_fits() breaking. Adding the columns resolves the issue.

Issue #1095, opened 2021-06-30
Incorrect World Coordinate System values


vvilhelmus:
#### Problem description

Any Kepler star observed in two TESS sectors are not represented with the correct RA/Dec coordinates. Both given coordinates are significantly off from the correct coordinates.

#### Example

```python
# insert code here ...
import lightkurve as lk
import matplotlib.pyplot as plt
import matplotlib.gridspec as gs
import numpy as np

#download the files
search_result = lk.search_tesscut('Kepler-89')
s14tpf = search_result[0].download(cutout_size=11)
s15tpf = search_result[1].download(cutout_size=11)

#generate the masks
mymask = np.zeros((s14tpf[0].shape[1:]), dtype='bool')
mymask[5:7,5:7] = True
s14bgmask = ~s14tpf.create_threshold_mask(threshold=0.001, reference_pixel=None)
s14bgmask[4:8, 4:8] = False
s15bgmask = ~s15tpf.create_threshold_mask(threshold=0.001, reference_pixel=None)
s15bgmask[4:8, 4:8] = False

#create the figure
totalfig = plt.figure(figsize=(15, 12))
grid = gs.GridSpec(2,2)

# Plot custom aperture mask sector 14
ax1 = plt.subplot(grid[0, 0], projection=s14tpf.wcs)
s14tpf.plot(ax=ax1, title='Custom aperture mask', xlabel='Right Ascension', ylabel='Declination', aperture_mask=mymask, mask_color='w')

# Plot background aperture mask sector 14
ax2 = plt.subplot(grid[0, 1], projection=s14tpf.wcs)
s14tpf.plot(ax=ax2, title='Background aperture mask', xlabel='Right Ascension', ylabel='Declination',  aperture_mask=s14bgmask)

# Plot custom aperture mask sector 15
ax3 = plt.subplot(grid[1, 0], projection=s15tpf.wcs)
s15tpf.plot(ax=ax3, title='Custom aperture mask', xlabel='Right Ascension', ylabel='Declination', aperture_mask=mymask, mask_color='w')

# Plot background aperture mask sector 15
ax4 = plt.subplot(grid[1, 1], projection=s15tpf.wcs)
s15tpf.plot(ax=ax4, title='Background aperture mask', xlabel='Right Ascension', ylabel='Declination',  aperture_mask=s15bgmask)

totalfig.text(0.5, 0.91, 'Kepler-89 in Sector 14 cutout', ha='center', size=18)
[...]

PR #1298, opened 2023-04-17
WCS plotting

Nschanche:
This addresses lightkurve issue #1095. When providing an axes with WCS coordinates, TPFs were plotted incorrectly. This is because the img_extent keyword, which was set to display the appropriate row and pixel values, also modified the plotted WCS coordinates. The displayed coordinates no longer were correct for the target star. The aperture mask plotting also needed to be changed to account for whether or not the column/row values were used for the axes.

### DIFF
--- a/src/lightkurve/targetpixelfile.py
+++ b/src/lightkurve/targetpixelfile.py
@@ -1155,8 +1155,8 @@ class TargetPixelFile(object):
             #If an axes is passed that used WCS projection, don't use img_extent
             #This addresses lk issue #1095, where the tpf was incorrectly plotted
             try: 
-            	if hasattr(ax, "wcs"):
-            		img_extent = None
+                if hasattr(ax, "wcs"):
+                    img_extent = None
             except NameError:
             	pass
 
--- a/src/lightkurve/targetpixelfile.py
+++ b/src/lightkurve/targetpixelfile.py
@@ -1172,8 +1172,7 @@ class TargetPixelFile(object):
             )
             ax.grid(False)
             
-            #if not hasattr(ax, "wcs"):
-            #Make sure the ticks show integer values for col/row
+            # Make sure the ticks show integer values for col/row
             ax.yaxis.get_major_locator().set_params(integer=True) 
             ax.xaxis.get_major_locator().set_params(integer=True) 
 
@@ -1184,7 +1183,7 @@ class TargetPixelFile(object):
                 for j in range(self.shape[2]):
                     if aperture_mask[i, j]:
                         if hasattr(ax, "wcs"):
-                            #When using WCS coordinates, do not add col/row to mask coords
+                            # When using WCS coordinates, do not add col/row to mask coords
                     	    xy = (j - 0.5, i - 0.5)
                         else:
                     	    xy = (j + self.column - 0.5, i + self.row - 0.5)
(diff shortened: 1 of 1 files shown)


rebekah9969 (commented):
This looks like a great fix. Nice catch with looking for attributes.

src/lightkurve/targetpixelfile.py
```diff
@@ -1150,6 +1151,16 @@ def plot(
                 self.row - 0.5,
                 self.row + self.shape[1] - 0.5,
             )
+            
+            #If an axes is passed that used WCS projection, don't use img_extent
+            #This addresses lk issue #1095, where the tpf was incorrectly plotted
+            try: 
+            	if hasattr(ax, "wcs"):
+            		img_extent = None
```
christinahedges, line 1159: I think there is a double tab here

src/lightkurve/targetpixelfile.py
```diff
@@ -1160,15 +1171,25 @@ def plot(
                 **kwargs,
             )
             ax.grid(False)
+            
+            #if not hasattr(ax, "wcs"):
```
christinahedges, line 1175: Not needed?

src/lightkurve/targetpixelfile.py
```diff
@@ -1160,15 +1171,25 @@ def plot(
...
+            #Make sure the ticks show integer values for col/row
```
christinahedges, line 1176:
```suggestion
            # Make sure the ticks show integer values for col/row
```

src/lightkurve/targetpixelfile.py
```diff
@@ -1160,15 +1171,25 @@ def plot(
...
+            ax.xaxis.get_major_locator().set_params(integer=True) 
 
         # Overlay the aperture mask if given
         if aperture_mask is not None:
             aperture_mask = self._parse_aperture_mask(aperture_mask)
             for i in range(self.shape[1]):
                 for j in range(self.shape[2]):
                     if aperture_mask[i, j]:
+                        if hasattr(ax, "wcs"):
+                            #When using WCS coordinates, do not add col/row to mask coords
```
christinahedges, line 1187:
```suggestion
                            # When using WCS coordinates, do not add col/row to mask coords
```

src/lightkurve/targetpixelfile.py
christinahedges: Nice comment here

src/lightkurve/targetpixelfile.py
Nschanche: I don't know why it looks like that here. I checked and it is a single tab.

christinahedges:
@Nschanche can you add a demo to this issue with a regular plot and then a WCS plot

Nschanche:
@christinahedges this plot shows the same target in two subsequent sectors. In this case, the plot figure and axes are first instantiated with WCS coordinates `ax1 = plt.subplot(grid[0, 0], projection=s14tpf.wcs)` before plotting the TPF on the predefined axis `s14tpf.plot(ax=ax1)` 

The result of the previous lightkurve code seen here, where the coordinates do not match each other or the known coordinates of the target.  

WCScoord_issue

The proposed fix makes the following plots. The target now has the same coordinates in both sectors that match the known coordinates for the star. 

WCScoord_fixed

christinahedges:
@Nschanche nice work! This looks good to me I will merge!

PR #1299, opened 2023-04-18
Ensure FITS filestream is closed on generic read

bmorris3:
The generic light curve reader opens FITS files to determine how to parse the contents. The call to `fits.open` leaves a file stream open, which can prevent later and/or downstream calls from opening that same file. This workaround ensures the filestream is closed, and simply makes a copy of the list `hdulist`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,3 +79,6 @@ testpaths = [
     "tests",
     "src",
 ]
+filterwarnings = [
+    "error::ResourceWarning",
+]
--- a/src/lightkurve/io/generic.py
+++ b/src/lightkurve/io/generic.py
@@ -1,6 +1,7 @@
 """Read a generic FITS table containing a light curve."""
 import logging
 import warnings
+from copy import deepcopy
 
 from astropy.io import fits
 from astropy.table import Table
@@ -34,7 +35,8 @@ def read_generic_lightcurve(
     if isinstance(filename, fits.HDUList):
         hdulist = filename  # Allow HDUList to be passed
     else:
-        hdulist = fits.open(filename)
+        with fits.open(filename) as hdulist:
+            hdulist = deepcopy(hdulist)
 
     # Raise an exception if the requested extension is invalid
     if isinstance(ext, str):
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -615,7 +615,7 @@ class LightCurve(TimeSeries):
     )
     def hdu(self):
         with fits.open(self.filename) as hdulist:
-            hdulist = deepcopy(hdulist)
+            hdulist = hdulist.copy()
         return hdulist
 
     @property
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -614,7 +614,9 @@ class LightCurve(TimeSeries):
         warning_type=LightkurveDeprecationWarning,
     )
     def hdu(self):
-        return fits.open(self.filename)
+        with fits.open(self.filename) as hdulist:
+            hdulist = deepcopy(hdulist)
+        return hdulist
 
     @property
     @deprecated("2.0", warning_type=LightkurveDeprecationWarning)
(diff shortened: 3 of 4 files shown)


bmorris3:
With further testing, it looks like this particular failure be the result of a mismatch between the astropy and lightkurve versions. If I update to astropy dev, no fix is needed. I'll keep investigating.

Nschanche:
This may not work as a change. I think the data stream needs to remain open as we still need access to the data. 
For example, line 49 calls `Table.read(hdulist[ext], format="fits")`. With the proposed change, this fails with the error, "ValueError: I/O operation on closed file".

christinahedges:
@Nschanche @bmorris3 there are parts of the `LightCurve` and `TargetPixelFile` classes that rely on these files being open, as @Nschanche points out.

We could change the functionality so that the `hdu` attribute always reopens the file, rather than leaving it open. @Nschanche what do you think?

bmorris3:
I was wrong that my implementation was making a copy of the `hdulist`, which I've corrected in the latest commit ([e50c2e2](https://github.com/lightkurve/lightkurve/pull/1299/commits/e50c2e213404b4c6d16c7e93218f2577fadb6214)). 

I see the `hdu` attribute challenge in the LC/TPF classes. I'll make another commit in a minute with a similar fix there. It doesn't look like an _open_ file stream is ever needed, since lightkurve doesn't modify the original file on disk, only makes copies.

bmorris3:
I think in [ab001e7](https://github.com/lightkurve/lightkurve/pull/1299/commits/ab001e76c0930ce9e15855a8fca9eea162031e44) I've closed the remaining open file streams. If https://github.com/lightkurve/lightkurve/pull/1299#issuecomment-1514931927 isn't correct and you do need open file streams anywhere, I'm happy to revise again.

Nschanche:
Yes, I agree they don't need to be open, the data just needs to be accessible. The deep copy you added looks good to me.

bmorris3:
Thanks @Nschanche! Let me know if there's anything else you'd like added @christinahedges.

christinahedges:
Hey @bmorris3 I am a little confused on this PR and I want to be sure I understand, can you post a comment here with an example showing what the issue is you're encountering, and why you're suggesting this is a fix?

dfm:
I'm finding that this PR introduces one regression (something to do with pickling) that I'll try to track down, but otherwise I think this should be merged to catch memory leaks!

bmorris3:
@christinahedges I opened this PR because if you `read` a local FITS file with lightkurve, it keeps the file stream open. If other packages/scripts read that same FITS file after lightkurve has opened but not closed the file stream, they can crash. 

This exact problem was occurring specifically in 
https://github.com/spacetelescope/lcviz/pull/10, which caused two tests to fail. We opted to merge that PR after commenting out the failing tests, with the plan to uncomment them when this lightkurve PR is merged.

dfm (approved):

bmorris3:
@christinahedges Does that sound ok?

christinahedges:
This looks good to me, I think @dfm is right that we need to eventually have a better solution for carrying around this fits file.

orionlee:
@bmorris3 What are the use cases that leaving the file stream open become a problem?

My understanding is that normally, once a `LightCurve` object is longer in use, the file stream will be closed.

The workaround here, with `deepcopy(hdulist)`, causes memory leak in typical (if not all) cases. See https://github.com/lightkurve/lightkurve/issues/1388#issuecomment-1817683109.

---

One scenario that leaving the file stream open would cause problem is in opening corrupted FITS file on Windows platform. For that specific case, another solution can be used (ensure the generic reader close `hdulist` upon any exception during read).

bmorris3:
@orionlee The problematic case is explained here: https://github.com/lightkurve/lightkurve/pull/1299#issuecomment-1530107904.

orionlee:
@bmorris3 Just to be sure, are you referring to the test failures in https://github.com/lightkurve/lightkurve/pull/1299#issuecomment-1530107904 , specifically, the `ResourceWarning` about unclosed file? E.g.

```python
78: PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO [closed]>

    File "C:\dev\lightkurve\src\lightkurve\io\kepler.py", line 34, in read_kepler_lightcurve     
      lc = read_generic_lightcurve(
  ResourceWarning: unclosed file <_io.BufferedReader name='C:\\Users\\SL\\.astropy\\cache\\download\\url\\0dd9ac829437a6265e20d71bc15c9e18\\contents'>
```

PR #1307, opened 2023-05-01
Improving periodogram.flatten() for SNRPeriodogram purposes.

ojhall94:
This work in progress pull request is making a number of changes to the Periodogram module, in order to improve the functionality of the .flatten() function. List of changes made and to be made:

- [ ] Copy `Periodogram.flatten()` to `LombScarglePeriodogram`. **Reasoning**: the `BLSPeriodogram` has its own `flatten()` function, and `Periodogram.flatten()` was only intended for use on `LombScarglePeriodograms`. `Periodogram.flatten()` should be deprecated.
- [ ] Add a `diagnose_flattening()` function to the `SNRPeriodogram`. **Reasoning**: there is little transparency on what the `Periodogram.flatten()` function does to the periodogram power, unless the user diagnoses this themselves. This makes that process more straightforward.
- [ ] Make a more intelligent `LombScarglePeriodogram.flatten()` function that accounts for whether power excess is at low or high frequency, so that it requires less trial and error by users.
- [ ] Update the unit tests to account for the new features and changed locations of the `.flatten()` function.
- [ ] Add a deprecation warning to the `Periodogram.flatten()` function.

Discussion points:
- Should we keep `Periodogram.flatten()` in case users load in their own periodograms? Or should we encourage users to read those in as a `LombScarglePeridogram`?

### DIFF
--- a/src/lightkurve/periodogram.py
+++ b/src/lightkurve/periodogram.py
@@ -383,56 +383,6 @@ class Periodogram(object):
             ax.set_title(title)
         return ax
 
-    def flatten(self, method="logmedian", filter_width=0.01, return_trend=False):
-        """Estimates the Signal-To-Noise (SNR) spectrum by dividing out an
-        estimate of the noise background.
-
-        This method divides the power spectrum by a background estimated
-        using a moving filter in log10 space by default. For details on the
-        `method` and `filter_width` parameters, see `Periodogram.smooth()`
-
-        Dividing the power through by the noise background produces a spectrum
-        with no units of power. Since the signal is divided through by a measure
-        of the noise, we refer to this as a `Signal-To-Noise` spectrum.
-
-        Parameters
-        ----------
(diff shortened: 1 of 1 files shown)


ojhall94:
After discussing with @christinahedges we decided to keep the `.flatten()` code in the `Periodogram` module, as people may want to use it using their own data that isn't strictly generated using a `LombScarglePeriodogram` (such as `TESS-SIP`, which generates a `Periodogram` object but doesn't use Lomb Scargle).

ojhall94:
Here's an example of the new `SNRPeriodogram.diagnose_flatten()` function. Users can use this to check whether the `.flatten()` function is removing power from the modes of oscillation they're looking at and adjust the `.flatten()` parameters accordingly.

The `.flatten()` keyword argument `filter_width` has been increased to 0.1 for better performance - but this isn't intelligent and needs to be further improved.

image

PR #1311, opened 2023-05-01
Update the stylefile 💅

danhey:
The lightkurve stylefile seems overdue for an update. The current stylefile has a 0.5 linewidth, which is a bit too thin according to journal recommendations, and the grey text color is hard to read. I've also updated the color scheme ordering to be more accessible to color blindness (see attached image).

Screenshot 2023-05-01 at 1 32 32 PM

I don't consider this to be a final version! Very open to suggestions / improvements. Maybe a new font?


christinahedges:
@megbedell and l added some changes to the `matplotlib` style file to make the lightkurve figures a bit more clean and easy to read. We changed the aspect ratios and the fonts, would be great to see what people think especially @danhey. Here are some examples of figures.
image
image

keyuxing:
> @megbedell and l added some changes to the `matplotlib` style file to make the lightkurve figures a bit more clean and easy to read. We changed the aspect ratios and the fonts, would be great to see what people think especially @danhey. Here are some examples of figures. image image

I think these changes are great! Hope to see it be merged soon.

PR #1313, opened 2023-05-01
fix outlier removal bug

### LINKED ISSUE #1312: remove_outliers() fails on NDArrays
`lc.sigma_clip()` fails on masked NDarrays. This is not caught by the unit tests.

#### Problem description
The current unit test implementation is:
```python
def test_remove_outliers():
    # Does `remove_outliers()` remove outliers?
    lc = LightCurve(time=[1, 2, 3, 4], flux=[1, 1, 1000, 1])
    lc_clean = lc.remove_outliers(sigma=1)
    assert_array_equal(lc_clean.time.value, [1, 2, 4])
```
If we try this with a real light curve it fails, because `lc.flux` is now a maskedNDarray:
```python
lc = lk.search_lightcurve('KIC 757450', author='Kepler', cadence='long').download_all().stitch()
lc.remove_outliers(sigma=3)
```

#### Example
As above

#### Expected behavior
The sigma clip should proceed without error.

#### Environment

-  platform (e.g. Linux, OSX, Windows): OSX
-  lightkurve version (e.g. 1.0b6): 2.4.0dev

danhey:
Fixes #1312 by ensuring all input to sigma_clip are no longer masked arrays

### DIFF
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -1376,7 +1376,7 @@ class LightCurve(TimeSeries):
         # a local import here.
         from astropy.stats.sigma_clipping import sigma_clip
 
-        # sigma_clip won't work with masked ndarrays.
+        # astropy.stats.sigma_clip won't work with masked ndarrays so we convert to regular arrays
         flux = self.flux.copy()
         if isinstance(flux, Masked):
             flux = flux.filled(np.nan)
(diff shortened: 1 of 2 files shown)


dfm:
LGTM - The only thing I'm not clear on is what `sigma_clip` does with `NaN` values. Perhaps we should test for that?

dfm (approved):

Nschanche (approved):
Looks good to me. I like the simple test.

PR #1318, opened 2023-05-01
Changed "Frame" to "Cadence" for the animate function

rebekah9969:
Here i changed the label from "Frame" to "Cadence" in the animate function.


christinahedges:
@rebekah9969 thank you for the PR! I believe "Frame" here is used because cadence numbers don't index from zero, and can sometimes have gaps. If you think cadence is better for animate, can you change this to the cadence number?

PR #1322, opened 2023-05-01
Add a `PeriodogramCollection` object

ojhall94:
There are established use cases where a `PeriodogramCollection` may be useful. These are, for example:
- Averaging the power spectra of time-separated sectors of TESS data to avoid aliasing
- Studying signals of variability independently in time separated sectors of TESS data (i.e. looking at spot rotation).
- Separately treating periodograms of light cuves of the same target with different exposure times (i.e. stars observed by Kepler and TESS, stars observed in TESS FFIs across the extended mission).

A `PeriodogramCollection` can be established by a user passing in a list of `lightkurve.Periodogram` objects, or by using the function `lightkurve.LightCurveCollection.to_periodograms(**kwargs)`, which passes all the keyword arguments of `lightcurve.to_periodogram()`.

An important discussion point is how to combine consecutive sectors of Kepler and TESS data, and whether to do that by default. Maybe there should be a `LightCurveCollection.partial_stitch()` method?

To do:
- Build `PeriodogramCollection.flatten()` function
- Build `PeriodogramCollection.normalize()` function
- Build `PeriodogramCollection.average()` function
- Build unit tests
- Establish workflow for partial stitching of light curves
- Build code into `LightCurveCollection` to partially stitch light curves.

### DIFF
--- a/src/lightkurve/collections.py
+++ b/src/lightkurve/collections.py
@@ -17,7 +17,8 @@ __all__ = ["LightCurveCollection", "TargetPixelFileCollection"]
 
 
 class Collection(object):
-    """Base class for `LightCurveCollection` and `TargetPixelFileCollection`.
+    """Base class for `LightCurveCollection`, `PeriodogramCollection`
+    and `TargetPixelFileCollection`.
 
     A collection can be indexed by standard Python list syntax.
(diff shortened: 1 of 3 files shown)


ojhall94:
Woops, I accidentally branched off of the wrong thing when I started this. I'll close this PR and start again.

PR #1323, opened 2023-05-01
Add a `PeriodogramCollection` object

ojhall94:
There are established use cases where a `PeriodogramCollection` may be useful. These are, for example:
- Averaging the power spectra of time-separated sectors of TESS data to avoid aliasing
- Studying signals of variability independently in time separated sectors of TESS data (i.e. looking at spot rotation).
- Separately treating periodograms of light cuves of the same target with different exposure times (i.e. stars observed by Kepler and TESS, stars observed in TESS FFIs across the extended mission).

A `PeriodogramCollection` can be established by a user passing in a list of `lightkurve.Periodogram` objects, or by using the function `lightkurve.LightCurveCollection.to_periodograms(**kwargs)`, which passes all the keyword arguments of `lightcurve.to_periodogram()`.

An important discussion point is how to combine consecutive sectors of Kepler and TESS data, and whether to do that by default. Maybe there should be a `LightCurveCollection.partial_stitch()` method?

To do:
- Build `PeriodogramCollection.flatten()` function
- Build `PeriodogramCollection.normalize()` function
- Build `PeriodogramCollection.average()` function
- Build unit tests
- Establish workflow for partial stitching of light curves
- Build code into `LightCurveCollection` to partially stitch light curves.

### DIFF
--- a/src/lightkurve/collections.py
+++ b/src/lightkurve/collections.py
@@ -17,7 +17,8 @@ __all__ = ["LightCurveCollection", "TargetPixelFileCollection"]
 
 
 class Collection(object):
-    """Base class for `LightCurveCollection` and `TargetPixelFileCollection`.
+    """Base class for `LightCurveCollection`, `PeriodogramCollection`
+    and `TargetPixelFileCollection`.
 
     A collection can be indexed by standard Python list syntax.
     Additionally, it can be indexed by a subset of `numpy.ndarray` syntax:
@@ -85,7 +86,7 @@ class Collection(object):
         Parameters
         ----------
         obj : object
-            Typically a LightCurve or TargetPixelFile object
+            Typically a LightCurve, Periodogram or TargetPixelFile object
         """
         self.data.append(obj)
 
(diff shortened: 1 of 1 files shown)


christinahedges:
This is really awesome! Can't wait to see how this evolves. @ojhall94 what do you envision partial stitching might be?

ojhall94:
> This is really awesome! Can't wait to see how this evolves. @ojhall94 what do you envision partial stitching might be?

So, when I've worked with TESS LCs in the past I've just stitched any sectors that were consecutive and treated those as independent light curves, so I think that would be a sensible place to start. I'm not sure if "partial stitch" is the right phrasing for that kind of behaviour though. Perhaps implementing a split() function would allow for more bespoke manipulation of the light curves for power users?

ojhall94:
Averaging periodograms requires them to be on the same grid of frequencies, which at the moment isn't working well! I'm not sure if this is a deeper `periodogram` issue or not. At some point I'd like to figure this out!

PR #1328, opened 2023-05-02
Improving the lightkurve binning function by replacing the aggregate_downsample function

tylerapritchard:
This is a work-in-progress PR where we're looking at improving the binning functionality/accuracy/speed by replacing the `astropy` `aggregate_downsample `call with a custom version that allows per-column aggregate functions, and by adding additional optimization to `reduceat`

### DIFF
--- a/src/lightkurve/downsample.py
+++ b/src/lightkurve/downsample.py
@@ -1,0 +1,272 @@
+'''astropy-timeseries-downsample function imported as a starting point
+to improve the lightkurve binning function'''
+
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+import warnings
+
+import numpy as np
+
+from astropy import units as u
+from astropy.time import Time, TimeDelta
+from astropy.timeseries.binned import BinnedTimeSeries
+from astropy.timeseries.sampled import TimeSeries
+from astropy.utils.exceptions import AstropyUserWarning
+
+__all__ = ["aggregate_downsample"]
+
+
+def reduceat(array, indices, function):
+    """
+    Manual reduceat functionality for cases where Numpy functions don't have a reduceat.
+    It will check if the input function has a reduceat and call that if it does.
+    """
+    if len(indices) == 0:
+        return np.array([])
+    elif hasattr(function, "reduceat"):
+        return np.array(function.reduceat(array, indices))
+    else:
+        result = []
+        for i in range(len(indices) - 1):
+            if indices[i + 1] <= indices[i] + 1:
+                result.append(function(array[indices[i]]))
+            else:
+                result.append(function(array[indices[i] : indices[i + 1]]))
+        result.append(function(array[indices[-1] :]))
+        return np.array(result)
+
+
+def _to_relative_longdouble(time: Time, rel_base: Time) -> np.longdouble:
+    # Convert the time objects into plain ndarray
+    # so that they be used to make various operations faster, including
+    # - `np.searchsorted()`
+    # - time comparison.
+    #
+    # Relative time in seconds with np.longdouble type is used to:
+    # - a consistent format for search, irrespective of the format/scale of the inputs,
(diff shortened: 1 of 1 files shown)


mirca:
I just wanted to point out that there are some issues with the computation of ``bins`` from the method `'knuth'` in https://github.com/lightkurve/lightkurve/blob/209d0c8095becc397fd8abc33933ca7be7b5d3ce/src/lightkurve/lightcurve.py#L1498

See https://github.com/astropy/astropy/issues/8326#issuecomment-521981400  and https://github.com/astropy/astropy/issues/11357. This might be something we would want to consider changing within a newer implementation of `bin` for `lightkurve`.

tylerapritchard:
This now should be functional and working to 0th order, implemented essentially the current defaults in the current version.  

Timing tests suggest a ~2.5x speedup:
Initial test: Total time: 13.8932 s
Current test: Total time: 5.51083 s

dfm:
See [this repo](https://github.com/dfm/fast-astropy-timeseries-binning) for some faster accumulation functions!

dfm:
Using that C library linked above, I get another order of magnitude speed up on my system! I'll look into contributing that upstream to astropy.

dfm:
Here are the benchmarks:

- on `main` (5.84s):

Screenshot 2023-05-02 at 4 03 52 PM

- on `tylerapritchard/replace_aggregate_downsample` (2.82s):

Screenshot 2023-05-02 at 4 04 27 PM

- incorporating the (highly incomplete) C library above (164ms):

Screenshot 2023-05-02 at 4 05 09 PM

tylerapritchard:
Made further tweaks with @dfm! Current changes from the previous default behavior:

- Previously, if the flux_err of a lightcurve was full of NaNs, in a binned lightcurve the flux error was replaced with the standard deviation of the binned flux points.  This could be misleading because this was labeled an "error" but could contain astrophysical signals due to intrinsic flux variance.  To expose this more clearly to users, when binning a lightkurve we're now addidng a column - 'flux_bin_std' that contains this value - and when binning errors that are all NaN values will now return NaN values.  This could break users current functionality! 

- Previously, lightcurve.bin executed two aggregate_downsample calls due to needing to treat errors differently than flux values, as [Astropy.Timeseries.aggregate_downsample ](url) requires the same function in the binning to be used across all columns.  We have created a minimally modified version of this function that allows the user to supply a function per column while attempting to maintain the current functionality of aggregate_downsample otherwise.  

- We are using this modified function to allow us to call aggregate_downsample a single time, with different binning logic for different columns.  We have supplied potentially sensible defaults for columns 
          - Current defaults are: 
          - flux - mean - SAME as current default
          - flux_err - average root sum squares - SAME as the current default 
          - flux_bin_std - NEW column, standard deviation 
          - quality: bitwise or of the quality flags - NEW (previously the mean of the quality flags)
          - cadenceno: median cadence value NEW (previously the integer rounded mean of the cadence values)

Some things that might be worth some discussion?  (and what else?)
- Are these sensible defaults?  Cadenceno and quality flag combining were new decisions, and could certainly be different.  
- Is the new treatment of lightkurves without errorbars, nad the potential breaking of currently extant workflows, acceptable? 
- One of our concerns with the current bin function was that it was opaque to users.  What else do we need to make more clear? How can we do that? There are the lightkurve.meta keys, what can/should we add there?  
- A more proper, general, treatment of errors would involve time weighted errors - this could be done with minimal overhead but would be a more compicated change, we'll open a separate issue about that .

manodeep:
(Apologies for the drive-by comment but I stumbled across this thread from astropy. Also hi Tyler :) ) 

If y'all are using/(planning to use) the C++ templates, you might benefit from a performance boost if you add 
`#pragma omp simd reduction(+: <varname>)` on top of the for-loop line and then compile with `-fopenmp-simd`

Issue #1278, opened 2023-02-01
TESS + csv data : to/from fits file errors


orionlee:
#### Problem description

When one combines lightcurves from TESS and other sources, e.g., from a csv, for the resulting stitched lightcurve object.
1. export failure: `lc._to_fits()` raised `AttributeError` 
2. one can workaround the error by recreating the stitched lightcurve (see example below) and export to fits
3. import failure: `lk.read()` raised `LightkurveError` for the fits file exported at 2, complaining it it not a supported data product.
  i. the exported fits is verified to be a proper fits file.

#### Example

```python
import numpy as np
from astropy.time import Time
import lightkurve as lk

lc1 = lk.search_lightcurve("pi Men", mission="TESS", cadence="short")[0].download()

# simulate lighcurve from other sources, e.g., read from csv
lc2 = lk.LightCurve(time=Time(2457000 + np.linspace(1200, 1230, num=100), format="jd"), flux=np.ones(100))

lc = lk.LightCurveCollection([lc1.normalize(), lc2]).stitch()

# lc is of type TessLightCurve, which is probably the root cause of the AttributeError
print(type(lc))

# AttributeError: object has no attribute centroid_col
lc.to_fits("lk_export.fits", overwrite=True)
```

``` python
# Workaround the AttributeError by not using TessLightCurve
lc = lk.LightCurveCollection([lc1.normalize(), lc2]).stitch()
lc = lk.LightCurve(time=lc.time, flux=lc.flux)

# the export is then successful
lc.to_fits("lk_export.fits", overwrite=True)
```

```python
# read the exported fits file using the workaround above
# LightkurveError: Not recognized as a supported data product
lc_imported = lk.read("lk_export.fits")
lc_imported.scatter();
```

```python
from astropy.io import fits

# ensure the exported fits is a proper FITS file
with fits.open("lk_export.fits") as hdul:
    print("hdul[0]")
    display(hdul[0].header)
    print("hdul[1]")
    display(hdul[1].header)
    print("hdul[1].data")
    print(hdul[1].data.shape)
    print(hdul[1].data)
```

#### Environment

-  platform (e.g. Linux, OSX, Windows): Windows 11
-  lightkurve version (e.g. 1.0b6): 2.3.0
-  installation method (e.g. pip, conda, source): conda

orionlee:
**Root cause for `to_fits()` error:**
- the stitched lightcurve is of type `TessLightCurve`, but it really is one anymore.
[...]

PR #1330, opened 2023-05-02
Addressing github issue 1278

isabelangelo:
Addressing issue [#1278](https://github.com/lightkurve/lightkurve/issues/1278) , `stitch()`  now saves all columns from all component lightcurves in the stitched lightcurve. If a column is missing from one of the component lightcurves, all its will be filled with -999. This also helps with saving and reading in .fits files since it preserves relevant columns across Kepler/TESS/simulated lightcurve objects.

### DIFF
--- a/src/lightkurve/collections.py
+++ b/src/lightkurve/collections.py
@@ -178,6 +178,11 @@ class LightCurveCollection(Collection):
         before stitching. For example, passing "lambda x: x.normalize().flatten()"
         will normalize and flatten each light curve before stitching.
 
+        All columns from all component lightcurves are saved. When a column is
+        missing from a component lightcurve, the column is filled with dummy 
+        value -999 in the stiched lightcurve. 
+        (Addressing github issue #1278)
+
         Parameters
         ----------
         corrector_func : function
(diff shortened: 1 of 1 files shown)


src/lightkurve/collections.py
```diff
@@ -196,32 +201,13 @@ def stitch(self, corrector_func=lambda x: x.normalize()):
...
-                lc.remove_columns(columns_to_remove.intersection(lc.columns))
-                for lc in lcs
-            ]
-
         # Need `join_type='inner'` until AstroPy supports masked Quantities
-        return vstack(lcs, join_type="inner", metadata_conflicts="silent")
+        lc_vstack = vstack(lcs, join_type="outer", metadata_conflicts="silent")
+        for col in lc_vstack.columns:
+          if col not in ['time','flux','flux_err']:
+              lc_vstack[col] = lc_vstack[col].value.filled(-999)
```
dfm, line 208: I don't love this approach of using `-999` as a magic number! What about using something like:

```python
lc_vstack[col].fill_value = np.nan
```

in the cases when there is missing data?

isabelangelo: I definitely agree, we did try that but the problem is that it only works when the column datatype is float, and some of the columns in the lightcurve objects have int datatypes. But hard coding in the -999 didn't feel great either.

Issue #1206, opened 2022-04-30
TASOC DR5 issues: flux correction/normalization;  choosing correction method


orionlee:
#### Problem description

I came across a few the new [TASOC DR5](https://archive.stsci.edu/hlsp/tasoc) lightcurves, and believe `lightkurve`'s handling should be updated.

Flux issues:
1. The default `flux` should be systematics-corrected `FLUX_CORR` column, rather than the current `FLUX_RAW` (Output flux from TASOC photometry module, before systematics-correction)
2. Even if we ignore (1), as it is, when plotting TASOC lightcurve (based on `FLUX_RAW`), the plots incorrectly label the flux as normalized flux (it is in `electron/s`)
3. If we agree on 1 (changing default flux), do we need to retain logic to handle pre-DR5 lightcurves?

Correction method:
4. For 30 minute cadence lightcurves, there are 2 versions: `_cbv` (correction by CBV) and `_ens` (correction by ensemble photometry). In lightkurve `SearchResult`, the 2 versions appears as 2 rows that look identical (in the default display). To identify the versions, one need to check the actual filename pattern, e.g., in  `search_result.table['dataURI']` (or 'productFilename').

I'm uncertain what exactly `lightkurve` should do, if any (hack `author` to be `TASOC-ENS` / `TASOC-CBV`?).

#### Example

```python
import lightkurve as lk

# 3 entries returned: 2-minute cadence, 30-minute cadence CBV, 30-minute cadence ENS
# the 2 30-minute cadence entries look identical in default SearchResult display
sr = lk.search_lightcurve(f"TIC91369561", author="TASOC", sector=1, mission="TESS")
display(sr)

lc = sr[0].download()
# y-axis is incorrectly labeled as normalized flux
lc.scatter()

# inspect the data and you'll see flux is based on `FLUX_RAW`
lc[:5]
```

#### Environment

-  lightkurve version (e.g. 1.0b6): 2.2

PR #1333, opened 2023-05-02
Changed flux_raw to flux_corr for TASOC files

rebekah9969:
modified tasco.py file to use corrected data instead of raw.


rebekah9969:
#1206

rebekah9969:
The checks fail because they cannot download data from MAST. I want to re-run these checks to see if this is because of a connection issue at the time or something else.

Issue #1125, opened 2021-07-28
Flux values returned from PLDCorrector appear not to be deterministic


decaelus:
#### Problem description
If I run PLDCorrector on the same target pixel file to generate a light curve more than once, I get different value, different enough that fitting algorithms (e.g., scipy curve_fit) return (slightly) different answers. 

#### Example

```python
import lightkurve as lk
search_result = lk.search_tesscut('Kepler-13A')
tpf = search_result.download(sector=15, cutout_size=10)
pld = tpf.to_corrector('pld')
flattened_lc = pld.correct().flatten()
print(flattened_lc.flux[0])
```
I get this output on multiple runs:
0.9999522333480251 electron / s
0.9999522333482418 electron / s
0.999952233348407 electron / s

#### Expected behavior
Maybe this behavior is exactly what's expected, but I would have expected exactly the same answer every time. This non-determinism may make replicability problematic.

#### Environment

-  platform (e.g. Linux, OSX, Windows): OSX
-  lightkurve version (e.g. 1.0b6): 2.0.10
-  installation method (e.g. pip, conda, source): conda

orionlee:
I can reproduce it using a stripped down variation of the tutorial https://docs.lightkurve.org/tutorials/2-creating-light-curves/2-3-k2-pldcorrector.html

```python
tpf = lk.search_targetpixelfile('K2-199', author='K2', campaign=6).download()
pld = tpf.to_corrector('pld')
for i in range(0, 5):
    lc = pld.correct()
    print(lc.flux[0])
```

Output:
```
229698.43291257642 electron / s
229692.75462726582 electron / s
229706.7975446729 electron / s
229712.53869939488 electron / s
229709.87251854019 electron / s
```

I don't know about `PLDCorrector` to offer specific help though.

christinahedges:
Hi @decaelus that's a great catch. I think this is actually expected, the `PLDCorrector` relies on a PCA algorithm from `fbpca`. I believe determining the PCA components is a somewhat stochastic process, and you can tune the number of iterations that the algorithm takes to find the PCA components. You can see this if you look inside the `PLDCorrector` object at the regressors. 

Because these discrepancies are small, we've assumed that this adequate for most people's use. But, we could increase the number of iterations of the `fbpca` algorithm to try to reduce this.

Are you finding this significantly impacts your work?

decaelus:
Thanks for clarifying. I wasn't clear that there was a non-deterministic element to the correction.

PR #1334, opened 2023-05-02
Expose n_iters to the pca method in DesignMatrix

mirca:
In this way, users can better control the accuracy of the PCAs computed via `fbpca`.  This is related to #1125.

### DIFF
--- a/src/lightkurve/correctors/designmatrix.py
+++ b/src/lightkurve/correctors/designmatrix.py
@@ -249,7 +249,7 @@ class DesignMatrix:
         dm.df = new_df
         return dm
 
-    def pca(self, nterms=6):
+    def pca(self, nterms=6, n_iter=10):
         """Returns a new `.DesignMatrix` with a smaller number of regressors.
 
         This method will use Principal Components Analysis (PCA) to reduce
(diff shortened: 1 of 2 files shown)


src/lightkurve/correctors/designmatrix.py
```diff
@@ -274,7 +278,7 @@ def pca(self, nterms=6):
         # produces more stable results.
         from fbpca import pca  # local import because not used elsewhere
 
-        new_values, _, _ = pca(self.values, nterms, n_iter=10)
+        new_values, _, _ = pca(self.values, nterms, n_iter)
```
dfm, line 281:
```suggestion
        new_values, _, _ = pca(self.values, nterms, n_iter=n_iter)
```

This needs to be a keyword argument, because `raw` is the next positional argument and that's not what we want!

dfm (changes requested):

dfm (approved):

dfm:
LGTM!

christinahedges:
Closes #1125 
The issue above talks about there being a slight randomization to the correction. Increasing this parameter will reduce the random noise here.

The [tutorials](http://docs.lightkurve.org/tutorials/2-creating-light-curves/2-3-removing-scattered-light-using-regressioncorrector.html) might benefit from being updated to reflect this at some level, but I'm not too concerned.

christinahedges:
@tylerapritchard can you add a CHANGELOG update for this and then we will merge and update the API docs.

Issue #1067, opened 2021-05-18
CDIPS product minor bugs / compatibility issues


orionlee:
#### Problem description
For lightcurves from CDIPS, I found the following discrepancies from what one would normally expect in using a `LightCurve` object:
1. `lc.time`'s values are in BJD, but is stored and treated as if they are BTJD
2. `lc.label` is some CDIPS-specific number (from header 'OBJECT'), rather than the TIC id
3. `lc.quality` is a 1 character string rather than integer.

I think:
- (1) and (2) would be considered bugs. 
- While there is not much we can do directly about (3), it causes complication, e.g., calling `LightCurveCollection.stitch()` on a mix of SPOC and CDIPS products would cause errors
```TableMergeError: The 'quality' columns have incompatible types: ['int32', 'str32']```
The error is clear enough that users can probably know what to do. But we might want to make stitch() more tolerant when merging non-essential columns fails.

#### Example

TIC 74482749 has a CDIPS product in sector 8
```python
import lightkurve as lk

sr = lk.search_lightcurve("TIC74482749", mission="TESS")
lcc = sr.download_all()
lcc[lcc.sector == 8][0].scatter();
```

One can see the label and the time aren't quite right

![image](https://user-images.githubusercontent.com/250644/118732272-d941ce00-b7ee-11eb-81c3-fec709556581.png)

#### Environment

-  lightkurve version (e.g. 1.0b6): 2.0.9

barentsen:
Thank you for opening this issue!

1. Agreed, this is a bug!  It looks like line 57 in `io/cdips.py` incorrectly sets `time_format='btjd'`.
2. There is no strict rule as to what `lc.label` should or shouldn't contain, but it makes sense to use TIC identifiers for TESS data when possible and/or available.  Note that official TESS data products use the `TICID` header keyword to record the integer part of the TIC ID number, so it would make sense for light curve objects to have `lc.meta["TICID"]` populated whenever possible.
3. Agreed that `stitch()` ought to be tolerant to columns with incompatible types.  The method does contain logic right now to exclude columns with incompatible types, but I guess it does not work perfectly!

scfleming:
For #2, note that those are Gaia DR2 IDs, not TIC IDs like most other HLSP have decided to base their object IDs off of.

orionlee:
For the `lc.label`, I feel it'd be better if it is one of the following:
1. a TIC-based label, like SPOC products, or
[...]

PR #1335, opened 2023-05-05
Modified CDIPS output to present flux values and BTJD time

rebekah9969:
PR in response to issue #1067 
I have modified the cdips.py code to show the correct BTJD. I have also made IFL1 the default flux and IFE1 the default flux error. 
Note that a user can still change this value back to mag or anyone of the other flux values, they just need to specify that column when downloading.
Showing the flux_error is now set as a default. 

There is still an issue about what to put for the quality values. As is this is set to G for good values and X for bad. The data is filtered to only show the good data, so G is always only what is shown. Would it be better to set this to 0, this would make it more consistent with other light curve data.

### DIFF
--- a/src/lightkurve/io/cdips.py
+++ b/src/lightkurve/io/cdips.py
@@ -13,8 +13,8 @@ from .generic import read_generic_lightcurve
 log = logging.getLogger(__name__)
 
 def read_cdips_lightcurve(filename,
-                            flux_column="IRM1",
-                            include_inst_errs=False,
+                            flux_column="IFL1",
+                            include_inst_errs=True,
                             quality_bitmask=None):
     """Returns a TESS CDIPS `~lightkurve.lightcurve.LightCurve`.
 
@@ -24,6 +24,11 @@ def read_cdips_lightcurve(filename,
     are removed according to Bouma et al. 2019, and no other quality filtering
     is allowed. The `quality_bitmask` parameter is ignored but accepted for
     compatibility with other data format readers.
+
+    There are several kinds of flux and  magnitudes provided. For consistancy
+    we have chosen to display as default 'IFL1', the flux in aperture 1. 
+    This is given in ADU. 
+    The flux_err is also provided as 'IFE1' in ADU.
 
     More information: https://archive.stsci.edu/hlsp/cdips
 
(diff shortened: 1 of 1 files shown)


src/lightkurve/io/cdips.py
```diff
@@ -63,6 +69,11 @@ def read_cdips_lightcurve(filename,
                                  quality_column=quality_column,
                                  time_format='btjd')
 
+    #The time displayed is in BJD not BTJD. We can fix this by subtracting
+    # 2457000
+    lc["time"] = lc["time"]-2457000
```
christinahedges, line 74: This should be taken care of with the `time_format` argument by setting it to `jd`

src/lightkurve/io/cdips.py
```diff
@@ -74,8 +85,19 @@ def read_cdips_lightcurve(filename,
...
+        lc.meta["TARGETID"] = tic
+        lc.meta["TICID"] = tic
+        lc.meta["OBJECT"] = f"TIC {tic}"
+        # for Lightkurve's plotting methods
+        lc.meta["LABEL"] = f"TIC {tic}"
+
+    
     lc.meta["AUTHOR"] = "CDIPS"
-    lc.meta['TARGETID'] = lc.meta.get('TICID')
     lc.meta['QUALITY_BITMASK'] = 36
```
christinahedges, line 101: This needs to be updated

christinahedges (commented):
You need to make sure that the flux/flux errors have the "correct" astropy unit from the HLSP

src/lightkurve/io/cdips.py
rebekah9969: This has now been adjusted. Thank you.

src/lightkurve/io/cdips.py
rebekah9969: Is the point here to change the quality bitmask to the TESS default i.e., 175?

rebekah9969: Additionally CDIPS light curves have already had quality filtering applied, and do not provide the bitflags necessary for a user to apply a new bitmask.

src/lightkurve/io/cdips.py
rebekah9969: Note however that the time is given in BJD, but this does not seem to be an option.

kecnry:
I recently ran into this same problem (BTJD discrepancy) - is there any chance this PR would make it in a release soon?

Nschanche:
@rebekah9969 was the bit mask discussion resolved? If so, is this ready to merge?

scfleming:
To provide some additional context, we are soon working on adding interactive previews back into Portal using a package Kyle is writing that uses lightkurve to load the data. While not an explicit blocker, if any residual PRs can be merged into a version of lightkurve that would add to the number of supported datasets at initial release. If things go well on our end, that could be as soon as July 30th as part of our MAST monthly patch for July.

Issue #1230, opened 2022-06-24
TESS Quality Flag default bitmask update


orionlee:
#### Problem description

This is a follow up of issue #1163 / PR #1218 , regarding what tweaks should be done for the default TESS quality bitmask.

There is some recommendation in [TESS Data Product documentation](https://outerspace.stsci.edu/display/TESS/2.0+-+Data+Product+Overview#:~:text=As%20a%20starting%20point%2C%20the%20following%20flags%20indicate%20that%20the%20cadence%20will%20likely%20be%20of%20lesser%20quality
).

I tabulate the suggestion in the doc and compare it against the current `default` and `hard` bitmasks

| Bit Description                                      | bin digit # | TESS Doc | `default` | `hard` |
| ---------------------------------------------------- | ----------- | -------- | ------- | ---- |
| AttitudeTweak = 1                                    | 1           | ✓        | ✓       | ✓    |
| SafeMode = 2                                         | 2           | ✓        | ✓       | ✓    |
| CoarsePoint = 4                                      | 3           | ✓        | ✓       | ✓    |
| EarthPoint = 8                                       | 4           | ✓        | ✓       | ✓    |
| Argabrightening = 16                                 | 5           | ✓        |         |      |
| Desat = 32                                           | 6           | ✓        | ✓       | ✓    |
| ApertureCosmic = 64                                  | 7           |          |         | ✓    |
| ManualExclude = 128                                  | 8           | ✓        | ✓       | ✓    |
| Discontinuity = 256                                  | 9           |          |         |      |
| ImpulsiveOutlier = 512                               | 10          | ✓        |         |      |
| CollateralCosmic = 1024                              | 11          |          |         | ✓    |
| Straylight = 2048 #: set by MIT manually             | 12          |          |         | ✓    |
| Straylight2 = 4096 #: set by Ames/SPOC automatically | 13          | ✓        |         | ✓    |
| PlanetSearchExclude = 8192                           | 14          |          |         |      |
| BadCalibrationExclude = 16384                        | 15          | ✓        |         |      |
| InsufficientTargets = 32768                          | 16          |          |         |       |

If we follow the recommendation, the following bits will be added to `default` (and indirectly added to `hard`
- bit 5, Argabrightening = 16
- bit 10, ImpulsiveOutlier = 512	
- bit 13, Straylight2 = 4096 #: set by Ames/SPOC automatically, already in hard
- bit 15, BadCalibrationExclude = 16384

#### Environment

[...]

PR #1336, opened 2023-05-05
Updated TESS default bitmask

Nschanche:
Issue #1230


christinahedges:
@Nschanche could you please make a stand alone version of the test that fails, write up what it's trying to test in 2 sentences, explain what a value of "1" or "0.95" means in this context, and then we'll decide what to do about the failing test.

christinahedges (commented):
This looks good to me, only one comment, if you fix that I approve!

PR #1345, opened 2023-06-13
Added fixes for some bugs in the `to_fits` methods of LightCurve classes

christinahedges:
This PR fixes bugs in the `LightCurve`, `KeplerLightCurve`, and `TESSLightCurve` `to_fits` methods and the tests. This ensures users can pass extra keyword arguments.

@Nschanche points out that it's odd that, by default, `to_fits` doesn't save every column. Some of the columns can be a bit superfluous, so I'm not sure of a neat way forward here. What do people think, should `to_fits` save all columns by default?

### DIFF
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -40,7 +40,8 @@ __all__ = ["LightCurve", "KeplerLightCurve", "TessLightCurve", "FoldedLightCurve
 
 log = logging.getLogger(__name__)
 
-_HAS_VAR_BINS = 'time_bin_end' in aggregate_downsample.__kwdefaults__
+_HAS_VAR_BINS = "time_bin_end" in aggregate_downsample.__kwdefaults__
+
 
 def _to_unitless_day(data):
     if isinstance(data, Quantity):
@@ -166,7 +167,6 @@ class LightCurve(TimeSeries):
     __array_priority__ = 100_000
 
     def __init__(self, data=None, *args, time=None, flux=None, flux_err=None, **kwargs):
-
         # the ` {has,get,set}_time_in_data()`: helpers to handle `data` of different types
         # in some cases, they also need to access kwargs["names"] as well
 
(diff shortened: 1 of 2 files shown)


christinahedges:
@Nschanche when you review this you'll see I've accidentally linted this file....but I don't know that I hate that...

Nschanche:
> 
> @Nschanche points out that it's odd that, by default, `to_fits` doesn't save every column. Some of the columns can be a bit superfluous, so I'm not sure of a neat way forward here. What do people think, should `to_fits` save all columns by default?

There are a number of keywords that are dropped by to_fits as it is now, including timecorr, sap_flux, sap_bkg, sap_bkg_err, psf_centr1+2 and errors, and pos_cor1+2 (the last two are addressed in this pull request). I tend to think everything should be saved by default, but we could make it an explicit choice by, e.g., adding a flag such as all=True that would save every column, but stick with the current subset by default.

kelle:
I'd like to check in on this PR since I'm currently working with some folks to use the `lightcurve.to_fits` method to create FITS files from a heterogeneous sample of lightcurves.  We're also struggling with similar philosophical issues in the `fits_tabular` writer in `specutils` (https://github.com/astropy/specutils/pull/1113).

In my opinion, `to_fits` should be able to maintain all columns as long as the file generated continues to be readable by the generic fits reader with default settings.  I think a keyword `all_columns` makes sense. With default = False.

As I discuss in #1446, it would also be nice to have the `lightcurve.to_fits` method not add the extra Kepler-related header keywords.

kelle:
@Nschanche @christinahedges  - Pinging on this again.  Any reason this can't be merged after fixing conflicts?

kelle:
I opened a teeny PR (#1454) that just fixes the unit issue.

Nschanche:
@christinahedges I think content-wise this looks fine, just needs some tidying up of comments. A small conflict also needs to be resolved before merging.

Nschanche:
OK, for this PR, it sounds like we still need to 

- [ ] add a flag to save all keywords (default to False)

- [ ] Make sure the primary header follows the HEASARC template [data/lc-ext0-header.txt](https://github.com/lightkurve/lightkurve/blob/main/src/lightkurve/data/lc-ext0-header.txt)

- [ ] Add tests to make sure new features work. Also ensure Kepler-specific keywords are not added for TESS data

Anything else I'm missing @kelle ?

kelle:
Honestly, that's A LOT. I would break this up into multiple PRs

Making sure the keywords follow HEASARC is its own thing. 

For this particular PR, I think the save all flag is the highest priority. The keywords can be a different thing.

PR #1366, opened 2023-09-14
Basic outline for new sky catalog search function

rebekah9969:
This PR will create a new function called get_skycatalog. 

This is very initial code and the function will likely be captured within another file or function like search.py
LOTS OF REFORMATTING

The goal of this function is to provide the user with a list of stars/objects within a defined region around a requested target.

The returned list contains the proper motion corrected R.A and Dec. positions of the objects in addition to several other parameters.

This tool should be applied to the following TESS products
- LightCurve Objects
- TargetPixelFile Objects
- TESSCut Objects

Eventual use case will be:
mytpf.get_skycatalog()
Returning a table of objects and their parameters. 

Several adjustments to be made inclide

- [ ] A check to make sure that the file type input is correct i.e., LC, TPF, TESSCut
- [ ] Adjust the search region to be by default slightly larger than the input TPF or TESSCut size
- [ ] Make the catalog search conduct a box search instead of a radius search
- [ ] Based on the mission keyword make the catalog search be either TIC or KIC etc

### DIFF
--- a/src/lightkurve/get_skycatalog.py
+++ b/src/lightkurve/get_skycatalog.py
@@ -1,0 +1,139 @@
+"""
+THIS IS A WORK IN PROGRESS
+
+This tool will provide the user with a list of stars/objects within a defined region around a requested target.
+
+The returned list contains the proper motion corrected R.A and Dec. postions of the objects in addition to several other parameters.
+
+This tool can be applied to the following TESS products
+- LightCurve Objects
+- TargetPixelFile Objects
+- TESSCut Objects
+
+Eventual use case will be:
+mytpf.get_skycatalog()
+Returning a table
+
+
+This file will likely be merged into the search.py code and structure 
+It is placed here as an initial working document 
+"""
+
+import sys, os
+
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+
+from astropy.coordinates import SkyCoord
+from astropy.coordinates import ICRS, Galactic, FK4, FK5  # Low-level frames
+from astropy.coordinates import Angle, Latitude, Longitude  # Angles
+
+import astropy.units as u
+from astropy.time import Time
+
+from astroquery.mast import Observations
+from astroquery.mast import Catalogs
+
+import lightkurve as lk
+from lightkurve import *
+
+def get_skycatalog(tpf_or_lc, 
+                search_radius_deg=None, 
+                lim_mag=None, 
+                catalog=None):
+    
+    #Need to add a check to make sure file type is correct 
+
+    #Need to make this a box search
+    #If a TPF supplied make the box search slightly larger than the size of the TPF
+     
+    
+    
+    #Get the time that the data was collected
+    new_time = tpf_or_lc.time[0]
+    #Get the R.A., Dec., and the equinox 
+    ra = tpf_or_lc.ra
+    dec = tpf_or_lc.dec
+    #Format it so that it works with the catalog query
+    cords = str(ra)+" "+str(dec)
+    #Get the equinox from the header
+    equinox = tpf_or_lc.meta["EQUINOX"]
+    
+    
+    #This code is specific for a TESS object a TPF, TESSCut or a LC - TPF.starcat_all(rad,lim_mag) and it return the answer
+    #We could modify it in the future so that you could just put in the object R.A and Dec.
+
+    
+    #Setting defaults for search radius, limiting mag, and catalog
+
+    #We should set a default value if a radius is specified
+    #If it is a TPF and the radius is not specified then get this info based on the size plus x amount of pixels
+    
+    if search_radius_deg==None:
+        srd=0.1
+    else:
+        srd = search_radius_deg
+        
+    if lim_mag == None:
+        lm = 18
+    else: 
+        lm = lim_mag
+
+    #Want the catalog to have either the TIC or KIC inputs as default
+    #Need to use tpf.mission to determine which mission it is from
+    #Then set default
+    #Problem is that there is no KIC option in catalog query - we could try something else?
+    
+    if catalog == None:
+        ct = "TIC"
+    else:
+        ct = catalog
+        
+    #Allowed catalog values - want to add a check here
+   
+    cv = ["HSC","Galex","Gaia","TIC","CTL","Panstarrs", "DiskDetective", "Plato"]
+    
+    catalogTIC = Catalogs.query_object(str(cords), radius=srd, catalog=ct)
+    mag_cut = np.where(catalogTIC["Tmag"]<float(lm))
+    mag_lim = catalogTIC[mag_cut]
+    
+    ra_list = mag_lim["ra"]
+    dec_list = mag_lim["dec"]
+    pm_ra = mag_lim["pmRA"]
+    pm_dec = mag_lim["pmDEC"]
+
+    pm_unit = u.milliarcsecond / u.year
+    
+    new_ra = []
+    new_dec = []
+
+    #Below is then code pulled in/adapted from #1332 by Veselin
+    #_get_nearby_gaia_objects
+    #_get_corrected_coordinates
+    for a in range(len(ra_list)):
+    
+        if ra is None or dec is None or pm_ra is None or pm_dec is None or (np.all(pm_ra == 0) and np.all(pm_dec == 0)) or equinox is None:
+            return ra, dec, False
+
+        c = SkyCoord(ra_list[a] * u.deg, 
+                     dec_list[a] * u.deg, 
+                     pm_ra_cosdec=pm_ra[a] * pm_unit, 
+                     pm_dec=pm_dec[a] * pm_unit,
+                     frame='icrs',
+                     obstime=Time(equinox, format="decimalyear", scale="tt") + 0.5)
+    
+            
+        new_c = c.apply_space_motion(new_obstime=new_time)
+        
+        new_ra.append(new_c.ra.to(u.deg).value)
+        new_dec.append(new_c.dec.to(u.deg).value)
+    
+    #Will need to convert this into astropy to make lk compatable
+    df = mag_lim.to_pandas()
+    
+    df.insert(3, 'ra_corrected', new_ra, True)
+    df.insert(4, 'dec_corrected', new_dec, True)
+        
+    return df
+


src/lightkurve/get_skycatalog.py
```diff
@@ -0,0 +1,121 @@
...
+
+    #Below is then code pulled in/adapted from #1332 by Veselin
+    #_get_nearby_gaia_objects
+    #_get_corrected_coordinates
+    for a in range(len(ra_list)):
+    
+        if ra is None or dec is None or pm_ra is None or pm_dec is None or (np.all(pm_ra == 0) and np.all(pm_dec == 0)) or equinox is None:
+            return ra, dec, False
+
+        c = SkyCoord(ra_list[a] * u.deg,
```
christinahedges, line 98: SkyCoord is an object that will tolerate iterables

src/lightkurve/get_skycatalog.py
```diff
@@ -0,0 +1,179 @@
...
+    return catalog
+
+
+def query_skycatalog(
+    coord: SkyCoord,
+    epoch: Time,
+    catalog_name: str,
+    radius: float = 20.0,
+    magnitude_limit: float = 18.0,
+    equinox: Time = Time(2451545.0, format="jd", scale="tt"),
```
christinahedges, line 85:
```suggestion
    equinox: Time = Time(2000, format="jyear"),
```

src/lightkurve/get_skycatalog.py
```diff
@@ -0,0 +1,179 @@
...
+    catalog["DEJ2000"] = c1.dec.to(u.deg).value
+
+    return catalog
+
+
+def query_skycatalog(
+    coord: SkyCoord,
+    epoch: Time,
+    catalog_name: str,
+    radius: float = 20.0,
```
christinahedges, line 83: You could accept an astropy quantity here
```suggestion
    radius: Union[float, u.Quantity] = u.Quantity(20, 'arcsecond'),
```
And then in the code you would ensure radius has units of arcsecond
```python
try: 
    radius = u.Quantity(radius, 'arcsecond')
except u.UnitConversionError:
    raise
```

src/lightkurve/get_skycatalog.py
```diff
@@ -0,0 +1,179 @@
...
+        catalog = filters.query_region(
+            coord, catalog=catalog_name, radius=Angle(radius, "arcsec")
+        )[catalog_name]
+        catalog.rename_columns(("DR3Name", "pmDE", "Gmag"), ("ID" "pmDEC", "Mag"))
+
+        # apply_propermotion
+        catalog = apply_propermotion(catalog, equinox=equinox, epoch=epoch)
+    else:
+        raise ValueError(
+            "This is not a valid catalog name please refer to Lightkurve docs for a list of catalogs that can be used"
```
christinahedges, line 177:
```suggestion
           f "Can not parse catalog name '{catalog_name}'"
```

src/lightkurve/get_skycatalog.py
rebekah9969: Great suggestion! 
Thanks

src/lightkurve/targetpixelfile.py
```diff
@@ -2227,6 +2270,33 @@ def extract_aperture_photometry(
         return KeplerLightCurve(
             time=self.time, flux=flux, flux_err=flux_err, **keys, meta=meta
         )
+    
+    @property
+    def get_skycatalog(self):
```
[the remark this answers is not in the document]
rebekah9969: Name changed - will check timing

src/lightkurve/targetpixelfile.py
```diff
@@ -2227,6 +2270,33 @@ def extract_aperture_photometry(
...
+        catalog = query_skycatalog(SkyCoord(self.ra, self.dec, unit="deg"), epoch=Time(np.mean(self.time.value), scale='tdb', format='bkjd'), catalog='kepler')
```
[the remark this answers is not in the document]
rebekah9969: Done and also have to make this lower case.

src/lightkurve/targetpixelfile.py
```diff
@@ -2227,6 +2270,33 @@ def extract_aperture_photometry(
...
+
+        # apply self.wcs to add column and row positions
+        w = self.wcs
```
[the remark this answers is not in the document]
rebekah9969: Done

christinahedges:
This is looking dangerously close, you need to add 

- An entry to the changelog
- Update the API reference in the docs folder (`docs/source/reference/targetpixelfile.rst`) to ensure that your new functionality shows up
- Optional; add the functionality to a tutorial page

rebekah9969:
Created a GitHub gist notebook demonstrating how this works.
https://gist.github.com/rebekah9969/3ba956149325acab679e66ae39babb56

rebekah9969:
The following warning appears when running 
tpf.skycatalog
` ErfaWarning: ERFA function "pmsafe" yielded 472 of "distance overridden (Note 6)"
  warnings.warn('ERFA function "{}" yielded {}'.format(func_name, wmsg),`

Looking at https://pyerfa.readthedocs.io/en/latest/api/erfa.pmsafe.html note 6

"6) An extremely small (or zero or negative) parallax is overridden
   to ensure that the object is at a finite but very large distance,
   but not so large that the proper motion is equivalent to a large
   but safe speed (about 0.1c using the chosen constant).  A warning
   status of 1 is added to the status if this action has been taken."

I think that this happens because some stars have zero proper motion - these are listed as "--" in the catalog and as such converted into 0.0 to preserve the original R.A and Dec. 
I think that this is probably something we can ignore and catch.

src/lightkurve/targetpixelfile.py
```diff
@@ -2045,7 +2051,60 @@ def plot_pixels(
...
+            catalog_name = "epic"
+            radius = 4 * (np.max([self.shape[1], self.shape[2]]) + 10) / 2 * u.arcsec
+        elif self.mission.lower() == "tess":
+            catalog_name = "tic"
+            radius = 21 * (np.max([self.shape[1], self.shape[2]]) + 10) / 2 * u.arcsec
+        else:
+            raise ValueError("Must pass a valid Target Pixel File object.")
+
+        catalog = query_skycatalog(
+            coord=SkyCoord(self.ra, self.dec, unit="deg"),
```
orionlee, line 2077: Should the target coordinate (`self.ra, self.dec`) be proper motion  corrected before sending the query out?

src/lightkurve/utils.py
```diff
@@ -871,3 +875,195 @@ def finalize_notebook_url(notebook_url):
...
+            - [value for key, value in result.items() if "_Mag" in key][0][ref_index]
+        )
+        / -2.5
+    )
+
+    # apply_propermotion
+    result = _apply_propermotion(result, equinox=equinox, epoch=epoch)
+
+    # Now sort the table based on separation
+    result.sort(["Separation"])
```
orionlee, line 1067: Should the separation be recalculated after the proper motion has been applied?

src/lightkurve/targetpixelfile.py
```diff
@@ -2045,7 +2051,60 @@ def plot_pixels(
...
+
+        # Cut down to targets within 1 pixel of edge of TPF
+        col_min = self.column - 1
+        row_min = self.row - 1
+
+        col_max = self.column + self.shape[2]
+        row_max = self.row + self.shape[1]
+
+        # Filter
+        catalog = catalog[
```
orionlee, line 2098: Here the logic is to filter out stars outside of the tpf.

If the result is intended to be used by `tpf.interact_sky()` as well, it might not be desirable.
In `tpf.interact_sky()` use case, from time to time I'd zoom outside of the tpf, usually in cases there are some pecularity in some pixels at the edge of the tpf, I'd zoom out and see if there could be due to some stars just outside of the tpf.

orionlee (commented):
Some comments from using and implementing `tpf.interact_sky()`.

src/lightkurve/targetpixelfile.py
```diff
@@ -2045,7 +2051,60 @@ def plot_pixels(
 
         return ax
 
+    @property
+    def skycatalog(self):
```
orionlee, line 2055: A style comment: exposing the list of neighboring TICs/KICs/EPICs via a property `tpf.skycatalog` could be somewhat misleading from programming point of view. 

Conventionally, data exposed via properties tend to be in-memory or easily computed. The `tpf.skycatalog` here is much more involved.

An alternative is to provide it as a method `tpf.query_skycatalog(...)` with parameters' default values do what the properties currently does.

src/lightkurve/targetpixelfile.py
rebekah9969: This is a very good point. We want all objects over a certain brightness within the TPF to be listed in the catalog. The input coordinate must allow for this. If the J2000 R.A and Dec from the TPF is used or even the values corrected for proper motion - the list of returned objects may not cover the TPF. As such i have now adjusted the code such that when a TPF is used the central coordinates of the TPF are used within the catalog search. 

We cannot use the corrected co-ordinates as suggested above since it is from this catalog search we get the proper motion values, and again this would not be beneficial in returning all objects within the TPF FOV.

src/lightkurve/targetpixelfile.py
rebekah9969: Correct the logic is to filter out stars outside of the TPF. 
One of the practical uses for this would be to recreate the scene via PRF modelling or examining nearby contamination.

The result is not intended for interact sky.

src/lightkurve/targetpixelfile.py
rebekah9969: @christinahedges what do you think with regards to this?

src/lightkurve/targetpixelfile.py
orionlee: Good point that TPF center should be used for catalog search.

src/lightkurve/targetpixelfile.py
orionlee: For the use cases for supporting contamination calculation / PRF, etc,
1. it might make sense to let users easily customize the catalog to 
  a. include stellar companion identified in follow up, and/or
  b. exclude duplicates, artifacts, etc.
One way to do it is through a method tpf.query_skycatalog(...), something like the following:
```python
# add a stellar companion identified in follow up
cat = tpf.query_skycatalog(additions=[
    dict(id="stellar companion 1", sep=0.3*u.arcsec, angle=Angle(131.3 * u.deg), delta_mag=1),
])

# TIC 237184773 is split to 1715469662, 1715469667
cat = tpf.query_skycatalog(subtractions=[237184773])
```

2. Are we concerned about the cases a bright star that is just outside of the tpf, but it significantly affects many pixels inside the tpf?
This is probably too much of an edge case. I suppose users can workaround it somewhat easily if they can add additional stars, say, via `tpf.query_skycatalog(additions=...)`.

src/lightkurve/utils.py
```diff
@@ -871,3 +875,224 @@ def finalize_notebook_url(notebook_url):
...
+    c1_pm_correct = SkyCoord(result["RA"], result["DEC"], unit="deg")
+
+    # Then calculate the separation based on pm corrected values
+    sep_pm_correct = coord_pm_correct.separation(c1_pm_correct)
+
+    # Provide the separation in the output table
+    result["Separation"] = sep_pm_correct
+
+    # Calculate the relative flux
+    result["Relative_Flux"] = np.exp(
```
orionlee, line 1087: I believe the relative flux calculation is buggy. `np.exp()` should probably be replaced by `10 **`. 

E.g., assume the target star is of magnitude 10, the nearby star is of magnitude 9, i.e., the relative flux is about ~2.512.

```
np.exp((9 - 10) / -2.5) : 1.492
10 ** ((9 - 10) / -2.5) : 2.512
```

orionlee (changes requested):

PR #1370, opened 2023-10-19
Update searching-for-data-products

Nschanche:
Modified text to reflect the results of the Kepler search in the tutorial "Searching & downloading Kepler, K2, and TESS data"


christinahedges:
LGTM!

PR #1374, opened 2023-10-23
Remove redundant conditional 

### LINKED ISSUE #1192: Same condition on lines 176 and 178 in regressioncorrector.py
What does it should be?

https://github.com/lightkurve/lightkurve/blob/7489f7a6d3eecf7c792ebd85d86c8030c02d9dc6/src/lightkurve/correctors/regressioncorrector.py#L176

https://github.com/lightkurve/lightkurve/blob/7489f7a6d3eecf7c792ebd85d86c8030c02d9dc6/src/lightkurve/correctors/regressioncorrector.py#L178

```
if prior_sigma is not None:
            sigma_w_inv = sigma_w_inv + np.diag(1.0 / prior_sigma ** 2)
if prior_sigma is not None:
            B = B + (prior_mu / prior_sigma ** 2)
```

arda-guler:
It seems like this small issue was noticed a while ago but never fixed in the main repository. Fixes #1192 .
The same conditional appears just two lines above the line I deleted.

### DIFF
--- a/src/lightkurve/correctors/regressioncorrector.py
+++ b/src/lightkurve/correctors/regressioncorrector.py
@@ -177,7 +177,6 @@ class RegressionCorrector(Corrector):
 
         if prior_sigma is not None:
             sigma_w_inv = sigma_w_inv + np.diag(1.0 / prior_sigma ** 2)
-        if prior_sigma is not None:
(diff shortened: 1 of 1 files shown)


christinahedges:
This looks good to me!

Issue #1073, opened 2021-05-20
search_lightcurve() by TIC id ended up doing a cone search


orionlee:
#### Problem description
When searching for lightcurves by TIC id, in some cases (*), the current logic will end up doing:
1. an exact match search with a incorrect id, with no results
2. a cone search as a fallback.

Users do end up getting the results they wanted. But there are two side effects:
1. lightkurve end up making extra MAST queries
2. if target is in a crowded area, the users might get extra results of other TICs.

(*) The issue happened when the tic id has fewer than 9 digits.

I don't know if the issue also applies to:
1. other search methods for TESS products (possibly)
2. Kepler / K2 products (probably not)

#### Example

```python
import lightkurve as lk
import logging

# don't care, but need to run it to get a rerference of lk.search.log
sr = lk.search_lightcurve("TIC123456789", mission="TESS") 

# actual test
lk.search.log.setLevel("DEBUG")
sr = lk.search_lightcurve("TiC74534430", mission="TESS")
```

The log shows that internally, the exact target search failed:
```python
Started querying MAST for observations with the exact target_name='074534430'.
No observations found. Now performing a cone search instead.
Started querying MAST for observations within 0.0001 arcsec arcsec of objectname='TiC74534430'.
MAST found 7 observations. Now querying MAST for the corresponding data products.
MAST found 6 matching data products.
```

#### Environment

-  platform (e.g. Linux, OSX, Windows): Windows
-  lightkurve version (e.g. 1.0b6): 2.0.9
-  installation method (e.g. pip, conda, source): conda

orionlee:
The issue lies in  `_query_mast()`: https://github.com/lightkurve/lightkurve/blob/3a0e979041e5944cd32e7552923090bce44d8c63/src/lightkurve/search.py#L1116-L1118

- For a TIC with ID fewer than 9 digits, e.g., `TIC74534430`, the logic will zero-pad it, `074534430` in the example, as the exact match term in search.
- MAST will return no results.
- Not zero-padding it, `74534430` in the example, will work.
- Given Kepler / K2 ID search has the same logic, I don't know if it is the case that 
  - MAST handles TESS differently than Kepler/K2, or
  - MAST at some point changed the behavior so that zero-padding must not be done.

orionlee:
[...]

PR #1383, opened 2023-11-03
Read in new HLSPs, fix search

christinahedges:
Right now there are some problems with search which are only becoming apparent as we have more and more data from TESS;

* Right now if you search for a TIC, KIC or EPIC explicitly it will only show you TESS, Kepler, or K2 data. This is helpful because it does an object query not a cone search. I'm changing this so that it will only do an object query if the ID and **author** is set
* I added an author alias for "SPOC" as "TESS" to be inline with "Kepler" and "K2" being the mission product authors
* I added a mission product column to help people figure out which products are "official"
* I added start and end times for the datasets as `astropy.time.Time` objects, and implemented a slightly hacky fix for the incorrect Kepler dates.
* I split the mission column into `mission` and `sequence` so that it's easier for people to narrow down based on a mission or a particular sequence of quarter, campaign, sector etc.

For readers I'm adding in a reader for the new KBONUS HLSP.

This is a weird one, because there are extensions in the file for each quarter, so we need a way to return a LightCurveCollection as well as a LightCurve. I've added a special case for if quarters are passed.

@jorgemarpa is going to write a tutorial on how to use this reader that I can add to the docs.

### DIFF
--- a/src/lightkurve/io/kbonus.py
+++ b/src/lightkurve/io/kbonus.py
@@ -1,0 +1,123 @@
+"""Reader function for KBONUS-BKG community light curve products."""
+from ..lightcurve import KeplerLightCurve
+from ..collections import LightCurveCollection
+from .generic import read_generic_lightcurve
+from astropy.io import fits
+import numpy as np
+
+import logging
+
+log = logging.getLogger(__name__)
+
+
+def _warn_bonus_file(filename):
+    with fits.open(filename, lazy_load_hdus=True) as hdulist:
+        if hdulist[0].header["KEPMAG"] < 12:
+            log.warning(
+                "Kepler magnitude is bright (less than 12), indicating the target is saturated. KBONUS-BKG data is invalid for saturated targets."
+            )
+        quarter_map = {
+            ext.header["quarter"]: idx
+            for idx, ext in enumerate(hdulist)
+            if "QUARTER" in ext.header
+        }
+        psffrac = np.asarray(
+            [
+                ext.header["PSFFRAC"]
+                for idx, ext in enumerate(hdulist)
+                if "QUARTER" in ext.header
+            ]
+        )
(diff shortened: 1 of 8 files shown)


scfleming:
"I added start and end times for the datasets as astropy.time.Time objects, and implemented a slightly hacky fix for the incorrect Kepler dates." --> which Kepler dates are incorrect?  If it's a database issue I'd love to see if that's something we could improve at MAST if there are inaccurracies in any fo the reported dates/times.

christinahedges:
@scfleming it's the `t_min` and `t_max` for Kepler that is incorrect. They are the start and end of the mission, not the start and end of the dataset.

orionlee:
> Right now if you search for a TIC, KIC or EPIC explicitly it will only show you TESS, Kepler, or K2 data. This is helpful because it does an object query not a cone search

A related issue: currently, there is a known bug of Search for TIC doing a cone search #1073

src/lightkurve/io/kbonus.py
```diff
@@ -0,0 +1,105 @@
...
+import numpy as np
+
+import logging
+
+log = logging.getLogger(__name__)
+
+
+def _warn_bonus_file(filename):
+    with fits.open(filename, lazy_load_hdus=True) as hdulist:
+        if hdulist[0].header["KEPMAG"] < 12:
```
jorgemarpa, line 15: this will crash when "KEPMAG" is empty for sources not in the KIC. We should also compare to GMAG

jorgemarpa (commented):
partial review

src/lightkurve/io/kbonus.py
christinahedges: great, is GMAG always set?

jorgemarpa: Yes, GMAG always has values.

PR #1402, opened 2024-01-30
Enhanced tpf.interact(): return pixel selection optionally.

### LINKED ISSUE #1386: Enhance `TargetPixelFile.interact()` to return the selected pixels programmaticaly
#### Problem description
Currently, when an user selects custom pixels via `tpf.interact()`, the user cannot use the selected pixels (as an aperture mask) in subsequent work outside of `interact()`

The user can only indirectly access it by exporting the lightcurve and read it back.

Proposal: make `tpf.interact()` to optionally return the the selected pixels as an aperture mask, if users specify a new `return_selected_mask` paramaeter

#### Example

```python
# return_selected_mask is the new parameter
_,  interact_mask = tpf_trunc.interact(return_selected_mask=True);
```

Here an user selects custom pixels to highlight a background spike due to SSO.
image

Users can then access the mask in subsequent cells in the notebook, creating a lightcurve, highlight the pixels in `plot()`, `interact_sky()`, `plot_pixels(), `etc.
```
# make a copy of the selection to avoid accidental pixel selection changesin the interact.
sso_mask = interact_mask.copy()
```

#### Environment

-  lightkurve version (e.g. 1.0b6): v2.4.2

orionlee:
Close #1386

it requires a slight public API change for `tpf.interact()` (a new optional parameter and the correspond return object).

TODOs:
- [x] make the API more generic to leave room for future additions: return a `dict` object that holds the selected pixels mask, rather than the mask directly.
- [x] correspond tests
- [x] docstring update

Misc. changes from review:
- [ ] rename `return_selection_mask` to `return_selected_mask`.

Changelog:
```rst
- Added ``return_selection_mask`` parameter to ``tpf.interact()``, to optionally 
  return the pixel selection as an aperture mask. [#1402]
```

### DIFF
--- a/src/lightkurve/interact.py
+++ b/src/lightkurve/interact.py
@@ -1009,6 +1009,7 @@ def show_interact_widget(
     vmax=None,
     scale="log",
     cmap="Viridis256",
+    return_selected_mask=False,
 ):
     """Display an interactive Jupyter Notebook widget to inspect the pixel data.
 
@@ -1073,6 +1074,9 @@ def show_interact_widget(
         Maximum color scale for tpf figure
     cmap: str
         Colormap to use for tpf plot. Default is 'Viridis256'
+    return_selected_mask: bool
+        Optional, if set to `True`, also return the pixel selection as an aperture mask.
+    TODO: 1) document return result, 2) update for the docstring in tpf.interact(), 3) remind users to copy the result once satisfied.
     """
     if _BOKEH_IMPORT_ERROR is not None:
         log.error(
@@ -1113,6 +1117,9 @@ def show_interact_widget(
         aperture_mask[0, 0] = True
 
     lc.meta["APERTURE_MASK"] = aperture_mask
+
+    # a copy of current pixel current selection for the use of caller
+    selected_mask_to_return = aperture_mask.copy()
 
     if transform_func is not None:
         lc = transform_func(lc)
@@ -1222,10 +1229,12 @@ def show_interact_widget(
                     ylims = _to_unitless(ylim_func(lc_new))
                 fig_lc.y_range.start = ylims[0]
                 fig_lc.y_range.end = ylims[1]
+                np.copyto(selected_mask_to_return,  lc_new.meta["APERTURE_MASK"])  # Update selected_mask_to_return
             else:
                 lc_source.data["flux"] = lc.flux.value * 0.0
                 fig_lc.y_range.start = -1
                 fig_lc.y_range.end = 1
+                selected_mask_to_return.fill(False)  # Update selected_mask_to_return
 
             message_on_save.text = " "
             export_button.button_type = "success"
@@ -1312,7 +1321,11 @@ def show_interact_widget(
         doc.add_root(widgets_and_figures)
 
     output_notebook(verbose=False, hide_banner=True)
-    return show(create_interact_ui, notebook_url=notebook_url)
+    show_result = show(create_interact_ui, notebook_url=notebook_url)
+    if return_selected_mask:
+        return show_result, selected_mask_to_return
+    else:
+        return show_result
 
 
 
--- a/src/lightkurve/interact.py
+++ b/src/lightkurve/interact.py
@@ -1009,7 +1009,7 @@ def show_interact_widget(
     vmax=None,
     scale="log",
     cmap="Viridis256",
-    return_selected_mask=False,
+    return_selection_mask=False,
 ):
     """Display an interactive Jupyter Notebook widget to inspect the pixel data.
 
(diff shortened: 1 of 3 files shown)


orionlee:
The PR is pending feedback on the API changes. @keatonb , thoughts?

- it adds an optional `return_result` parameter to `tpf.interact()
- if `return_result` is `True`, it returns a `dict` . The key `"selection_mask"` points to the selected pixels mask.


See the example in https://github.com/lightkurve/lightkurve/issues/1386#issue-1989151049

keatonb:
I think this is a very nice idea! I'm not a core lightkurve developer, so I don't have feedback to provide on API decisions. It's not intuitive to me that `return_result` would mean aperture mask. What other attributes does the `return_result` have? The corresponding light curve?

orionlee:
> It's not intuitive to me that `return_result` would mean aperture mask.

keatonb, thanks for the feedback.

`return_result` is indeed cryptic, and would require docstring and examples to mitigate it  if this path is chosen. Currently, it returns a `dict` with only 1 item `"selection_mask"`.

This is done to leave room say, if we want to return additional data without changing the API again, conceivably, the x / y limits (that an user zooms), the single cadence / time the user taps.

The alternative would make the API change specific to the mask, e.g., a parameter `return_selected_mask` that returns the mask `ndarray`.

Nschanche:
I agree the dictionary could make sense as it may be easier add additional parameters down the line, but I think returning an ndarray makes more sense in lightkurve. I would lean toward a parameter defaulting to false return_mask=False.

orionlee:
The API is reverted back to directly returning the mask array. 

Docstring and tests are added. The PR is ready to be merged, pending any further feedback.

Example:
```python
interact_mask = tpf.interact(return_selection_mask=True);

# make a copy of the selection to avoid accidental pixel selection changes in the interact.
my_custom_mask = interact_mask.copy()
```

src/lightkurve/targetpixelfile.py
```diff
@@ -1365,6 +1376,13 @@ def interact(
             >>> transform_func = lambda lc: lc.normalize()  # doctest: +SKIP
             >>> tpf.interact(ylim_func=ylim_func, transform_func=transform_func)  # doctest: +SKIP
 
+        the lightcurve after each pixel selection::
+
+            >>> interact_mask = tpf.interact(return_selection_mask=True)  # doctest: +SKIP
+            >>> # Once the desired pixels have been selected, save the result in
+            >>> # a separate variable to "freeze" the selection.
+            >>> my_custom_mask = interact_mask.copy()  # doctest: +SKIP
```
Nschanche, line 1384: Does this need to be copied to edit interact_mask? I don't think that is the case, so I might skip the my_custom_mask example in the header.

src/lightkurve/interact.py
```diff
@@ -1000,6 +1000,7 @@ def show_interact_widget(
     vmax=None,
     scale="log",
     cmap="Viridis256",
+    return_selection_mask=False,
```
Nschanche, line 1003: I think return_selected_mask is more apt (rather than selection) as it is actively selected when using interact.

tests/test_interact.py
```diff
@@ -61,6 +61,18 @@ def test_graceful_exit_outside_notebook():
...
+@pytest.mark.skipif(bad_optional_imports, reason="requires bokeh")
+def test_return_selection_mask():
+    """Test to `return_selection_mask=True` can run without any syntax error."""
+    import bokeh
+
+    tpf = TessTargetPixelFile(example_tpf)
+    mask_to_use = tpf.create_threshold_mask()
+    selection_mask = tpf.interact(aperture_mask=mask_to_use, return_selection_mask=True)
+    # the returned mask should be the same as the supplied one initially
+    assert_array_equal(selection_mask, mask_to_use)
```
Nschanche, line 73: Looks like a good test to me.

Nschanche (commented):
Great added functionality!

src/lightkurve/interact.py
orionlee: We could rename `return_selection_mask` to `return_selected_mask`.
I'll wait for any other suggested changes from review feedback, and update the codes in 1 batch.

src/lightkurve/targetpixelfile.py
orionlee: `interact_mask.copy()` is not strictly necessary, but is often needed in practice. That's why I include it in the example.

Reason:
`interact_mask` reflects the live / active selection. In practice, the selection can easily be inadvertently changed, surprising users. Scenario:
1. run `interact()` in a cell, and select a custom aperture
2. use the custom aperture (in `interact_mask`) to create a lightcurve.
3. do something else in other cells.
4. try to use the custom aperture again.

In my experience, after step 3 (do something else in other cells), in many cases the selection in `interact()` would be changed (to no pixel selected usually) [*]. In such scenario, if users do not pay attention and use the live `interact_mask`, they would have used a different aperture from what they initially selected.

The step of `my_custom_mask = interact_mask.copy()` in the example is one way to "freeze" / save the custom aperture mask while the users are still conscious of what they have just selected, avoiding any surprise from inadvertent changes.

---
[*] After I use `tpf.interact()`, scroll to other parts of the notebook, and scroll back to `tpf.interact()` cell, in many cases the selection is somehow lost:

image

Issue #1401, opened 2024-01-23
`centroid_quadratic()` incorrectly failed in a boundary case, when `mask` is specified and fluxes are all negative


orionlee:
#### Problem description

`lk.utils.centroid_quadratic()` would fail and return `(nan, nan)` in a specific boundary case:
- `mask` argument is specified, and
- the flux in `data` is all negative.

While all negative flux should not happen in normal flux data, they could possibly arise when the data is from a difference flux.

#### Example

See: https://gist.github.com/orionlee/fb9a855c3f4fe935ef653d72910768ac

#### Expected behavior

It should still return a reasonable centroid.

#### Environment

-  lightkurve version (e.g. 1.0b6): 2.4.2

orionlee:
The root cause is when `mask` is specified, `centroid_quadratic()` sets flux of the masked pixels to 0.

https://github.com/lightkurve/lightkurve/blob/9dbb63433fcaa391f45a3fccda65d9ac1bb96340/src/lightkurve/utils.py#L632-L633

In the case where flux is all negative and `mask` is specified, the 3x3 patch chosen will all be the masked pixels, with uniform value of 0.

Nschanche:
Hi @orionlee , thanks for the great notebook demonstrating the issue. I think this could be resolved with an extra step to ensure that 0 will not be the max location:

```
    if mask is not None:
        data = data * mask
    data_masked = data.copy()
    data_masked[data_masked == 0] = np.nan
    arg_data_max = np.nanargmax(data_masked)
'''

Nschanche:
I've put in PR #1414, which sets the masked values to nan before identifying the pixel with max flux. This appears to work with the examples in the gist @orionlee provided.

PR #1414, opened 2024-03-08
Issue #1401 - centroid_quadratic

Nschanche:
This PR addresses #1401 which raises the issue that estimate_centroid fails when all flux values are negative and a mask is provided. 
When a mask is provided, this PR sets the mask value as nans rather than 0 before locating the brightest pixel. When all fluxes are negative, this allows for a negative max value location.

### DIFF
--- a/src/lightkurve/utils.py
+++ b/src/lightkurve/utils.py
@@ -629,15 +629,22 @@ def centroid_quadratic(data, mask=None):
         data = data.value
     # Step 1: identify the patch of 3x3 pixels (z_)
     # that is centered on the brightest pixel (xx, yy)
+        
+     # Issue 1401 demonstrates that using 'data' to find the max will break when all flux is negative
+    # replacing 0 with nan allows the correct max to be identified. 
     if mask is not None:
         data = data * mask
-    # Issue 1401 demonstrates that using 'data' to find the max will break when all flux is negative
-    # replacing 0 with nan allows the correct max to be identified. 
-    data_masked = data.copy()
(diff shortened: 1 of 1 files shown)


orionlee:
Comments:
1. The fix looks good to me with one small concern. In the edge case that 
  a. fluxes are all negative, and
  b. the (correctly) identified 3x3 patch around the brightest pixel (`Z_`) happen to contain masked pixels.

The centroid returned would be slightly skewed because the masked pixel will have value 0 in the otherwise all negative 3x3 patch. 
E.g., if the lower right corner pixel of the 3x3 patch is masked, the patch could be:
```
-3, -3, -3
-4, -9, -3
-3, -2, 0(because of masking)
```

The resulting centroid would be skewed towards lower right, as 0 is significantly larger than all other negative values.


Suggestion: instead of setting masked pixels to `nan`, set them to some value smaller than the existing minimum. (The codes are shorter too).
```python
    # Issue 1401 demonstrates that using 'data' to find the max will break when all flux is negative
    # set masked pixels to be smaller than the existing minimum (instead of 0) to resolve it.
    if mask is not None:
        data = data.copy()
        data[~mask] = np.nanmin(data) - 1
    arg_data_max = np.nanargmax(data)
```

2. It'd be great if corresponding tests are added.

orionlee:
@Nschanche I think new PR 1426 could supersede this one. It is more robust. Tests are included as well.

Nschanche:
Nice, I took a look at #1426 and it looks good.

Issue #1302, opened 2023-04-24
Removing NaN fluxs values in TESS target pixel files for DesignMatrix


JTregloanReed:
#### Problem description
I am downloading a dataset, which sadly contains some NaN values in some of the flux cadences (the entire cadence has NaN flux values). To preform a Regression Corrector I need to use the TPF in the DesignMatrix, which outputs an error due to the data containing NaNs. I've tried the remove_nans function but I get the "AttributeError: 'Quantity' object has no 'remove_nans' member" error.

Is there a way to remove NaN flux values (or set them to zero) within the target pixel file?

#### Example
```python
import lightkurve as lk
import numpy as np
import astropy.units as u
import matplotlib.pyplot as plt
from lightkurve import TessTargetPixelFile
import multiprocessing as mp

# Search for target 

TARGET = 'TIC 413376180'

search_result = lk.search_targetpixelfile(TARGET,author='TESS-SPOC')

print(search_result)


# Download the TPF file and set cutout size
tpf = search_result.download(quality_bitmask='default')
#https://docs.lightkurve.org/reference/api/lightkurve.TargetPixelFileCollection.sector.html

#tpf = search_result.download(quality_bitmask='default', sector= 3)
# Set aperture to SPOC pipeline aperture. ##############################################
tpf[0].plot(aperture_mask=tpf.pipeline_mask, mask_color='red');
plt.show()

raw_lc = tpf.to_lightcurve(aperture_mask = tpf.pipeline_mask)

reg_aper = tpf.create_threshold_mask()
dm = lk.DesignMatrix(tpf.flux[:, ~reg_aper], name='regressors').pca(5).append_constant()
rc = lk.RegressionCorrector(raw_lc)
corrector_lc = rc.correct(dm)

corrected_lc = raw_lc - rc.model_lc + np.percentile(rc.model_lc.flux, 1)

```

#### Expected behavior
I would of thought it would of created a regression corrector, after ignoring any NaN values.

#### Environment

-  platform (Linux, Ubuntu 20.04):
-  lightkurve version (e.g. 2.4):
-  installation method (e.g. pip):

Nschanche:
Hi there, thanks for your comment! 
[...]

PR #1415, opened 2024-03-13
Issue1302 qualityflag

Nschanche:
Added new quality flags to the default exclude.  See issue #1302.


Nschanche:
This would be addressed by PR #1336, which updates the default quality mask.

PR #1419, opened 2024-04-16
Workaround for poetry install error "Package docutils (0.21.post1) not found"

orionlee:
Currently, `poetry install` would fail with error "Package docutils (0.21.post1) not found".

CIs would also thus fail: [sample runs](https://github.com/orionlee/lightkurve/actions/runs/8710041215/job/23891040112#step:4:177)

Applying the workaround from:
https://github.com/python-poetry/poetry/issues/9293#issuecomment-2048205226

Note: the root cause seems to be pypi / docutils . When they fix the underlying issue, the workaround would no longer be needed.


orionlee:
I am merging the workaround, as the problem still persists.

PR #1428, opened 2024-05-13
Fix bokeh 3.4.0 deprecation warnings for `circle()`, `cross()`

### LINKED ISSUE #1424: `tpf.interact()` and `lc.interact_bls()`: handle  bokeh deprecation of `circle()` starting v3.4.0
#### Problem description

For bokeh-based `tpf.interact()` and `lc.interact_bls()`, the implementation should be updated to handle:

`BokehDeprecationWarning: 'circle() method with size value' was deprecated in Bokeh 3.4.0 and will be removed, use 'scatter(size=...) instead' instead.`

Additionally, `tpf.interact_sky()` emits 2 warnings
```
BokehDeprecationWarning: 'circle() method with size value' was deprecated in Bokeh 3.4.0 and will be removed, use 'scatter(size=...) instead' instead.
BokehDeprecationWarning: 'cross() method' was deprecated in Bokeh 3.4.0 and will be removed, use "scatter(marker='cross', ...) instead" instead.
```

#### Environment

-  lightkurve version (e.g. 1.0b6): 2.4.2

orionlee:
Close #1424 

Manually tested on bokeh 3.4.1 and bokeh 2.4.3 . No regression found.

changelog:
```rst
- Fixed interact features, e.g. ``tpf.interact()``, to be compliant
  with Bokeh v3.4.0. The minimum Bokeh version is raised to v2.3.2 accordingly. [#1428]
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@
   lightcurve is normalized with ``transform_func``. [#1387]
 - Fixed ``lightkurve.utils.centroid_quadratic()`` in edge cases, e.g., fluxes are
   all negative with mask specified, NaN in the identified brightest 3X3 patch. [#1426]
+- Fixed interact features, e.g. ``tpf.interact()``, to be compliant
+  with Bokeh v3.4.0. The minimum Bokeh version is raised to v2.3.2 accordingly. [#1428]
 
 
 2.4.2 (2023-11-03)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,8 +38,8 @@ pandas = ">=1.1.4"
 uncertainties = ">=3.1.4"
 patsy = ">=0.5.0"
 fbpca = ">=1.0"
-# bokeh v2+ requirement due to https://github.com/lightkurve/lightkurve/issues/1261
-bokeh = ">=2.0.0"
+# bokeh v2.3.2+ requirement due to https://github.com/lightkurve/lightkurve/pull/1428
+bokeh = ">=2.3.2"
 memoization = { version = ">=0.3.1", python = ">=3.8,<4.0" }
 scikit-learn = ">=0.24.0"
 
--- a/src/lightkurve/interact.py
+++ b/src/lightkurve/interact.py
@@ -336,7 +336,7 @@ def make_lightcurve_figure_elements(lc, lc_source, ylim_func=None):
         nonselection_line_color="gray",
         nonselection_line_alpha=1.0,
     )
-    circ = fig.circle(
+    circ = fig.scatter(
         "time",
         "flux",
         source=lc_source,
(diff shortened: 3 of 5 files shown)


orionlee:
@Nschanche A small fix that has no surprise. I intend to merge this PR in the next few days.

Nschanche:
The required bokeh version for lightkurve appears to be 2.0.0. Adding multiple marker types seems to have been added to bokeh in 2.4.3, so we should probably change the requirement in the toml file before merging.

orionlee:
Good catch! The minimum bokeh requirement is raised to 2.3.2. 
`tpf.interact()`, `tpf.interact_sky()` and `lc.interact_bls()` were manually tested.

- the scatter with marker api is available from [v2.3.0](https://github.com/bokeh/bokeh/blob/2.3.0/bokeh/models/glyphs.py) 
- but it turns out that there is a bigger issue with old bokeh, which uses the (now) removed `jinja2.Markup`, causing import failure of bokeh `show_notebook()`. It is fixed in [v2.3.2](https://github.com/bokeh/bokeh/pull/11174),

orionlee:
An additional commit is made to tweak how `lightkurve` reports bokeh import issues. In addition to emitting a friendly message, it now also throws the actual import error, to assist users to diagnose the problem.

---
Currently, if there is any error during bokeh import, `lightkurve` will silently consume the error, assuming bokeh is not installed, and emit an error message asking users to install bokeh when the relevant function is called, e.g., `tpf.interact()`.

It could lead to confusion when users do have bokeh installed, but there is some other issues during import. 

E.g., if users have bokeh 2.3.0 installed along with some recent version of jinja2, calling `tpf.interact()` would fail with a head-scratching `NameError`:
```python
-> 1347 output_notebook(verbose=False, hide_banner=True)
   1348 return show(create_interact_ui, notebook_url=notebook_url)

NameError: name 'output_notebook' is not defined
```

The commit changes it so that it displays the actual import error when calling `tpf.interact()`.
The import error can be served as hints that there is some incompatibility between bokeh and jinja2 in the installation.
```python
The interact() tool requires the `bokeh` Python package; you can install bokeh using e.g. `conda install bokeh`.
ImportError                               Traceback (most recent call last)
Cell In[2], line 10
...
...
File C:\pkg\_winNonPortables\miniforge3\envs\lk_dev\lib\site-packages\bokeh\core\templates.py:42
     39 from os.path import dirname, join
     41 # External imports
---> 42 from jinja2 import Environment, FileSystemLoader, Markup
     44 #-----------------------------------------------------------------------------
     45 # Globals and constants
     46 #-----------------------------------------------------------------------------
     48 __all__ = (
     49     'AUTOLOAD_JS',
     50     'AUTOLOAD_NB_JS',
   (...)
     63     'SCRIPT_TAG',
     64 )

ImportError: cannot import name 'Markup' from 'jinja2' (C:\pkg\_winNonPortables\miniforge3\envs\lk_dev\lib\site-packages\jinja2\__init__.py)

```

Issue #1430, opened 2024-05-14
Slow looping when plotting the aperture over the TPF of saturated stars


astrobatty:
#### Problem description

Aperture mask plotting over large TPFs is very slow, because `lightkurve` loops over each masked pixel and use `matplotlib Rectangle` to highlight pixels. This happens [here](https://github.com/lightkurve/lightkurve/blob/c3cb1b37df1ea25b5ef2fe5e4a2ccaa20d5aea21/src/lightkurve/targetpixelfile.py#L1196).

#### Example

```python
import lightkurve as lk

tpf = lk.search_targetpixelfile('Beta Gem', sector=20).download()
tpf.plot(aperture_mask='pipeline')
```

#### Expected behavior

According to [this](https://stackoverflow.com/questions/54463269/avoid-slow-looping-when-plotting-irregular-raster-plot-using-patches-rectangle) Stack Overflow issue, `lightkurve` should use `PolyCollection` instead of `Rectangle`.

#### Environment

-  platform independent
-  lightkurve version (e.g. 1.0b6): 2.4.2
-  installation method (e.g. pip, conda, source): any

orionlee:
An estimate of the actual impact using a bright star (TIC 233798897, Tmag 4.2) sector 76 SPOC TPF.
- 11x33 pixels in the TPF, with 78 pixels in SPOC aperture

On a recent laptop:
- `tpf.plot(aperture_mask=None)      ` - CPU times: total: 62.5 ms, Wall time: 89.5 ms
- `tpf.plot(aperture_mask="pipeline")` - CPU times: total: 391 ms, Wall time: 720 ms

---

![image](https://github.com/lightkurve/lightkurve/assets/250644/8eef33a4-abd2-4b08-946b-1f77a2f1d522)

Nschanche:
Thanks @astrobatty for pointing this out. The loop is really slowing things down in this case. Changing the function to

```
        if aperture_mask is not None:
            aperture_mask = self._parse_aperture_mask(aperture_mask)
            in_aperture = np.where(aperture_mask)
            if hasattr(ax, "wcs"):
                ap_row = in_aperture[0] - 0.5
                ap_col = in_aperture[1] - 0.5
            else:
                ap_row = in_aperture[0] + self.row - 0.5
                ap_col = in_aperture[1] + self.column - 0.5    
            for ii in range(len(ap_row)):
                
[...]

PR #1434, opened 2024-05-24
Rectangles

Nschanche:
This PR is to speed up the aperture mask plotting. Lightkurve issue #1430 pointed out that for extended sources, over-plotting the aperture mask onto the image was very slow. This PR reduces the plotting speed from 1min 24s ± 1.96 s per loop to 1.33 s ± 15.1 ms for the case of the star 'Beta Gem'.


tylerapritchard (approved):
looks good to me!

PR #1443, opened 2024-06-28
fix KEPSEISMIC time format from bkjd to mjd

kecnry:
From the metadata, it seems the correct format is MJD (based on `MJD-BEG` and `MJD-END` entries).  Without this PR, calling `lc.time.iso` gives timestamps in the year 2159.


scfleming:
To provide some additional context, we are soon working on adding interactive previews back into Portal using a package Kyle is writing that uses lightkurve to load the data. While not an explicit blocker, any residual PRs can be merged into a version of lightkurve would add to the number of supported datasets at initial release. If things go well on our end, that could be as soon as July 30th as part of our MAST monthly patch for July.

PR #1451, opened 2024-07-17
Read LCs and TPFs from the Cloud

snbianco:
This PR allows users to read data products from the cloud using S3 URIs.

- Users can read data products (light curve files and TPFs) from the AWS S3 cloud.
- New functions to read a collection of light curves or a collection of TPFs.
  - Optional `stitch` parameter in `read_lc_collection` to return a single light curve from a collection. 
- Added relevant test cases to `test_read.py`.
- Fixed out-of-date astropy import in `test_periodogram.py` and `test_synthetic_data.py`

### DIFF
--- a/src/lightkurve/io/read.py
+++ b/src/lightkurve/io/read.py
@@ -202,7 +202,7 @@ def _read_collection(path_list, products, *, stitch=False, **kwargs):
     else:
         return TargetPixelFileCollection(prod_list)
     
-def read_lc_collection(path_list, *, stitch=True, **kwargs):
+def read_lc_collection(path_list, *, stitch=False, **kwargs):
     return _read_collection(path_list, [TessLightCurve, KeplerLightCurve], stitch=stitch, **kwargs)
 
 def read_tpf_collection(path_list, **kwargs):
--- a/src/lightkurve/io/read.py
+++ b/src/lightkurve/io/read.py
@@ -1,18 +1,22 @@
 """Functions for reading light curve data."""
 import logging
+import gzip
 
 from astropy.io import fits
 from astropy.utils import deprecated
 import s3fs
 
+from lightkurve.targetpixelfile import KeplerTargetPixelFile, TessTargetPixelFile
+
 from ..lightcurve import KeplerLightCurve, TessLightCurve
+from ..collections import LightCurveCollection, TargetPixelFileCollection
 from ..utils import LightkurveDeprecationWarning, LightkurveError
 from .detect import detect_filetype
 
 log = logging.getLogger(__name__)
 
 
-__all__ = ["open", "read"]
+__all__ = ["open", "read", "read_lc_collection", "read_tpf_collection"]
 
 
 @deprecated("2.0", alternative="read()", warning_type=LightkurveDeprecationWarning)
@@ -84,8 +88,14 @@ def read(path_or_url, **kwargs):
             # path_or_url is an S3 cloud URI
             fs = s3fs.S3FileSystem(anon=True)
             with fs.open(path_or_url, 'rb') as f:
-                with fits.open(f) as temp:
-                    filetype = detect_filetype(temp)
+                if f.key.endswith('.gz'):
+                    # Unpack if file is in gzip format (Kepler TPFs)
+                    with gzip.open(f) as g:
+                        with fits.open(g) as temp:
+                            filetype = detect_filetype(temp)
+                else:
+                    with fits.open(f) as temp:
+                        filetype = detect_filetype(temp)
         else:
             with fits.open(path_or_url) as temp:
                 filetype = detect_filetype(temp)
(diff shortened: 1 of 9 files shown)


snbianco:
@ttdu

dr-rodriguez:
@barentsen this PR is ready for review

ttdu:
I'm not sure how active Geert is in this repo anymore; perhaps @tylerapritchard would be the right person to tag for review?

tylerapritchard:
Thanks Sam, Thomas!  This looks great in general, reviewing the details now.  So far, one quick question:
When opening a file you have code blocks where you open a remote file using s3fs, and then unzip the file, that are along the lines of:
```python
        elif (isinstance(path, str) and path.startswith('s3://')):
            # Filename is an S3 cloud URI
            fs = s3fs.S3FileSystem(anon=True)
            self.fs_file = fs.open(path, 'rb')
            if self.fs_file.key.endswith('gz'):
                # Unpack if file is in gzip format (Kepler TPFs)
                self.gz_file = gzip.open(self.fs_file)
                self.hdu = fits.open(self.gz_file)
            else:
                self.hdu = fits.open(self.fs_file)
```
in [src/lightkurve/targetpixelfile.py](https://github.com/lightkurve/lightkurve/pull/1451/files#diff-e3549d78be9d53bdbf0a975e794b975a2862e0e28a12ee9a19078f7a5a71a6ab), [lightkurve.io.read](https://github.com/lightkurve/lightkurve/pull/1451/files#diff-f4a70db96429a6148c3f986e5aecd65a86bdc073d530d31499c9e53b762492b4), [lightkurve.io.generic](https://github.com/lightkurve/lightkurve/pull/1451/files#diff-322bb11b4998e4bc05708b04f55b005c9703d75ddddbf9cd9ac54c2da6d8d678)

Astropy has the ability to [remote read cloud-hosted files using fsspec](https://docs.astropy.org/en/stable/io/fits/#working-with-remote-and-cloud-hosted-files) , including gzip files.  I think this could be simplified then to something like:

```
self.hdu = fits.open(path, use_fsspec=True, fsspec_kwargs={"anon": True})
```

Now, I suspect that your code might be more efficient than the astropy code - have you done any comparisons by any chance?  

In general we try to minimize duplicated functionality (and lines of code) to make maintenance easier and maintain readability,  but if the efficiency savings is substantial (or there's some complication with the astropy remote-read compression) than I think this is worth it.  fsspec uses s3fs so we're not adding a dependency either way.

tylerapritchard:
Can you add a summary (or summaries) of your changes to the change log in  [CHANGES.rst](https://github.com/lightkurve/lightkurve/blob/main/CHANGES.rst) please?

snbianco:
Using Astropy to open the files is definitely a better solution, thank you! There's not a significant difference in efficiency, and using Astropy keeps things a lot cleaner. I'll push a commit with those changes shortly!

src/lightkurve/io/read.py
```diff
@@ -153,3 +161,39 @@ def read(path_or_url, **kwargs):
...
+        except Exception as e:
+            log.warning(
+                f'Unable to read {path}: {e}. This file will not be added to the collection.'
+            )
+
+    if not prod_list:
+        log.warning(
+            'The resulting collection contains no products.'
+        )
+        return LightCurveCollection([]) if is_lc else TargetPixelFileCollection([])
```
tylerapritchard, line 188: can we remove this return statement?  

Since we defined prod_list as empty above, can we not just check if it is empty and print a warning here.  

Then, use the return below?

tylerapritchard (commented):
This is great!  I've left some suggestions on _read_collection that I think are worth taking a look at.  I think the changes might make it slightly more robust, but I could be wrong.

tylerapritchard:
I'm sorry, those comments make more sense if you view them by line in the review. I didn't realize they would get jumbled up chronologically.  

High level point was - I'm going to suggest that instead of sending an array of product types we instead send a single LightCurve or TargetPixelFile class and use isinstance() to compare our products against this (as both TESS/Kepler inherit from the top-level LightCurve/TargetPixelFile classes).

tylerapritchard:
I also have a usage question:  I'm assuming that the motivation for this functionality is cloud-based science platforms like TIKE, where the remote-data read is very fast and local storage is expensive. To use this functionality we need a S3URI, which is non-trivial to get out of Lightkurve natively.  

If this is the case, how are you envisioning users getting these S3 URIs?  I see three main methods:

1. Users write astroquery.mast.observations queries  
2. [lksearch](https://github.com/lightkurve/lksearch) for this & document how to access this information. 
3. We could build a similar functionality as we have in lksearch into lightkurve, where when we say lightkurve.searchresult.download(), and we have a configuration parameter set, we return a S3 path to the LC/TPF constructor instead of downloading the file and returning a local path.  This could then become the default behaviour on TIKE where you configure the configuration parameters, but not the default installed behaviour.  

I think we can merge this as is, particularly if you are not interested in the third option or if I misunderstood the use case for this feature, but I think this is worth considering.

ttdu:
You're exactly correct about our intentions; thanks for suggesting those options!

From a user perspective, option 3 hides cloud complexity; we want TIKE to "just work", so this is my strong preference. Sam, do you have any concerns about this option?

snbianco:
I agree that option 3 is ideal! I'll work on implementing it, but if it does prove to be non-trivial, I'll suggest that we move that to a follow-up PR.

Option 1 could also work in the meantime with the new streamlined version of `Observations.get_cloud_uris`. It makes it a bit easier to make a query and get the URIs for returned products.

tylerapritchard:
I think your intuition that this could be a separate PR is right on, thank you so much, this is great!  If you agree with my ```_read_collections```  suggestions above and want to implement, I think we can merge that as is and then start a new issue/PR for the potential change to ```lightkurve.searchresult.download()``` . 

Part of the reason why I bring this up now is that I also think that we are dangerously close to releasing a new lightkurve version, and would love if we could get at the very least this PR but ideally this and 3) merged before that. 

From our work putting together lksearch & its DOWNLOAD_CLOUD configuration parameter, I suspect that this isn't quite trivial but is hopefully straightforward.  I think this might actually be easier in lk as we don't need to add it to the results table, just check if the config parameter is set on calling SearchResult._download_one() and if then if the S3 URI exists given a call to Observations.get_cloud_uris from the supplied table row.

snbianco:
Yes, I thought your comments were great feedback and really helped simplify things! I just pushed those changes along with some method comments.

I'm hoping to take another look at what option 3 would require tomorrow, and I will make an issue/PR for it as well.

tylerapritchard:
This looks good to me and passes all checks.  Merging - thanks, this will be a great addition!  

Sounds good on the plan for a follow-up PR.  If the implementation of the URI portion is on a longer timescale, I think we can at the very least update our lksearch documentation showing how to get a S3 uri with this feature highlighting how we can now remote-read these S3 uri's in lightkurve as a potential source of documentation.

Issue #1023, opened 2021-04-05
Some of our dependencies should be optional (e.g. `scikit-learn`, `oktopus`)


christinahedges:
I believe we currently have dependencies like `scikit-learn` and `oktopus` that are great packages and super useful, but only in certain contexts (`CBVCorrector` for example is the only module that needs `scikit-learn` and the PRF fitting module which is quite out of date is the only part of lightkurve requiring 'oktopus'.)

I think these should be optional dependencies, as these packages can make it harder to install `lightkurve`.

barentsen:
I agree!

sakhee-17:
I am running into an error while installing lightkurve version 1.11.3. I am working in a virtual environment on macOS 10.15. 

Running this command:
```
python3 -m pip install lightkurve==1.11.3
```
generated this error:

```
ERROR: Could not find a version that satisfies the requirement oktopus (from lightkurve)
ERROR: No matching distribution found for oktopus
```
These two versions of oktopus it had tried installing were:
```
Using cached oktopus-0.1.2.tar.gz (10 kB)
Using cached oktopus-0.1.1.tar.gz (9.9 kB)
```

barentsen:
@sakhee-17 Thank you for reporting this issue!  I will investigate...

PR #1452, opened 2024-07-17
Make `oktopus` optional and/or get numpy-2-compatible `autograd`; fix deprecated `astropy.stats.bls` and `numpy` imports

dhomeier:
Addresses #1440 and #1023 by only conditionally importing `prf.tpfmodels`, allowing lightkurve to run with numpy 2.0.
Just restricted `oktopus` dependency to older Python versions where numpy 2 is not available for now; to be discussed how to best mark this as a dependency for optional functionality.
Also the import warning should perhaps only be issued when actually trying to use `tpfmodels` functionality, rather than on any import of lightkurve, but I am not sure how to best set this up.

This also removes the `BoxLeastSquares` imports from `astropy.stats`, which have been deprecated since astropy 3.2 and finally removed in 6.0, but where still present in some tests (which do not even use the class).

Tested with numpy 2.0.0 and current versions of other deps in Python 3.12 on macOS 15, direct numpy failures remain like:
> FAILED tests/correctors/test_pldcorrector.py::test_kepler_pld_corrector - AttributeError: module 'numpy' has no attribute 'product'
FAILED tests/test_targetpixelfile.py::test_tpf_ones[quadratic] - AttributeError: `np.asfarray` was removed in the NumPy 2.0 release. Use `np.asarray` with a proper dtype instead.. Did you mean: 'asarray'?
FAILED tests/test_targetpixelfile.py::test_tpf_from_images - AttributeError: `np.float_` was removed in the NumPy 2.0 release. Use `np.float64` instead.. Did you mean: 'float16'?

which are straightforward to fix, with some help from Ruff. Could not test this in the CI since some dependency evidently still pins astropy < 5.3 and numpy < 1.25.


Nschanche:
Hi @dhomeier , this looks great. I'll take a look at some of the additional issues you mentioned. @tylerapritchard would you like to also review this PR?

dhomeier:
I guess a less intrusive way to handle missing oktopus might be to redefine all (13) classes in `tpfmodels.py` to raise `NotImplementedError` or `ModuleNotFoundError` in the `NO_OKTOPUS` case.

.github/workflows/lightkurve-tests.yml
```diff
@@ -5,6 +5,7 @@
 # 2) pytest on Linux, Python 3.8, with --remote-data enabled (i.e. get data from MAST).
 # 3) pytest on Linux, Python 3.9.
 # 4) pytest on Linux, Python 3.10.
+# 4) pytest on Linux, Python 3.12.
```
Nschanche, line 8: renumber the tests

src/lightkurve/correctors/cbvcorrector.py
```diff
@@ -707,7 +707,7 @@ def _correct_initialization(self, cbv_type='SingleScale', cbv_indices='ALL',
                         cbv_idx_loop = cbvs.cbv_indices
                     # Trim to nCBVs in cbvs
                     cbv_idx_loop = np.array([idx for idx in cbv_idx_loop if
-                        bool(np.in1d(idx, cbvs.cbv_indices))])
+                        bool(np.isin(idx, cbvs.cbv_indices))])
```
Nschanche, line 710: Thanks for updated all of the in1d calls for numpy 2.0.

src/lightkurve/correctors/pldcorrector.py
```diff
@@ -250,7 +250,7 @@ def create_design_matrix(
             # Create a DesignMatrix for each PLD order
             all_pld = []
             for order in range(1, pld_order + 1):
-                reg_n = np.product(list(multichoose(regressors_pld.T, order)), axis=1).T
+                reg_n = np.prod(list(multichoose(regressors_pld.T, order)), axis=1).T
```
Nschanche, line 253: I didn't see this in the deprecation list for numpy, good catch

Nschanche (commented):
Could you update CHANGES.rst with your updates as well?

dhomeier:
OK, will update the numbers – perhaps one of the 3.11 or 3.12 tests without oktopus should also be run with `--remote-data` since most of the tpfmodel tests are skipped otherwise?

I thought about better controlling oktopus as an optional dependency, so it would be installed only as an extra for e.g. `pip install lightkurve[tpfmodel]`. I figured how to set this up for a separate, [optional dependency group](https://python-poetry.org/docs/managing-dependencies/#optional-groups), but not how to translate that into pip syntax (if poetry can do this at all) – any ideas?

dhomeier:
@Nschanche I have rebased and updated the test labels and changelog. Any thoughts how to address https://github.com/lightkurve/lightkurve/pull/1452#issuecomment-2253423550?

dhomeier:
It turned out the restrictions on Numpy and Astropy were internal after all, created by poetry's weird handling of dependencies: it enforces the versions of _all_ packages in _all_ environments to be supported for the entire range of Python versions – i.e. basically everything is restricted by Python 3.8 at this point.
- [ ] For the moment this could be alleviated a bit by moving to `python = "^3.9"`, but this is still barring current Astropy (and thus in turn, any version compatible with Numpy 2.0), and will also block Numpy 2.1.
I am wondering if it would not be better to revert CI to a standard pip install that can test various version combinations, especially since poetry is not summoned in the documented install process at all except for editable installs.
Another option would be to sync with the Numpy support scheme (like Astropy) and limit Python versions to now ">=3.10,<3.14".

dhomeier:
@Nschanche, @tylerapritchard are there any plans how to move forward with compatibility with Numpy 2.x and/or testing on Python 3.11-3.12?

dhomeier:
Hi @Nschanche, at first look your commit appeared quite massive, and I was wondering if this just needed a rebase to `main`, but now everything looks ok and tests with numpy 2.2 are running, so this should indeed be ready; thanks!

src/lightkurve/prf/__init__.py
```diff
@@ -1,2 +1,6 @@
+import warnings
 from .prfmodel import *
-from .tpfmodel import *
+try:
+    from .tpfmodel import *
+except ModuleNotFoundError:
+    warnings.warn("Warning: the tpfmodel module is not available without oktopus")
```
Nschanche, line 6: Should we add a reference to this issue so users can more easily understand this warning?

tests/prf/test_tpfmodel.py
```diff
@@ -7,20 +7,28 @@
...
-from lightkurve.prf import TPFModel, PRFPhotometry
 from lightkurve.prf import SimpleKeplerPRF, KeplerPRF
 
+try:
+    from oktopus import PoissonPosterior
+    from lightkurve.prf import FixedValuePrior, GaussianPrior, UniformPrior
+    from lightkurve.prf import StarPrior, BackgroundPrior, FocusPrior, MotionPrior
+    from lightkurve.prf import TPFModel, PRFPhotometry
+    NO_OKTOPUS = False
+except ModuleNotFoundError:
```
Nschanche, line 18: I would also suggest adding a comment here that octopus is a required dependency for tpfmodels only, and is not supported by numpy2.x

Nschanche:
@dhomeier I thought I could edit the code to add those comments, but it appears I can't. Could you add them, and then I can merge this in?

Nschanche:
A new poetry lock file will also be needed with the changes to the requirements

dhomeier:
@Nschanche, @christinahedges, situation has changed quite significantly again, as autograd turns out to be no longer unmaintained, and last week's release 1.8.0 now supports Numpy 2! Oktopus by itself never seemed to have any conflicts. :D

So I have updated the install and test requirements accordingly to reenable `tpfmodel` with newer versions, see [unskipped test_tpfmodel](https://github.com/lightkurve/lightkurve/actions/runs/14840805655/job/41663051513?pr=1452#step:5:39).
I've also replaced two `math` trigonometric function calls with their `numpy` counterparts to get rid of some 1500 deprecation warnings (as of numpy 1.25); apparently length-1 arrays are passed as `rotation_angle`, but I did not track that further down.
Also added a 3.13 test; maybe the macOS and Windows tests should be upgraded to somewhat newer Python versions as well?

The framework for the conditional `oktopus` imports is still in place, if for some other reason its installation is not possible or desired (e.g. my miniconda env did not let me install 0.1.2 with numpy>=2, seems conda has registered its own conflict list).
That could of course be reverted, but it might be useful to leave in place, since making oktopus an optional (or perhaps test-only) dependency had already been requested in #1023. I've updated the `ModuleNotFoundError` warning accordingly, final wording yet to be discussed.

Nschanche:
> @Nschanche, @christinahedges, situation has changed quite significantly again, as autograd turns out to be no longer unmaintained, and last week's release 1.8.0 now supports Numpy 2! Oktopus by itself never seemed to have any conflicts. :D
> 
> So I have updated the install and test requirements accordingly to reenable `tpfmodel` with newer versions, see [unskipped test_tpfmodel](https://github.com/lightkurve/lightkurve/actions/runs/14840805655/job/41663051513?pr=1452#step:5:39). I've also replaced two `math` trigonometric function calls with their `numpy` counterparts to get rid of some 1500 deprecation warnings (as of numpy 1.25); apparently length-1 arrays are passed as `rotation_angle`, but I did not track that further down. Also added a 3.13 test; maybe the macOS and Windows tests should be upgraded to somewhat newer Python versions as well?
> 
> The framework for the conditional `oktopus` imports is still in place, if for some other reason its installation is not possible or desired (e.g. my miniconda env did not let me install 0.1.2 with numpy>=2, seems conda has registered its own conflict list). That could of course be reverted, but it might be useful to leave in place, since making oktopus an optional (or perhaps test-only) dependency had already been requested in #1023. I've updated the `ModuleNotFoundError` warning accordingly, final wording yet to be discussed.

That's great! This will simplify things. The PR looks good, although I'd recommend updating the ModuleNotFound comment with this PR along with the issue already referenced. I want to get this merged in asap to prepare for a new release next week.

dhomeier:
I am hesitant to put too much detail into the `ModuleNotFound` warning, though in principle #1023 is also relevant – but they are also all cross-referencing each other. Also, _technically_ I guess oktopus is still not an optional dependency – `tool.poetry.group.dev.dependencies` might do that? But really it should probably be some kind of `recommended` dep. Anyway, at this point even `pip install lightkurve` will still pull it in, although pip would also happily uninstall it afterwards.

Nschanche:
We've switched around one of the remote data tests, and merged it in now. Thanks for your work on this!

dhomeier:
Thanks @Nschanche, glad you worked out what left the remote data tests hanging. It seems `3.12-remote-data` is now running without `--remote-data`, is this still broken with 3.12 then?

PR #1455, opened 2024-07-26
Add max numpy version to `pyproject.toml

kelle:
There seem to be problems with using numpy 2.0 but I found that 1.24 works!  Modified `pyproject.toml` accordingly.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ include = ["AUTHORS.rst", "CHANGES.rst", "CONTRIBUTING.rst", "CITATION"]
 
 [tool.poetry.dependencies]
 python = ">=3.8"
-numpy = {">=1.18, <=1.24"}
+numpy = ">=1.18, <=1.24"
 astropy = ">=5.0"
 scipy = { version = ">=1.7", python = ">=3.8,<3.11" }
 matplotlib = ">=3.1"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ include = ["AUTHORS.rst", "CHANGES.rst", "CONTRIBUTING.rst", "CITATION"]
 
 [tool.poetry.dependencies]
 python = ">=3.8"
-numpy = ">=1.18"
+numpy = {">=1.18, <=1.24"}
 astropy = ">=5.0"
 scipy = { version = ">=1.7", python = ">=3.8,<3.11" }
 matplotlib = ">=3.1"


Nschanche:
Hi @kelle , what issues did you run into with numpy 2.0? I would like to try to make it compatible with the latest numpy before capping it. I'll take a look and see what I find.

dhomeier:
It's still breaking `oktopus`; do you think substituting its `autograd` dependency planned in https://github.com/lightkurve/lightkurve/issues/1440#issuecomment-2187344614 is a realistic short to mid-term option?

Nschanche:
> It's still breaking `oktopus`; do you think substituting its `autograd` dependency planned in [#1440 (comment)](https://github.com/lightkurve/lightkurve/issues/1440#issuecomment-2187344614) is a realistic short to mid-term option?

I took a stab at this, and it wasn't a simple substitution so it won't be an immediate fix. We are discussing if we should continue to try to fix oktopus, or remove it entirely and replace the functionality. Either way, it won't be a short term change.

pyproject.toml
```diff
@@ -24,7 +24,7 @@ include = ["AUTHORS.rst", "CHANGES.rst", "CONTRIBUTING.rst", "CITATION"]
 
 [tool.poetry.dependencies]
 python = ">=3.8"
-numpy = ">=1.18"
+numpy = ">=1.18, <=1.24"
```
dhomeier, line 27:
```suggestion
numpy = ">=1.18, <2.0"
```
should be sufficient for oktopus, at least it is still importing and passing tests with 1.26.4

orionlee:
If PR #1452 (making `oktopus` optional) is meraged, is the numpy < 2.0 restriction still needed?

Somehow in CI, numpy 1.24.4 is chosen by `poetry install`, in all platforms / Python versions (3.8 - 3.10) tested. [Example CI run](https://github.com/lightkurve/lightkurve/actions/runs/10889616921/job/30216640313).

dhomeier:
> If PR #1452 (making `oktopus` optional) is meraged, is the numpy < 2.0 restriction still needed?
> 
No
> 
> Somehow in CI, numpy 1.24.4 is chosen by `poetry install`, in all platforms / Python versions (3.8 - 3.10) tested. [Example CI run](https://github.com/lightkurve/lightkurve/actions/runs/10889616921/job/30216640313).
> 
- See https://github.com/lightkurve/lightkurve/pull/1452#issuecomment-2291350465 – as far as I could figure out, poetry always tries to find _one_ version for each dependency that can be installed on the entire range of supported Python versions. Which will block an increasing number of updates while 3.8 and even 3.9 is supported (Astropy dev for example has just dropped 3.10).

PR #1460, opened 2024-08-20
Configuration parameter for using cloud URIs rather than downloading products

snbianco:
Just a draft PR for now since I have to add more tests, but I wanted to get some eyes on this and ask some questions.

This PR:
- Adds a configuration parameter `search_result_use_cloud_uri` that controls whether cloud URIs will be used to fetch products rather than actually downloading files.
- Adds a new function `SearchResult.get_cloud_uris` that gets the cloud URIs for a table of products.
- Adds an attribute `SearchResult.use_cloud_uri` that essentially does the same thing as the config parameter, but for an instance of `SearchResult`. If this attribute is set to `True`, then files will not be downloaded with `download` and `download_all` and instead, products will be fetched from the cloud.

Some considerations/other questions:
- How should I handle FFI cutouts? These use TESSCut and are not available on the cloud. For now, I'm logging a warning message and skipping those products.
- How should I handle when files are not found in the cloud? Output a warning with the products that couldn't be found? Fall back to downloading?

### DIFF
--- a/src/lightkurve/__init__.py
+++ b/src/lightkurve/__init__.py
@@ -65,6 +65,9 @@ class Conf(_config.ConfigNamespace):
     search_result_display_extra_columns
         List of extra columns to be included when displaying a SearchResult object.
 
+    search_result_use_cloud_uri
+        If set to True, use an S3 URI rather than download a search result.
+
     cache_dir
         Default cache directory for data files downloaded, etc. Defaults to ``~/.lightkurve/cache`` if not specified.
 
@@ -81,6 +84,13 @@ class Conf(_config.ConfigNamespace):
         [],
         "List of extra columns to be included when displaying a SearchResult object.",
         cfgtype="string_list",
+        module="lightkurve.search"
+    )
+
+    search_result_use_cloud_uri = _config.ConfigItem(
+        False,
+        "If set to True, use an S3 URI rather than download a search result.",
+        cfgtype="boolean",
         module="lightkurve.search"
     )
 
--- a/src/lightkurve/search.py
+++ b/src/lightkurve/search.py
@@ -14,12 +14,14 @@ from astropy.io import ascii
 from astropy.table import Row, Table, join
 from astropy.time import Time
 from astropy.utils import deprecated
+from astroquery.mast import Observations, TesscutClass
+from astroquery.exceptions import NoResultsWarning, ResolverError
 from memoization import cached
 from requests import HTTPError
 
 from . import PACKAGEDIR, conf, config
 from .collections import LightCurveCollection, TargetPixelFileCollection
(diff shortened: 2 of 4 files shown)


snbianco:
@tylerapritchard @ttdu

scfleming:
> How should I handle FFI cutouts? These use TESSCut and are not available on the cloud. For now, I'm logging a warning message and skipping those products.

What do you mean by this exactly?  Both the cube files and the individual FFIs are on the cloud. Maybe you are referring to something else though?

snbianco:
> > How should I handle FFI cutouts? These use TESSCut and are not available on the cloud. For now, I'm logging a warning message and skipping those products.
> 
> What do you mean by this exactly? Both the cube files and the individual FFIs are on the cloud. Maybe you are referring to something else though?

I actually didn't know that those files were in the S3 bucket! I'll have to look more into the `TesscutClass` in Astroquery to see if it's actually optimized to use them.

christinahedges:
@snbianco @scfleming we are currently attempting to break functionality out of lightkurve into smaller more maintainable sub-packages. One of them includes [lksearch](github.com/lightkurve/lksearch) which is designed to query the data and then return it to the user. 

There are currently ways in `lksearch` to return cloud data, and the plan is when `lksearch` is in a good shape we will use it to replace the `search.py` submodule in `lightkurve`. 

At the moment, the plan is not to update `search.py` and to wait until `lksearch` is in a shape to replace it.

Please feel free to discuss or give your thoughts here!

snbianco:
> @snbianco @scfleming we are currently attempting to break functionality out of lightkurve into smaller more maintainable sub-packages. One of them includes [lksearch](github.com/lightkurve/lksearch) which is designed to query the data and then return it to the user.
> 
> There are currently ways in `lksearch` to return cloud data, and the plan is when `lksearch` is in a good shape we will use it to replace the `search.py` submodule in `lightkurve`.
> 
> At the moment, the plan is not to update `search.py` and to wait until `lksearch` is in a shape to replace it.
> 
> Please feel free to discuss or give your thoughts here!

This makes sense to me. There was some discussion about this on #1451 with @tylerapritchard, but as mentioned on that PR, there are other methods to fetch URIs and load in products from TIKE until `lksearch` is ready. Should I close this PR?

scfleming:
That broadly makes sense to me. Timing-wise, we will be moving more and more products onto the cloud, so whether it's best to wait until lksearch is in better shape or to add that support into search.py depends on where the lksearch development timeline intersects with more and more TESS and TESS HLSP data are on cloud.  That's probably measured in Quarters, not weeks, if that helps at all!

orionlee:
+1 for breaking out MAST search  / download into its own sub package.

PR #1463, opened 2024-09-16
better workaround for docutils installation with poetry/pypi

orionlee:
This is to replace the workaround in PR #1419  for pypi/poetry issue with docutils 0.21.post1
- this workaround is localized to docutils version, without the possible installation-wise side effects.
- the previous workaround in PR #1419 changed the pypi repository url to https://pypi.org/simple/ for entire installation, possibly inviting unforeseen side-effects.


orionlee:
I'll merge the PR in a week, unless I hear otherwise.

PR #1468, opened 2024-10-07
Fixed select_flux() inconsistency between flux and flux_err in some cases

### LINKED ISSUE #1467: `LightCurve.select_flux()` would cause inconsistent units between flux and flux_err in some cases, e.g., `flux_bkg` from GSFC-ELEANOR-LITE
#### Problem description

`LightCurve.select_flux()` would cause inconsistent units between flux and flux_err in some cases, e.g., `flux_bkg` from GSFC-ELEANOR-LITE. The inconsistency would lead to errors in some usage.

#### Example

```python
import lightkurve as lk

lc = lk.search_lightcurve("TIC 336732616", author="GSFC-ELEANOR-LITE", sector=1)[0].download()
print("Units of flux, flux_err:", lc.flux.unit, ",", lc.flux_err.unit, "; origin:", lc.flux_origin);

lc = lc.select_flux("flux_bkg");
print("Units of flux, flux_err:", lc.flux.unit, ",", lc.flux_err.unit, "; origin:", lc.flux_origin);

lc = lc["time", "flux", "flux_err"]  # ValueError
```

The output is:

```python
Units of flux, flux_err: electron / s , electron / s ; origin: corr_flux
Units of flux, flux_err: % , electron / s ; origin: flux_bkg
```

with a `ValueError`
```python
Cell In[1178], line 7
      4 lc = lc.select_flux("flux_bkg");
      5 print("Units of flux, flux_err:", lc.flux.unit, ",", lc.flux_err.unit, "; origin:", lc.flux_origin);
----> 7 lc = lc["time", "flux", "flux_err"]

File C:\pkg\_winNonPortables\miniforge3\envs\my_lk_plus\lib\site-packages\astropy\timeseries\sampled.py:293, in TimeSeries.__getitem__(self, item)
    289         out._groups = groups.TableGroups(
    290             out, indices=self.groups._indices, keys=self.groups._keys
    291         )
    292         return out
--> 293 return super().__getitem__(item)

File C:\pkg\_winNonPortables\miniforge3\envs\my_lk_plus\lib\site-packages\astropy\table\table.py:2070, in Table.__getitem__(self, item)
   2068     return self.Row(self, item.item())
   2069 elif self._is_list_or_tuple_of_str(item):
-> 2070     out = self.__class__(
   2071         [self[x] for x in item], copy_indices=self._copy_indices
   2072     )
   2073     out._groups = groups.TableGroups(
   2074         out, indices=self.groups._indices, keys=self.groups._keys
   2075     )
   2076     out.meta = self.meta.copy()  # Shallow copy for meta

File C:\pkg\_winNonPortables\miniforge3\envs\my_lk_plus\lib\site-packages\lightkurve\lightcurve.py:400, in LightCurve.__init__(self, data, time, flux, flux_err, *args, **kwargs)
    398 # Ensure flux and flux_err have the same units
    399 if self["flux"].unit != self["flux_err"].unit:
--> 400     raise ValueError("flux and flux_err must have the same units")
    402 self._new_attributes_relax = False
    403 self._required_columns_relax = False

ValueError: flux and flux_err must have the same units
```

#### Expected behavior

`select_flux()` should not cause inconsistency between flux / flux_err.

#### Environment

-  lightkurve version (e.g. 1.0b6): 2.5.0

orionlee:
Close #1467 

Changelog to be added:
```rst
- Fixed ``LightCurve.select_flux()`` in edge cases, where the unit of
  the new ``flux`` column is different from that of the ``flux_err`` column. [#1467]
```

### DIFF
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -561,7 +561,7 @@ class LightCurve(TimeSeries):
             else:
                 # fill in a dummy all-nan flux_err column
                 # ensure the unit of new flux_err is consistent with that of flux.
-                flux_err_col_vals = np.full_like(lc["flux"].value, np.nan)
+                flux_err_col_vals = np.full(lc["flux"].shape, np.nan)
                 if lc["flux"].unit is not None:
                     flux_err_col_vals = flux_err_col_vals * lc["flux"].unit
                 lc["flux_err"] = flux_err_col_vals
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -559,7 +559,12 @@ class LightCurve(TimeSeries):
             if flux_err_column in lc.columns:
                 lc["flux_err"] = lc[flux_err_column]
             else:
-                lc["flux_err"][:] = np.nan
+                # fill in a dummy all-nan flux_err column
+                # ensure the unit of new flux_err is consistent with that of flux.
+                flux_err_col_vals = np.full_like(lc["flux"].value, np.nan)
+                if lc["flux"].unit is not None:
+                    flux_err_col_vals = flux_err_col_vals * lc["flux"].unit
+                lc["flux_err"] = flux_err_col_vals
 
         lc.meta['FLUX_ORIGIN'] = flux_column
         normalized_new_flux = lc["flux"].unit is None or lc["flux"].unit is u.dimensionless_unscaled
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -566,6 +566,15 @@ class LightCurve(TimeSeries):
                     flux_err_col_vals = flux_err_col_vals * lc["flux"].unit
                 lc["flux_err"] = flux_err_col_vals
 
+        # Ensure resulting flux / flux_err have the same
+        # Do the check here after the columns are selected so as to uniformly handle
+        # different cases.
+        if lc["flux"].unit != lc["flux_err"].unit:
+            raise ValueError(
+                f"Columns '{flux_column}' and '{flux_err_column}' have different units: "
+                f"{lc.flux.unit} and {lc.flux_err.unit} respectively."
+            )
+
         lc.meta['FLUX_ORIGIN'] = flux_column
         normalized_new_flux = lc["flux"].unit is None or lc["flux"].unit is u.dimensionless_unscaled
         # Note: here we assume unitless flux means it's normalized
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,8 +1,12 @@
 2.5.1  (unreleased)
 =====================
-- Fixed pixel to world coordinate transformation in ``TargetPixelFile.get_coordinates()`` 
-  in line 488 ("ra, dec = w.wcs_pix2world(X.ravel(), Y.ravel(), 1)"), where for consistency with 
-  Gaia the origin should be 0 instead of 1.
+
+- Fixed pixel to world coordinate transformation in ``TargetPixelFile.get_coordinates()``
+  in line 488 ("ra, dec = w.wcs_pix2world(X.ravel(), Y.ravel(), 1)"), where for consistency with
+  Gaia the origin should be 0 instead of 1. [#1465]
+- Fixed ``LightCurve.select_flux()`` in edge cases, where the unit of
+  the new ``flux`` column is different from that of the ``flux_err`` column. [#1467]
+
 
 2.5.0 (2024-08-29)
 =====================
--- a/tests/io/test_eleanor.py
+++ b/tests/io/test_eleanor.py
@@ -3,7 +3,7 @@ import pytest
 from astropy.io import fits
 from astropy.utils.data import get_pkg_data_filename
 import numpy as np
-from numpy.testing import assert_array_equal
+from numpy.testing import assert_array_equal, assert_equal
 
 from lightkurve import search_lightcurve
 from lightkurve.io.eleanor import read_eleanor_lightcurve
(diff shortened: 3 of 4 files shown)


src/lightkurve/lightcurve.py
```diff
@@ -559,7 +559,12 @@ def select_flux(self, flux_column, flux_err_column=None):
             if flux_err_column in lc.columns:
                 lc["flux_err"] = lc[flux_err_column]
             else:
-                lc["flux_err"][:] = np.nan
+                # fill in a dummy all-nan flux_err column
+                # ensure the unit of new flux_err is consistent with that of flux.
```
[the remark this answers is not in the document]
orionlee: Hmm, if the array is integers, e.g., `np.full_like(np.array([1,2,3]), np.nan)`, I get `array([-2147483648, -2147483648, -2147483648])`

So I suppose to be safe, we should either use `np.full()`, or specify `dtype` when calling `full_like`, e.g., `np.full_like(np.array([1,2,3]), np.nan, dtype=float)`

orionlee:
Also, an edge case this PR currently does not address is that if the new flux has associated error column but they have different units, e.g, column `flux_bkg` in % and column `flux_bkg_err` in e-/s, the lc returned from `.select_flux("flux_bkg")` will have inconsistent units.

We might consider to issue a warning for such scenarios.

Nschanche:
> Also, an edge case this PR currently does not address is that if the new flux has associated error column but they have different units, e.g, column `flux_bkg` in % and column `flux_bkg_err` in e-/s, the lc returned from `.select_flux("flux_bkg")` will have inconsistent units.
> 
> We might consider to issue a warning for such scenarios.

The only scenario where I think it makes sense to only have a warning is the one you mention, with a mix between a percentage and a unit. However, I think users are more likely to ignore the warning and treat it as though they are the same unit.  I think it is safer and more consistent if it breaks if there is a mismatch.

orionlee:
> I think it is safer and more consistent if it breaks if there is a mismatch.

Do you mean that `.select_flux()` should raise an exception for cases like this? If users wants to proceed, they'd need to first fix the inconsistency in the underlying columns before calling `.select_flux()`.

It's a slight change in API behavior, but that's probably okay, given it's a really edge case. 
I can bundle the change in this PR if there's an agreement in the behavior.

Nschanche:
> > I think it is safer and more consistent if it breaks if there is a mismatch.
> 
> Do you mean that `.select_flux()` should raise an exception for cases like this? If users wants to proceed, they'd need to first fix the inconsistency in the underlying columns before calling `.select_flux()`.
> 
> It's a slight change in API behavior, but that's probably okay, given it's a really edge case. I can bundle the change in this PR if there's an agreement in the behavior.

That's what I think. @tylerapritchard, do you want to weigh in?

rebekah9969:
@Nschanche and @orionlee  - I agree that the code should break if there is a mismatch and so force the user to address that the units are not the same and fix that before moving forward. If we fix it for them i worry they would continue to miss the issue and make other assumptions.

orionlee:
@rebekah9969 To clarify (in case there is confusion), fixing the unit mismatch is never on the table in the first place. There are two cases:

1. (the base case) current `flux` / `flux_err` is in e-/s, while `new_flux` is in %, but there is no `new_flux_err`  (such as flux_bkg from GSFC-ELEANOR-LITE).
The current `lc.select_flux()` will result in the new `flux` in % while, the all-nan `flux_err` (filled by `lightkurve`) in e-/s  . The PR fixes the mismatch created by lightkurve itself.

2. (the really edge, hypothetical, case) current `flux` / `flux_err` is in e-/s, while `new_flux` is in %, but `new_flux_err` is in a different unit, say, e-/s.
The current `lc.select_flux()` will result in the new `flux` in % and `flux_err` in e-/s, without checking the units are inconsistent. The proposal is to raise an error instead, to force users address the mismatch.

Nschanche:
> @rebekah9969 To clarify (in case there is confusion), fixing the unit mismatch is never on the table in the first place. There are two cases:
> 
> 1. (the base case) current `flux` / `flux_err` is in e-/s, while `new_flux` is in %, but there is no `new_flux_err`  (such as `flux_bkg` from GSFC-ELEANOR-LITE).
>    The current `lc.select_flux()` will result in the new `flux` in % while, the all-nan `flux_err` (filled by `lightkurve`) in e-/s  . The PR fixes the mismatch created by lightkurve itself.
> 2. (the really edge, hypothetical, case) current `flux` / `flux_err` is in e-/s, while `new_flux` is in %, but `new_flux_err` is in a different unit, say, e-/s.
>    The current `lc.select_flux()` will result in the new `flux` in % and `flux_err` in e-/s, without checking the units are inconsistent. The proposal is to raise an error instead, to force users address the mismatch.

I think that we do want to throw the error from option 2 in this unlikely event.

orionlee:
The codes now raise a `ValueError` if the flux and flux_err columns to be selected have different units.

The PR is ready to merge, pending review feedback. Changelog entry to be added is in the first comment.

src/lightkurve/lightcurve.py
```diff
@@ -559,7 +559,21 @@ def select_flux(self, flux_column, flux_err_column=None):
...
+                flux_err_col_vals = np.full(lc["flux"].shape, np.nan)
+                if lc["flux"].unit is not None:
+                    flux_err_col_vals = flux_err_col_vals * lc["flux"].unit
+                lc["flux_err"] = flux_err_col_vals
+
+        # Ensure resulting flux / flux_err have the same
```
Nschanche, line 569: same <units>

src/lightkurve/lightcurve.py
```diff
@@ -559,7 +559,21 @@ def select_flux(self, flux_column, flux_err_column=None):
...
+        # Do the check here after the columns are selected so as to uniformly handle
+        # different cases.
+        if lc["flux"].unit != lc["flux_err"].unit:
+            raise ValueError(
+                f"Columns '{flux_column}' and '{flux_err_column}' have different units: "
```
Nschanche, line 574: Good informative error

Nschanche (commented):
I think this looks good, just need to add change log. Then I will merge in.

orionlee:
Changelog added.

PR #1472, opened 2024-11-06
TessCut: add header values to make it closer to SPOC-produced TPFs

### LINKED ISSUE #1470: Better parity between TessCut TPF and SPOC-generated TPF, for `interact_sky()`, `interact()`, etc.
#### Problem description

For a given TIC, there are some small differences between TPFs from TESS SPOC and those from TessCut that would cause some small issues when using TPFs from TessCut. For examples, for TPFs from TessCut:

1. no proper motion data is available, so the crosshair shown in `tpf.interact_sky()` might be noticeably off for targets with huge proper motions (the issue was indirectly raised in a few PRs)

2. some metadata is missing, e.g., the TIC, TESSMAG, etc. As a result, the title in `tpf.interact()` is missing the TIC:
image

#### Environment

-  lightkurve version (e.g. 1.0b6): 2.5.0

orionlee:
Close #1470


orionlee:
Limitation: the filling of header values only kicks in the typical usage, when users gets the TPF from `search_tesscut().download()`, e.g.,
```python
tic, sector = 261136679, 8  #  pi Men 
sr_tc = lk.search_tesscut(f"TIC{tic}", sector=sector)
tpf_tc = sr_tc[-1].download(cutout_size=(11, 11))
```

It would not kick in in other cases, when lightkurve has no explicit context that the TPF is about a specific TIC, e.g., 
- users open a TPF directly from a path/URL, `lk.read(path_to_tpf)` 
- users search the target using non-TIC name, e.g., `lk.search_tesscut(f"pi Men")`

orionlee:
@Nschanche if I remember correctly, PR 1366 does not address the issue to be resolved here (for TessCut TPFs, the target's coordinate has no proper motion correction because PM values are not in header).

Nschanche:
I think this functionality would be really useful. However, I would expect this to work when name resolving an object, not just when using a TIC. If the input name is a string like "pi Men", one option is that we could first resolve the coordinates using astroquery.mast.MastClass().resolve_object("AU Mic"). We could then use the resulting coordinates for the Catalog.query_region(), setting the search radius to something small.

orionlee:
Nschanche, I see what you mean for cases when a non TIC name is specified. 

A small concern for those cases is the ambiguities they could bring:
1. a name might not actually map to a TIC, which could manifest in 2 cases:
  a. no TIC found in the given coordinate + search radius
  b. multiple TICs are found.  We could, say, use the closest one, but it might not be the one the user actually has in mind, and thus introduces other errors.
  c. a single TIC is found in the given coordinate + search radius, but it is a different star. 
2. the users might not have TICs in mind at all. E.g., if users specify a Gaia DR3 source, would they want a PM-corrected coordinate based on Gaia DR3 result?

Bottomline: if we want to support non TIC name as well, we need to flush out the specific behavior. If the codes pick an incorrect TIC that happens to have large PM, we could end up introducing more user visible errors.

Nschanche:
Looping @tylerapritchard in as he did a lot of work on this for lksearch. Tyler, did you find a good solution to solve issues with name resolve?

tylerapritchard:
Yeah, we had similar concerns as @orionlee when we were working on lksearch.catalogsearch.  We wanted a non-positional crossmatch that would allow users to get the TIC for something like " Pi Men." We ended up creating [lksearch.catalogsearch.match_names_catalogs](https://lightkurve.github.io/lksearch/tutorials/catalog-searches.html#Looking-up-Catalog-IDs) which uses the simbad "name resolver" to crossmatch common names with catalog names.  

This does have a number of limitations - primarily that this crossmatch only exists if things have gotten published in a way that simbad can read, so it doesn't exist for lots of objects.  But it does get the biggest/most "popular" ones.

orionlee:
Fundamentally, when users specify a star by non TIC name, we don't know whether the users have TICs in mind at all.

One solution, along the line of what tylerapritchard suggests, if it's a non TIC name, we first search it on SIMBAD, and use the coordinate and proper motions from SIMBAD.
The current codes (using MAST name resolver) will serve as the last resort (if SIMBAD has no result).

It won't be perfect, but it should work reasonably well for most cases, and is easy for users to understand.

Issue #1438, opened 2024-06-11
search_lightcurve() and download_all() return different author


GijsMulders:
#### Problem description

When searching for and downloading TESS SPOC long cadence lightcurves, the author field is different

#### Example

```python
import lightkurve as lk
search_result= lk.search_lightcurve('TIC {}'.format(176218128), 
    mission='TESS',
    author= ['TESS-SPOC'],  
    exptime=600,
    sector=29)
dls= search_result.download_all()
print (search_result.author[0], dls[0].author, search_result.author[0] == dls[0].author)
```
<img width="929" alt="Screenshot 2024-06-11 at 17 29 14" src="https://github.com/lightkurve/lightkurve/assets/13457058/cdb9ce77-a829-475f-a572-4465255f88e5">

#### Expected behavior

I would expect both lightcurves to be called TESS-SPOC but the downloaded one is only called SPOC

#### Environment

-  platform OSX
-  lightkurve version 2.4.2:
-  installation method (e.g. pip, conda, source): conda, pip

mriddyagrawal:
Surprisingly, this is only happening for SPOC and TESS-SPOC, not for something else like QLP. Furthermore, this phenomenon only shows the author indifferently, the light curve downloaded is different and does correspond to each product. 

Maybe there is an error in the code that accepts them as different authors in one place and same in another.

Nschanche:
Hi @GijsMulders and @mriddyagrawal, the lightkurve io for any SPOC product defines the author property as 'SPOC'. However, I agree it is confusing to have this mismatch with the search results. I put in PR #1487 to set the author to TESS-SPOC when reading in the HLSP files. I don't see anything in the fits header that distinguishes between spoc/tess-spoc directly, so this just checks if tess-spoc is in the filename. Therefore, if users rename the file, the author keyword will still be set to SPOC.

PR #1487, opened 2025-03-19
update author metadata

Nschanche:
Previously, the TESS reader set the 'author' property to SPOC for the TESS-SPOC HLSP in addition to the SPOC mission product. This did not match the provenance found in the search result table. This change checks the source and sets the author property accordingly. See issue #1438.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,7 @@
 - Modified RegressionCorrector.correct() to add the model flux error in quadrature. [#1439]
 - Updated the default TESS quality mask to include Argabrightening, Impulsive Outlier, and Bad Calibration Exclude flags (see issue #1230). [#1336]
 - Updated tutorial notebook links [#1484]
-- Author keyword now specifies between SPOC and TESS-SPOC [#1486]
+- Author keyword now specifies between SPOC and TESS-SPOC [#1487]
 
 2.5.0 (2024-08-29)
 =====================


orionlee:
Having a distinct `TESS-SPOC` author in Lightcurve meta data would be more convenient (and will be consistent with the author from `SearchResult`).

But the change is slightly backward incompatible. The changelog should probably call it out.

E.g., I have some codes that differentiate between SPOC and TESS-SPOC by a combination of author and cadence that would be broken once this change is released.

```python
lc.meta.get("AUTHOR") == "SPOC" and lc.meta.get('NUM_FRM', 0) * lc.meta.get('FRAMETIM', 0) <= 120
```

Nschanche:
> Having a distinct `TESS-SPOC` author in Lightcurve meta data would be more convenient (and will be consistent with the author from `SearchResult`).
> 
> But the change is slightly backward incompatible. The changelog should probably call it out.
> 
> E.g., I have some codes that differentiate between SPOC and TESS-SPOC by a combination of author and cadence. The codes would be broken once this change is released.
> 
> ```python
> lc.meta.get("AUTHOR") == "SPOC" and lc.meta.get('NUM_FRM', 0) * lc.meta.get('FRAMETIM', 0) <= 120
> ```

Thanks for the callout, I have this comment in the change log: - Author keyword now specifies between SPOC and TESS-SPOC [#1487]

Issue #1271, opened 2023-01-10
`lc.bin(bins=###)` failed for FoldedLightcurve with AttributeError 'TimeDelta' object has no attribute 'mjd'


AlexCohenDambros:
#### Problem description

```lc_fold.bin(bins = ###)``` The problem specifically occurs when using the 'bins' parameter in the bin function, a type error occurs, it expects an mjd and receives a jd.

I already tried to change the format but the object created by the 'fold' function returns as TimeDelta and this type does not have the mjd format

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-16-dcdfea69ac06> in <module>
     17 
     18 # ========= Defining global curves =========
---> 19 lc_global = lc_fold.bin(bins = 2001).normalize() - 1
     20 lc_global = (lc_global / np.abs(lc_global.flux.min()) ) * 2.0 + 1
     21 

2 frames
/usr/local/lib/python3.8/dist-packages/astropy/time/core.py in __getattr__(self, attr)
   1657         else:
   1658             # Should raise AttributeError
-> 1659             return self.__getattribute__(attr)
   1660 
   1661     def __dir__(self):

AttributeError: 'TimeDelta' object has no attribute 'mjd' 

#### Example

```python
import lightkurve as lk
import numpy as np

lcs = lk.search_lightcurve('KIC 757450', author='Kepler', cadence='long').download_all()
period, t0, duration_hours = 8.88492, 134.452, 2.078

lc_raw = lcs.stitch()

lc_clean = lc_raw.remove_outliers(sigma=3)

temp_fold = lc_clean.fold(period, epoch_time=t0)

fractional_duration = (duration_hours / 24.0) / period
phase_mask = np.abs(temp_fold.phase.value) < (fractional_duration * 1.5)
transit_mask = np.in1d(lc_clean.time.value, temp_fold.time_original.value[phase_mask])

lc_flat, trend_lc = lc_clean.flatten(return_trend=True, mask=transit_mask)

lc_fold = lc_flat.fold(period, epoch_time=t0)

# ========= Defining global curves =========
lc_global = lc_fold.bin(bins = 2001).normalize() - 1
lc_global = (lc_global / np.abs(lc_global.flux.min()) ) * 2.0 + 1

phase_mask = (lc_fold.phase > -4*fractional_duration) & (lc_fold.phase < 4.0*fractional_duration)
lc_zoom = lc_fold[phase_mask]


[...]

Issue #1429, opened 2024-05-13
Many functions on phase-normalized FoldedLightCurve failed with `UnitConversionError`:  `copy()`, `select_flux()`, `normalize()`, etc.


orionlee:
#### Problem description

Many operations on a phase-normalized FoldedLighCurve object, created from `lc.fold(...,  normalize_phase=True)`, would fail with `UnitConversionError: only quantities with time units can be used to instantiate Time instances.`

The affected functions all use all use `lc.copy()` internally, which triggered the error. They include (not exhaustive): `lc.copy()`, `lc.select_flux()`, `lc.remove_outliers()`, `lc.normalize()`, lightcurve math (e.g., `lc * 1.1`), etc.

Some operations are **not affected**, including `lc.remove_nans()` , `lc.truncate()` (operations that create a new lighcurve object using numpy boolean indexing. The new objects are shallow copies.)
 
#### Example

```python
import lightkurve as lk
# insert code here ...
lc = lk.search_lightcurve("TIC261136679", mission="TESS", cadence="short")[-1].download()  # pi Men

lc_f = lc.fold(epoch_time=1325.5, period=6.26782, normalize_phase=True)  

# They all fail with UnitConversionError, caused by `self.copy()` in implementation
lc_f.copy()
lc_f.select_flux("sap_flux") 
lc_f.remove_outliers()
lc_f * 1.1
lc_f.normalize()
```

The stacktrace:
```python
UnitConversionError                       Traceback (most recent call last)
File C:\pkg\_winNonPortables\miniforge3\envs\lk_dev\lib\site-packages\astropy\time\utils.py:105, in quantity_day_frac(val1, val2)
    104 try:
--> 105     factor = val1.unit.to(u.day)
    106 except Exception:
    107     # Not a simple scaling, so cannot do the full-precision one.
    108     # But at least try normal conversion, since equivalencies may be set.

File C:\pkg\_winNonPortables\miniforge3\envs\lk_dev\lib\site-packages\astropy\units\core.py:1165, in UnitBase.to(self, other, value, equivalencies)
   1164 else:
-> 1165     return self._get_converter(Unit(other), equivalencies)(value)

File C:\pkg\_winNonPortables\miniforge3\envs\lk_dev\lib\site-packages\astropy\units\core.py:1094, in UnitBase._get_converter(self, other, equivalencies)
   1092                 return lambda v: b(converter(v))
-> 1094 raise exc

[...]

Issue #1423, opened 2024-04-29
`lc.fold()` parameter `wrap_phase` does not accept plain float number when `normalize_phase=True`


orionlee:
#### Problem description

For [`lc.fold()`](https://docs.lightkurve.org/reference/api/lightkurve.LightCurve.fold.html), `wrap_phase` parameter should accept float number (similar to other parameters).
Currently `wrap_phase` parameter does accept float `number` except when the folded LightCurve is normalized with `normalize_phase=True`.

#### Example

Make the Folded LightCurve wraps at phase 0.25
```python
import lightkurve as lk
# insert code here ...
lc = lk.search_lightcurve("TIC261136679", mission="TESS", cadence="short")[-1].download()  # pi Men

lc.fold(epoch_time=3011.551, period=6.26791, wrap_phase=6.26791*0.25).scatter();  # okay

lc.fold(epoch_time=3011.551, period=6.26791, normalize_phase=True, wrap_phase=0.25*u.dimensionless_unscaled).scatter();  # okay

# failed, it should have assumed the unit is u.dimensionless_unscaled
lc.fold(epoch_time=3011.551, period=6.26791, normalize_phase=True, wrap_phase=0.25).scatter();

```

Error / stacktrace
```python
File C:\pkg\_winNonPortables\miniforge3\envs\my_lk_plus\lib\site-packages\lightkurve\lightcurve.py:1048, in LightCurve.fold(self, period, epoch_time, epoch_phase, wrap_phase, normalize_phase)
   1040     elif self.time.format == "btjd":
   1041         warnings.warn(
   1042             "`epoch_time` appears to be given in JD, "
   1043             "however the light curve time uses BTJD "
   1044             "(i.e. JD - 2457000).",
   1045             LightkurveWarning,
   1046         )
-> 1048 ts = super().fold(
   1049     period=period,
   1050     epoch_time=epoch_time,
   1051     epoch_phase=epoch_phase,
   1052     wrap_phase=wrap_phase,
   1053     normalize_phase=normalize_phase,
   1054 )
   1056 # The folded time would pass the `TimeSeries` validation check if
   1057 # `normalize_phase=True`, so creating a `FoldedLightCurve` object
   1058 # requires the following three-step workaround:
   1059 # 1. Give the folded light curve a valid time column again
   1060 with ts._delay_required_column_checks():

[...]

Issue #1422, opened 2024-04-29
`lc.bin()` failed for normalized FoldedLightCurve with AttributeError: 'Quantity' object has no 'format' member


orionlee:
#### Problem description

When trying to perform `bin()` for a normalized FoldedLightCurve, created from `lc.fold(...,  normalize_phase=True)`, one would get `AttributeError: 'Quantity' object has no 'format' member`.

The underlying issue is that `bin()` implementation assumes `lc.time` is `Time` / `TimeDelta`, but for normalized FoldedLightCurve, `lc.time` is `Quantity` with unit `u.dimensionless_unscaled` (conceptually floats).

#### Example

```python
import lightkurve as lk
# insert code here ...

lc = lk.search_lightcurve("TIC261136679", mission="TESS", cadence="short")[-1].download()  # pi Men

lc.fold(epoch_time=3011.551, period=6.26791).bin(n_bins=100).scatter();  # okay


# The following all failed with AttributeError: 'Quantity' object has no 'format' member

lc.fold(epoch_time=3011.551, period=6.26791, normalize_phase=True).bin(n_bins=100).scatter();

lc.fold(epoch_time=3011.551, period=6.26791, normalize_phase=True).bin(bins=100).scatter();

lc.fold(epoch_time=3011.551, period=6.26791, normalize_phase=True).bin(time_bin_size=0.05*u.dimensionless_unscaled).scatter();

```

Example error / stacktrace:
```python
...
File C:\pkg\_winNonPortables\miniforge3\envs\my_lk_plus\lib\site-packages\lightkurve\lightcurve.py:1519, in LightCurve.bin(self, time_bin_size, time_bin_start, time_bin_end, n_bins, aggregate_func, bins, binsize)
   1514         time_bin_start = TimeDelta(
   1515             time_bin_start, format=self.time.format, scale=self.time.scale
   1516         )
   1517     else:
   1518         time_bin_start = Time(
-> 1519             time_bin_start, format=self.time.format, scale=self.time.scale
   1520         )
   1522 # Backwards compatibility with Lightkurve v1.x
   1523 if time_bin_size is None:

File C:\pkg\_winNonPortables\miniforge3\envs\my_lk_plus\lib\site-packages\astropy\units\quantity.py:1108, in Quantity.__getattr__(self, attr)
   1103 """
   1104 Quantities are able to directly convert to other units that
   1105 have the same physical type.
   1106 """
   1107 if not self._include_easy_conversion_members:
[...]

PR #1491, opened 2025-04-23
Folded LightCurve fix

Nschanche:


Nschanche:
This WIP PR will address Issues #1429 and #1422. These issues arise because normalized phase folded light curves are dimensionless, and therefore functions like astropy's aggregate_downsample do not work.

Nschanche:
- [ ] Figure out wrap_phase #1423
- [ ] Add tests for this new functionality
- [ ] Check if this fixes #1271

orionlee:
It seems like the PR proposes a breaking behavior change on `lc.fold()` in general, such that `lc.time` is no longer the phase (be it in TimeDelta or normalized phase).

If that's the case, 
- I feel it should be done in a major version upgrade (say in v3), as opposed to a minor point release.
- if the issue should be resolved at upstream astorpy TimeSeries.
  I just double checked with the current Astropy 7.1 dev codes. The issue remains there. For a normalized folded TimeSeries, with time of `Quantity`: 
  a. `bin()`: underlying `aggregate_downsample()` raises AttributeError
  b. `copy()`: TimeSeries `copy()` raises ValueError

orionlee:
An alternative approach that could fix the issues without any behavior change or upstream astropy change.

For each affected method, if it's a FoldedLightCurve with normalized phase
- create a temporary copy and maps normalized phase in Quantity to TimeDelta (say, in JD, the unit does not matter)
- perform the operation on the temporary copy.
  - time in TimeDelta type is the workaround to make the operations work (that assumes time is of `Time` type)
- convert the result's time column from the TimeDelta back to Quantity.

An outline of such approach: 

```python
class FoldedLightcurve:
    # ...

    def _to_tmp_lc_with_fake_time(self):
        # create a copy with  time column (normalized phase)
        # changed from Quantity to a TimeDelta
        pass

    def _restore_normalized_phase_to_time(self):
        # create a copy: change time column (logically normalized phase)
        # from the temporary TimeDelta back to Quantity
        pass

    def select_flux(self, flux_column, flux_err_column=None):
        lc = self
        if isinstance(self.time, Quantity):  # case in normalized phase
            lc = self._to_tmp_lc_with_fake_time()  # now the time (normalized phase) is of type TimeDelta
        lc = lc.select_flux(flux_column, flux_err_column)
        lc = lc._restore_normalized_phase_to_time()
        return lc
```

Nschanche:
Thanks @orionlee, I like your second suggestion. It functionally does the same thing, but is less intrusive and less likely to cause problems for users. I will pivot to that structure and keep pushing on this WIP.

orionlee:
It's probably obvious but I'd state it as a reminder: if we follow my suggestion above, the override for `lc.bin()` needs to explicitly raise errors for binning with time based parameters (`time_bin_size`, etc.) for the normalized phase case.

orionlee:
Nschanche, I've reported the problem in upstream astropy TimeSeries: 
https://github.com/astropy/astropy/issues/18070

Nschanche:
@orionlee can you review the PR and let me know your thoughts? It may be confusing as normalized phase folded light curves are 'unnormalized' (reverted back to TimeDelta time series) before binning, so the time steps are required to be in time units. I didn't add an explicit warning about this, just left it in the doctoring. I also restricted inputs to the FoldedTimeSeries binning function to be either time units or number of bins.

orionlee:
@Nschanche I'll take a look by early next week.

src/lightkurve/lightcurve.py
```diff
@@ -3140,7 +3195,113 @@ def plot_river(self, **kwargs):
...
+                result.remove_column("time")
+                result.add_column(normalized_phase, name="time", index=0)
+                         
+        return result 
+    
+
+    def copy(self):
+
+    
+        if self.normalize_phase == True:
```
orionlee, line 3285: Consider to replace the implementation with something like this:
```python
    def copy(self):
        # the default copy() in astropy fails if the lightcurve
        # has normalized phase (time column is `Quantity` instead of `Time` like)
        # workaround it by temporarily changing time column to 
        # non-normalized `TimeDelta`
        try:
            if self.normalize_phase == True:
                self._replace_normalized_phase()
            result = super().copy()
        finally:
            if self.normalize_phase == True:
                self._normalize_the_phase()
                result._normalize_the_phase()

        return result
```

1. simplify `result._delay_required_column_checks():` block with `result._normalize_the_phase()` ( or `result._restore_normalized_phase()` if it's renamed )
2. Wrap around with a try finally block so that if something unexpectedly go wrong during actual copy(), `self` will be restored back to the original state.
3. The comment is also applicable to the similar code path in `bin()`.

src/lightkurve/lightcurve.py
```diff
@@ -3140,7 +3195,113 @@ def plot_river(self, **kwargs):
...
         )
         return ax
+    
 
+    def bin(
+        self,
+        time_bin_size=None,
+        time_bin_start=None,
+        aggregate_func=None,
+        bins=None,
```
orionlee, line 3205: - Shouldn't the parameters here be the same as the `bin()` of regular (non-folded) lightcurve?
  - shouldn't `n_bins` (and probably `binsize`) be supported as well?
  - if `time_bin_size` is supported even for a folded lightcurve with normalized phase, shouldn't `time_bin_end` be included as well?

src/lightkurve/lightcurve.py
```diff
@@ -3247,8 +3408,9 @@ def to_fits(
         }
 
         for kw in kepler_specific_data:
-            if ~np.asarray([kw.lower == k.lower() for k in extra_data]).any():
-                extra_data[kw] = kepler_specific_data[kw]
+            if hasattr(self, kw):
```
orionlee, line 3411: This change in `to_fit()` appears to be unrelated. Some kind of merge problems?

tests/test_lightcurve.py
```diff
@@ -257,7 +257,7 @@ def test_lightcurve_fold():
     assert_array_equal(np.sort(fold.time_original), lc.time)
     assert len(fold.time_original) == len(lc.time)
     fold = lc.fold(period=1, epoch_time=-0.1)
-    assert_almost_equal(fold.time[0], -0.5, 2)
```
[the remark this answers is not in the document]
Nschanche: Great, phase is an alias for time, so time should also be valid. I've added this test back in.

tests/test_lightcurve.py
```diff
@@ -280,6 +280,14 @@ def test_lightcurve_fold():
     with pytest.warns(LightkurveWarning, match="appears to be given in JD"):
         lc.fold(10, 2456600)
 
+    # Make sure normalizing and binning work for folded lightcurves
```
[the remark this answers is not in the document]
Nschanche: Updated comment with "Make sure binning a phase-normalized folded lightcurve works (#1422)"

tests/test_lightcurve.py
```diff
@@ -280,6 +280,14 @@ def test_lightcurve_fold():
...
+    fold = lc.fold(period=1.5, normalize_phase=False)
+    assert_almost_equal(np.max(fold.phase)-np.min(fold.phase), 1.5, 1)
+    assert len(fold.bin(bins=10)) == 10
+    fold = lc.fold(period=1.5, normalize_phase=True)
+    assert_almost_equal(np.max(fold.phase)-np.min(fold.phase), 1, 1)
+    assert len(fold.bin(bins=10)) == 10
```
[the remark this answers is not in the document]
Nschanche: Added
`    
fold_copy = fold.copy()
    assert_array_equal(fold.time, fold_copy.time)
    assert_array_equal(fold.flux, fold_copy.flux)
   `

src/lightkurve/lightcurve.py
```diff
@@ -3054,6 +3072,39 @@ def even_mask(self):
...
+    def _replace_normalized_phase(self):
+        # Some astropy functions, such as aggregate_downsample, require Time or TimeDelta
+        # As normalized phase-folded lightcurves are unitless, this breaks
+        # This will replace the normalized phase with phase in TimeDelta
+        if not hasattr(self, 'normalize_phase'):
+            warnings.warn("Not a phase-folded lightcurve, no phase available")
+
+        # If the lightcurve is phase folded but not normalized, just return self
+        if self.normalize_phase == False:
+            return self
```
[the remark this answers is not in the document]
Nschanche: Great, I've updated this

src/lightkurve/lightcurve.py
```diff
@@ -3054,6 +3072,39 @@ def even_mask(self):
...
+    
+    def _replace_normalized_phase(self):
+        # Some astropy functions, such as aggregate_downsample, require Time or TimeDelta
+        # As normalized phase-folded lightcurves are unitless, this breaks
+        # This will replace the normalized phase with phase in TimeDelta
+        if not hasattr(self, 'normalize_phase'):
+            warnings.warn("Not a phase-folded lightcurve, no phase available")
+
+        # If the lightcurve is phase folded but not normalized, just return self
+        if self.normalize_phase == False:
```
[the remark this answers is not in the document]
Nschanche: OK, this looks like a good warning to me. I've replaced it

src/lightkurve/lightcurve.py
```diff
@@ -3054,6 +3072,39 @@ def even_mask(self):
...
+            self.remove_column("time")
+            self.add_column(phase, name="time", index=0)
+        #return self
+
+
+    def _normalize_the_phase(self):
+        # Some astropy functions, such as aggregate_downsample, require Time or TimeDelta
+        # As normalized phase-folded lightcurves are unitless, this breaks
+        # This will re-normalized the phase
+        with self._delay_required_column_checks():
```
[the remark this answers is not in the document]
Nschanche: Check added

src/lightkurve/lightcurve.py
Nschanche: I don't know why this change ended up here. I will revert.

src/lightkurve/lightcurve.py
Nschanche: I updated the simplification suggested in 1 for this and for bin()

src/lightkurve/lightcurve.py
```diff
@@ -3140,7 +3206,125 @@ def plot_river(self, **kwargs):
...
+            time_bin_start = self.time[0]
+        if np.isscalar(time_bin_size):
+            if isinstance(self.time, TimeDelta):
+                time_bin_start = TimeDelta(
+                    time_bin_start, format=self.time.format, scale=self.time.scale
+                )
+            else:
+                warnings.warn("The function cannot be run on a lightcurve with normalized phase.")
+
+        if time_bin_size != None:
```
orionlee, line 3278: For `time_bin_size` semantics in the case of normalized phase folded LC, I suggest the following changes to make it more intuitive to users.

```python

# lc_f is a folded lightcurve with normalized phase with 10 day period

# supported by the PR. No change
lc_fb = lc_f.bin(time_bin_size=1.0*u.day)  

# not supported by the PR
# suggestion: support it, 
#  internally converts it from normalized phase to time (with lc_f.period * time_bin_size)
lc_fb = lc_f.bin(time_bin_size=0.1 * u.dimensionless_unscaled)

# PR treats it the same as 0.1 * u.day, the behavior of generic LightCurve.fold()
# it's probably surprising to users
# suggestion: treat it as 0.1 * u.dimensionless_unscaled  (in normalized phase)
lc_fb = lc_f.bin(time_bin_size=0.1)  
```

The implementation can be something like this:
```python
        if time_bin_size is not None:
            if not isinstance(time_bin_size, Quantity):
                # case time_bin_size is a plain float / int
                if self.normalize_phase:
                    time_bin_size = time_bin_size * u.dimensionless_unscaled
                else:
                    time_bin_size = time_bin_size * u.day  # duplicating the logic in parent .bin()
            if self.normalize_phase and time_bin_size.unit == u.dimensionless_unscaled:
                # aggregate_downsample() does not work on normalized phase
                # convert time_bin_size from normalized phase to time
                time_bin_size = self.period * time_bin_size.value
```

src/lightkurve/lightcurve.py
```diff
@@ -3140,7 +3206,125 @@ def plot_river(self, **kwargs):
...
+                ValueError('bins must be an integer describing the total number of bins.')
+
+
+        if self.normalize_phase == True:
+            self._replace_normalized_phase()
+
+        if time_bin_start is None:
+            time_bin_start = self.time[0]
+        if np.isscalar(time_bin_size):
+            if isinstance(self.time, TimeDelta):
```
orionlee, line 3271: The `TimeDelta` check is not necessary, as `self.time` has been temporarily converted to `TimeDelta` (if needed) by the earlier line `self._replace_normalized_phase()`

CHANGES.rst
```diff
@@ -11,6 +11,7 @@
 - Updated the default TESS quality mask to include Argabrightening, Impulsive Outlier, and Bad Calibration Exclude flags (see issue #1230). [#1336]
 - Updated tutorial notebook links [#1484]
 - Author keyword now specifies between SPOC and TESS-SPOC [#1487]
+- Modified copy() and bin() functions for FoldedLightCurve objects [#1491]
```
orionlee, line 14: 1. Reword it to state the purpose (of fixing them for case FoldedLightCurve with normalized phase)
2. If `time_bin_end`,  `n_bins`, and `binsize` will no longer be supported for `FoldedLightCurve`, state it here as well given it is a breaking change.

orionlee (changes requested):
See the new comments and the conversations I unresolved. The main things:

1. concerns of breaking changes by dropping `n_bins`, `time_bin_end` and `binsize` for FoldedLightCurve *without* normalized phase (i.e., time in `TimeDelta`)
2. make `time_bin_size` more intuitive for FoldedLightCurve with normalized phase: supporting `u.dimensionless_unscaled`
3. Tests for `lc.copy()` remain to be missing.

tests/test_lightcurve.py
```diff
@@ -280,6 +281,19 @@ def test_lightcurve_fold():
...
+    assert isinstance(fold.time, TimeDelta)
+    assert_almost_equal(np.max(fold.phase)-np.min(fold.phase), 1.5, 1)
+    assert len(fold.bin(bins=10)) == 10
+    fold = lc.fold(period=1.5, normalize_phase=True)
+    assert isinstance(fold.time, u.Quantity)
+    assert_almost_equal(np.max(fold.phase)-np.min(fold.phase), 1, 1)
+    binned = fold.bin(bins=10)
+    assert len(binned) == 10
+    # ensure fold was not changed
+    assert len(fold) == 100
```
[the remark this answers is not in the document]
Nschanche: Gotcha, I've added this assertion.

src/lightkurve/lightcurve.py
```diff
@@ -3054,6 +3074,48 @@ def even_mask(self):
         See the documentation of `odd_mask` for examples.
         """
         return ~self.odd_mask
+    
+    def _replace_normalized_phase(self):
+        # Some astropy functions, such as aggregate_downsample, require Time or TimeDelta
+        # As normalized phase-folded lightcurves are unitless, this breaks
+        # This will replace the normalized phase with phase in TimeDelta
+        if not hasattr(self, 'normalize_phase'):
```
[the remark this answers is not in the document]
Nschanche: I think I somehow didn't commit the copy tests, which were quite simple
    fold_copy = fold.copy()
    assert_array_equal(fold.time, fold_copy.time)
    assert_array_equal(fold.flux, fold_copy.flux)

src/lightkurve/lightcurve.py
Nschanche: Actually, looking into time_bin_end more closely, it looks like it doesn't really get passed in the the regular bin() function. I do want to remove it as an option (I will add that in the change file).

CHANGES.rst
Nschanche: Added an additional comment for these deprecated keywords.

src/lightkurve/lightcurve.py
```diff
@@ -3140,7 +3204,143 @@ def plot_river(self, **kwargs):
...
+        Binning always occurs in time units (not normalized phase units)
+
+        If the original light curve contains flux uncertainties (``flux_err``),
+        the binned lightcurve will report the root-mean-square error.
+        If no uncertainties are included, the binned curve will return the
+        standard deviation of the data.
 
+        Parameters
+        ----------
+        time_bin_size : `~astropy.units.Quantity` or `~astropy.time.TimeDelta`, optional
```
orionlee, line 3229: The docstring for time_bin_size should be updated to explain the semantics for normalized phase (that it can accept either time or normalized phase).

src/lightkurve/lightcurve.py
orionlee: Hmm, it does look `time_bin_end` parameter never gets passed in regular bin(), an issue dated from the introduction of the feature in #1042 (v2.1.0). 

I'm confused about what you intend to do:
1. On the one hand, for regular lightcurve, codes are added in this PR to properly pass along `time_bin_end` to the underlying `aggregate_downsample()`.
2. On the other hand, for FoldedLightCurve, `time_bin_end` is added as a parameter but is not passed along.

orionlee (changes requested):

src/lightkurve/lightcurve.py
Nschanche: I made the last commit before finishing my edits. I intend to remove time_bin_end as an optional keyword for FoldedLightCurves:
    def bin(
        self,
        time_bin_size=None,
        time_bin_start=None,
        aggregate_func=None,
        bins=None,
        n_bins=None,
    ):

src/lightkurve/lightcurve.py
Nschanche: Good catch

tests/test_lightcurve.py
```diff
@@ -287,12 +287,22 @@ def test_lightcurve_fold():
...
     assert isinstance(fold.time, u.Quantity)
+    assert fold.time.unit == u.dimensionless_unscaled
     assert_almost_equal(np.max(fold.phase)-np.min(fold.phase), 1, 1)
     binned = fold.bin(bins=10)
     assert len(binned) == 10
     # ensure fold was not changed
     assert len(fold) == 100
+    assert isinstance(fold.time, u.Quantity)
+
+    # Make sure 'copy()' works as expected
```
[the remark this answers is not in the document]
Nschanche: I have added this additional assertions.

Nschanche:
@orionlee this test keeps timing out, but I think this addresses your concerns. I want to merge this asap as we want to release a new lightkurve version before the new TIKE image is released next week. Any final thought on this (or other PRs you strongly think should be imminently merged)?

orionlee:
I think this PR is okay to be merged. It'd be even better if change log mentions fixing the `bin(bin=###)` case ( issue #1271, it is for all FoldedLightCurve, not just those with normalized phase). Something like:

```rst
- Fixed the bug in ``bin()`` function with ``bin` parameter for ``FoldedLightCurve`` objects. [#1491]
```

---

I'd love to have PR #1472 merged, but it's not a showstopper.

orionlee:
Also, a note about github issue management: Once this PR is merged, you probably need to manually close all the issues.

If you change the wording in the comment to something like "Fixes hashtag<the issue number>", github will automatically link the issues and close them once the PR is merge. See:  https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

src/lightkurve/lightcurve.py
```diff
@@ -3140,7 +3204,141 @@ def plot_river(self, **kwargs):
...
+                    time_bin_size = time_bin_size * u.dimensionless_unscaled
+                else:
+                    time_bin_size = time_bin_size * u.day
+            if isinstance(time_bin_size, Quantity):
+                if time_bin_size.unit == u.dimensionless_unscaled:
+                    time_bin_size = time_bin_size.value * self.period
+ 
+
+        if bins != None:
+            # astropy's aggregate_downsample doesn't work for phase data:
```
orionlee, line 3297: I just realized the comment `astropy's aggregate_downsample doesn't work for phase data` is incorrect.  The problem of `bin(bins=###)` is due to lightkurve's own `Lighcurve.bin()` codes that assume time column is of `Time` type.

https://github.com/lightkurve/lightkurve/blob/b81ebc207a5cad3040659fd141f75fa7136a678f/src/lightkurve/lightcurve.py#L1543-L1544

It'd be better to reword it to something like "parent bin() does not work for phase data"

PR #1492, opened 2025-05-07
update numpy array comparisons to use isin

emmanuel-ferdman:
# PR Summary
This small PR resolves the NumPy deprecation warnings of `np.in1d` which you can find in the [CI logs](https://github.com/lightkurve/lightkurve/actions/runs/14870093767/job/41756341467#step:5:61):
```python
DeprecationWarning: `in1d` is deprecated. Use `np.isin` instead.
```
It also fixes a few typos in those relevant files.

### DIFF
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -818,7 +818,7 @@ class LightCurve(TimeSeries):
         Returns
         -------
         new_lc : `LightCurve`
-            Light curve which has the other light curves appened to it.
+            Light curve which has the other light curves append to it.
         """
         if inplace:
             raise ValueError(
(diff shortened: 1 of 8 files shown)


src/lightkurve/lightcurve.py
```diff
@@ -1935,7 +1935,7 @@ def _create_plot(
             flux_err = np.full(len(flux), np.nan)
 
         # Second workaround for AstroPy v5.0.0 issue #12481:
-        # matplotlib does not work well with `MaskedNDArray` arrays.
+        # matplotlib does not work well with `MaskedANDArray` arrays.
```
Nschanche, line 1938: I believe this is actually meant to be the original (MaskedNDArray)

emmanuel-ferdman: @Nschanche yes of course! thanks for that. Updated 😄

Issue #1371, opened 2023-10-20
lc.to_fits() + lk.io.read() does not preserve correct FLUX_ORIGIN meta data for SAP_FLUX data 


pmaxted:
If SAP_FLUX data are downloaded and saved as a FITS file then re-loading the data using lk.io.read() assigns the incorrect value to lc.meta['FLUX_ORIGIN']

#### Example

```python
>>> import lightkurve as lk
>>> s = lk.search_lightcurve('HD1',author='TESS-SPOC',exptime=200)
>>> lc = s.download_all(flux_column='sap_flux').stitch()
>>> print(lc.meta['FLUX_ORIGIN'])
>>> print(lc.flux[42])
>>> lc.to_fits('test.fits',overwrite=True)
>>> lctest = lk.io.read('test.fits')
>>> print(lctest.meta['FLUX_ORIGIN'])
>>> print(lctest.flux[42])

sap_flux
0.9995154142379761
pdcsap_flux
0.9995154142379761 electron / s
```

#### Expected behavior
lc.to_fits() needs to write the FLUX_ORIGIN value to the header of the output FITS file so that lk.io.read() can read this keyword and assign the correct value to this item in the meta data.

#### Environment

-  platform:  Windows
-  lightkurve version: '2.4.1'
-  installation method: pip

orionlee:
The issue is essentially a symptom of the underlying problem, discussed in #896

PR #1521, opened 2025-11-14
Change readin for to_fits roundtrip

Nschanche:
When saving a lightkurve object using to_fits, the origin of the flux column is lost (Issue #1371). This is because it is saved in the lighkurve object's metadata (meta.flux_origin) and is not preserved in the fits file. This PR addresses the uncertainty in the to_fits flux origin by populating meta.flux_origin with "to_fits()".

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ beautifulsoup4 = ">=4.6.0"
 requests = ">=2.22.0"
 urllib3 = { version = ">=1.23", python = ">=3.8,<4.0" }  # necessary to make requests work on py310
 tqdm = ">=4.25.0"
-pandas = ">=1.3.6"
+pandas = ">=1.3.6, <3.0.0"
 uncertainties = ">=3.1.4"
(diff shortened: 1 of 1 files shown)


src/lightkurve/lightcurve.py
```diff
@@ -2639,7 +2645,7 @@ def _make_lightcurve_extension(extra_data=None):
                     )
                 )
             if ~np.asarray(
-                [flux_column_name in k.upper() for k in extra_data.keys()]
+                [flux_column_name == k.upper() for k in extra_data.keys()]
```
d-giles: Should we use `flux_column_name.upper()` to be case invariant?

Nschanche:
There was some discussion regarding this PR and the potential for loss of information. For example, if users do not change the 'flux' column from the default PDCSAP, save the file with to_fits(), and read it in again, the flux origin information is lost and it is no longer clear where the flux came from. However, it was agreed upon that no information was better than confusing or misleading information, so this PR essential provides no information regarding the flux origin.

PR #1528, opened 2025-12-03
Code to fix SAP_QUALITY column occuring when not reading a K2, TESS, or Kepler file - Issue 1445 & 1393

rebekah9969:
Code to fix SAP_QUALITY column occuring when not reading a K2, TESS, or Kepler file


rebekah9969:
New Gist doc created to show user how to use the read_generic_lightcurve function. 
https://gist.github.com/rebekah9969/22b8dc8e8eabd5f7482d5523c073540e

PR #1535, opened 2026-01-23
Add support for generic FITS light curves (JDREF, etc.)

adrianlucy:
Despite its name, `read_generic_lightcurve` hasn't worked on generic FITS Standard-compliant light curves, and has relied on timestamps being in the special TESS or Kepler formats BTJD or BKJD, triggering on the value of`BJDREFI`. (Note that`BJDREFI` isn't defined in the FITS Standard, which would use e.g. `JDREFI`, together with`TIMESYS='TDB'` and `TREFPOS='BARYCENTER'` to remove the need for the ambiguously defined `B`.)

Of course, `lightkurve` must support TESS and Kepler -derived products, so I still check for the BTJD and BKJD special cases first.

But then, once those cases are ruled out, I look for time reference value keywords in the priority order defined by the [FITS Standard v4.0](https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf), section 9.2.2, stopping when we find a sufficient result:
- Both MJDREFI and MJDREFF (i.e., MJDREF split into integer and fractional keywords)
- MJDREF
- Both JDREFI and JDREFF
- JDREF
- (Preserve your pre-existing checks for nonstandard variant values of `TIMESYS`=`'local'`, `'mjd'`, or `'jd'`)
- ~As a last resort, the FITS Standard says to default to assuming MJD with MJDREF=0. I do so here with a warning message.~ **[Edit: not done, per feedback below we decided to be more realistic about the likelihood of deviations from the FITS Standard on this topic. I have instead written a more informative error message.]**

I implement this with a `reference_time_value` that defaults to 0.0, remains 0.0 for the TESS and Kepler special cases, and otherwise is changed to the appropriate reference time extracted from the keywords listed above. This `reference_time_value` is then added to the timestamps (except in the `TIMESYS='local'` case) to produce absolute timestamps in the specified format.

Finally, while we're making `read_generic_lightcurve` more generic, we might as well be safe and check for the uncommon case where the `TUNITn` of the time column is a recognized unit other than days, and convert to days if so.

If it helps to have a test file, here's one that uses JDREF: https://stsci.box.com/s/kpfm5qlk28k33yu05t4bpns8pn3s2a9s

Thanks for taking a look! MAST is highly motivated to help get something like this change adopted if possible, to help support upcoming cubesat missions producing light curves.


orionlee:
Consider adding some tests to avoid regression in the future. The example FITS file on box.com is probably small enough (~380 Kb) that can be added to the repository as-is.

tests/io/test_generic.py
```diff
@@ -0,0 +1,75 @@
...
+    assert type(lc).__name__ == "LightCurve"
+
+    # Open the file in astropy.fits
+    with fits.open(filename) as hdul:
+        data = hdul[1].data
+        header0 = hdul[0].header
+        header1 = hdul[1].header
+
+    # Is the time axis calculated correctly?
+    assert lc.time[3].value == (data['TIME'][3] + header1['JDREF'])
```
d-giles: Instead of comparing a single value, please use `assert all(lc.time.value == (data['TIME'] + header1['JDREF']))`

tests/io/test_generic.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # Is time format correctly inferred as JD?
+    assert lc.time.format.lower() == "jd"
+
+    # Is the time scale correctly read in as TDB?
+    assert lc.time.scale.lower() == header1['TIMESYS'].lower()
+    
+    # Are other data and metadata correctly read in?
+    assert float(lc.flux[5].value) == data['FLUX'][5]
```
d-giles: Compare all, like above

tests/io/test_generic.py
```diff
@@ -0,0 +1,75 @@
...
+    assert float(lc.ctr_err[5].value) == data['CTR_ERR'][5]
+    assert lc.label == header0['OBJECT']
+    assert len(lc) == len(data)
+    assert lc.flux.unit == u.Unit(header1['TUNIT5'])
+
```
d-giles, line 46: I'd like to see permutations tested as well, including cases where it should fail. The `lightkurve.read` module can read in an HDUList, so you can create a new one with a modified header. For example:

```
new_header = lc.hdu[1] + fits.Header({"JDREFI": 2400000, "JDREFF": 0.5})
# del(new_header["JDREF"]  # optional in this case since the REFI and REFF take priority over the lone keyword
new_table = fits.BinTableHDU(data=lc.hdu[1].data, header=new_header)
new_list = fits.HDUList(hdus=[lc.hdu[0], new_table])
new_lc = read(new_list)
```

tests/io/test_generic.py
```diff
@@ -0,0 +1,75 @@
...
+
+@pytest.mark.filterwarnings("error::ResourceWarning")
+@pytest.mark.filterwarnings("error::pytest.PytestUnraisableExceptionWarning")
+@pytest.mark.filterwarnings("ignore::astropy.units.UnitsWarning")  # Suppress astropy complaint about slashes in unit strings
+def test_read_generic_mjdrefif():
```
d-giles: Same comments for this suite of tests as the other. Test all values, check permutations.

tests/io/test_generic.py
```diff
@@ -0,0 +1,75 @@
...
+from lightkurve.io import read
+from astropy.io import fits
+from astropy import units as u
+import os
+from .. import TESTDATA
+
+@pytest.mark.filterwarnings("error::ResourceWarning")
+@pytest.mark.filterwarnings("error::pytest.PytestUnraisableExceptionWarning")
+@pytest.mark.filterwarnings("ignore::astropy.units.UnitsWarning")  # Suppress astropy complaint about slashes in unit strings
+def test_read_generic_jdref():
```
d-giles: Accessing the lc.hdu directly is deprecated, so if you implement my suggestion directly you'll want to add a filter for the LightkurveDeprecationWarning

d-giles:
Hey @adrianlucy, sorry for the delay, I've added some comments mostly requesting a bit more coverage. Thanks for your work on this!

tests/io/test_generic.py
adrianlucy: This function was really just a test of a select subset of permutations, so I'm now thinking of just covering all the permutations in one test function with one FITS file whose header I progressively modify, if that makes sense?

d-giles: Sounds good to me!

adrianlucy: Thanks! And just to make sure I understand what you'd like added: between the two current test functions I've tested JDREF, and then MJDREFI+MJDREFF with unit change (+ different timescale + erroneous JDREF). So if I also added the following permutations, would that cover everything you want?
- MJDREF (and throw in a an erroneous lone MJDREFF without MJDREFI)
- JDREFI + JDREFF (and throw in an erroneous JDREF)
- Failure state

Anything I'm missing there?

tests/io/test_generic.py
adrianlucy: Thanks! Should we be worried about that feature being removed in the future? As an alternative I could possibly try to create modified temporary files in the pytest `tmp_path` fixture: https://docs.pytest.org/en/stable/how-to/tmp_path.html Or something like that.

tests/io/test_generic.py
d-giles: I think that covers it

tests/io/test_generic.py
d-giles: I wouldn't worry about it for now, you could always modify the hdulist read in directly from the file instead of the lc.hdu. Despite being deprecated, a separate issue has revealed that the hdu is fairly enmeshed with some of the lightkurve functionality.

src/lightkurve/io/generic.py
```diff
@@ -115,14 +132,19 @@ def read_generic_lightcurve(
...
             elif hdulist[ext].header.get("TIMESYS") == "mjd":
                 time_format = 'mjd'
             elif hdulist[ext].header.get("TIMESYS") == "jd":
                 time_format = 'jd'
+            # Raise error if reference time (or related variants) not found
             else:
-                raise ValueError(f"Input file has unclear time format: {filename}")
+                raise ValueError(f"No reference time found in {filename}. Pass in a common time_format manually, or edit the FITS header.")
         if 'time' not in locals():
+            if (tab["time"].unit is not None) and (tab["time"].unit != u.d) and (type(tab["time"].unit) is not u.UnrecognizedUnit):
```
d-giles, line 144: I don't think I see this tested anywhere, can you add a test with units switched to hours or something?

adrianlucy: If I understand your request correctly, this is already tested here: https://github.com/adrianlucy/lightkurve/blob/31883d815d33f1e3b984bf87d1f8afe4e145902e/tests/io/test_generic.py#L65-L69 The second FITS file has time column units in seconds.

d-giles: My mistake!

adrianlucy: No worries, thanks for all your help!! Will get to all this ASAP.

adrianlucy:
I'll do my best to make these changes to the tests sometime today. Thanks again!

tests/io/test_generic.py
adrianlucy: In the final implementation, I did basically the same thing but got the data and header to pass into BinTableHDU/HDUList from `astropy.fits` instead of from `lc.hdu`.

tests/io/test_generic.py
adrianlucy: In the end I didn't need to use `lc.hdu`, I constructed the hdul directly using `astropy.fits` to pass into `lightkurve`, so no need to filter the deprecation warning.

adrianlucy:
Hi @d-giles and @Nschanche, I believe I've added all the agreed-upon permutations, and now assert equality between entire arrays instead of single elements. Thank you again! Sorry for the slight delay, had a few challenges that have now been resolved. (Testing equality between arrays with NaN/masked values, issues closing the hdul with the original `lc.hdu` implementation of the permutation tests, and `astropy.fits` being excessively clever in the units-change permutation test by overriding header units with units from the passed-in `data` object. All fixed now, tests are passing! Let me know if you see any problems.)

FWIW, if possible, it would be extremely beneficial for this to be merged in time for your upcoming release! We expect a mission that will benefit from this change may have a data release in ~July or so.

tylerapritchard:
lgtm!  Thanks Adrian! Dan!

PR #1543, opened 2026-03-10
Fix normalize() crash with masked flux arrays

### LINKED ISSUE #1505: LightCurve.normalize does not handle MaskedColumn with fill_value of nan
**Problem description:**

`LightCurve.normalize()` returns `nan` if flux is a `MaskedColumn` containing masked elements with `fill_value == numpy.nan`. 

This is because the median and standard deviation are computed with `numpy.nanmedian` and `numpy.nanstd` ([source](https://github.com/lightkurve/lightkurve/blob/e360709b0c5743a3adb2233b17a714b18190bc6d/src/lightkurve/lightcurve.py#L1144)). There are corresponding methods in the `numpy.ma` module (`numpy.ma.median`, `numpy.ma.std`) which ignore masked entries and return a finite value. 

I'm aware of previous discussions related to masked data (e.g. #1109), but I'm not sure whether there has been a resolution or if this is intended behavior. 

**MWE:**
```
import lightkurve as lk
import numpy as np

search_result = lk.search_lightcurve("TIC 417660696", author="QLP", sector=75)
lc = search_result[0].download()

print(np.nanmedian(lc.det_flux))  # currently implemented by `LightCurve.normalize()`
print(np.ma.median(lc.det_flux))
```
Output:
```
nan
0.99999934
```

**System and versions:**

- OS: MacOS Sequoia 15.6
- Numpy: 2.0.2 or 2.3.1
- Lightkurve: 2.5.1

m4cd4r4:
Fixes #1505.

np.nanmedian and np.nanstd don't handle MaskedColumn correctly when the fill value is nan - they return nan instead of computing over the unmasked values.

Fixed by converting to a plain ndarray first via np.ma.filled(self.flux, np.nan), which works for both masked and non-masked inputs.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,8 @@
   fallback cone search [#1541]
 - Modified how SAP_QUALITY is treated when reading/writing TESS and Kepler data [#1538]
 - Added ability to use nifty-ls for periodograms [#1550]
+- Fixed ``LightCurve.normalize()`` crash on masked flux arrays where ``np.nanmedian``
+  returned ``nan`` instead of computing over unmasked values [#1543]
 
 2.5.1  (2025-05-20)
 =====================
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -1250,7 +1250,10 @@ class LightCurve(TimeSeries):
             from zero.
         """
         validate_method(unit, ["unscaled", "percent", "ppt", "ppm"])
-        flux_data = np.ma.filled(self.flux, np.nan) if hasattr(self.flux, "filled") else np.asarray(self.flux)
+        if hasattr(self.flux, "mask") and hasattr(self.flux, "filled"):
+            flux_data = self.flux.filled(np.nan)
+        else:
+            flux_data = self.flux
         median_flux = np.nanmedian(flux_data)
         std_flux = np.nanstd(flux_data)
 
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -1250,8 +1250,9 @@ class LightCurve(TimeSeries):
             from zero.
         """
         validate_method(unit, ["unscaled", "percent", "ppt", "ppm"])
-        median_flux = np.nanmedian(self.flux)
-        std_flux = np.nanstd(self.flux)
+        flux_data = np.ma.filled(self.flux, np.nan) if hasattr(self.flux, "filled") else np.asarray(self.flux)
+        median_flux = np.nanmedian(flux_data)
+        std_flux = np.nanstd(flux_data)
 
         # If the median flux is within half a standard deviation from zero, the
         # light curve is likely zero-centered and normalization makes no sense.
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -1250,12 +1250,12 @@ class LightCurve(TimeSeries):
             from zero.
         """
         validate_method(unit, ["unscaled", "percent", "ppt", "ppm"])
-        if hasattr(self.flux, "mask") and hasattr(self.flux, "filled"):
-            flux_data = self.flux.filled(np.nan)
+        if hasattr(self.flux, "mask"):
+            median_flux = np.ma.median(self.flux)
+            std_flux = np.ma.std(self.flux)
         else:
-            flux_data = self.flux
-        median_flux = np.nanmedian(flux_data)
-        std_flux = np.nanstd(flux_data)
+            median_flux = np.nanmedian(self.flux)
+            std_flux = np.nanstd(self.flux)
 
         # If the median flux is within half a standard deviation from zero, the
         # light curve is likely zero-centered and normalization makes no sense.
--- a/src/lightkurve/lightcurve.py
+++ b/src/lightkurve/lightcurve.py
@@ -1250,12 +1250,12 @@ class LightCurve(TimeSeries):
             from zero.
         """
         validate_method(unit, ["unscaled", "percent", "ppt", "ppm"])
-        if hasattr(self.flux, "mask"):
-            median_flux = np.ma.median(self.flux)
-            std_flux = np.ma.std(self.flux)
+        if hasattr(self.flux, "mask") and hasattr(self.flux, "filled"):
+            flux_data = self.flux.filled(np.nan)
         else:
-            median_flux = np.nanmedian(self.flux)
-            std_flux = np.nanstd(self.flux)
+            flux_data = self.flux
+        median_flux = np.nanmedian(flux_data)
+        std_flux = np.nanstd(flux_data)
 
         # If the median flux is within half a standard deviation from zero, the
         # light curve is likely zero-centered and normalization makes no sense.
--- a/tests/test_lightcurve.py
+++ b/tests/test_lightcurve.py
@@ -942,6 +942,26 @@ def test_normalize():
     assert len(warn_record) == 0
     assert lc.meta["NORMALIZED"]
 
+    # Issue #1505: normalize() must handle MaskedColumn flux without
+    # collapsing to nan. np.nanmedian on a MaskedColumn returns nan because
+    # numpy's partition step ignores the mask before nan-skipping runs.
+    t = Time([2458000.0 + i * 0.02 for i in range(6)], format="jd", scale="tdb")
+    masked_lc = LightCurve(time=t)
+    masked_lc["flux"] = MaskedColumn(
+        data=[1.00, 1.01, np.nan, 1.02, np.nan, 1.03],
+        mask=[False, False, True, False, True, False],
+        fill_value=np.nan,
+    )
+    masked_lc["flux_err"] = masked_lc["flux"]
+    normalized = masked_lc.normalize()
+    # Median of unmasked values [1.00, 1.01, 1.02, 1.03] is 1.015, so the
+    # normalised median should be 1.0 (not nan, which is the bug).
+    assert_allclose(np.nanmedian(normalized.flux), 1.0, atol=1e-6)
+    expected = np.array([1.00, 1.01, np.nan, 1.02, np.nan, 1.03]) / 1.015
+    assert_allclose(
+        np.ma.filled(normalized.flux, np.nan), expected, atol=1e-6, equal_nan=True
+    )
+
 
 def test_invalid_normalize():
     """Normalization makes no sense if the light curve is negative,


Nschanche:
Hi @m4cd4r4, can you provide an example where the nanmedian/std was not working?

m4cd4r4:
Hi @Nschanche, here's a self-contained reproducer on the pre-PR code. It uses the same `MaskedColumn` + `fill_value=nan` pattern from the original report in #1505.

```python
import numpy as np
from astropy.table import MaskedColumn
from astropy.time import Time
from lightkurve import LightCurve

t = Time([2458000.0 + i * 0.02 for i in range(6)], format='jd', scale='tdb')
flux = MaskedColumn(
    data=[1.00, 1.01, np.nan, 1.02, np.nan, 1.03],
    mask=[False, False, True, False, True, False],
    fill_value=np.nan,
)

lc = LightCurve(time=t)
lc['flux'] = flux
lc['flux_err'] = flux

print("np.nanmedian(lc.flux):", np.nanmedian(lc.flux))  # -> nan
print("lc.normalize().flux:", lc.normalize().flux)       # -> all nan
```

Output on current main:
```
np.nanmedian(lc.flux): nan
lc.normalize().flux: [nan nan nan nan nan nan]
```

The root cause: `np.nanmedian(MaskedColumn)` returns `nan` when unmasked valid values exist. numpy's nan-aware reductions and numpy.ma's mask handling don't compose cleanly on `MaskedColumn` - the mask influences the internal partition step before the nan-skip logic runs.

I also pushed a fix for the CI failures - the previous version used `np.ma.filled()` which stripped the unit from plain `Quantity` flux (the non-masked path). The new version calls `.filled(np.nan)` on the flux object directly, which preserves both units and subclass type. Non-masked flux now passes through unchanged.

Nschanche:
OK, I see the issue. Instead of filling the masked array and defining a new temporary array, we could use np.ma.median(self.flux) in the normalize function, rather than np.nanmeddian. 

Regardless, a test should be added to assert the fix works for masked columns.  I suggest the test would go in test_lightcurve, in the test_normalize function. The example you provided would be a good test, asserting the correct median value.

m4cd4r4:
Thanks @Nschanche - pushed [ba15e4d](https://github.com/lightkurve/lightkurve/pull/1543/commits/ba15e4df) with both changes.

The refactor is split: `np.ma.median`/`np.ma.std` for the masked branch, `np.nanmedian`/`np.nanstd` kept for the unmasked branch. Reason for keeping both: `np.ma.std` on a plain `Quantity` (no mask) returns a value with sqrt-of-units (e.g. `electron(1/2) / s(1/2)` instead of `electron / s`), which would corrupt the std_flux passed into the zero-centered warning message. `np.nanmedian`/`np.nanstd` go through Quantity cleanly.

Test added in `test_normalize` using the reproducer from the comment above. Asserts the post-normalise median is 1.0 (not nan) and that unmasked positions divide cleanly by 1.015.

Full test_lightcurve.py run: 85 passed, 36 skipped, 3 xfailed.

Nschanche:
Interesting, it looks like np.ma.std() fails when there is a unit associated with the array. Using your example, 

```

t = Time([2458000.0 + i * 0.02 for i in range(6)], format='jd', scale='tdb')
masked_lc = lk.LightCurve(time=t)
masked_lc['flux'] = MaskedColumn(data = [1.0, 1.01, np.nan, 1.02, np.nan, 1.03] * u.electron / u.second, mask = [False, False, True, False, True, False], fill_value=np.nan)
    
masked_lc['flux2'] = MaskedColumn(data = [1.0, 1.01, np.nan, 1.02, np.nan, 1.03] * u.dimensionless_unscaled, mask = [False,False, True, False, True, False], fill_value=np.nan)

In [52]: np.ma.std(masked_lc.flux2)
Out[52]: <Quantity 0.01118034>

In [53]: np.ma.std(masked_lc['flux'])
File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/quantity.py:1653, in Quantity._to_own_unit(self, value, check_precision, unit)
   1652 try:
-> 1653     _value = value.to_value(unit)
   1654 except AttributeError:
   1655     # We're not a Quantity.
   1656     # First remove two special cases (with a fast test):
   1657     # 1) Maybe masked printing? MaskedArray with quantities does not
   1658     # work very well, but no reason to break even repr and str.
   1659     # 2) np.ma.masked? useful if we're a MaskedQuantity.

AttributeError: 'numpy.float64' object has no attribute 'to_value'

UnitConversionError                       Traceback (most recent call last)
File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/quantity.py:982, in Quantity.to_value(self, unit, equivalencies)
    981 try:
--> 982     scale = self.unit._to(unit)
    983 except Exception:
    984     # Short-cut failed; try default (maybe equivalencies help).

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/core.py:1159, in UnitBase._to(self, other)
   1157         return self_decomposed.scale / other_decomposed.scale
-> 1159 raise UnitConversionError(f"'{self!r}' is not a scaled version of '{other!r}'")

UnitConversionError: 'Unit(dimensionless)' is not a scaled version of 'Unit("electron / s")'

UnitConversionError                       Traceback (most recent call last)
Cell In[53], line 1
----> 1 np.ma.std(masked_lc['flux'])

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/numpy/ma/core.py:6882, in _frommethod.__call__(self, a, *args, **params)
   6878 if method is None:
   6879     # use the corresponding np function
   6880     method = getattr(np, method_name)
-> 6882 return method(marr, *args, **params)

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/numpy/ma/core.py:5478, in MaskedArray.std(self, axis, dtype, out, ddof, keepdims)
   5464 """
   5465 Returns the standard deviation of the array elements along given axis.
   5466 
   (...)
   5474 numpy.std : Equivalent function
   5475 """
   5476 kwargs = {} if keepdims is np._NoValue else {'keepdims': keepdims}
-> 5478 dvar = self.var(axis, dtype, out, ddof, **kwargs)
   5479 if dvar is not masked:
   5480     if out is not None:

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/numpy/ma/core.py:5431, in MaskedArray.var(self, axis, dtype, out, ddof, keepdims)
   5429     danom = umath.absolute(danom) ** 2
   5430 else:
-> 5431     danom *= danom
   5432 dvar = divide(danom.sum(axis, **kwargs), cnt).view(type(self))
   5433 # Apply the mask if it's not a scalar

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/numpy/ma/core.py:4364, in MaskedArray.__imul__(self, other)
   4362     self._mask += m
   4363 other_data = getdata(other)
-> 4364 other_data = np.where(self._mask, other_data.dtype.type(1), other_data)
   4365 self._data.__imul__(other_data)
   4366 return self

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/quantity.py:1851, in Quantity.__array_function__(self, function, types, args, kwargs)
   1849 function_helper = FUNCTION_HELPERS[function]
   1850 try:
-> 1851     args, kwargs, unit, out = function_helper(*args, **kwargs)
   1852 except NotImplementedError:
   1853     return self._not_implemented_or_raise(function, types)

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/quantity_helper/function_helpers.py:548, in where(condition, *args)
    545 if isinstance(condition, Quantity) or len(args) != 2:
    546     raise NotImplementedError
--> 548 args, unit = _quantities2arrays(*args)
    549 return (condition,) + args, {}, unit, None

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/quantity_helper/function_helpers.py:363, in _quantities2arrays(unit_from_first, *args)
    359 # We use the private _to_own_unit method here instead of just
    360 # converting everything to quantity and then do .to_value(qs0.unit)
    361 # as we want to allow arbitrary unit for 0, inf, and nan.
    362 try:
--> 363     arrays = tuple((q._to_own_unit(arg)) for arg in args)
    364 except TypeError:
    365     raise NotImplementedError

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/quantity_helper/function_helpers.py:363, in <genexpr>(.0)
    359 # We use the private _to_own_unit method here instead of just
    360 # converting everything to quantity and then do .to_value(qs0.unit)
    361 # as we want to allow arbitrary unit for 0, inf, and nan.
    362 try:
--> 363     arrays = tuple((q._to_own_unit(arg)) for arg in args)
    364 except TypeError:
    365     raise NotImplementedError

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/quantity.py:1669, in Quantity._to_own_unit(self, value, check_precision, unit)
   1667 try:
   1668     as_quantity = Quantity(value)
-> 1669     _value = as_quantity.to_value(unit)
   1670 except UnitsError:
   1671     # last chance: if this was not something with a unit
   1672     # and is all 0, inf, or nan, we treat it as arbitrary unit.
   1673     if not hasattr(value, "unit") and can_have_arbitrary_unit(
   1674         as_quantity.value
   1675     ):

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/quantity.py:985, in Quantity.to_value(self, unit, equivalencies)
    982     scale = self.unit._to(unit)
    983 except Exception:
    984     # Short-cut failed; try default (maybe equivalencies help).
--> 985     value = self._to_value(unit, equivalencies)
    986 else:
    987     value = self.view(np.ndarray)

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/quantity.py:891, in Quantity._to_value(self, unit, equivalencies)
    888     equivalencies = self._equivalencies
    889 if not self.dtype.names or isinstance(self.unit, StructuredUnit):
    890     # Standard path, let unit to do work.
--> 891     return self.unit.to(
    892         unit, self.view(np.ndarray), equivalencies=equivalencies
    893     )
    895 else:
    896     # The .to() method of a simple unit cannot convert a structured
    897     # dtype, so we work around it, by recursing.
    898     # TODO: deprecate this?
    899     # Convert simple to Structured on initialization?
    900     result = np.empty_like(self.view(np.ndarray))

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/core.py:1195, in UnitBase.to(self, other, value, equivalencies)
   1193     return UNITY
   1194 else:
-> 1195     return self._get_converter(Unit(other), equivalencies)(value)

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/core.py:1124, in UnitBase._get_converter(self, other, equivalencies)
   1121             else:
   1122                 return lambda v: b(converter(v))
-> 1124 raise exc

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/core.py:1107, in UnitBase._get_converter(self, other, equivalencies)
   1105 # if that doesn't work, maybe we can do it with equivalencies?
   1106 try:
-> 1107     return self._apply_equivalencies(
   1108         self, other, self._normalize_equivalencies(equivalencies)
   1109     )
   1110 except UnitsError as exc:
   1111     # Last hope: maybe other knows how to do it?
   1112     # We assume the equivalencies have the unit itself as first item.
   1113     # TODO: maybe better for other to have a `_back_converter` method?
   1114     if hasattr(other, "equivalencies"):

File ~/opt/anaconda3/envs/py3/lib/python3.9/site-packages/astropy/units/core.py:1085, in UnitBase._apply_equivalencies(self, unit, other, equivalencies)
   1082 unit_str = get_err_str(unit)
   1083 other_str = get_err_str(other)
-> 1085 raise UnitConversionError(f"{unit_str} and {other_str} are not convertible")

UnitConversionError: '' (dimensionless) and 'electron / s' are not convertible

```

whereas np.nanstd works in both cases:

```
In [59]: np.nanstd(masked_lc.flux)
Out[59]: <MaskedQuantity 0.01118034 electron / s>

In [60]: np.nanstd(masked_lc.flux2)
Out[60]: <MaskedQuantity 0.01118034>
```

I'm not familiar with why this might be the case. @d-giles or @orionlee have you come across this? I suppose we could just use np.nanstd/np.nanmedian rather than np.ma.std/np.ma.median.

orionlee:
I'm not familiar with this specific issue, but it looks like astropy `MaskedQuantity`  does not work with `np.ma.std`. 
I looked into it a bit further, it looks like astropy `Masked` / `MaskedQuantity` are not guaranteed to work with `np.ma` functions in the first place ([doc](https://docs.astropy.org/en/latest/utils/masked/index.html#numpy-functions-work-as-expected)): `Masked` / `MaskedQuantity` are meant to work with `np` functions instead.

It's probably safer and more defensive to use np.nanstd/np.nanmedian over non-masked values (e.g. calling `np.nanstd(masked_values.filled(np.nan))` if needed). 

---

Also, some of the sample codes in the thread here have some mix up on the 2 different types of masked columns in astropy.

1. the older `astropy.table.MaskedColumn` , which is a wrapper over `numpy.ma.MaskedArray`.  `lightkurve` does not use it in general. ([doc](https://docs.astropy.org/en/latest/table/masking.html)).
2. the newer `astropy.utils.masked.Masked` / `MaskedQuantity`, which does not use `numpy.ma.MaskedArray` . The masked columns in `lightkurve` from reading Kepler/TESS FITS files are of these types. Issues such as the one discussed here are usually associated with these newer types. ([doc](https://docs.astropy.org/en/latest/utils/masked/index.html))

In the sample codes, they seem to create `MaskedColumn` (type 1 column); but in reality, the column somehow got converted to a `MaskedQuantity` (type 2 column). (I think some logic in `lightkurve` does the conversion but am not 100% sure.) ( The issue with `np.ma.std` has to do with `MaskedQuantity` )

```python
masked_lc['flux2'] = MaskedColumn(data = [1.0, 1.01, np.nan, 1.02, np.nan, 1.03] * u.dimensionless_unscaled, mask = [False,False, True, False, True, False], fill_value=np.nan)
```

Illustrated in the following snippets: 
```python
flux2_col = MaskedColumn(data = [1.0, 1.01, np.nan, 1.02, np.nan, 1.03] * u.dimensionless_unscaled, mask = [False,False, True, False, True, False], fill_value=np.nan)
print("Column alone:", type(flux2_col))

tab = Table()
tab["flux2"] = flux2_col
print("Column in Table:", type(tab["flux2"]))

t = Time([2458000.0 + i * 0.02 for i in range(6)], format='jd', scale='tdb')
masked_lc = lk.LightCurve(time=t)
masked_lc["flux2"] = flux2_col  # converted to MaskedQuantity
print("Column in lk.LightCurve:", type(masked_lc["flux2"]))
```

Output: 
```python
Column alone: <class 'astropy.table.column.MaskedColumn'>
Column in Table: <class 'astropy.table.column.MaskedColumn'>
Column in lk.LightCurve: <class 'astropy.utils.masked.core.MaskedQuantity'>
```

orionlee:
I looked into the original QLP example in [issue 1505](https://github.com/lightkurve/lightkurve/issues/1505#issuecomment-3247422159), I think changing`normalize()` would not fix the actual issue. 

1. `normalize()` only works on column `lc.flux` . `lightkurve` logic  forces `flux` column to be of type `Quantity` / `MaskedQuantity`: `MaskedColumn` is converted to `Quantity` / `MaskedQuantity`.  So while the current  `normalize()` has problems with `MaskedColumn`, the issue will not happen in practice, because `lc.flux ` is always `Quantity` / `MaskedQuantity`

2. for QLP, `lc.select_flux("det_flux")` is potentially buggy, however. `det_flux` is of type `MaskedColumn`. After `select_flux()`, it gets converted to `Quantity` (i.e.,  `lc.select_flux("det_flux").flux` is of type `Quantity`), but the mask is lost. So the actual values could be different. after `lc.select_flux()`

3. I did encounter `LightCurve.normalize()` inexplicably returns `nan` at times, but not in the condition described in the issue 1505. (I do not know the precise condition.)

Example for (2):
```python
import numpy as np
from astropy.table import MaskedColumn
from astropy.time import Time
from lightkurve import LightCurve

# simulating QLP lightcurve
t = Time([2458000.0 + i * 0.02 for i in range(6)], format='jd', scale='tdb')
flux = MaskedColumn(
    data=[1.00, 1.01, np.nan, 1.02, 999, 1.03],
    mask=[False, False, True, False, True, False],
    fill_value=np.nan,
)

lc = LightCurve(time=t)
lc["det_flux"] = flux

print(type(lc["det_flux"]))   # MaskedColumn
print(lc["det_flux"])

lc2 = lc.select_flux("det_flux")
print(type(lc2.flux))  # Quantity
print(lc2.flux)  # the mask is lost, lc2.flux[4] is incorrectly at 999
```

Output:
```python
<class 'astropy.table.column.MaskedColumn'>
det_flux
--------
     1.0
    1.01
      --
    1.02
      --
    1.03
<class 'astropy.units.quantity.Quantity'>
[  1.     1.01    nan   1.02 999.     1.03]
```

orionlee:
@Nschanche, the original issue #1505 was fixed in PR #1513 by forcing flux to be `Quantity` in  `select_flux()` . From I could tell PR 1513 does fix the normalize() issue in QLP lightcurves in practice. I think issue 1505 could have been closed.

The `select_flux()` issue raised in my [above message](https://github.com/lightkurve/lightkurve/pull/1543#issuecomment-4356808160) is a rare case that is much less likely happen in practice (a non `nan` value in a `MaskedColumn` is masked).

m4cd4r4:
Thanks @orionlee for the deep dive. You're right that #1513 covers the QLP path in practice - I missed that when scoping this. Reframing as defensive `normalize()` hardening: `MaskedColumn` / `MaskedQuantity` can still reach `normalize()` via user-constructed LightCurves or paths that don't go through `select_flux()`, so it seems worth making the function robust on its own.

Pushed 0e638ed which reverts the masked branch from `np.ma.median/std` back to `np.nanstd(values.filled(np.nan))` - the pattern you recommended, and one that sidesteps the `np.ma.*` vs `MaskedQuantity` incompatibility you documented. The regression test from the previous commit is unchanged and still exercises the masked path. Local `pytest -k normalize` is green.

Happy to close if you and @Nschanche still consider this not worth carrying.

m4cd4r4:
Closing this one.

As @orionlee worked through above, the original issue (#1505) is already handled in practice by #1513 forcing `flux` to `Quantity` in `select_flux()`, so `normalize()` doesn't need to change for the QLP case. The remaining `MaskedColumn` / `MaskedQuantity` path this PR hardens is a rare defensive case and isn't worth carrying as an open PR against the maintainers' time.

Thanks @orionlee and @Nschanche for the thorough dig into it. Happy to revisit with a focused PR if that edge case ever turns up in real use.

PR #1552, opened 2026-03-29
Fix #1551: Add missing TESS Quality flags 17 and 18

S-Bhattacharya240611:
This PR fixes Issue #1551 by implementing the missing TESS quality flags for Rolling Band artifacts.


Nschanche:
Hi @S-Bhattacharya240611 thanks for opening this PR. As I mention in #1551, the hard coded values in the quality mask represent the SPOC standards. We do not maintain these for all of the High Level Science Products, as there is a lot of variation amongst products. You can define your own custom mask using known quality flags, which would be the recommended option when using other products. 

Also, in this PR, I think the flags are off by one bit. You would want 65536 and 131072 for flags 17 and 18.

PR #1556, opened 2026-04-23
Update URL

Nschanche:
The SkyBot service changed the URL for queries, so this fixes the search using the new URL.

### DIFF
--- a/src/lightkurve/utils.py
+++ b/src/lightkurve/utils.py
@@ -742,7 +742,6 @@ def _query_solar_system_objects(
     elif location.lower() == "tess":
         location = "C57"
 
-    #url = "http://vo.imcce.fr/webservices/skybot/skybotconesearch_query.php?"
     url = "https://ssp.imcce.fr/webservices/skybot/api/conesearch.php?"
     url += "-mime=text&"
     url += "-ra={}&".format(ra)
--- a/src/lightkurve/utils.py
+++ b/src/lightkurve/utils.py
@@ -776,6 +776,7 @@ def _query_solar_system_objects(
                 df = pd.concat([df, res])
     if df is not None:
         df.reset_index(drop=True)
+        df = df.groupby('Name', as_index=False).first() # Only keep unique results
     return df
 
 
--- a/src/lightkurve/utils.py
+++ b/src/lightkurve/utils.py
@@ -742,7 +742,8 @@ def _query_solar_system_objects(
     elif location.lower() == "tess":
         location = "C57"
 
-    url = "http://vo.imcce.fr/webservices/skybot/skybotconesearch_query.php?"
+    #url = "http://vo.imcce.fr/webservices/skybot/skybotconesearch_query.php?"
+    url = "https://ssp.imcce.fr/webservices/skybot/api/conesearch.php?"
     url += "-mime=text&"
     url += "-ra={}&".format(ra)
     url += "-dec={}&".format(dec)
--- a/src/lightkurve/utils.py
+++ b/src/lightkurve/utils.py
@@ -22,6 +22,7 @@ from astropy.visualization import (
     LinearStretch,
 )
 from astropy.time import Time
+import time
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


orionlee:
Code Suggestions:
1. update the skybot URLs in the docstring of `utils._query_solar_system_objects()`, and public `query_solar_system_objects()` methods in `lightcurve` and `targetpixelfile` modules.
2. I wonder if we should revive the two `test_SSOs()` tests (in [test_lightcurve.py](https://github.com/lightkurve/lightkurve/blob/f5c3f9fc2cf41ba04230ed3c936d89e1c019408d/tests/test_lightcurve.py#L1552) and the TPF counterpart). They are currently marked as skipped. Should we change it to `@pytest.mark.xfail` instead (so that they will run)?
If we do revive the tests, one complication is that the tests currently do not appear to work. For the epoch / coordinate from the test `asteroid_TPF`, Skybot returns empty result, and the codes will raise an exception. It looks like a different epoch / coordinate is needed. It looks like `asteroid_TPF` file needs to be replaced.

src/lightkurve/utils.py
orionlee: Make the `pd.read_csv` robust against empty result.

In the current iteration of Skybot, if the query (of an epoch/coordinate) has no result, it returns empty content, which would cause `pd.read_csv()` raise exception with a `pandas.errors.EmptyDataError`.

So when an user supplies a set of epochs, if any one of the epochs has no result, lightkurve will still raise an exception (even when there are SSOs in some of the epochs).

src/lightkurve/utils.py
```diff
@@ -775,6 +775,7 @@ def _query_solar_system_objects(
                 df = pd.concat([df, res])
     if df is not None:
         df.reset_index(drop=True)
+        df = df.groupby('Name', as_index=False).first() # Only keep unique results
```
orionlee: For the behavior change that only returns unique SSOs, rather than repeating the same object for each timestep of a query, it might have unintended consequences, breaking the user expectation

While I appreciate that in many cases, users only care about the unique SSOs, there could be cases  that users want to know the timestamps as well. E.g., in the past, I have occasionally used the timestamps to get an estimate of when an SSO enters / leaves a field.

orionlee (changes requested):

src/lightkurve/utils.py
Nschanche: Yeah, in general it is just a table with the same entry many times. But if users expect that, than this could be an issue. I can remove to keep the behavior the same.

orionlee: I vote for keeping the existing behavior. 
Returning unique SSOs is probably a more helpful default behavior in hindsight. But I don't think a change should be made at this point (in bug fixing releases).

Nschanche:
@orionlee I agree the test would be nice. It was commented out because there were a lot of HTTP errors, which there still are. I added a line to make up to 5 attempts before breaking, which seems to work in my testing, but not sure if it is the cleanest solution. Also, as you noted there no longer appears to be a returned object when using that TPF. I did play around with making another single cadence TPF that does return an asteroid that I could swap the file out for. The tests work when I make the swap locally.

orionlee:
For testing, I'm on the fence on whether to re-enable the test as well, given Skybot is not the most reliable, and is not a critical dependency in the grand scheme.  

One compromise could be that we mark it as xfail for network errors, something like
```python
# not sure if it should be HTTPError, IOError, or something else.
@pytest.mark.xfail(reason="Skybot web service fails intermittently", raises=HTTPError)
```

But the test will mask the rare cases such as the original issue here (where the HTTP URL has changed) as well.

---

On adding retry (up to 5 attempts) in the making the Skybot HTTP calls: it probably won't hurt. Though if we do it, 
- usually some backoff/jitter is done before retry,
- replace the `print` line with log

Nschanche:
> For testing, I'm on the fence on whether to re-enable the test as well, given Skybot is not the most reliable, and is not a critical dependency in the grand scheme.
> 
> One compromise could be that we mark it as xfail for network errors, something like
> 
> ```python
> # not sure if it should be HTTPError, IOError, or something else.
> @pytest.mark.xfail(reason="Skybot web service fails intermittently", raises=HTTPError)
> ```
> 
> But the test will mask the rare cases such as the original issue here (where the HTTP URL has changed) as well.
> 
> On adding retry (up to 5 attempts) in the making the Skybot HTTP calls: it probably won't hurt. Though if we do it,
> 
> * usually some backoff/jitter is done before retry,
> * replace the `print` line with log

I think we should skip the test. Skybox is brittle and outside our control. And as the orbital parameters for the solar system objects are refined, the results of the test may change, as it appears to have happened with the current tpf.

orionlee:
Ok. Let's not touch the test.

orionlee (approved):

PR #1570, opened 2026-06-24
Tweak poetry install in "Developing for Lightkurve" docs

orionlee:
Include `--with dev` in poetry install instruction, as it's needed to run tests, build doc, etc.

### DIFF
--- a/docs/source/development/contributing.rst
+++ b/docs/source/development/contributing.rst
@@ -58,8 +58,9 @@ You can set up the environment as follows:
 .. code-block:: bash
 
     $ cd lightkurve
-    $ python -m pip install poetry
-    $ poetry install
+    $ curl -sSL https://install.python-poetry.org | python -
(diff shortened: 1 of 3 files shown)


docs/source/development/contributing.rst
```diff
@@ -58,8 +58,9 @@ You can set up the environment as follows:
 .. code-block:: bash
 
     $ cd lightkurve
-    $ python -m pip install poetry
-    $ poetry install
+    $ curl -sSL https://install.python-poetry.org | python -
```
Nschanche: Should this be python3?

$ curl -sSL https://install.python-poetry.org | python3 -

orionlee: Yeah, it should be `python3`, at least for Linux / Mac. Updated.

On Windows with Python from conda, I have to use `python` instead, but this is probably too much details.

PR #1576, opened 2026-06-24
Tests LS periodogram with `nifty-ls` suport

### LINKED ISSUE #1572: No lightkurve CI tests for nifty_ls functionality
in Lightkurve 2.6.0 we added support for nifty_ls (#1550 ) ,and have merged at least 1 patch since since (#1568 )- we should add some nifty-ls regression tests for future CI runs.  

@orionlee has demoed some good tests in #1568 - we should evaluate whether we want to drop them in as is into the remote-data tests, or come up with a modified version that mocks some arrays/data to test the functionality without remote calls.

orionlee:
Close #1572 

- Added sanity tests for `nifty-ls` support. 
  The general idea is to a) do sanity tests for basic `ls_methods`  of `fast`, `fastchi2`, and b) ensure the `nifty-ls` counterparts behave the same.
- The tests are run on CI for Linux and Mac. Windows skipped to avoid complication of installing nifty-ls on Windows.
- Also make `nifty-ls` an explicit optional dependency such that end users can install it as extras: `pip install lightkurve[nifty-ls]`.

---
FYI @tylerapritchard @Nschanche

### DIFF
--- a/.github/workflows/lightkurve-tests.yml
+++ b/.github/workflows/lightkurve-tests.yml
@@ -60,7 +60,7 @@ jobs:
         python -m venv ./poetry
         ./poetry/bin/python -m pip install --upgrade pip
         ./poetry/bin/pip install poetry==2.2.1
-        ./poetry/bin/poetry install --with dev
+        ./poetry/bin/poetry install --with dev --extras "nifty-ls"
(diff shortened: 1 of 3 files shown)


tylerapritchard:
Thanks - LGTM!

PR #1577, opened 2026-06-25
CI: auto-retry failed tests for  remote-data test

orionlee:
An attempt to reduce the number of misleading CI failures due to intermittent network errors in running remote-data test.

General idea: at the end of a remote-data test, if there are test failures, re-run the failed tests using [`pytest --last-failed`](https://docs.pytest.org/en/latest/how-to/cache.html#usage) again.
- No impact to the test codes. The change is centralized in `lightkurve-tests.yml`
- The re-run happens at the end of a remote-data test (as opposed to at the end of individual fail tests). This will (hopefully) give extra buffer time for any intermittent network / server issues resolve before the re-run.

Notes:
- in the test output log , search for `Re-run failed tests` to see if there are failed tests that were re-run.

Auto-retry effectiveness:
- for the [remote-data test](https://github.com/lightkurve/lightkurve/actions/runs/28147132864/job/83498865636?pr=1577) for this PR, the auto-retry didn't exactly help.
  - the test was run 3 times (i.e, 2 manually-triggered re-runs),
  - the first 2 runs failed even with the retry, due to timeout from `test_metrics.py::test_underfit_metric_neighbors()`, `test_cbvcorrector.py::test_CBVCorrector_retrieval()` (my impression is that they're common culprits)
  - 3rd run completed, without the need of retry.


Nschanche:
Yes, the number of timeout errors is really a problem, but I have found it will fail multiple times if I manually kick it off, which is an even longer delay. I guess this could at least reduce, but not eliminate, the total number of manual re-runs.

PR #1581, opened 2026-08-20
Add automatic Fourier term selection for Lomb-Scargle periodograms

bibinthomas123:
This PR adds select_nterms() to LombScarglePeriodogram to automatically select the number of Fourier terms using AIC or BIC.

The method evaluates models with 1 through max_nterms Fourier terms and selects the model with the lowest information criterion.

This provides a way to determine model complexity without requiring users to manually choose nterms.

Tests were added for the new functionality.


bibinthomas123:
@orionlee can you kindly check this code and let me know if there is any changes to be made or this is a good addition or not to the module

orionlee:
Lightkurve team at NASA TESS GI office (@Nschanche , @christinahedges , etc.) would be in a better position to address this.  
My understanding is that Lightkurve has been in maintenance mode, accepting bug fixes only. New features are put on hold until the next major version.

orionlee:
bibinthomas123, setting aside of whether new features are accepted for now, how do you envision the new function `select_nterms()` be used in practice?

Fundamentally, the `nterms` selected by the new function is dependent on the best frequency determined by the current model. So if the best frequency used is sub-optimal, the resulting model will still be sub-optimal, no matter what `nterms` is used.


Take an example of a Beta-Lyare type variable (TIC 127767174). If an user does the following steps:
1. Create a LS Periodogram with default parameters (i.e., `nterms == 1`). The best frequency returned is about 1 / 0.598 d, which is about half of the actual best frequency (~1 / (2 x 0.598) d, or 1 / 1.196 d)
2. Using the new `select_nterms()`, the `nterms` is determined to be 3
3. The resulting model is clearly sub-optimal (even by eye test) because the frequency used is sub optimal.

The models based on the sub-optimal frequency of 1 / 0.598 d.
image

bibinthomas123:
You're right  the current implementation assumes "frequency_at_max_power" is already the right frequency and only optimizes model complexity ("nterms") at that fixed frequency.

The Beta Lyrae example shows where that breaks down: when the strongest peak is a harmonic of the true fundamental, "select_nterms()" can't recover the correct period because the frequency never changes.

That was actually intentional  "select_nterms()" was meant to answer a narrower question: *given a frequency, how many Fourier terms does the data actually support?* Not to jointly pick the frequency and "nterms" at the same time.

The cleaner fix is probably to keep that separation and add a higher-level method that evaluates candidate frequencies and harmonics together with "nterms". I'll rethink the API with that distinction in mind.

bibinthomas123:
I’ll modify select_nterms() so that frequency is an explicit input/assumption, and treat joint frequency + nterms selection as a separate problem.

Nschanche:
Hi @bibinthomas123, thanks for your interest in Lightkurve and for engaging with this PR! @orionlee is right, in that we are moving lkv2 more or less into maintenance mode and are not adding major features. 
That being said, in general lightkurve is meant to be a toolbox that can be used to easily access and work with TESS/Kepler/K2 data. While we provide the structure to create periodograms, it is left up to the user to use the results as they see fit for their science case. So while we acknowledge the framework you showed to be valuable for your work, it is not necessarily something we would incorporate into the lightkurve package. Please reach out if you have any further questions or comments!

PR #1584, opened 2026-08-24
Replace cadence with frame in TargetPixelFile.get_coordinates()

### LINKED ISSUE #1404: tpf.get_coordinates() keyword argument name should be frame, not cadence.
.. for consistency with keyword argument names in tpf.plot(), for example.

btfcookies:
Resolves #1404 
`get_coordinates` was the only place in TargetPixelFile that called a frame index as 'cadence'. Every other location in the file - `tpf.plot()`, `tpf.animate()`, `tpf.interact` - uses the same concept as `frame`. 

## Changes
- `TargetPixelFile.get_coordinates()` now takes frame instead of cadence. Docstrings updated to match
- Added `test_get_coordinates_frame_keyword` which covers the new keyword, its equivalence to the positional and "all" forms, and the deprecation path. 

## Testing
`test_get_coordinates_frame_keyword` asserts that:
- `get_coordinates(frame=0)` has shape `tpf.shape[1:]`
- it equals both `get_coordinates(0)` and `get_coordinates()[0]`
- `get_coordinates(cadence=0)` raises LightkurveDeprecationWarning and still returns the same arrays as `frame=0`


Nschanche:
Hi @btfcookies , thanks for your engagement with lightkurve! This PR would introduce a complication for existing pipelines that implement lightkurve, and therefore we do not want to make this change at this time. I will close the related issue as well.

PR #1586, opened 2026-09-01
Use the column count for the default cutout column centre

dylanpulver:
`TargetPixelFile.cutout()` derives its default centre from `imshape = self.flux.shape[1:]` as

```python
x, y = imshape[0] // 2, imshape[1] // 2
```

but the edge computation a few lines below proves `imshape` is `(row, column)`: `col` is clamped against `imshape[1]` and `row` against `imshape[0]`. So the column centre gets the row count and the row centre gets the column count. Square TPFs cancel it; on a rectangular one the cut is taken around the wrong pixel, comes back the wrong size, and the `RA_OBJ`/`DEC_OBJ` written into the new header are averaged over the wrong region.

Rectangular TPFs are a normal product — `cutout_size` is documented as "int, float or tuple" in `SearchResult.download()`, and any earlier `cutout(size=(a, b))` gives one too.

Measured on the repo's own `tests/data/tess25155310-s01-first-cadences.fits.gz` (11x11):

```
rect = tpf.cutout(center=(5, 5), size=(9, 3))   ->  (5, 3, 9)
rect.cutout()                                    ->  (5, 2, 3)     expected (5, 3, 5)
```

`imshape` is `(3, 9)`, so `x, y` come out `(1, 4)` instead of `(4, 1)`.

Why nothing caught it: every TPF passed to `cutout()` in `test_cutout` is square (3x3 and 11x11), and the one non-square case gives an explicit `size` to a square TPF, so `imshape[0] == imshape[1]` in every existing call. `cutout()` also has no caller anywhere in `src/` — the only occurrence is a comment in `interact.py` — so nothing in the library exercises the default centre either.

Ran, same venv, Python 3.13, macOS arm64: `pytest` with the four S3-backed tests in `tests/io/test_read.py` deselected in both legs (they hang against `s3://stpubdata` here, unrelated to this change and not marked `remote_data`) — 261 passed / 7 failed before, 262 passed / 7 failed after. The 7 failures are byte-identical in both legs and are all `ModuleNotFoundError: No module named 'IPython'` in the `interact` tests, which is this venv, not the tree.

Mutants, each caught by the new test: (A) revert the swap; (B) `imshape[1] // 2` for both, which fixes the column and breaks the row; (C) `imshape[0] // 2` for both.

Not tested: `--remote-data`, `--doctest-modules`, and any platform other than macOS arm64 / Python 3.13.

Implementation, tests and this description were drafted with Claude Opus 5.


tests/test_targetpixelfile.py
```diff
@@ -737,6 +737,26 @@ def test_cutout():
         assert ntpf.targetid == tpf.targetid
 
 
+def test_cutout_default_center_non_square():
```
Nschanche, line 745: This test is quite confusing, partially because you're starting out with a square TPF. You could start with asteroid_test.fits which is already in the test folder, and is 11x10 (col x row)

dylanpulver: Agreed on starting from a non-square fixture — the intermediate cutout was doing nothing but manufacturing a rectangle. Switched, but not to `asteroid_test.fits`, and the reason is worth stating.

`asteroid_test.fits` is 10 rows x 11 columns. The bug swaps `imshape[0] // 2` with `imshape[1] // 2`, and there those are `10 // 2 = 5` and `11 // 2 = 5` — identical, so the swap cancels. I checked: on a build with the fix reverted, a test written on that fixture still passes. It reads as a rectangular case but cannot detect this bug.

Detection needs the row and column half-counts to differ, i.e. the two dimensions to differ by at least 2. I measured every non-square TPF in `tests/data`:

| fixture | rows x cols | halves | detects |
|---|---|---|---|
| `asteroid_test.fits` | 10 x 11 | 5, 5 | no |
| `test-tpf-kplr-tabby-100-cadences.fits` | 10 x 11 | 5, 5 | no |
| `test-tpf-kplr-tabby-first-cadence.fits` | **7 x 8** | **3, 4** | **yes** |

So the test now uses `test-tpf-kplr-tabby-first-cadence.fits`, which is natively rectangular as you asked and also pins the fix — with the source change reverted it fails on all 25 pixels of the cutout. `tests/test_targetpixelfile.py` is 52 passed, 1 skipped.

Happy to add a short comment naming the 7x8 fixture as the deliberate choice if you would rather that reasoning live in the file than in this thread.

PR #1587, opened 2026-09-03
update lightkurve to read TARS light curves

awboyle:
This pull request allows lightkurve to download and read light curves from the TARS HLSP (https://archive.stsci.edu/hlsp/tars), which contains ~40 million light curves derived from TESS full frame images from sectors 1 -- 96. The issue and changes are as follows: 

TARS files were not recognized by detect_filetype(), so read() fell back to the generic reader — which crashed because it doesn't accept the quality_bitmask keyword that SearchResult.download() always passes. Additionally, TARS files contain only TIME and FLUX columns, with no quality flags for a bitmask to act on.

**Fix**

 Added a dedicated TARS reader following the same pattern as the other TESS HLSPs (QLP, TGLC, CDIPS, etc.):

  - src/lightkurve/io/tars.py (new): read_tars_lightcurve() wraps read_generic_lightcurve() and accepts quality_bitmask. Since TARS files carry no per-cadence quality flags, the returned light curve is assigned a quality column of zeros (all cadences treated as good, so quality_bitmask has no effect — this is documented in the docstring). Also sets AUTHOR="TARS", NORMALIZED=True (TARS fluxes are normalized), and TARGETID/OBJECT/LABEL as TIC <ticid> for consistency with SPOC/QLP.
  - src/lightkurve/io/detect.py: detect TARS files via the HLSPID = 'TARS' keyword in the primary header.
  - src/lightkurve/io/read.py: dispatch filetype "TARS" to TessLightCurve.read(..., format="tars").
  - src/lightkurve/io/__init__.py: register the reader with the astropy io registry.
  - docs/source/reference/io.rst: list tars.read_tars_lightcurve under TESS Data Products.

**Testing**

  - tests/io/test_tars.py (new, @pytest.mark.remote_data): verifies filetype auto-detection, that fluxes match the raw FITS data, that the zero-filled quality column is assigned, and the full search_lightcurve → download() path for TIC 51791910 (sector 55). TIC 51791910 is a randomly chosen test star.
  - Verified the original failing case end-to-end: download_all() now returns a LightCurveCollection of TessLightCurve objects (sectors 55/82/92 for TIC 51791910) with btjd time format and normalized flux.


orionlee:
@Nschanche At this point, are PRs for supporting new HLSPs accepted?

For TARS lightcurves, without any code changes, users can still read individual TARS FITS files with `lk.read()`. But 
- automatic detection, necessary for the `SearchResult.download_all()` workflow, would not work for SearchResult with TARS products.
- some nice-to-have headers, such as `LABEL`, `NORMALIZED`, would be missing.

Nschanche:
Hi @orionlee yes we will still support new HLSPs, so no problem adding a PR for this.

orionlee:
The PR looks solid to me.

Nschanche:
Agreed, thanks for the PR @awboyle, it will be great to add access to your light curves!

awboyle:
Great, thank you both!!

Issue #1496, opened 2025-05-17
`FoldedLightCurve.bin()` needs to be added to documentation


orionlee:
@Nschanche  PR #1491 overrides `FoldedLightCurve.bin()` with a different function signature. So the docstring needs to be exposed.

https://github.com/lightkurve/lightkurve/blob/main/docs/source/reference/foldedlightcurve.rst

oaksprout:
Would you want this one done? — Lightkurve's FoldedLightCurve.bin() override, added in a merged PR, is missing from the public API documentation page (#1496). If yes, we'll do it and open a PR; if not, say so (or just ignore this) and we won't touch it.

How it works: an AI agent does the work against your write-up; a human reviews it, submits it, and answers the review. You keep full control — nothing lands unless you merge it.

We're testing whether work offered this way is useful to maintainers — a blunt "no" is as helpful as a yes.

PR #1589, opened 2026-09-14
docs: add FoldedLightCurve.bin to API reference

shaikn6:
## What
Add `FoldedLightCurve.bin` to the API reference page (`docs/source/reference/foldedlightcurve.rst`).

## Why / evidence
`FoldedLightCurve` overrides `bin()` with phase-folded-specific behaviour and its
own docstring (`src/lightkurve/lightcurve.py`, `class FoldedLightCurve`), but that
method is not listed on the `FoldedLightCurve` API reference page, so it never
gets its own rendered doc page. Issue #1496 reports this; maintainer @orionlee
confirmed the hosted page shows it but the change to `foldedlightcurve.rst` was
never merged back.

Confirmed the override is real and the rst omits it:

    $ python -c "import ast; t=ast.parse(open('src/lightkurve/lightcurve.py').read()); print([m.name for n in t.body if getattr(n,'name',None)=='FoldedLightCurve' for m in n.body if isinstance(m, ast.FunctionDef)])"
    ['__init__', 'phase', 'cycle', 'odd_mask', 'even_mask', ..., 'to_fits', 'bin', 'copy']
    $ grep -c bin docs/source/reference/foldedlightcurve.rst   # 0 before this change

## Change
Adds a new `Modified methods` section listing `FoldedLightCurve.bin`, mirroring
the existing `Modified plotting methods` autosummary block. Docs only, one file,
9 added lines.

## Verified
- AST check above confirms `bin` is defined directly on `FoldedLightCurve`.
- The added `.. autosummary:: :toctree: api/` block is copied verbatim from the
  adjacent working section, so it renders identically.

Ref #1496


orionlee:
@Nschanche can probably shed some light into this:
- On the official doc website, `FoldedLightCurve.bin()` is now properly documented ( [link](https://lightkurve.github.io/lightkurve/reference/api/lightkurve.FoldedLightCurve.bin.html#lightkurve.FoldedLightCurve.bin) )
- I don't see any changes in `main` that could make it happen. Perhaps the changes have been made in some private branch not merged to `main`?

Nschanche:
@tylerapritchard updated the API I believe. Any thoughts on this Tyler?

shaikn6:
@orionlee @Nschanche that matches what I found writing this PR — `FoldedLightCurve.bin` is a real override (own docstring, distinct from the base `LightCurve.bin`), the hosted docs page shows it, but `docs/source/reference/foldedlightcurve.rst` on `main` never got the autosummary entry added (0 hits for `bin` before this PR). So it's not a private-branch situation — the page content and the rst source just drifted apart. This PR adds the missing entry to close that gap. Let me know if you want it structured differently (e.g. folded into the existing "Modified plotting methods" block instead of a new section).

